PR #481, opened 2014-01-12
Proper Numeric support

### LINKED ISSUE #476: Wrong method calling on extended bridged classes
```ruby
class String
	def foo
		puts "foo"
	end
end

class Foo
  def foo
    puts "baz"
  end
end

class Object
	def foo
		puts "bar"
	end
end

"lol".foo
Foo.new.foo
```

This prints `foo` and `baz` in Ruby, but `bar` and `baz` in Opal.

meh:
Early pull request **TO NOT MERGE** so we can fix the main issue with this.

We need a way to either bridge a class with a different base class (might fix #476 as well), or hack around the ancestors for `Number` so it is seen as an instance of `Numeric`.

This compliancy is fairly important for any number related code, I personally need to use it in opal-browser and another project.

This issue came up while working on the `Math` module, both the lack of `Complex` and `Rational` was throwing off specs.


meh:
Replaced by #1064.

PR #589, opened 2014-09-11
Included modules as ancestors 🔗

elia:
### This PR is here for everyone to review or suggest stuff

The idea is to inject the a copy of the module in the ancestor chain of a class each time the module is included.
The copy is named "iclass" as included class and has as prototype the previous parent prototype of the class (the superclass or another module).


wied03:
@elia - Given ancestors is now passing the specs, can this be closed?

elia:
@wied03 - good call

PR #682, opened 2015-01-06
Use case insensitive search when finding Encodings

wasnotrice:
Other rubies allow both `Encoding.find('binary')` and
`Encoding.find('BINARY')`. Opal was only allowing the uppercase version.
This change permits the lowercase search string as well.


elia:
Thanks!

PR #687, opened 2015-01-08
fix rescue’s default behavior

huochunpeng:
When no type is given, “rescue” should catch StandardError by default,
*NOT* all types of Exceptions.

I am surprised `rubyspec`'s `corelib/language/rescue_spec.rb` did not test this.

### DIFF
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -98,8 +98,10 @@ module Opal
       def compile
         push "if ("
         if rescue_exprs.empty?
-          # if no expressions are given, then catch all errors
-          push "true"
+          # if no expressions are given, then catch StandardError only
+          push "Opal.rescue($err, ["
+          push expr(Sexp.new([:const, :StandardError]))
+          push "])"
         else
           push "Opal.rescue($err, ["
           rescue_exprs.each_with_index do |rexpr, idx|
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -96,21 +96,17 @@ module Opal
       children :args, :body
 
       def compile
-        push "if ("
+        push "if (Opal.rescue($err, ["
         if rescue_exprs.empty?
           # if no expressions are given, then catch StandardError only
-          push "Opal.rescue($err, ["
           push expr(Sexp.new([:const, :StandardError]))
-          push "])"
         else
-          push "Opal.rescue($err, ["
           rescue_exprs.each_with_index do |rexpr, idx|
             push ', ' unless idx == 0
             push expr(rexpr)
           end
-          push "])"
         end
-        push ") {"
+        push "])) {"
 
         if variable = rescue_variable
           variable[2] = s(:js_tmp, '$err')
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -16,7 +16,11 @@ module Opal
       end
 
       def compile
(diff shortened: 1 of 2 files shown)


huochunpeng:
There is some weird bug on my last commit, fails on mri 2.0 and 2.2, but not mri 2.1 and 1.9 / rubinius / jruby. (did I hit a mri regression?)
The compiled javascript is broken on `spec/corelib/language/precedence_spec.rb`

I can confirm this is the only test having trouble.
```
  it "? : has higher precedence than rescue" do
    (true ? oops : 0 rescue 10).should == 10
  end
```

huochunpeng:
anyway, found another bug related.
try in http://opalrb.org/try/

    a = true ? 2 : 0 rescue 10
    puts a

nothing printed due to a missing `return` in compiled js.

huochunpeng:
one more serious bug on `rescue`.
still exists in my branch.

```
begin
  raise IOError, 'foo'
rescue RangeError              # this one is correct
rescue TypeError               # miss a return
rescue IOError                 # following two lines disappear in js
  puts "I got #{$!.message}"
end
```
```
/* Generated by Opal 0.7.0.beta2 */
(function(Opal) {

  Opal.add_stubs(['$raise']);
  try {
  return self.$raise($scope.get('IOError'), "foo")
  } catch ($err) {if (Opal.rescue($err, [$scope.get('RangeError')])) {
    return nil
    }else if (Opal.rescue($err, [$scope.get('TypeError')])) {
    nil
    }else { throw $err; }
  }
})(Opal);
```

elia:
Got caught by the latter a few days ago, anyway I'm (slowly) working to bring in CRuby tests, maybe they cover these cases.

elia:
Thanks! merged in f1c7e73e77346b282416c7ea9ada0373d4917375

@huochunpeng I also fixed the last bug you pointed out in #1269

huochunpeng:
@elia, the bug on one line rescue is still not resolved. You may want to open another ticket.
```
a = true ? 2 : 0 rescue 10                         # fail
a = (true ? 2 : 0) rescue 10                       # pass
a = begin; true ? 2 : 1; rescue; 10; end           # fail
a = begin; (true ? 2 : 1); rescue; 10; end         # fail
```

To illustrate
```
a = true ? 2 : 0 rescue 10
```
Compile result doesn't give variable "a" any val because of missing return in `try` block.
```
(function(Opal) {
  var $a, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, a = nil;

  return a = (function() {try {(function() {if ((($a = true) !== nil && (!$a.$$is_boolean || $a == true))) {
    return 2
    } else {
    return 0
  }; return nil; })() } catch ($err) { if (Opal.rescue($err, [$scope.get('StandardError')])) {return 10}else { throw $err; } }})()
})(Opal);
```

```
a = (true ? 2 : 0) rescue 10
```
Extra brackets yield correct result
```
(function(Opal) {
  var $a, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, a = nil;

  return a = (function() {try {return ((function() {if ((($a = true) !== nil && (!$a.$$is_boolean || $a == true))) {
    return 2
    } else {
    return 0
  }; return nil; })()) } catch ($err) { if (Opal.rescue($err, [$scope.get('StandardError')])) {return 10}else { throw $err; } }})()
})(Opal);
```

elia:
@huochunpeng thanks! 
yes, a separate issue would be better, do you mind opening it with a link to this one in the description?

PR #697, opened 2015-02-06
Utility to run a block in context of `this`

sleewoo:
When passing a block it will run in the context it was defined, normally.
This method wraps the block into a proc that when called will run the block in context of incoming `this`

https://gist.github.com/sleewoo/8d785061fc0cfbe5675d

### DIFF
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -178,6 +178,11 @@ module Kernel
       a.map! {|x| Native(`x`)}
       instance = Native(`this`)
       %x{
+        // if window is current scope, run the block in the scope it was defined
+        if(this.toString() === '[object Window]') {
+          return block.apply(self, a);
+        }
+        // otherwise run the block in the current scope
         return (function() {
           var s = block.$$s;
           block.$$s = null;
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -160,6 +160,39 @@ module Kernel
       }
     }
   end
+  
+  # allows to pass a proc that runs in context of `this`
+  #
+  # @example
+  #   options = {
+  #     persist: false,
+  #     render: function { |item, escape|
+  #       '<div>' + escape.call(item.name) + '</div>'
+  #     }
+  #   }
+  #
+  def function *args, &block
+    args.any? && raise(ArgumentError, '`function` does not accept arguments')
+    block || raise(ArgumentError, 'block required')
+    proc do |*a| # no way to expect a block here as this is always called from JavaScript
+      a.map! {|x| Native(`x`)}
+      instance = Native(`this`)
+      %x{
+        return (function() {
+          var s = block.$$s;
+          block.$$s = null;
+          try {
+            var r = block.apply(instance, a);
+            block.$$s = s;
+            return r;
+          } catch (e) {
+            block.$$s = s;
+            throw e;
+          }
+        })();
+      }
+    end
+  end
 end
 
 class Native::Object < BasicObject


elia:
I like the idea, but I'd like to discuss with @adambeynon and @meh about:

- the name `function` (not sure if it's clear what it does)
- how `@this` is made available (not very fond of instance vars)

and then it would need some specs :)

sleewoo:
well, it does not set `@this`, only local variables used.
used `function` to be "closer" to javascript :)

Ajedi32:
I'm pretty familiar with Javascript and very familiar with Ruby, but I found this to be rather confusing. Maybe that's because I'm having trouble bridging the gap between the different constructs in the Ruby and JS worlds.

What does it mean to run a Ruby block in a context of `this` (which is a concept only in JavaScript, and has no meaning in the Ruby world)? Does that affect anything on Ruby's side of things? Or is it just a matter of changing what the x-string `` `this` `` refers to within that method? How does this differ from `instance_eval`, which can already execute Ruby blocks in a different context in Ruby?

sleewoo:
It is basically used to work with existing js libs that accepts functions as callbacks.
If providing a proc to such a lib, it will run in the context the proc was defined, normally.

To run it in the context of calling lib, we have to wrap it into another proc that will run the original proc in correct context.

in terms of js, it just wraps a function in another function and use `apply` on original one.

it works just like `instance_exec`, in fact code stolen from [Opal's corelib](https://github.com/opal/opal/blob/dc8f5fa5545ccb46b48753e1ea6e471591346dc6/opal/corelib/basic_object.rb#L58) :)

sleewoo:
worth to note, it will also convert received arguments into native ones, so you do not need to do this manually nor to use backticks...

contrived example
using a proc:
```ruby
Element.get('/some/url', proc { |xhr|
  if Native(`xhr`)[:responseText] == 'OK' ...
  # or
  if `xhr.responseText` == 'OK' ...
})
```

using this wrapper:
```ruby
Element.get('/some/url', proc { |xhr|
  if xhr[:responseText] == 'OK' ...
})
```

Ajedi32:
Ah, I think I get it now. A lot of libraries rely on the value of "this" in the callback functions, but you're saying Opal ignores this value, somehow? So this is then designed to make the value available to Ruby.

I guess I just don't understand Opal's internals well enough to work out what this does from the commit itself...

stdlib/native.rb
```diff
@@ -160,6 +160,44 @@ def Array(object, *args, &block)
...
+      a.map! {|x| Native(`x`)}
+      instance = Native(`this`)
+      %x{
+        // if window is current scope, run the block in the scope it was defined
+        if(this.toString() === '[object Window]') {
+          return block.apply(self, a);
+        }
+        // otherwise run the block in the current scope
+        return (function() {
+          var s = block.$$s;
```
Ajedi32: What is `$$s` here? Is that the value of `self` within the block?

stdlib/native.rb
```diff
@@ -160,6 +160,44 @@ def Array(object, *args, &block)
...
+  def function *args, &block
+    args.any? && raise(ArgumentError, '`function` does not accept arguments')
+    block || raise(ArgumentError, 'block required')
+    proc do |*a| # no way to expect a block here as this is always called from JavaScript
+      a.map! {|x| Native(`x`)}
+      instance = Native(`this`)
+      %x{
+        // if window is current scope, run the block in the scope it was defined
+        if(this.toString() === '[object Window]') {
+          return block.apply(self, a);
```
Ajedi32: What is "self" in this context? Since this is within an x-string, wouldn't it refer to a local JavaScript variable, and not the Ruby `self`?

sleewoo: yes, here it will run the original block in the context it was defined, and `self` here refers to that context.
`// if window is current scope, run the block in the scope it was defined`

stdlib/native.rb
sleewoo: i have very vague ideas why this is done, i just stolen it from stdlib's `instance_exec` :)
https://github.com/opal/opal/blob/dc8f5fa5545ccb46b48753e1ea6e471591346dc6/opal/corelib/basic_object.rb#L58

stdlib/native.rb
Ajedi32: Right, but `self` is a keyword in Ruby, not a local variable in JavaScript (unless that's how Opal implements it?). So since this is in an x-string and you're referencing JavaScript `self`, not Ruby `self`...

Sorry for all the question. Again, I'm not familiar with Opal's internals.

sleewoo: Opal makes use of `self` inside javascript to refer to the scope the method/proc/class etc. was defined
http://opalrb.org/try
see the compiled javascript

Ajedi32: Ah! Good idea, I didn't think to just try the compiler.

It looks like within functions, `self` is set to `this`.

meh:
This was definitely a good idea, I fixed the code and renamed the function to `Native.proc`.

PR #701, opened 2015-02-23
Implement String#[] with a string argument

vais:
DONE:
pass "String#[] with String returns other_str if it occurs in self"
pass "String#[] with String returns nil if there is no match"

TODO:
still fails "String#[] with String doesn't call to_str on its argument"
still fails "String#[] with String returns a subclass instance when
given a subclass instance”

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -98,7 +98,6 @@ class String
     when String
       return self.include?(index) ? index : nil
     when Regexp
-      #WTF - not implemented yet, but somehow the tests are passing
       #TODO implement by testing the regexp, then setting index and
       #length accordingly to let the code below do the rest of work
     when Range, Numeric
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -95,6 +95,10 @@ class String
 
   def [](index, length = undefined)
     %x{
+      if (typeof index === 'string') {
+        return self.indexOf(index) === -1 ? nil : index;
+      }
+
       var size = self.length;
 
       if (index.$$is_range) {
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -94,11 +94,19 @@ class String
   end
 
   def [](index, length = undefined)
+    case index
+    when String
+      return self.include?(index) ? index : nil
+    when Regexp
+      #WTF - not implemented yet, but somehow the tests are passing
+      #TODO implement by testing the regexp, then setting index and
+      #length accordingly to let the code below do the rest of work
+    when Range, Numeric
+      #implemented in JS below
+    else
+      raise TypeError, "type mismatch: #{index.class} given"
+    end
     %x{
-      if (typeof index === 'string') {
-        return self.indexOf(index) === -1 ? nil : index;
-      }
-
       var size = self.length;
 
       if (index.$$is_range) {
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -94,17 +94,6 @@ class String
   end
 
   def [](index, length = undefined)
-    case index
-    when String
-      return self.include?(index) ? index : nil
-    when Regexp
-      #TODO implement by testing the regexp, then setting index and
-      #length accordingly to let the code below do the rest of work
-    when Range, Numeric
-      #implemented in JS below
-    else
-      raise TypeError, "type mismatch: #{index.class} given"
-    end
     %x{
       var size = self.length;
 
@@ -138,24 +127,39 @@ class String
         return self.substr(index, length);
       }
 
-      if (index < 0) {
-        index += self.length;
-      }
 
-      if (length == null) {
-        if (index >= self.length || index < 0) {
+      if (index.$$is_number) {
+        if (index < 0) {
+          index += size;
+        }
+
+        if (length == null) {
+          if (index >= size || index < 0) {
+            return nil;
+          }
+
+          return self.substr(index, 1);
+        }
+
+        if (index > size || index < 0) {
           return nil;
         }
 
-        return self.substr(index, 1);
+        return self.substr(index, length);
       }
 
-      if (index > self.length || index < 0) {
-        return nil;
+
+      if (index.$$is_string) {
+        return self.indexOf(index) !== -1 ? index : nil
       }
 
-      return self.substr(index, length);
+
+      if (index.$$is_regexp) {
+        #{raise NotImplementedError}
+      }
     }
+
+    raise TypeError, "type mismatch: #{index.class} given"
   end
 
   def capitalize
--- a/opal/corelib/string/inheritance.rb
+++ b/opal/corelib/string/inheritance.rb
@@ -15,6 +15,8 @@ class String
 end
 
 class String::Wrapper
+  `def.$$is_string = true`
+
   def self.allocate(string = "")
     obj = super()
     `obj.literal = string`
--- a/spec/filters/bugs/string.rb
+++ b/spec/filters/bugs/string.rb
@@ -30,8 +30,6 @@ opal_filter "String" do
   fails "String#[] with Regexp, index raises a TypeError when the given index can't be converted to Integer"
   fails "String#[] with Regexp, index raises a TypeError when the given index is nil"
   fails "String#[] with Regexp, index sets $~ to MatchData when there is a match and nil when there's none"
-  fails "String#[] with String doesn't call to_str on its argument"
-  fails "String#[] with String returns a subclass instance when given a subclass instance"
 
   fails "String#dup does not copy constants defined in the singleton class"
   fails "String#dup does not modify the original string when changing dupped string"
--- a/spec/filters/bugs/string.rb
+++ b/spec/filters/bugs/string.rb
@@ -156,10 +156,6 @@ opal_filter "String" do
   fails "String#slice with Regexp, index raises a TypeError when the given index can't be converted to Integer"
   fails "String#slice with Regexp, index raises a TypeError when the given index is nil"
   fails "String#slice with Regexp, index sets $~ to MatchData when there is a match and nil when there's none"
-  fails "String#slice with String returns other_str if it occurs in self"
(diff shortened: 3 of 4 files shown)


meh:
For the failures, look [here](https://github.com/opal/opal/blob/master/opal/corelib/string.rb#L189).

meh:
Wait, actually that doesn't make much sense since you're not calling it at all.

Also instead of `typeof` use `$$is_string` please, sometimes literals become objects, but they're still valid.

vais:
> Wait, actually that doesn't make much sense since you're not calling it at all.

I think you are referring to `to_str`? This is the test and the failure (even though I am not calling `to_str`, the test fails because it is expecting a `TypeError`):
```ruby
  it "doesn't call to_str on its argument" do
    o = mock('x')
    o.should_not_receive(:to_str)

    lambda { "hello".slice!(o) }.should raise_error(TypeError)
  end
```
```
  1. String#[] with String doesn't call to_str on its argument

    Expected TypeError but no exception was raised (h was returned)
    SpecExpectationNotMetError: Expected TypeError but no exception was raised (h was returned)
    at OpalClass.def.$initialize.$arity [as $new] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3361:15)
    at Opal.defn.$arity (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3037:31)
    at OpalClass.$fail_with (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:29347:35)
    at Function.Opal.defn.$arity [as $should] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:29379:64)
    at Object.$a.$$p.TMP_65 (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:78659:87)
    at Object.Opal.defn.TMP_2 [as $instance_eval] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:2382:22)
    at module_constructor.Opal.defs.TMP_5 [as $protect] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:25612:76)
    at $a.$$p.TMP_19 (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:25267:88)
    at Opal.yieldX (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:837:18)
    at self.$each.$$p (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3880:28)
```
The other test:
```ruby
  it "returns a subclass instance when given a subclass instance" do
    s = StringSpecs::MyString.new("el")
    r = "hello".slice!(s)
    r.should == "el"
    r.should be_an_instance_of(StringSpecs::MyString)
  end
```
fails as follows:
```
  2. String#[] with String returns a subclass instance when given a subclass instance

    Expected "h" to equal "el"
    SpecExpectationNotMetError: Expected "h" to equal "el"
    at OpalClass.def.$initialize.$arity [as $new] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3361:15)
    at Opal.defn.$arity (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3037:31)
    at OpalClass.$fail_with (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:29347:35)
    at $SpecPositiveOperatorMatcher.def.$== (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:24266:46)
    at Object.$a.$$p.TMP_67 (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:78665:25)
    at Object.Opal.defn.TMP_2 [as $instance_eval] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:2382:22)
    at module_constructor.Opal.defs.TMP_5 [as $protect] (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:25612:76)
    at $a.$$p.TMP_19 (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:25267:88)
    at Opal.yieldX (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:837:18)
    at self.$each.$$p (/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150223-3080-g8ca8a:3880:28)
```
Any ideas?

meh:
Ah ok, yeah then use what I linked, you need to use `Opal.coerce_to!`. Don't worry about the subclassing issue.

vais:
Also, @meh, I am a little confused regarding `$$is_string` versus this example in string.rb:
```ruby
  def index(what, offset = nil)
    if String === what
      what = what.to_s
```
Which is the preferred way: `$$is_string` or `String === what`?
Next question: so, there are native strings and ruby Strings? Are there any caveats for calling js methods (e.g. `indexOf()`) on a ruby String? E.g. the use of `to_s` in the example above.

meh:
Depends if you're writing Ruby or inlined JavaScript.

Ruby `String` is bridged to the native JavaScript `String`, it's the same thing. What I was talking about is how JavaScript handles literals and object literals.

```
> String.prototype.hue = function(){ return this }
[Function]
> typeof("")
'string'
> typeof("".hue())
'object'
```

vais:
I see, thanks, @meh, makes sense. I have some ideas to improve on this PR now...

meh:
Go ahead, I'll wait before merging.

vais:
Ok, @meh, I think this is ready to merge - it's passing all "String#[] with String" and "String#slice with String" specs. I will work on removing the remaining `String#[]` and `String#slice` failures from `spec/filters/bugs/string.rb` another day.

vais:
Hmm... judging by this quick-and-dirty test, I just killed the performance of this method  (`string` is my PR branch, `master` is the original code):
```
kelp:opal vais$ git branch
  master
* string
kelp:opal vais$ bundle exec opal-repl
>> S = "hello world"; t = Time.now; 1000000.times{S["wo"]}; Time.now - t
=> 2.106
>> S = "hello world"; t = Time.now; 1000000.times{S[2]}; Time.now - t
=> 3.729
>> exit
kelp:opal vais$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
kelp:opal vais$ bundle exec opal-repl
>> S = "hello world"; t = Time.now; 1000000.times{S["wo"]}; Time.now - t
=> 1.394
>> S = "hello world"; t = Time.now; 1000000.times{S[2]}; Time.now - t
=> 0.971
```
So, the new code is 2 to 3 times slower. Bummer. Is that because I am implementing it in ruby instead of js, or does it have more to do with the type checking that has been added? (I.e. `case index; when String...`) Any suggestions?

meh:
Could you move those things down to JavaScript?

The `case` does add some overhead which we try to avoid as much as possible in the corelib.

Basically:

```javascript
if (index.$$is_string) {
  return #{include? ...};
}

if (index.$$is_regexp) {
  #{raise NotImplementedError};
}

// old code
```

vais:
@meh performance is back to acceptable level now. All "String#[] with String" and "String#slice with String" specs are still passing after this refactor to JS.

To make *String#[] with String returns a subclass instance when given a subclass instance* pass, I had to add `def.$$is_string = true` to `String::Wrapper` class in string/inheritance.rb. I think the fact that a subclass of a string was returning false for `$$is_string` was actually a bug, so I fixed it. I hope that's ok?

Regexp will throw `NotImplementedError` for now, and I updated the bugs filter file accordingly.

Sorry for the ugly JS diff - I re-arranged the JS code for clarity - better an ugly diff one time than ugly code forever ;)

adambeynon:
:+1: Nice one. Looks good to me now, @meh good for merge?

On a side note, maybe we should improve `contributing.md` to recommend javascript for core lib when possible, as well as how the rubyspec filters work (and why we use them).

vais:
Thanks @adambeynon! I think the idea that JS should be used for the core is pretty obvious once you learn what's what even a little bit. But I agree that the filters subject deserves more attention in the docs because that's really the gateway to contributing to Opal. Once you know about the filters, you know what is and what's not implemented in Opal and how you can help. There's definitely a lot of work still to be done!

meh:
Yeah, looks good, merging.

elia:
@vais would you mind opening another PR with the changelog entry? (feel free to add your name as the author)

vais:
@elia here is the PR with the changelog entry: https://github.com/opal/opal/pull/705 Thanks!

Issue #698, opened 2015-02-13
Opal reports incorrect version numbers


bwl21:
```
puts RUBY_ENGINE_VERSION
````
Yields `0.6.1` even if performed in http://opalrb.org/try/# which shows

 `/* Generated by Opal 0.7.0.beta4 */`

vais:
@adambeynon is there a reason why `RUBY_ENGINE_VERSION` [here](https://github.com/opal/opal/blob/106cfa129ef4bb1b8d3b511dd7def0fbb2e91527/opal/corelib/variables.rb#L23) is not the same as `Opal::VERSION` [here](https://github.com/opal/opal/blob/8cf47331284160762b8fe0b9279aafbfe618be3a/lib/opal/version.rb#L2)? I believe that's what @bwl21 is referring to. I.e. are they different on purpose or is that a bug?

adambeynon:
@bwl21, @vais Yes this is a bug. We can't just alias them because `Opal::VERSION` is only present when the parser/compiler is loaded, which is not ideal for production.

I will fix this, but perhaps some sort of release policy is needed with a checklist of things to do for each gem release.

bwl21:
would it help to specify in https://github.com/opal/opal/blob/master/opal/corelib/variables.rb

RUBY_ENGINE_VERSION = VERSION

vais:
@bwl21 I guess not, based on @adambeynon's response, that constant is not available at runtime without the compiler loaded. I am still trying to wrap my mind around how opal works. I wonder if that can be part of some sort of build system... because updating it manually, even with a release checklist in place, this kind of stuff will fall through the cracks eventually.

PR #703, opened 2015-02-23
Fix issue #698 (Opal reports incorrect version numbers)

vais:


vais:
Issue #698 is still open, so, to fix it, I went ahead and added a test to spec/opal/core/language/predefined_spec.rb (which is where all the rest of the "variables" defined in opal/variables.rb are tested).

Then, when I tried to make my test fail, I saw a different test fail - looks like this issue has already been addressed by specs/lib/version_spec.rb.

For what it's worth, I think my solution is simpler and more consistent with the already existing tests for variables.rb, but I may be missing something. Just putting this PR out there anyway - feel free not to merge this, I won't get offended :)

elia:
@vais your PR is much better, I'll remove my spec after merging ;)

vais:
Thanks, @elia!

PR #709, opened 2015-02-25
String#[] and String#slice implementation now fully compliant with rubyspec

vais:


vais:
Regarding *Implement "String#[] with Range calls to_int on range arguments*:

Unfortunately *fails "String#[] with Range calls to_int on range arguments”* and *fails "String#slice with Range calls to_int on range arguments”* filters cannot be removed because of an issue with the spec itself: it expects `Range` constructor to call `<=>` on the mock object it uses for the range literal, but in Opal the `Range` constructor does not receive the mock, but receives a `Numeric` instead. Then it calls `<=>` on that `Numeric`, not the mock, so the mock’s expectation of getting called is not fulfilled. This makes the spec fail like so:
```
  1. An exception occurred during: Mock.verify_count String#[] with Range calls to_int on range arguments

    Mock 'from' expected to receive '<=>' exactly 2 times but received it 0 times
```
Here is the spec (the offending line is `from.should_receive(:<=>).twice.and_return(0)`)
```ruby
  it "calls to_int on range arguments" do
    from = mock('from')
    to = mock('to')

    # So we can construct a range out of them...
    from.should_receive(:<=>).twice.and_return(0)

    from.should_receive(:to_int).twice.and_return(1)
    to.should_receive(:to_int).twice.and_return(-2)

    "hello there".send(@method, from..to).should == "ello ther"
    "hello there".send(@method, from...to).should == "ello the"
  end
```
Not sure who the culprit is here: the spec or the way Opal range literals are implemented. Any suggestions?

meh:
I'm pretty sure it's Opal, can you open another bug for that?

Also please rebase the branch to master so I can merge it.

vais:
@meh I see @elia already merged, but for future reference, would you guys prefer PRs with a single giant squashed commit or multiple granular commits showing each change? Or is that even what you are referring to when you say "...please rebase the branch to master so I can merge it..."? Hope I am understanding you correctly, LMK please.

meh:
@vais no I meant someone pushed stuff to master and your branch was behind, so it couldn't be automatically merged because there were conflicts. @elia did it by hand.

Also I personally prefer commits that talk about specific additions changes in one, as in commits like "fix this spec and this spec" are fine "changed a wrong + somewhere" I'd rather have it squashed with the feature that error came from. But that's just me.

meh:
Addendum, your commits were fine.

vais:
I see what you mean now, thanks, @meh! And thanks @elia for taking the time to resolve conflicts to merge this PR.

Issue #699, opened 2015-02-17
Array#product


JikkuJose:
The `method` `Array#product` seem to be failing.

vais:
@JikkuJose I don't think this method has been implemented yet. There is a [spec for it](https://github.com/opal/rubyspec/blob/master/core/array/product_spec.rb) but it is also listed in [spec/filters/bugs/array.rb](https://github.com/opal/opal/blob/master/spec/filters/bugs/array.rb#L84), which means that this is something that needs to be worked on? @elia is everything in https://github.com/opal/opal/tree/master/spec/filters/bugs basically a "to do" item for Opal?

adambeynon:
@JikkuJose yeap, that's

adambeynon:
Sorry. Closed by mistake. Yes, that's a method yet to be implemented.

JikkuJose:
I just implemented `product` in a module and extended an already available class with it. Just wanted to know whether its recommended, I seem to feel the code seems be pretty slow. Its likely that I am making some stupid mistake, but felt like asking if there is anything recommended to use/not in Opal from Ruby?

vais:
@JikkuJose are you implementing it in Ruby or JS? I imagine something like `Array#product` is best implemented in JS for performance reasons.

JikkuJose:
@vais I implemented an algorithm for Conway Game of Life, but to my disappointment it ran less performant than what I expected. Its possible my implementation is pretty sloppy. Its done in Ruby. You can take a look at the [implementation](https://github.com/JikkuJose/Conway-Game-of-Life/blob/master/app/conway/algorithm.rb).

PR #711, opened 2015-02-27
Array#product implementation fully compliant with rubyspec

vais:


vais:
Fix for issue https://github.com/opal/opal/issues/699

vais:
Regarding the [Pass "Array#product does not attempt to produce an unreasonable number of products"](https://github.com/vais/opal/commit/ef7294b522cad20e2e804803a32aa132743d1f1a#diff-981b14160ae83ab1ab0c0936525d462dR1161) commit: [this max number](https://github.com/vais/opal/commit/ef7294b522cad20e2e804803a32aa132743d1f1a#diff-981b14160ae83ab1ab0c0936525d462dR1161) is completely arbitrary, not sure how else to handle it. In MRI there is basically a check to see if multiplication would cause integer overflow, and if it would, then the number is considered "unreasonable", and MRI will refuse to calculate the product. There is no such thing in JS. I basically pulled a max number out of a hat. It should probably be much smaller. But it's small enough for the spec to get the `RangeError` that it's expecting, so the spec passes.

vais:
@meh please take a moment to close [the issue](https://github.com/opal/opal/issues/699) when you get a chance.

Ajedi32:
@vais For future reference, I'm pretty sure GitHub will close related issues automatically when the pull request gets merged if you put "Fix #issue-id" somewhere in one of your commit messages.

vais:
@Ajedi32 oh, sweet, I didn't know that, thanks!

Issue #666, opened 2014-12-11
Regexp match with a string argument doesn't work


easydatawarehousing:
When trying a regular expression match using a string argument the match fails. When creating the regexp matcher via Regexp.new(string) it works as expected. In IRB both versions work as expected.
Is this a bug or a feature ?

    puts '123'.match(Regexp.new('\\d+')) # => 123
    puts '123'.match('\\d+')             # => opal:nil / irb:123

Opal version used: opal 0.7.0.beta4

PR #712, opened 2015-02-27
Fix issue #666

vais:
Note that it says “…without escaping” in the spec:
```ruby
describe "String#match" do
  #…
  it "converts string patterns to regexps without escaping" do
    'hello'.match('(.)\1')[0].should == 'll'
  end
end
```


elia:
Ref #666

PR #713, opened 2015-02-28
Module#const_get accepts a scoped constant name

vais:
Fix issue #694


Ajedi32:
#694 still needs to be closed btw.

GitHub's auto-close feature only works if the keyword (e.g. "Fix" in this case) is immediately followed by the issue number. So `Fix #694` will work, but `Fix issue #694` won't. (see https://help.github.com/articles/closing-issues-via-commit-messages/)

Issue #571, opened 2014-08-23
String#scan yields "undefined" instead of "null" such that a subsequent compact fails


bwl21:
```ruby
a = "abc".scan(/(a)|(b)/).map{|a|a.compact}
puts `JSON.stringify(a)`
```
yields in http://opalrb.org/try/#
```json
[["a",null],[null,"b"]]
```

while in pry
```ruby
[16] pry(main)> puts "abc".scan(/(a)|(b)/).map{|a|a.compact}.to_json
[["a"],["b"]]
```

In fact the scan above yields
```json
[["a",undefined],[undefined,"b"]]
````
Therefore the ```compact``` does not work properly.

as @meh mentioned elsewhere, "undefined" cannot be treated as "nil" or "null" but somehow this should be fixed.

PR #714, opened 2015-02-28
String#scan: use MatchData captures to deal with undefined values

vais:
Fix issue #571

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -646,10 +646,9 @@ class String
       while ((match = pattern.exec(self)) != null) {
         var match_data = #{MatchData.new `pattern`, `match`};
         if (block === nil) {
-          match.length == 1 ? result.push(match[0]) : result.push(match.slice(1));
-        }
-        else {
-          match.length == 1 ? block(match[0]) : block.apply(self, match.slice(1));
+          match.length == 1 ? result.push(match[0]) : result.push(#{`match_data`.captures});
+        } else {
+          match.length == 1 ? block(match[0]) : block.apply(self, #{`match_data`.captures});
         }
       }
 


vais:
Does anyone know what needs to be done to get the `corelib/core/string/scan_spec` to run? I suspect there are lots of issues with `String#scan` that we don't even know about, since the spec is not being run at all. Currently, it is commented out with a note that parsing the file causes an infinite loop:
```ruby
...
corelib/core/string/rstrip_spec
# corelib/core/string/scan_spec - infinite loop on parsing
corelib/core/string/setbyte_spec
...
```
I tried running it, and yes, it does cause an infinite loop :) I also tried commenting out various parts of it with no luck: looks like as long as there are any `it` blocks active, the node process goes into a loop. Any pointers?

Ajedi32:
The *parser* goes into an infinite loop? That's really strange. Is there an issue open for that?

vais:
@Ajedi32 yep, it does, and no, there is no issue for that, but there is a comment to that effect in opal/spec/rubyspecs file that I quoted above.

Ajedi32:
@vais Yeah, I realize there's a comment. It just seemed to me like that would be the kind of thing you'd expect to find in a bug tracker, not just a comment. Weird...

vais:
@Ajedi32 hopefully @adambeynon, @meh, or @elia can comment on this - I have no idea about the history of this problem, why there is not an issue open for it, or whether it can even be fixed in Opal or requires modifying rubyspec somehow.

As of now I can't even get anyone to merge this PR - I can only guess because nobody feels comfortable merging a change that is not covered by tests, no matter how small or straight-forward that change may be.

I do hope this gets merged though - it fixes a pretty bad bug (leaking `undefined` into Ruby-land), and I tested in manually every which way I could.

meh:
Honestly, I wouldn't want to touch the parser with a 42 light years pole, that's @adambeynon turf, I'll merge it even without tests.

elia:
Hey @vais, you're doing an awesome job with these PRs. I happen to wait before merging in order to properly review and understand the change first. But usually at that point @meh already took the time to do it. 😁

vais:
Thank you @meh and @elia!

Hopefully @adambeynon can provide some insight on what's wrong with that spec. Looks like this is a very old issue, since I could not even figure out by looking at the git blame who and when added that infinite loop comment next to string/scan_spec. If @adambeynon does not know either, then we are truly lost :)

meh:
Also I second @elia, I appreciate the work you're putting in these pull requests as well.

vais:
@meh thanks for your encouragement!

PR #715, opened 2015-03-01
String#succ and String#next implementation fully compliant with rubyspec

### LINKED ISSUE #535: Range with numeric strings with trailing zeros freezes browser
In Ruby, you can use ranges not only for Numbers but also for strings, like this.
```
("a".."b").to_a
=> ["a", "b"]
 ```

The same works with numeric strings, like this:
```
("01".."05").to_a
 => ["01", "02", "03", "04", "05"] 
```
This also works fine in Opal unless you cross the border between when a trailing zero is required and when it's not required anymore.
`("09".."10").to_a` should give you ` => ["09", "10"]`

However, this makes the browser (Firefox 29) freeze (until it complains about an unresponsive script).

vais:
Fix #535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 *   `Module#const_get` accepts a scoped constant name
 
+*   `String#succ` and `String#next` implementation fully compliant with rubyspec
+
 ## 0.7.1 2015-02-14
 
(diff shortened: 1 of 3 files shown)


Ajedi32:
Nice work on all these methods by the way. The closer we can get Opal to matching MRI's behavior, the better. :+1:

vais:
Thanks, @Ajedi32!

PR #717, opened 2015-03-02
String#===, String#==, and String#eql? implementation fully compliant with rubyspec

vais:
*"String#=== returns false if obj does not respond to to_str"* and  *"String#== returns false if obj does not respond to to_str"* actually pass, except for the one line that tests a symbol argument:
```ruby
'hello'.send(@method, :hello).should be_false
```
Same goes for *"String#eql? when given a non-String returns false"* - only the following line fails:
```ruby 
'hello'.should_not eql(:hello)
```
There is currently no way to pass these assertions because strings and symbols are identical in Opal.

I really would like to see those `fail` lines gone from filters/bugs/string.rb, and I see 3 possible courses of action (in the order from least invasive to most invasive to Opal):

1. Wrap `not_compliant_on :opal do ...` around the offending assertions in the rubyspec. 
2. Set a new additional property on the string object at compile time to indicate that is was originally meant to be a symbol. This way we can tell symbols from strings at runtime if/as needed with minimum overhead.
3. Make Symbol a subclass of String and set `$$is_symbol` flag on it? I don't like this one, but throwing it out there anyway.

What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 *   `String#succ` and `String#next` implementation fully compliant with rubyspec
 
+*   `String#===`, `String#==`, and `String#eql?` implementation fully compliant with rubyspec
+
 ## 0.7.1 2015-02-14
 
 *   CLI options `-d` and `-v` now set respectively `$DEBUG` and `$VERBOSE`
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -81,7 +81,16 @@ class String
   end
 
   alias eql? ==
-  alias === ==
+
+  def ===(other)
+    %x{
+      if (#{Opal.respond_to? `other`, :to_str}) {
+        return #{other == self};
+      } else {
+        return false;
+      }
+    }
+  end
 
   def =~(other)
     %x{
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -75,22 +75,19 @@ class String
   end
 
   def ==(other)
-    return false unless String === other
-
-    `#{to_s} == #{other.to_s}`
+    %x{
+      if (other.$$is_string) {
+        return self.toString() === other.toString();
+      }
+      if (#{Opal.respond_to? `other`, :to_str}) {
+        return #{other == self};
+      }
+      return false;
+    }
   end
 
   alias eql? ==
-
-  def ===(other)
-    %x{
-      if (#{Opal.respond_to? `other`, :to_str}) {
-        return #{other == self};
-      } else {
-        return false;
-      }
-    }
-  end
+  alias === ==
 
   def =~(other)
     %x{
--- a/spec/filters/bugs/string.rb
+++ b/spec/filters/bugs/string.rb
@@ -1,7 +1,5 @@
 opal_filter "String" do
-  fails "String#=== returns false if obj does not respond to to_str"
-  fails "String#=== returns obj == self if obj responds to to_str"
-  fails "String#=== returns obj == self if obj responds to to_str"
+  fails "String#=== returns false if obj does not respond to to_str" #passes except for the line with symbol: fails "'hello'.send(@method, :hello).should be_false"
 
   fails "String#=~ raises a TypeError if a obj is a string"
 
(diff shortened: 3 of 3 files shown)


vais:
Also, a quick question: is the `toString()` call in `return self.toString() === other.toString();` strictly necessary in the code below?
```javascript
if (other.$$is_string) {
  return self.toString() === other.toString();
}
```
I kinda cargo-culted it in there without fully understanding the reason for it. Something to do with strings being native but sometimes wrapped? Please pardon my ignorance and let me know if `return self === other` would have worked just as well, and, if not, why?

meh:
```
> String.prototype.hue = function() { return this }
[Function]
> "lol".hue() === "lol"
false
> "lol".hue() === "lol".hue()
false
> "lol".hue().toString() === "lol"
true
```

So yes, it is necessary, I'm kind of uncertain it will work with the inheritance trick, but that trick has to be fixed in the runtime sooner or later.

meh:
Also move those failures into unsupported, since we will never have a `String` and `Symbol` difference.

Ajedi32:
> ```
> > String.prototype.hue = function() { return this }
> [Function]
> > "lol".hue() === "lol"
> false
> > "lol".hue() === "lol".hue()
> false
> > "lol".hue().toString() === "lol"
> true
> ```

Woah. That's... really, really weird.

TIL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#Distinction_between_string_primitives_and_String_objects

vais:
@meh the problem is those specs are not granular enough: a single `it` block has multiple assertions, all of which I got to pass except for the one that compares with a symbol. Moving the entire thing into unsupported seems heavy-handed and we lose the safety net that the other passing assertions provide when changing the code in the future. That's why I proposed the three alternatives. Do you think none of the three is possible?

Ajedi32:
> Also move those failures into unsupported, since we will never have a `String` and `Symbol` difference.

Rather than throw out the entire test, isn't there some way to only disable the comparison between string and symbol part? It sounds like the textual descriptions of each test are perfectly valid and there should definitely be tests to ensure that those behaviors remain correct.

vais:
@Ajedi32 here is an example from opal/spec/corelib/core/string/shared/equal_value.rb:
```ruby
  it "returns false if obj does not respond to to_str" do
    'hello'.send(@method, 5).should be_false
    'hello'.send(@method, :hello).should be_false
    'hello'.send(@method, mock('x')).should be_false
  end
```
The first and the third assertion is passing now, but the second is not. What @meh is proposing as far as I understand is to move the entire "returns false if obj does not respond to to_str" into unsupported, since the filtering in mspec works at the level of `it` blocks - there is no way to filter out something inside the `it` block. What can be done is to update the rubyspec itself with `not_compliant_on :opal do...`, for which there are many precedents in the rubyspec currently. The problem is I am not sure what the status of Opal's version of the rubyspec is. It is a fork of a dead upstream project. Should we be able to commit into this fork? If so, that's the way to go, and hopefully when @elia hears back from @headius, we can push these changes upstream, whatever the upstream for rubyspec ends up being.

elia:
@vais :+1: completely agree

headius:
Hears back from me? Is there an email I missed?

vais:
@headius https://github.com/opal/rubyspec/pull/1 raised the question what to do if we need to fix bugs or otherwise evolve rubyspec and @elia cc'd you to get your opinion on starting a shared fork. Here is the original message, sorry for the confusion:

> I wonder how the @jruby guys are handling this, whether they stopped using rubyspec or forked (as we're doing) and if it could make sense to have a shared fork…

> cc @headius @adambeynon @meh

Really interested in hearing your thoughts on the future of rubyspec now that hopefully dust has settled on all that drama and we need to continue getting stuff done with it. Thanks!

meh:
@vais yeah I know, it was just the easiest way to go, I'm fine with whatever.

vais:
@meh I see. Well, it doesn't look like this matter of being able to contribute to rubyspec and/or adding some form of symbol support to Opal will be resolved any time soon. In the meantime, I don't see any reason not to merge this PR as is, with the `fail` lines in place.

And thanks for getting back to me on the `toString()` question, it helps!

meh:
@vais I thought I had merged this, my memory sucks, sorry.

PR #721, opened 2015-03-04
Fix recursive array failures for various Array methods

vais:


elia:
awesome

meh:
:pandaface: :+1:

vais:
Wow, that was quick! Thanks guys.

meh:
@vais you were just lucky, I had just woken up and checking emails :P

vais:
@meh I appreciate the vote of confidence, but I actually would like some feedback. I am still not sure what constitutes idiomatic "ruby javascript", so if I am doing something not idiomatic, I'd like to know. I realize time is short, but if you have a chance to comment on my use of ruby from js and vice versa, please do. I see a lot of different styles and approaches used in corelib.

meh:
@vais I do review everything before merging, if I say nothing it's because it was good or the only nitpick was related to the coding style.

Also yeah, some code in the corelib is older and I hadn't had the time to go janitor mode in a while.

vais:
@meh that's great to hear, thanks! But keep in mind that I can help you with the janitorial tasks as well - I've got a little OCD when it comes to code :) That's why if you have any nitpicks with the coding style, time permitting, do let me know.

PR #722, opened 2015-03-04
fixing source maps for 0.7

elia:

PR #723, opened 2015-03-04
source map server now returns .js.rb and .js.opal source files 

catmando:
Files start as foo.js.rb (or foo.js.opal)

the source map will reference the file as foo.rb

the server needs to recreate the name (either adding .js.rb or .js.opal) before sending it to sprockets to get the file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 *   `String#succ` and `String#next` implementation fully compliant with rubyspec
 
+*   `String#===`, `String#==`, and `String#eql?` implementation fully compliant with rubyspec
+
 ## 0.7.1 2015-02-14
 
 *   CLI options `-d` and `-v` now set respectively `$DEBUG` and `$VERBOSE`
--- a/lib/opal/sprockets/server.rb
+++ b/lib/opal/sprockets/server.rb
@@ -45,7 +45,8 @@ module Opal
 
         return [200, {"Content-Type" => "text/json"}, [source.to_s]]
       when %r{^(.*)\.rb$}
-        source = File.read(sprockets.resolve(path_info))
+        source = File.read(sprockets.resolve(path_info.gsub(/rb$/,"js.rb"))) rescue nil
+        source ||= File.read(sprockets.resolve(path_info.gsub(/rb$/,"js.opal"))) rescue nil
         return not_found(path_info) if source.nil?
         return [200, {"Content-Type" => "text/ruby"}, [source]]
       else
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -75,9 +75,15 @@ class String
   end
 
   def ==(other)
-    return false unless String === other
-
-    `#{to_s} == #{other.to_s}`
+    %x{
+      if (other.$$is_string) {
+        return self.toString() === other.toString();
+      }
+      if (#{Opal.respond_to? `other`, :to_str}) {
+        return #{other == self};
+      }
+      return false;
+    }
   end
 
   alias eql? ==
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -703,10 +703,9 @@ class String
       while ((match = pattern.exec(self)) != null) {
         var match_data = #{MatchData.new `pattern`, `match`};
         if (block === nil) {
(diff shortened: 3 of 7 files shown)


elia:
Thanks, I'll merge the specs manually in #722 and push it when it's ready

catmando:
I was finding other problems when I tried running on my own tests against master, so I went back and used v0.7.1 for my initial work on this problem.  If possible it would be great if the source_map changes could be integrated into a V0.7.1 patch so it could be released.

elia:
0.7.2 will be based on master so it should be fine. 

This unless you switched back to 0.7.1 for something serious…

catmando:
yeah rspec doesn't work (at least from rails)  I just added a problem for that.

elia:
Thanks:)

catmando:
I assume the build failure was not me?  I only ran the rspec tests and they all passed on both v0.7.1 and master

jaredcwhite:
I actually save many of my files as *.opal, as in some_file.opal. Currently this works. I'm concerned this pull-request doesn't take that into account. (Unless *.opal has been deprecated and I'm not aware).

elia:
@catmando the pull is against master, hence the failure
@jaredcwhite we talked about removing .opal support but didn't take any action yet, surely it's more difficult to keep support for two extensions (plus optional .js before it)

catmando:
@jaredcwhite

It does explicitly check for .js.opal... here is the test spec:

+    it 'serves the original source files ending with .js.opal' do
+      source_path = '/assets/source_map/source_files/a_file_ending_in_opal.rb'
+      
+      get source_path
+      
+      expect(last_response).to be_ok
+      expect(last_response.body).to start_with("#a_file_ending_in_opal")
+    end 

Note that there is no test in the specs for files ending in .opal without the .js first.  I believe files like foo.opal will continue to work, as nothing changed to effect that, but I am not sure if it will deliver the original source file to the debugger.

If you are successfully using foo.opal and it is source mapping this today, let me know and I will add it to the test spec, and make sure it works.

If you are doing foo.js.opal, then this definitely works.

catmando:
@elia you mentioned supporting foo.rb (without the .js. in between)... source maps will NOT work today in this case.  I think they could be made to work if that is desired.  

Currently this is what happens:

foo.rb, goes through sprockets and gets compiled to a javascript file with the same name (i.e. foo.rb).  

The source map references the source file also called foo.rb.

So the source server will never see the request, because there is no way to tell foo.rb from foo.rb.

I could see a couple of solutions:

1) during the translation of foo.rb also update the name to foo.js.  Thus we have to file names.  Not sure if this is even possible.

2) during the generation of the source map change the name of the mapped file to foo.rb.source, and update the source server to strip the .source off.  This would actually be an improvement from today's situation.

As always be happy to take this on, and apologies if this is not the best forum for this discussion.

jaredcwhite:
@ella @catmando I guess I don't mind changing from foo.opal to foo.js.opal if that ends up be required -- I just want to make sure I'm not forced to do foo.js.rb at some point because it's less confusing to me to be able to tell at-a-glance whether some code is client-side vs. server-side (server is rb, client is opal)

elia:
@catmando this is the discussion forum I prefer ;)

Feel free to add specs for .js.opal and .opal in a PR targeting my wip branch (or just drop the code anywhere and I'll add it myself)

catmando:
The js.opal and js.rb specs were added with this pull request.  The question is do we expect foo.rb and foo.opal to also be handled properly?

Imho I suggest that .opal be deprecated, and that only foo.js.rb be recognized for proposes of source mapping. 

@jaredcwhite:  files ending in .js.rb are easily identified as client side, due to the js in the name...

ryanstout:
I'm only half following this conversation, but in volt we use .rb and have code that runs in both mri and opal. So .js.rb would break a lot of things for us and cause major issues. :-)

elia:
Goal here is to work with everything so I'm stripping every extension when storing and retrieving sourcemaps and ruby files. The browser will see ruby files as .rb and maps as .map.

catmando:
Okay!  Given this I will add specs and make sure source maps work for files ending in .rb.

elia:
On master source-maps are now served inline with embedded sources, that means that we don't do the additional requests anymore and thus the extension problems are gone for good. WIN WIN 😄

PR #725, opened 2015-03-04
Fix issue with @delayed ignoring falsy values in Promise.

ryanstout:
This fixes an issue where the following prints 5 instead of nil

```ruby
Promise.new.resolve(5).then { nil }.then {|v| puts v.inspect }
```

Thanks!

### DIFF
--- a/spec/opal/stdlib/promise/then_spec.rb
+++ b/spec/opal/stdlib/promise/then_spec.rb
@@ -60,4 +60,12 @@ describe 'Promise#then' do
       p.then {}
     }.should raise_error(ArgumentError)
   end
+
+  it 'should pass a delayed falsy value' do
+    p = Promise.new.resolve(5).then { nil }
+
+    p.then do |value|
+      expect(value).to eq(nil)
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


meh:
Nice catch, I just have a nitpick.

Can you use `@delayed = false` and `@delayed = [value]` and `@delayed[0]`?

ryanstout:
@meh sure, just curios, why do it that way?

meh:
No real reason, I just like it more, it's a pattern I use when I need a `Maybe` from Haskell.

ryanstout:
@meh cool, I'll change it, one sec.

ryanstout:
so thats going to change behavior a bit, maybe I should make @delayed = nil to start, that way the places where you do: resolve(@delayed) still get nil

meh:
Those places should have `resolve(@delayed[0])`, otherwise they're getting a resolve without a value.

ryanstout:
ok, cool, I'll do that.

ryanstout:
ok, updated.

meh:
Ah right, can you also a test for that? If you don't know how it's fine, I'll do it.

ryanstout:
@meh yea, you want me to add a spec for it?  All of the existing specs passed.

ryanstout:
ok, added a spec for it.  Let me know if that looks good.  Thanks

meh:
Looks good, thanks!

PR #726, opened 2015-03-05
Make `rake mspec_node` runnable on Windows

vais:
With a little care, the same Ruby code can run on Unix and on Windows.

### DIFF
--- a/lib/opal/path_reader.rb
+++ b/lib/opal/path_reader.rb
@@ -9,7 +9,7 @@ module Opal
     def read(path)
       full_path = expand(path)
       return nil if full_path.nil?
-      File.read(full_path)
+      File.open(full_path, 'rb:UTF-8'){|f| f.read}
     end
 
     def expand(path)
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -50,14 +50,13 @@ task :mspec_node do
   mkdir_p File.dirname(filename)
   File.write filename, <<-RUBY
     require 'spec_helper'
-    #{requires.join("    \n")}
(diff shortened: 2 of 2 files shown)


elia:
Thanks! are you on windows? (just idle curiosity)

vais:
@elia sometimes :)

vais:
@elia basically right now with this little change all tests are running on Windows, and all except two are passing. The two that are failing are failing due to some kind of string encoding issue, or so it looks like. I will look into that when I get a chance tomorrow.

My plan is to make it possible to do Opal development on Windows as long as that does not require major changes. It is not a priority, just a pet peeve of mine - this meme that Ruby somehow does not work on Windows. I have used Ruby on Windows for years, and it works well.

On an unrelated subject (well, maybe not so unrelated), I believe the claim that Opal runs on IE 6 is no longer true. I have seen things like `Object.keys` used in corelib (I think in the Array or Hash implementation), and that only became available in IE 9 I think. Not sure what if anything needs to be done about that, but wanted to hear your thoughts on what browsers should (ideally vs practically) be supported and how to make sure they are as work on Opal continues.

elia:
Making development possible on Windows is definitely a good thing™.

Moved the last question to issue #727

vais:
Thanks you @elia

vais:
Update: the two specs that were failing for me on Windows were failing due to an older version of Ruby installed on the Windows machine. More info here: https://github.com/opal/opal/pull/738

Using `ruby 2.0.0p643 (2015-02-25) [i386-mingw32]` and up, all specs currently pass on Windows.

PR #731, opened 2015-03-06
Experiment: define object_id for strings, numbers, and booleans

### LINKED ISSUE #729: object_id meaningless for strings, numbers, and booleans
This is best explained with an example ([try opal](http://opalrb.org/try/?code:s%20%3D%20%22hello%22%0Ap%20s.object_id%0Ap%20s.object_id%0Ap%20s.object_id%0A%0Ai%20%3D%20420%0Ap%20i.object_id%0Ap%20i.object_id%0Ap%20i.object_id)):
```ruby
s = "hello"
p s.object_id
p s.object_id
p s.object_id

i = 420
p i.object_id
p i.object_id
p i.object_id

# => 6931
# => 6932
# => 6933
# => 6934
# => 6935
# => 6936
```

vais:
Fix #729. This is just an experiment. For strings, numbers, and
booleans, `object_id` can be simply their string representation with a
static prefix to distinguish a string “123” from a number 123 whose
`object_id` would also be “123”, from an array whose `object_id` happens to be
123 because that’s what `Opal.uid` gave it. Hope this makes sense. No
idea about the performance implications.

### DIFF
--- a/opal/corelib/boolean.rb
+++ b/opal/corelib/boolean.rb
@@ -1,6 +1,11 @@
 class Boolean
   `def.$$is_boolean = true`
 
+  def __id__
+    `"boolean#" + self`
+  end
+  alias object_id __id__
+
   class << self
     undef_method :new
   end
--- a/opal/corelib/numeric.rb
+++ b/opal/corelib/numeric.rb
@@ -5,6 +5,11 @@ class Numeric
 
   `def.$$is_number = true`
 
+  def __id__
+    `"numeric#" + self`
+  end
+  alias object_id __id__
+
   def coerce(other, type = :operation)
     %x{
       if (other.$$is_number) {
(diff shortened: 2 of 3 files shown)


vais:
Hmm... even though this would make `object_id` less misleading, it would still not help with things like `('stuff'.equal? 'stuff').should == false`. Assertions [like this one](https://github.com/opal/rubyspec/blob/master/core/kernel/equal_spec.rb#L14) would ultimately have to be wrapped in `not_supported_on :opal` in rubyspec.

Ajedi32:
FYI, in MRI I think certain objects have well-defined ids. For example, in all instances of Ruby:

```ruby
false.object_id
#=> 0
true.object_id
#=> 2
nil.object_id
#=> 4
0.object_id
#=> 1
1.object_id
#=> 3
2.object_id
#=> 5
-1.object_id
#=> -1
```

vais:
Closed in favor of https://github.com/opal/opal/pull/733

PR #733, opened 2015-03-07
Experiment 2: define object_id for strings, numbers, and booleans

### LINKED ISSUE #729: object_id meaningless for strings, numbers, and booleans

vais:
Fix #729. A different approach:

For strings, object_id can be simply their string representation. For
booleans, numbers, and nil, follow the MRI scheme:

* false: 0
* true: 2
* nil: 4
* numbers: (n * 2) + 1
* heap objects: 6, 8, 10, 12, 14, 16, etc…

### DIFF
--- a/opal/corelib/boolean.rb
+++ b/opal/corelib/boolean.rb
@@ -1,6 +1,11 @@
 class Boolean
   `def.$$is_boolean = true`
 
+  def __id__
+    `self.valueOf() ? 2 : 0`
+  end
+  alias object_id __id__
+
   class << self
     undef_method :new
   end
--- a/opal/corelib/numeric.rb
+++ b/opal/corelib/numeric.rb
@@ -5,6 +5,11 @@ class Numeric
 
   `def.$$is_number = true`
 
+  def __id__
+    `(self * 2) + 1`
+  end
+  alias object_id __id__
+
   def coerce(other, type = :operation)
     %x{
       if (other.$$is_number) {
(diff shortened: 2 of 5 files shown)


Ajedi32:
Looks pretty solid to me. The only real problem I have with this is that `"asdf".object_id == "asdf".object_id #=> true`, which doesn't match the behavior of MRI. As @elia said though, the fact that strings are immutable in Opal could justify that behavior.

So to summarize, after this is merged the only differences between the behavior of `object_id` in MRI and Opal will be:

* `"some_string".object_id == "some_string".object_id` is true in Opal, but false in MRI
* `"some_string".object_id` returns a string in Opal, as opposed to a number in MRI
* `1.1.object_id` returns a floating point number Opal, but an integer in MRI

The `object_id` in Opal for all other objects should effectively match the behavior of MRI.

vais:
@Ajedi32: You got it.

@meh, @elia: Do you guys have any other objections that are blocking this PR from being merged?

PR #735, opened 2015-03-07
Pass "Struct anonymous class instance methods includes Enumerable"

vais:
@meh, I need your help with this one - I don't know what I'm doing :) You [fixed Struct inheritance back in 2013](https://github.com/opal/opal/commit/9677ea31a2630b3690507b9722b5730f7d0ea544), and you have this guard clause in your code: `return if self == Struct`. I removed it and all tests pass, but I'm worried I'm missing something here.

Also, I'm doing `include Enumerable` in the `self.inherited` hook even though `Struct` already has `include Enumerable` at the class level. If I don't do that, `StructClasses::Car.include?(Enumerable).should == true` fails. `StructClasses::Car` is defined as follows:
```ruby
module StructClasses
  Car = Struct.new(:make, :model, :year)
end
```
So, is including `Enumerable` inside `self.inherited` the right way to go, or am I brute-forcing something that is really a bug to be fixed elsewhere?

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -334,7 +334,7 @@ class Module
 
   def include?(mod)
     %x{
-      for (var cls = self; cls; cls = cls.parent) {
+      for (var cls = self; cls; cls = cls.$$super) {
         for (var i = 0; i != cls.$$inc.length; i++) {
           var mod2 = cls.$$inc[i];
           if (mod === mod2) {
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -1,4 +1,8 @@
+require 'corelib/enumerable'
+
 class Struct
+  include Enumerable
+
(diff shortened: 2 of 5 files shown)


meh:
About the `Enumerable` part it sounds like a bug elsewhere.

About that line in inherited I'm not entirely sure what it was doing or why I added it. If all specs we have on `Struct` we have are passing may as well have been some workaround for something else.

vais:
@meh, thanks for getting back to me on this. The bug was in `Module#include?` and I fixed it in a new commit to this PR. This allowed me to remove the extraneous call to `include Enumerable` and all tests pass without it now.

In the process I discovered that we are not running any specs at all for `Module#include` and `Module#include?`, so I added the appropriate specs and marked them as `fail` in the bugs filter in this commit as well. This way I will remember to work on them next.

PR #738, opened 2015-03-10
String#intern and String#to_sym fully compliant with rubyspec

vais:
Special-casing of +(binary) and -(binary) [has been removed from Ruby](https://www.ruby-forum.com/topic/4418067).

These specs originally failed for me on Windows, but the reason turned out to have nothing to do with Windows, but rather with the version of Ruby I had installed on that Windows machine, which was `ruby 2.0.0p247 (2013-06-27) [i386-mingw32]`. After updating to `ruby 2.0.0p643 (2015-02-25) [i386-mingw32]`, the specs passed.

### DIFF
--- a/spec/filters/bugs/string.rb
+++ b/spec/filters/bugs/string.rb
@@ -57,9 +57,6 @@ opal_filter "String" do
 
   fails "String#index raises a TypeError if passed a Symbol"
 
-  # fails "String#intern does not special case certain operators"
-  # fails "String#intern special cases +(binary) and -(binary)"
-
   fails "String#lines accepts a string separator"
   fails "String#lines should split on the default record separator and return enumerator if not block is given"
   fails "String#lines splits using default newline separator when none is specified"
@@ -139,9 +136,6 @@ opal_filter "String" do
   fails "String#sum returns sum of the bytes in self if n less or equal to zero"
 
   fails "String#to_str returns a new instance of String when called on a subclass"
-
-  # fails "String#to_sym does not special case certain operators"
-  # fails "String#to_sym special cases +(binary) and -(binary)"
 
   fails "String#tr raises an ArgumentError a descending range in the replacement as containing just the start character"
   fails "String#tr raises an ArgumentError a descending range in the source as empty"
--- a/spec/filters/bugs/string.rb
+++ b/spec/filters/bugs/string.rb
@@ -57,8 +57,8 @@ opal_filter "String" do
 
   fails "String#index raises a TypeError if passed a Symbol"
 
-  fails "String#intern does not special case certain operators"
-  fails "String#intern special cases +(binary) and -(binary)"
+  # fails "String#intern does not special case certain operators"
+  # fails "String#intern special cases +(binary) and -(binary)"
 
   fails "String#lines accepts a string separator"
   fails "String#lines should split on the default record separator and return enumerator if not block is given"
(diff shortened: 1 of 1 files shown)


elia:
Any reason for commenting instead of removing?

vais:
No reason @elia, just forgot to remove them :)

vais:
@elia I'm kinda confused now. How does the version of Ruby installed on my machine have any effect on how Opal parses symbols? It obviously does have an effect, but then how does Try Opal work completely in the browser? I cannot reconcile these two facts in my brain. What am I missing?

elia:
Not 100% sure but probably this line is involved:

https://github.com/opal/opal/blob/d2e435b4c8f122525937dd5c376cc55677f78116/lib/opal/parser.rb#L167

adambeynon:
The special cases of `+(binary)` and `-(binary)` have caught me out before. The only way around this, without different ruby versions breaking it, is to keep symbols in the parse tree as strings, instead of symbols as we do here.

vais:
@elia, @adambeynon sorry guys, I am still confused:

Older version of Ruby:
```
C:\GitHub\opal>ruby -v -e "p :'+(binary)'"
ruby 2.0.0p247 (2013-06-27) [i386-mingw32]
:+
C:\GitHub\opal>ruby -Ilib bin\opal -v -e "p :'+(binary)'"
Opal v0.7.1
"+"
```
Newer version of Ruby:
```
C:\GitHub\opal>ruby -v -e "p :'+(binary)'"
ruby 2.0.0p643 (2015-02-25) [i386-mingw32]
:"+(binary)"
C:\GitHub\opal>ruby -Ilib bin\opal -v -e "p :'+(binary)'"
Opal v0.7.1
"+(binary)"
```
So, same Opal, different MRI Rubies, different results. This means that *the version of MRI installed on your machine affects how Opal parses/compiles code*. That's totally fine. What I am confused about is how does Try Opal work then? I mean it runs completely in the browser, right? It takes Ruby source, compiles it to JS, then evals the JS. But bin/opal seems to rely on MRI **to do the same job**. Why?

@adambeynon If I understood your comment correctly, I don't think we need to support older Rubies as far as running rubyspec is concerned. 2.0.0p643 and up works fine.

adambeynon:
@vais When you are compiling code on the command line through Opal, then the compiler will use the `String#to_sym` method from MRI, which changes behaviour based on the ruby version - some versions of MRI will return `:+(binary)`, and others will return `:+`.

When running in TryOpal, on the website, the "ruby runtime" is Opal, so `String#to_sym` will be the Opal version of that method, which always returns `self` - `"+"`, so it isn't ever trying to treat `:+` or `:-` as a special case.

vais:
I see, it all makes sense now, thank you @adambeynon!

What do you think about this PR though? I don't see any reason to support running Opal specs using MRI older than 2.0.0, do you?

elia:
Since we're targeting 2+ I think would be good to add a test *against* the special case. Which apparently is there, probably inside some version guard:

		-  fails "String#intern does not special case certain operators"

vais:
@elia I'm confused.

> I think would be good to add a test against the special case

That's exactly what we are doing - there are already specs that test against the special case [here](https://github.com/opal/rubyspec/blob/master/core/string/shared/to_sym.rb#L10) and [here](https://github.com/opal/rubyspec/blob/master/core/string/shared/to_sym.rb#L15), and prior to this PR we had these specs marked as failing. This PR makes it so that if you run the specs on Ruby __older than 2+__, you __will__ get a failure.

This PR also removes the following filters which no longer correspond to any specs presently in rubyspec:
```ruby
fails "String#intern special cases +(binary) and -(binary)"
fails "String#to_sym special cases +(binary) and -(binary)"
```
Perhaps these were in rubyspec at some point in the past (prior to the change in Ruby), but they have been removed and replaced with their *does __not__ special case* counterparts.

elia:
Hey @vais, sorry, I've been mislead by the fact that it was still an issue on windows, but reading again the issue made clear it depends on the patchlevel of 2.0. I guess we can't and shouldn't support each and every patchlevel so I think we're good :)

vais:
@elia no problem, thanks! I am easily confused. For example your profile picture, changed, and I was like who's that?! :)

elia:
lol

PR #745, opened 2015-03-12
Struct#[] and Struct#[]= fully compliant with rubyspec

vais:


meh:
Could you find the message printed by the `TypeError` and add it please?

vais:
Thanks @meh, it's fixed in the third commit. I will put proper MRI messages on errors from now on, that was just my laziness :) That, and a hole in rubyspec.

meh:
No worries :panda_face:

PR #750, opened 2015-03-12
Templates with locals

adambeynon:
**WIP Do not merge **

This pull request supports a new compiler option, `:template`. It is off by default (the most common case), but when turned on it makes any local variable reference check if it exists against a local hash (we assume a string key, as given variable names must be a string or symbol), and uses the value given if present, otherwise resorts to a method call.

The purpose of this switch is to allow ERB and HAML templates to have a hash of local variables passed in. The ERB and HAML builder processors and compilers set the `:template` flag manually.

## Future work

Really, this should be extended out to allow any file to evaluate in the context of a hash of local variables, which would allow us to mimic a binding (very basic mimicking). Also, compiled files should have the ability to take a custom `self` attribute, so a compiled file does not always have to run in the context of `Opal.top`, the main object. This will allow `eval()` to work with a custom object context.

### DIFF
--- a/lib/opal/builder_processors.rb
+++ b/lib/opal/builder_processors.rb
@@ -132,6 +132,14 @@ module Opal
         erb_compiler = erb_compiler_class.new(source, path)
         erb_compiler.prepared_source
       end
+
+      def compiled
+        @compiled ||= begin
+          compiler = compiler_for(@source, file: @filename, template: true)
+          compiler.compile
+          compiler
+        end
+      end
     end
 
     class ERBProcessor < Processor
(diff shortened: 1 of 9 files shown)


jaredcwhite:
Awesome stuff. One comment is perhaps use a keyword argument for the locals, or look for that inside of an optional hash, instead of just a single optional end argument. (aka how Rails handles locals)

For example:

`some_template.render(self, locals: {a_var: 'foo'})`

That way you can add additional arguments down the road and run up against backwards-compatibility issues.

PR #759, opened 2015-03-17
Fix failing RubySpec: Array#clone copies singleton methods

kaiwren:
`Array#clone` must copy singleton methods over. `Array#dup` should not.

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -522,6 +522,15 @@ class Array
 
   def clone
     copy = []
+    %x{
+      for (var key in self) {
+        if (key[0] == "$" && typeof(self[key]) === "function") {
+          if (self.hasOwnProperty(key) && self[key].$$stub === undefined) {
+            copy[key] = self[key];
+          }
+        }
+      }
+    }
     copy.initialize_clone(self)
     copy
   end
@@ -694,8 +703,6 @@ class Array
       return self.slice(number);
     }
   end
-
-  alias dup clone
 
   def each(&block)
     return enum_for :each unless block_given?
--- a/spec/filters/bugs/array.rb
+++ b/spec/filters/bugs/array.rb
@@ -1,5 +1,4 @@
 opal_filter "Array" do
-  fails "Array#clone copies singleton methods"
 
   fails "Array#combination generates from a defensive copy, ignoring mutations"
   fails "Array#combination yields a partition consisting of only singletons"


kaiwren:
Is there anything else I need to look into for this PR?

vais:
@elia @meh does this make sense: https://github.com/opal/opal/pull/759/files#diff-981b14160ae83ab1ab0c0936525d462dR526 I mean, is that how you would do it? I have no idea, sorry @kaiwren, otherwise I'd merge it in already :)

vais:
@kaiwren oh, is this the original source of inspiration for your implementation? https://github.com/opal/opal/blob/master/opal/corelib/kernel.rb#L42

elia:
Seems good to me 
@meh?

kaiwren:
@vais Yup, either that or something very similar IIRC. I am concerned about damaging performance though - `clone` is used a fair amount, so will defer to your collective judgements on this front.

vais:
@kaiwren if you rebase to master, you can try `bundle exec opal benchmark/run.rb` before and after the change to see if there's any significant negative impact. There's also `bundle exec rake mspec_node BM=10` which will run every spec 10 times and dump a json file with the timings per spec into the project's `tmp` directory.

kaiwren:
@vais Ok, noted, ty. Will try that.

kaiwren:
@vais `bundle exec rake mspec_node PATTERN=spec/corelib/core/array/permutation_spec.rb BM=10`
fails for me with
```
ruby -rbundler/setup -rmspec/opal/special_calls bin/opal -gmspec -Ispec -Ilib -smspec/helpers/tmp -smspec/helpers/environment -smspec/guards/block_device -smspec/guards/endian -rnodejs/io -rnodejs/kernel -Dwarning -A tmp/mspec_node.rb -c > tmp/mspec_node.js
jshint --verbose tmp/mspec_node.js
NODE_PATH=stdlib/nodejs/node_modules node tmp/mspec_node.js
........................................................................................................................
/Users/sidu/Work/opal/tmp/mspec_node.js:3615
      throw exception;
            ^
NoMethodError: undefined method `open' for File
    at OpalClass.$new (/Users/sidu/Work/opal/tmp/mspec_node.js:3965:15)
    at OpalClass.<anonymous> (/Users/sidu/Work/opal/tmp/mspec_node.js:3610:31)
    at OpalClass.Opal.defn.TMP_1 (/Users/sidu/Work/opal/tmp/mspec_node.js:2492:19)
    at OpalClass.method_missing_stub [as $open] (/Users/sidu/Work/opal/tmp/mspec_node.js:722:35)
    at $OSpecRunner.def.$did_finish (/Users/sidu/Work/opal/tmp/mspec_node.js:37190:61)
    at /Users/sidu/Work/opal/tmp/mspec_node.js:41067:44
    at Object.<anonymous> (/Users/sidu/Work/opal/tmp/mspec_node.js:41068:3)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
rake aborted!
```

Am I missing something?

vais:
@kaiwren yep, BM is broken right now.

@elia, looks like in order to fix [latest rubyspec spec_helper's use of realpath](https://github.com/opal/rubyspec/blob/master/spec_helper.rb#L32), you made [this change](https://github.com/opal/opal/commit/3b72c807983a4e03201863e61071a035a081ec1e#diff-31d0c7df77415700019f4ac1f555d007L66) to no longer require stdlib/nodejs's version of `File`, and instead aliased `realpath` to `expand_path` in corelib's `File`. This (the not requiring part) broke my use of stdlib/nodejs `File` for writing ruyspec benchmarking results to disk.

The thing is, there's actually nothing technically wrong with nodejs' `realpath`, it's doing what it's supposed to do. The issue is that inside spec_helper, `__FILE__` compiles to `corelib/spec_helper` instead of `spec/corelib/spec_helper`. My guess is that the reason it does that is because the load path is set to `spec`, so `__FILE__` is computed relative to the load path and not the current working directory?

Not sure where to go from here... Any suggestions?

elia:
>  Any suggestions?

Sadly no, it's tricky, can this be solved in the bm runner? (asking because primary focus is on the browser and not node)

vais:
Thanks, @elia. I'll sleep on it, maybe I can come up with something in the next few days.

vais:
@elia could not sleep, fixed our rubyspec-based benchmarking, https://github.com/opal/opal/pull/925 :stuck_out_tongue_winking_eye:

@kaiwren after you rebase with the latest master, adding `BM=10` should work. The results can be viewed by running `bundle exec rake bench:report`. If you run benchmarks multiple times (i.e. make changes, then run again, etc), the report will show you the results for each run side-by-side. For more on the rake bench task, see `bundle exec rake -T | grep bench`.

elia:
@vais that's because you already got some `#sleep`!

vais:
LOL, yes, the Opal version of `#sleep`: `while(true);`

kaiwren:
lol and ty. :)

elia:
Hey @kaiwren thanks for the PR and sorry for letting this become stale, the issue seems to be solved now (via https://github.com/opal/opal/pull/1099).

PR #760, opened 2015-03-17
Fixing failing RubySpec Singleton dump_spec

kaiwren:
Singleton#_dump returns an empty string
Singleton#_dump returns an empty string from a singleton subclass

### DIFF
--- a/spec/filters/bugs/singleton.rb
+++ b/spec/filters/bugs/singleton.rb
@@ -1,6 +1,4 @@
 opal_filter "Singleton" do
-  fails "Singleton#_dump returns an empty string"
-  fails "Singleton#_dump returns an empty string from a singleton subclass"
   fails "Singleton.instance returns an instance of the singleton's clone"
   fails "Singleton.instance returns the same instance for multiple class to instance on clones"
 end
--- a/stdlib/singleton.rb
+++ b/stdlib/singleton.rb
@@ -7,6 +7,10 @@ module Singleton
     raise TypeError, "can't dup instance of singleton #{self.class}"
   end
 
+  def _dump
+    ""
+  end
+
   module SingletonClassMethods
 
     def clone


meh:
Isn't `#_dump` for marshalling?

I don't think we'll ever be able to support that.

kaiwren:
I was wondering why this simple a spec wasn't passing. Makes sense.

I plan to work my way through the RubySpec as time permits - any suggestions on what I should pick up (or skip)?

meh:
Honestly I think they were in the wrong filter, should have gone into unsupported.

vais:
@kaiwren please update this PR as follows:

1. Remove `def _dump` from `singleton.rb`
2. Create `spec/filters/unsupported/marshal.rb`. File names in `spec/filters/unsupported` generally follow a theme/topic, e.g. "tainted", or "mutable_string", and any spec whose failure falls under that topic, regardless of the class the spec is targeting, gets recorded in that topic file.
3. Add the following content to `spec/filters/unsupported/marshal.rb`:
```ruby
opal_filter "Marshal" do
  fails "Singleton#_dump returns an empty string"
  fails "Singleton#_dump returns an empty string from a singleton subclass"
end
```
@meh do you agree?

vais:
@kaiwren 
> I plan to work my way through the RubySpec as time permits - any suggestions on what I should pick up (or skip)?

I will be working my way through `String` over the next few days. Anything else in `filters/bugs` is fair game. Regarding what specs to pick up, my strategy has been to try to fix some method I regularly use, then follow that rabbit hole wherever it goes :)

kaiwren:
@vias Will update this PR. I think I'll go after the specs in `Array` after this.

vais:
Thank you @kaiwren, and looking forward to more PRs for the corelib!

kaiwren:
Raised #836 to fix this

PR #763, opened 2015-03-18
String#squeeze fully compliant with rubyspec

vais:


meh:
Well, respect.

vais:
Thank you, @meh, that was a scary one, but nothing compared to [tr](https://github.com/opal/opal/blob/master/opal/corelib/string.rb#L1123) and [tr_s](https://github.com/opal/opal/blob/master/opal/corelib/string.rb#L1260) - @mojavelinux is either a genius or a madman (or both).

The plan is to move the `explode_sequences_in_character_set` helper out of `String#squeeze` and use it to implement `String#count` and `String#delete` and friends. Stay tuned :)

mojavelinux:
> @mojavelinux is either a genius or a madman (or both).

I'm going to go with both :)

Thanks for bringing more accuracy to Opal! It's fixes like these that give people pause to take notice of Opal as a true Ruby implementation.

vais:
Thank you @mojavelinux, this means a great deal coming from you. I hope to make my small dent in it :)

PR #766, opened 2015-03-19
String#count fully compliant with rubyspec

vais:


meh:
Can you move the helper function at the beginning of the class?

Or move the nethods using it closer together, whatever you like most.

vais:
@meh I will hoist it up to the top of the class. BTW I tried to find precedents of a class-level javascript helper functions, but found just one, not enough to conclude what constitutes "best practice" for such things. One question: should such functions be named using javascript's camelCase or ruby's snake_case?

vais:
@meh, in the spirit of "don't ask for permission, ask for forgiveness", here is how I chose to deal with this:

1. Put shared helper functions at the **bottom** of the class definition. The top of the class is a special place, and it looks weird to pollute it with random js code. Re-arranging method positions to be closer to the shared functions is also a little weird and does not scale in the long term. I think I chose the lesser of the three "evils" by putting large shared helper at the bottom of the class, out of sight.

2. Use Ruby's snake_case naming convention. I want to avoid the mental overhead of switching between naming conventions for local variables and function names. This is part of the reason I am interested in Opal in the first place. So, I'm gonna go ahead and use the Ruby way - after 10 years I'm getting tired of this shit if you know what I mean.

meh:
Sounds good :panda_face:

PR #770, opened 2015-03-22
RFC: String#casecmp fully compliant with rubyspec

vais:
Not sure about this one, need your feedback.

It looks like strings that contain non-ascii characters should not be lower-cased per rubyspec.

1. Does the above statement make sense?
2. Does my code for determining that a string is ascii-only make sense?

Thanks!


meh:
1. I'd say, believe in the specs passing.
2. Yeah, it's fine.

PR #776, opened 2015-03-25
String#rindex fully compliant with rubyspec

vais:


elia:
@vais any reason you're picking one letter var names? `var i, m, r, _m;`

vais:
Because [naming things is hard](http://martinfowler.com/bliki/TwoHardThings.html) :) I'll try to do a better job of it in the future - my names tend to be either way too long or way too short.

elia:
Of course the truth is there are 10 hard things is CS…

IMO you can safely go with "too long", itt easier to discard information than it is to create it afterwards

ylluminate:
:+1: LOL

PR #779, opened 2015-03-28
String#sub fully compliant with rubyspec

vais:


adambeynon:
Nice! Thanks :smile:

vais:
Thank you @adambeynon!
`String#gsub`, you're next :facepunch:

ylluminate:
@vais you are doing an absolutely FANTASTIC job with this stuff.  Can you document your workflow and process for biting these suckers off into a bullet list of steps you take?  I think a lot of us have wanted to really do what you're doing, but haven't really had the time to pass over the learning curve.  Even just putting it into an issue here would be great (doesn't matter where you stick it, just would like to see if more of us could pitch in after getting a nice overview of the process to follow).

vais:
Thanks for the compliment, @ylluminate, please provide feedback on https://github.com/opal/opal/pull/783

PR #780, opened 2015-03-29
Fix incorrect MatchData#pre_match and #post_match behavior

vais:
Despite passing rubyspec, the behavior of these methods does not match MRI:

MRI:
```
$ ruby -e "'hello'.sub(/ll/, ''); p $~.pre_match, $~.post_match"
"he"
"o"
```
Opal before the fix:
```
$ bundle exec opal -e "'hello'.sub(/ll/, ''); p $~.pre_match, $~.post_match"
""
"hello"
```
Opal after the fix:
```
$ bundle exec opal -e "'hello'.sub(/ll/, ''); p $~.pre_match, $~.post_match"
"he"
"o"
```

### DIFF
--- a/opal/corelib/match_data.rb
+++ b/opal/corelib/match_data.rb
@@ -6,8 +6,8 @@ class MatchData
     @regexp     = regexp
     @begin      = `match_groups.index`
     @string     = `match_groups.input`
-    @pre_match  = `#@string.substr(0, regexp.lastIndex - match_groups[0].length)`
-    @post_match = `#@string.substr(regexp.lastIndex)`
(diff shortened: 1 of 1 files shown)


elia:
Good catch, maybe we should open an issue/PR on ruby/rubyspec unless their updated fork already has a fix for it…

vais:
@elia yep. The sooner we start using their fork the sooner we'll know :) Current `MatchData` implementation is very suspect, including my recent `MatchData#offset` contribution - I know it's crap, but it passes the spec, and I only needed it to work enough so we could close `String#sub`. But it's not giving me the warm and fuzzy feeling.

PR #782, opened 2015-03-30
String#gsub fully compliant with rubyspec

vais:


elia:
awesome :beers:

vais:
Thanks, @elia! The mighty gsub did not take it laying down, but I prevailed in the end (I think :)

PR #783, opened 2015-03-30
Updated CONTRIBUTING.md to help people get started

vais:
Re: https://github.com/opal/opal/pull/779#issuecomment-87287434

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,13 @@ $ git clone git://github.com/opal/opal.git
 Get dependencies:
 
 ```
-$ bundle
+$ bundle install
+```
+
+RubySpec related repos must be cloned as a git submodules:
+
+```
+$ git submodule update --init
 ```
 
 Run the test suite:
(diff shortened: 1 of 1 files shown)


vais:
@ylluminate I was hoping for a bit of a discussion before this PR got merged :) But feel free to let me know in the comments here (or file an issue) if there's anything in particular that you'd like to see covered in CONTRIBUTING.md, and I'll do my best to address it. The problem with writing a getting started guide is that by the time you are able to write one, you already forgot what it's like to be the one getting started.

elia:
@vais @ylluminate the discussion is welcome even after the merge, we can always adjust

ylluminate:
@vais thanks for this.  This is an overly hectic week for me... :) I think you've done a great job of documenting what you have here.  The explanation of the tests is, I think, very enlightening and useful to find one's bearings.

I also think that perhaps some documentation may be useful as far as using Opal in projects from the local forked source and how to keep it isolated properly

For example, in `opal-phaser` we've found it useful to do something like this in the Gemfile:
```
gem 'opal', github: 'opal/opal', branch: 'master'
```

And then from `opal-phaser` I would execute:
```
bundle config local.opal ../opal
```

Now this may seem unrelated to `opal` proper, but to me it seems seems very valuable because one can use a local opal fork without having to do any bundle update, etc. incantations to get changes that you want to test.  Obviously it is wise to make the spec and run the tests (which is a must), but sometimes we want to test things in a quick and dirty way given the context of our own projects or such.  That might be useful to include in the documentation...  

Any other common shell mechanics (incantations) might also be useful for inbound would-be contributors.

Finally, there might be a documentation contribution section that could be useful showing how to properly document code both in the source and for the main opal site.

I will have to get my hands a bit more dirty myself in this before I can give more meaningful feedback.  Like I said, I think your current outline is extremely helpful because the tests are probably one of the most important things to understand right now at this point since bug fixing should be the highest priority.  And again, I think bringing things into a would-be user's context of their own project(s) is also valuable for really keying up the interest level for most folks.

vais:
@ylluminate, I have added information about a couple of shell commands I use all the time in the course of working on Opal (`opal-repl` and `opal -e ...`) in this PR to CONTRIBUTING.md: https://github.com/opal/opal/pull/785

Perhaps @elia can comment on your question regarding how one would hack on Opal from within a project that uses Opal - I have not had any experience with this yet.

PR #804, opened 2015-04-17
Fix deletion of rb/js/opal extension in Opal::Builder

RoxasShadow:
Well, I was having a look at the latest commits and I noticed that we `gsub` gaining no effect. So let's `gsub!`.

Also, just saying, at the moment machines running OSX 10.10.3 cannot `bundle` because of [this](http://stackoverflow.com/questions/29529455/missing-c-header-debug-after-updating-osx-command-line-tools-6-3).

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -101,7 +101,7 @@ module Opal
     end
 
     def process_require(filename, options)
-      filename.gsub(/\.(rb|js|opal)$/, '')
+      filename.gsub!(/\.(rb|js|opal)$/, '')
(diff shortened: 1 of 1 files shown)


meh:
Now, was it a bug, or the intended behavior was that the replacement shouldn't happen?

@adambeynon ^

RoxasShadow:
The original one is a line added by @elia at 4af32389 that actually does nothing, but since that commit introduced the deletion of those extensions, that replacement should actually happen imho.

elia:
Good catch! That was the intended behavior

PR #813, opened 2015-04-27
Fix String#scan, make #gsub, #rindex, #rpartition multiline by default

### LINKED ISSUE #457: Infinite loop on scan
```ruby
  def self.heredoc(string)
    indent = string.scan(/^[ \t]*(?=\S)/).min.size rescue 0

    string.gsub(/^[ \t]{#{indent}}/, '')
  end
```

This causes an infinite loop on Opal.

vais:
Fix #457


vais:
Thanks, @elia!

Issue #480, opened 2014-01-06
BasicObject#instance_eval(string, [, filename [, lineno]] ) is not implemented


siuying:
As we now have Kernel#eval, is it possible to add instance_eval(string) as well? (http://www.ruby-doc.org/core-2.1.0/BasicObject.html#method-i-instance_eval)

catmando:
I have a pull request in for this.  The PR will try to compile, and will fail if opal-parser is not loaded.

PR #815, opened 2015-04-27
Kernel#=== fully compliant with rubyspec

vais:


elia:
keep it going! ⚡️

vais:
I'm running out of steam, it's time for dinner now :) BTW, do you know what's the deal with [module_eval only taking a block and no arguments](https://github.com/opal/opal/blob/master/opal/corelib/module.rb#L406)? Is that something Opal-specific and on purpose? (as opposed to the official docs [stating that it takes a string or a block](http://ruby-doc.org/core-2.0.0/Module.html#method-i-module_eval)). I am getting a bunch of related errors trying to get `Kernel#String` to work. Before I jump in from the deep end and try to "fix" `Module#module_eval`, thought I'd ask first.

elia:
It's 2 things 

- Eval-ing with strings needs Opal-parser to parse the string as you probably already know
- then there's the fact that normal jus eval only evaluates in the context of the global object and iirc it's not possible to circumvent this limitation 

As usual I'd be happy to be proven wrong :D

elia:
Here's the only eval implementation that accepts a string

https://github.com/opal/opal/blob/master/stdlib/opal-parser.rb

vais:
Thanks, @elia. The problem I ran into is some of the rubyspecs lean heavily on meta-programming, e.g. Kernel#String specs :weary:

elia:
https://github.com/opal/opal/issues/480

vais:
@elia meta-programming on top of Opal == :sweat_smile::gun: moving on...

PR #819, opened 2015-04-29
Kernel#String fully compliant with rubyspec (minus crazy meta :to_s specs)

vais:


elia:
is it missing the rubyspecs entry?

vais:
Oops... thanks @elia!

PR #820, opened 2015-04-30
Run all rubyspecs for String

vais:


vais:
I am not feeling particularly warm and fuzzy about the way I hacked the `!` feature onto `spec/rubyspecs`, but this is basically what I need right now. This way I can look at the thing and say *"Ok, String is DONE, except for the stuff that's been blacklisted or that's in the bugs/unsupported filters"*. Otherwise, it's never clear where we stand vis-a-vis rubyspec - you have to manually compare the file list in `spec/corelib/core/whatever` **directory** with the file list in `spec/rubyspecs` **text file**.

I feel that the next step is to invert the approach altogether: instead of a whitelist of specs (with this new option to blacklist), I think that we're close enough to start using a blacklist instead. In other words, everything runs by default, except the specs that have been blacklisted. In fact, I think this is how this should have been from the beginning, cause it's much easier to track progress, and more intuitive for someone coming into the project - you goal is clear: remove specs from the blacklist, remove failures from the bugs filter. What we have now is kinda backwards: remove failures, then add more specs to the list (based on what?), then make more failures, etc. It's confusing.

adambeynon:
I think it is fine. The fact that we even use a text file is the result of a quick hacky way to get it working. I keep meaning to change it to a ruby dsl, which would make it easier to keep control of what we bring in from ruby spec. But, for now, Im fine with this approach.

vais:
@adambeynon so are you ok with merging this PR?

Issue #736, opened 2015-03-08
Guarding against performance regressions in Opal


vais:
Rubyspec does an excellent job guarding against regressions when new code is added or existing code is modified. It makes it possible for new people like me to jump in and help out, and as long as the spec suite passes, everyone else on the project can be comfortable in the knowledge that the new guy didn't break anything :-) I am using "the new guy" for dramatic effect, but the same certainly applies to everyone who touches the code in any way.

What Rubyspec unfortunately *does not* do for us is provide any assurance that while the code is correct, it did not inadvertently degrade (or utterly kill) the performance of Opal in some indirect way, like changing a method in one class that another seemingly unrelated class in corelib happens to call a lot.

I would love to have some sort of a safety net in place for this. I am thinking about a separate performance analysis tool that works by hooking into mspec so that each spec is run as a benchmark multiple times against Opal and then the same number of times against MRI. Then maybe play with some average stats for MRI baseline and see where Opal deviates abnormally.

Anyway, before jumping into the deep end, I thought it would be a good idea to ask you guys if there is already something that does this for us that I am simply not aware of, or if you have any ideas about how to put such a safety net in place, so when people contribute code to the corelib you can easily see what impact the new code has on Opal. I think having something like that would help both the maintainers and the contributors to Opal.

ryanstout:
@vais this sounds like an interesting idea. I was thinking about building out a simple benchmarking suite for volt. Maybe integrating capybara so we can test the browsers would be really useful.

elia:
Sounds great!

Maybe it's worth checking CRuby and the other implementations too see if they already have something like that in place and if it's something that can become like the standard test bench for many rubies.

elia:
Btw there's nothing like that that I'm aware of.

elia:
Could be a starting point: 

http://programmingzen.com/2007/12/03/the-great-ruby-shootout/

https://github.com/acangiano/ruby-benchmark-suite

Then we'll ask @acangiano to write a new shootout across al js implementations :P

vais:
@elia thanks for suggesting https://github.com/acangiano/ruby-benchmark-suite, that's a good starting place. It's not quite what I had in mind, but having some kind of benchmarking option is better than having none at all. I will look into what it would take to get it to run against Opal.

vais:
@ryanstout please report any performance bottlenecks you find in Volt back to Opal as issues. Focusing on a real world app performance issues beats working with any artificial benchmark.

[...]

PR #828, opened 2015-05-03
Enable benchmarking of rubyspecs

vais:
Re https://github.com/opal/opal/issues/736

This is the first phase of the project: just getting the raw performance data.

The way it works is you run the spec suite as you normally would, but set the `BM` environment variable to the number of iterations you want MSpec to perform on each `it` block. If this environment variable is present, Opal MSpec runner will record the number of milliseconds taken for each spec (across all iterations) and, at the end of the run, dump a JSON object with spec names for keys and milliseconds for values into a time-stamped file in the `tmp` directory.
```
$ bundle exec rake mspec_node PATTERN=spec/corelib/core/string/index_spec.rb BM=100
...
Finished
2000 examples, 0 failures (time taken: 2.8919999599456787)
```
Notice the `BM=100` above, and how it affects the number of examples: it's `2000 examples` instead of the normal `20 examples` for this spec file. It is actually 20 examples executed 100 times each, so don't be confused by that output.
```
$ cat tmp/bm_2015-05-03_15-53-48-662.json 
{
  "String#index raises a TypeError if passed nil": 45,
  "String#index raises a TypeError if passed a boolean": 41,
  "String#index calls #to_str to convert the first argument": 74,
  "String#index calls #to_int to convert the second argument": 72,
  "String#index raises a TypeError if passed a Fixnum": 29,
  "String#index with String behaves the same as String#index(char) for one-character strings": 340,
  "String#index with String returns the index of the first occurrence of the given substring": 50,
  "String#index with String doesn't set $~": 23,
  "String#index with String ignores string subclasses": 36,
  "String#index with String starts the search at the given offset": 110,
  "String#index with String starts the search at offset + self.length if offset is negative": 239,
  "String#index with String returns nil if the substring isn't found": 33,
  "String#index with Regexp behaves the same as String#index(string) for escaped string regexps": 1244,
  "String#index with Regexp returns the index of the first match of regexp": 51,
  "String#index with Regexp sets $~ to MatchData of match and nil when there's none": 25,
  "String#index with Regexp starts the search at the given offset": 73,
  "String#index with Regexp starts the search at offset + self.length if offset is negative": 238,
  "String#index with Regexp returns nil if the substring isn't found": 36,
  "String#index with Regexp returns nil if the Regexp matches the empty string and the offset is out of range": 23,
  "String#index with Regexp converts start_offset to an integer via to_int": 56
}
```
Phase 2 would be to create a rake task that can analyze data files from two or more runs. "Analyze" is a very open-ended proposition, I would appreciate feedback from the community on this one. What should the rake workflow be? What should the results of the analysis look like?

Phase 3 would be the ability to push (from a rake task as well?) the analysis results to a web app (hosted on Heroku?) for the purpose of tracking how Opal is doing vis-a-vis performance over time. Maybe integrate with GitHub too?

Phase 2 is fuzzy to me. Phase 3 is even fuzzier. Maybe the analysis in Phase 2 should not be done locally at all, and instead upload the files to a web app directly? In which case Phase 2 and 3 are the same thing?

So, any ideas are very much welcome, and code is even more welcome!

### DIFF
--- a/lib/mspec/opal/runner.rb
+++ b/lib/mspec/opal/runner.rb
@@ -227,7 +227,44 @@ class OSpecRunner
     MSpec.actions :start
   end
 
+  def bm!(repeat)
+    `self.bm = {}`
+    MSpec.repeat = repeat;
+    MSpec.register :before, self
+    MSpec.register :after,  self
+  end
+
+  def before(state = nil)
+    %x{
+      if (self.bm && !self.bm.hasOwnProperty(state.description)) {
+        self.bm[state.description] = {started: Date.now()};
+      }
+    }
+  end
+
+  def after(state = nil)
+    %x{
+      if (self.bm) {
+        self.bm[state.description].stopped = Date.now();
+      }
+    }
+  end
+
   def did_finish
+    %x{
+      var obj = self.bm, key, val, json, file;
+      if (obj) {
+        for (key in obj) {
+          if (obj.hasOwnProperty(key)) {
+            val = obj[key];
+            obj[key] = val.stopped - val.started;
+          }
+        }
+        json = JSON.stringify(obj, null, '  ');
+        file = #{Time.now.strftime('tmp/bm_%Y-%m-%d_%H-%M-%S-%L.json')};
+        #{File.open(`file`, 'w') {|f| f.write(`json`)}}
+      }
+    }
     MSpec.actions :finish
   end
 end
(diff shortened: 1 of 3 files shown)


elia:
Just one note, can you make a bit more clear what's going on in testing.rake by adding either a comment or using the name of a variable?

vais:
Sure, @elia, how about this?

elia:
👍🏼

PR #829, opened 2015-05-03
Support tilt integration without sprockets

jeremyevans:
This adds a tilt/opal.rb file that allows you to render opal templates
via Tilt without using sprockets. It changes lib/opal/sprockets/processor.rb
to load tilt/opal.rb and augment it with sprockets specific code.

This allows you to use opal with asset engines that don't use sprockets.

I'm not sure where to add specs for this, if you could point me to the appropriate place, I'll add some specs.  Ideally, these specs would load tilt/opal without loading lib/opal/sprockets/processor, to ensure that tilt/opal works standalone.  I've tested that manually, but I think there should be automated tests for it.

### DIFF
--- a/lib/opal/sprockets/processor.rb
+++ b/lib/opal/sprockets/processor.rb
@@ -8,12 +8,17 @@ require 'opal/sprockets/source_map_server'
 $OPAL_SOURCE_MAPS = {}
 
 module Opal
-  class Processor
+  class Processor < TiltTemplate
     class << self
       attr_accessor :source_map_enabled
     end
 
     self.source_map_enabled          = true
+
+    def self.inherited(subclass)
+      super
+      subclass.source_map_enabled = source_map_enabled
+    end
 
     def evaluate(context, locals, &block)
       return super unless context.is_a? ::Sprockets::Context
@@ -171,6 +176,9 @@ module Opal
     end
   end
 end
+
+Tilt.register 'rb',   Opal::Processor
+Tilt.register 'opal', Opal::Processor
 
 Sprockets.register_engine '.rb',  Opal::Processor
 Sprockets.register_engine '.opal',  Opal::Processor
--- a/lib/opal/sprockets/processor.rb
+++ b/lib/opal/sprockets/processor.rb
@@ -8,38 +8,12 @@ require 'opal/sprockets/source_map_server'
 $OPAL_SOURCE_MAPS = {}
 
 module Opal
-  # The Processor class is used to make ruby files (with rb or opal extensions)
-  # available to any sprockets based server. Processor will then get passed any
-  # ruby source file to build. There are some options you can override globally
-  # which effect how certain ruby features are handled:
-  #
-  #   * method_missing_enabled      [true by default]
-  #   * arity_check_enabled         [false by default]
-  #   * const_missing_enabled       [true by default]
-  #   * dynamic_require_severity    [:error by default]
-  #   * source_map_enabled          [true by default]
-  #   * irb_enabled                 [false by default]
-  #   * inline_operators_enabled    [false by default]
-  #
   class Processor
     class << self
-      attr_accessor :method_missing_enabled
-      attr_accessor :arity_check_enabled
-      attr_accessor :const_missing_enabled
-      attr_accessor :dynamic_require_severity
       attr_accessor :source_map_enabled
-      attr_accessor :irb_enabled
-      attr_accessor :inline_operators_enabled
     end
 
-    self.method_missing_enabled      = true
-    self.arity_check_enabled         = false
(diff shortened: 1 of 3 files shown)


elia:
I think that for now we can switch from the InstanceMethods module to a superclass to Opal::Processor that just does basic Tilt integration. That should be enough to allow automated testing.

That said in the long run Tilt and Sprockets will have quite different code paths as Sprockets 3 processors are no longer tied to Tilt.

As of where to put the specs [spec/lib/tilt/opal_spec.rb](https://github.com/opal/opal/tree/master/spec/lib) seems a good place.

That's good for this PR, but if this feature is actually used there's a couple more things that need to be fixed, namely:

1. the tilt processor should respect [default options][0] set on Opal::Processor
2. should probably use Opal::Builder instead of Opal::Compiler in order to include dependencies coming from required files

(btw congratz for being a hero! :mortar_board:)

[0]: https://github.com/jeremyevans/opal/blob/tilt/lib/opal/sprockets/processor.rb#L11-L41

jeremyevans:
Thanks for the quick response!

The reason I chose a module instead of a superclass is you'd have to use the superclass in the Tilt.register calls, which I think would make it so the sprockets-specific subclass wouldn't get called via Tilt.

As for the default options, I think we can bring all of them to the tilt/opal file, except the source_map_enabled, since I think that that is sprockets-specific.  I'll work on that and some specs shortly.

The issue with Opal::Builder is that it requires additional dependencies that Opal::Compiler does not require (27 additional files, including the hike and source_map gems).  Are all of those dependencies really needed if you just want to compile some ruby code to javascript?  Opal::Compiler seems to generate code without it.  From the README, it appears that Opal::Builder is for building the corelib/runtime, which in most applications I assume you can just use the minified javascript.  Maybe there is something I'm missing here?

My goal for this is to be able to use opal in Roda's assets plugin (which doesn't use sprockets), via something like:

    plugin :assets, :js=>%w'opal.min.js application.rb'

jeremyevans:
OK, I see what you mean with the require files, you mean the files that are required by the ruby script you are compiling.  Unfortunately, that code requires sprockets.  You'd have to reimplement `process_requires` and `process_required_trees` to not require sprockets.  I think it is possible to do some of that by adding settings to Opal::Processor, but as a new contributor I'm not sure I'd be the best person to design the API for that.

elia:
_last quick response of the day (2am here)_

> The reason I chose a module instead of a superclass is you'd have to use the superclass in the Tilt.register calls, which I think would make it so the sprockets-specific subclass wouldn't get called via Tilt.

I'm not sure if I am reading this correctly but I think registering the superclass to Tilt is actually fine and probably better. 

Choosing the Compiler over the Builder is totally fine but leaves you with dealing with dependencies by hand (Opal parses requires which both the Builer and the Processor manage automatically). 

As of options I think it's better to keep them all together on Opal::Processor. The Builder relies upon them too iirc but I understand that that defeats the whole purpose of this PR. The ultimate solution is to move them to their own place or directly on ::Opal while keeping a proxy on Opal::Processor. To me seems better to delay such changes and for now just use default options on tilt. 

Hopes this makes sense :)

jeremyevans:
OK, I've pushed another commit that makes tilt/opal add a superclass for Opal::Processor named Opal::TiltTemplate.  To get the Opal::Processor specs to pass, I had to register Opal::Processor in tilt if it is loaded.

I setup inheritance of settings, so that if you make a change to a Opal::TiltTemplate, and then subclass, the subclass gets a copy of the Opal::TiltTemplate settings.  This allows for per-subclass settings.

elia:
> I had to register Opal::Processor in tilt if it is loaded.

Wouldn't this make the tilt spec exercise the sprockets processor instead of the tilt-only one?

elia:
Merged, I already started work on extracting config and make everything more straightforward, will keep you posted

jeremyevans:
You are correct that the tilt spec will execute the sockets processor, but as it doesn't pass in a Sprockets::Context to render, it will call super to get the behavior in Opal::TiltTemplate.  If Opal::Processor does not register in Tilt, the Opal::Processor specs break.

For better confidence that Opal::TiltTemplate works independently, we could register a separate extension for Opal::TiltTemplate, or use a separate Tilt mapping (that might be a Tilt 2.0 feature).  The other way to do it would be to execute the tilt spec separately without loading Opal::Processor, but that would probably require some special handling by the spec runner.

Thanks for the quick feedback and review!

elia:
Fun fact is that hearing at the rubyheroes award that someone kept the issue count at zero was very inspiring in trying to do the same here 😅

Anyway i fixed it locally, just have a couple of server_spec examples that are still failing.

PR #833, opened 2015-05-05
String#initialize with an argument is not supported

vais:
I believe this is the right call, please review. My understanding is that all "String#initialize with an argument..." specs assume mutable strings.


elia:
I wonder if those related to to_s and to_str can be made pass in .new instead of #initialize. 

Also I think the trust/taint stuff should go in the taint unsupported file. Does it make sense?

vais:
> I wonder if those related to to_s and to_str can be made pass in .new instead of #initialize.

There's only one using `to_str`, and that for the incoming argument. The point is the next step is mutating the receiving string anyway, so it's a non-starter.

> Also I think the trust/taint stuff should go in the taint unsupported file. Does it make sense?

It makes more sense to me to keep them together. I mean if some were passing and some not, I would put the ones that are not passing into the appropriate category. But in this case, the entire class of "String#initialize with an argument" is not supported, so I felt that splitting up these related failures across multiple files would do more harm than good.

The only way I see these passing is if there's a way to substitute the string object that the variable is pointing to with a different string object. Is that possible using some kind of runtime magic trick?

PR #837, opened 2015-05-07
String#inspect fully compliant with rubyspec

vais:


vais:
Thanks, @meh!

meh:
No, thank you :panda_face:

PR #846, opened 2015-05-12
Fix #838

### LINKED ISSUE #838: define_singleton_method with Proc/lambda argument instead of block does not work
Using git version of Opal (currently v0.8.0.beta1). Also tested on Opal 0.7.2.
define_singleton_method is implemented, but only the one-argument symbol-and-block form, not two-argument symbol-and-method form. ( http://ruby-doc.org/core-1.9.3/Object.html#method-i-define_singleton_method )
Attempting to use define_singleton_method in the two-argument form produces the JavaScript console error: "ArgumentError: tried to create Proc object without a block".

```ruby
class DecoObject
  def this_works
    self.define_singleton_method(:cool!) do
      puts "Behold!"
    end
  end
  def this_does_not(lam)
    self.define_singleton_method(:and_that_is_bad, lam)
  end
  def neither_does_this(lam)
    self.send :define_singleton_method, :also_bad, lam
  end
end

deco = DecoObject.new()
deco.this_works
deco.cool!
l = lambda {puts "Cool! It works!"}
puts "Here we go..."
deco.this_does_not(l)
deco.and_that_is_bad
deco.neither_does_this(l)
deco.also_bad
```

vais:
@elia @meh guys, I'm having ["a senior moment"](http://www.thefreedictionary.com/senior+moment) - is there a better/more idiomatic way to write a method so it accepts **either** a proc argument **or** an anonymous block? :confused: 

fixes #838

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -116,7 +116,9 @@ module Kernel
     initialize_copy(other)
   end
 
-  def define_singleton_method(name, &body)
+  def define_singleton_method(name, body = nil, &block)
+    body ||= block
+
     unless body
       raise ArgumentError, "tried to create Proc object without a block"
     end


elia:
errr

vais:
@elia is that the sound of you having a "senior moment" too? LOL

adambeynon:
Can't test atm, but what about:

```ruby
def define_singleton_method(name, body = block, &block)
  # Proc === body
end
```

vais:
@adambeynon nice! This works in MRI, so thanks for clearing up the "senior moment" for me - I knew there had to be a cleaner way!

Unfortunately, at this time this approach does not work in Opal: `def define_singleton_method(name, body = block, &block)` gets you a `NoMethodError: undefined method 'block' for #<DecoObject:0x90>`

In the meantime, I think this PR is ok to merge as is to fix the issue.

adambeynon:
@vais I tried it on opalrb.org/try and it didn't work there, so it is something we need to fix anyway. I can give it a go later on today when I'm back in the office.

vais:
@adambeynon that would be awesome :+1: I see @elia already has an issue open for you :wink:

PR #848, opened 2015-05-12
Rubyspec: Array#combination

kaiwren:


vais:
Excellent, @kaiwren! But one last small problem: the variable `length` you use in your `for` loops is never defined, creating a global variable leak.

Side note: We really should find a way to integrate JSLint/JSHint into the build/test process for Opal at some point.
cc @elia @adambeynon @meh

elia:
@vais I can see it added to allow failure in travis with something like this:

```bash
rake dist
jslint build/*
```

I don't think it will be green for a while…

elia:
probably this is a good option: https://github.com/stereobooster/jshintrb
would be cool (maybe) to have it optionally enabled at compile time…

elia:
https://github.com/opal/opal/tree/jshint
https://travis-ci.org/opal/opal/builds/62364447

vais:
@kaiwren I'd like to merge this PR, could you do another push with `var length` declared?

kaiwren:
Apologies, fixed.

vais:
:beers:

PR #851, opened 2015-05-13
Make opal/builder depend on opal/paths

jeremyevans:
It doesn't seem to work if you require opal/builder without
opal/paths.


elia:
Thanks good catch!

PR #852, opened 2015-05-13
Replace ^ and $ with \A and \z in regular expressions

jeremyevans:
Unless you specifically want to match the start and end of any line
in the string, as opposed to the start and end of the string itself,
you should avoid using ^ and $.

I'm not sure if all of these changes are correct.  Maybe in some cases
where I made changes, you do want to match start/end of any line
in the string instead of the start/end of the string itself.

### DIFF
--- a/lib/mspec/opal/rake_task.rb
+++ b/lib/mspec/opal/rake_task.rb
@@ -163,7 +163,7 @@ module MSpec
 
       def paths_from_glob pattern
         Dir.glob(File.expand_path(pattern)).map do |s|
-          s.sub(/^#{basedir}\//, '').sub(/\.rb$/, '')
+          s.sub(/\A#{basedir}\//, '').sub(/\.rb\z/, '')
         end
       end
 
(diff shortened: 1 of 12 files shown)


elia:
Some of the updated files are also compiled to js (e.g. files in lib/opal/nodes) and js regexps doesn't support \A and \z. 

That's probably the reason for which the build is failing.

jeremyevans:
That makes sense.  I'll try to figure out which regexps are getting compiled to js, and which are staying in ruby, and push another commit.

jeremyevans:
Looks like the rspec tests are now fixed, but the mspec tasks are still failing. If I had to guess, it's because the compiler is being run under opal itself.

I think trying to hard code \A and \z instead of ^ and $ is not going to work.  We want to use \A and \z in ruby and ^ and $ in javascript.  I think I need to introduce constants that depend on RUBY_ENGINE in order to get the correct behavior when running both in ruby and javascript.  I'll send another patch that does that.

elia:
Great job!

Issue #850, opened 2015-05-13
Define_method fails when a Method/UnboundMethod object is supplied


wied03:
This problem seems too simple so I must be missing something, but when trying to use define_method (http://ruby-doc.org/core-2.2.2/Module.html#method-i-define_method), if you supply a Method/UnboundMethod instead of a block, which is valid behavior, you get a **wrong argument type (expected Proc/Method)** error.

The error message (module.rb line 290) notes that a method might be expected so it seems right but the if statement does not account for that.

This prevents rspec-mocks (via the opal-rspec project) from working in 0.8.

Observed in 0.8.0 beta 1 and I believe the problem is still in Master.

Replicated the UnboundMethod case using http://opalrb.org/try:
```
class A
  def fred
    puts "In Fred"
  end
  def create_method(name, &block)
    self.class.send(:define_method, name, &block)
  end
  define_method(:wilma) { puts "Charge it!" }
end
class B < A
  define_method(:barney, instance_method(:fred))
end
a = B.new
a.barney
a.wilma
a.create_method(:betty) { p self }
a.betty
```
Output:
```
TypeError: wrong argument type UnboundMethod (expected Proc/Method)
TypeError: wrong argument type UnboundMethod (expected Proc/Method)
    at t.$new (http://opalrb.org/javascripts/try.js:5:30463)
    at t.<anonymous> (http://opalrb.org/javascripts/try.js:5:26369)
    at t.c.$define_method.a [as $define_method] (http://opalrb.org/javascripts/try.js:5:5536)
    at eval (eval at <anonymous> (http://opalrb.org/javascripts/try.js:20:27290), <anonymous>:36:17)
    at eval (eval at <anonymous> (http://opalrb.org/javascripts/try.js:20:27290), <anonymous>:37:5)
    at $TryOpal.eval (eval at <anonymous> (http://opalrb.org/javascripts/try.js:20:27290), <anonymous>:45:3)
    at $TryOpal.def.$eval_code (http://opalrb.org/javascripts/try.js:20:27285)
    at $TryOpal.def.$run_code (http://opalrb.org/javascripts/try.js:20:27179)
    at e.$$p.$ (http://opalrb.org/javascripts/try.js:20:26490)
    at HTMLAnchorElement.o (http://opalrb.org/javascripts/try.js:20:2768)
```
I'll work on a fix/pull request.

wied03:
Created pull request https://github.com/opal/opal/pull/853

PR #853, opened 2015-05-13
define_method fixes for issue #850 (Method,UnboundMethod)

wied03:
Pass more Rubyspecs related to unbound/bound methods.


vais:
Nice work, @wied03! I just [fixed a similar issue](https://github.com/opal/opal/commit/4f4ee3b9d4dc731bc73b45776b7c99f17b54a368) the other day, and now I think my solution may have a problem :sweat_smile:

The use of `:default` is a little too hacky for my taste though. I totally see where you're coming from, having looked at the spec, but I'm starting to think the "right" way to solve this would be by using `*args` in the method signature, then looking at `args.length`. I believe that's what Ruby does, which is why it's possible to have this seeming contradiction in the spec. Resolving it with what ends up being a string works, but it's a smell that we've been moving in the wrong direction implementing this type of methods.

wied03:
Yeah, that makes sense. I'll work up something for that.

vais:
Kewl! One final nitpick, @wied03, Ruby raises `ArgumentError: wrong number of arguments (0 for 1..2)`. Could you do the same as a guard at the top of the method? Then you can be more explicit about `args.length` expectations throughout the rest of the method (e.g. `args.length > 1` becomes `args.length == 2`).
```
[1] pry(main)> module X
[1] pry(main)*   define_method
[1] pry(main)* end
ArgumentError: wrong number of arguments (0 for 1..2)
from (pry):2:in `define_method'
```
```
[2] pry(main)> module X
[2] pry(main)*   define_method(:a, :b, :c)
[2] pry(main)* end
ArgumentError: wrong number of arguments (3 for 1..2)
from (pry):3:in `define_method'
```
Generally you want to check your error messages against MRI. This happens to be rubyspec's weak spot.

elia:
Thanks @wied03! 

merged adding a couple of stylistic fixes, you can check them out in https://github.com/opal/opal/commit/f955e0db5ff54c9bf126c6e550c43a469362f171.
Feel free to ask if you want more info :wink:

PR #854, opened 2015-05-13
Update README to mention gitter per this interaction on IRC

fkchang:
```
<fkchang> jeremyevans: it seems that https://gitter.im/opal/opal get's a lot
          more chat love than irc these days, I'd bring your tilt discussion
          there                                                         [15:43]
<jeremyevans> fkchang: Thanks                                           [15:48]
<jeremyevans> Maybe the README could use an update                      [15:49]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,7 @@
 Opal is a ruby to javascript source-to-source compiler. It also has an
 implementation of the ruby corelib.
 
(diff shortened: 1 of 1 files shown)


elia:
Opalrb.org should be updated too, if you get the time of course :)

fkchang:
opalrb.org already has gitter on it

elia:
then we can just sit back and relax 😎

PR #855, opened 2015-05-14
Allow PathReader to deal with absolute and ./ and ../ relative paths

jeremyevans:
This allows you do do things like:

  Opal::Builder.build('/path/to/file.rb')
  Opal::Builder.build('./path/to/file.rb')
  Opal::Builder.build('../path/to/file.rb')

Previously, you couldn't build a file without adding a directory
containing the file to the paths to examine.  This makes Opal::Builder
operate more like ruby itself.

Note that this does not let you use absolute and ./ and ../ relative
paths in require, that requires additional changes, probably to
Builder#process_require.

### DIFF
--- a/lib/opal/path_reader.rb
+++ b/lib/opal/path_reader.rb
@@ -13,7 +13,11 @@ module Opal
     end
 
     def expand(path)
-      file_finder.find(path)
+      if Pathname.new(path).absolute? || path =~ %r{\A.?.#{File::SEPARATOR}}
+        path
+      else
+        file_finder.find(path)
+      end
     end
 
     def paths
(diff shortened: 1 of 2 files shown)


lib/opal/path_reader.rb
```diff
@@ -13,7 +13,11 @@ def read(path)
     end
 
     def expand(path)
-      file_finder.find(path)
+      if Pathname.new(path).absolute? || path =~ %r{\A.?.#{File::SEPARATOR}}
```
elia: @jeremyevans shouldn't this be `%r{\A\.?\.#{File::SEPARATOR}}`?

jeremyevans: Definitely, let me push up another commit

elia:
> This makes Opal::Builder operate more like ruby itself.

:+1:

PR #859, opened 2015-05-14
Don't swallow backtrace when reraising error

jeremyevans:
If you specify a message when raising an existing exception, ruby
does not use the exception's existing backtrace.


elia:
Thanks!

PR #860, opened 2015-05-15
Jshint branch rebased against master

vais:
@elia does it make sense to keep this branch up-to-date with master by rebasing periodically? I mean getting to green on jshint might take a while... What workflow would you propose if I were to take this on?


elia:
my idea, after the initial spike was to put it under allowed failures, wdyt? 

As of rebases are almost worthless via PR, feel free to rebase locally and then [force push](http://www.swtor.com/ability/1Ah4X4B/force-push) because by clicking the merge button github always adds a merge commit.

vais:
@elia I had a little time to think about this. I now believe the way to do this is to delete the jshint branch and instead add a very liberal .jshintrc to master along with the travis config changes you made for the jshint branch. Then I can make PRs over time fixing the various jshint errors, incrementally making the .jshintrc more and more strict. This way jshint is part of our development from day one and moving forward. I will prepare the first PR over the weekend. Again, the idea is to make it commitable to master from the start, and screw the screws tighter over time with more commits. Does this make sense to you?

PR #867, opened 2015-05-17
Handle null and undefined in if statements

jeremyevans:
Previously using null or undefined in an if statement would result
in something like: TypeError: Cannot read property '$$is_boolean'
of undefined.

### DIFF
--- a/lib/opal/nodes/helpers.rb
+++ b/lib/opal/nodes/helpers.rb
@@ -131,7 +131,7 @@ module Opal
             expr(sexp)
           end
         elsif [:lvar, :self].include? sexp.type
-          [expr(sexp.dup), fragment(" !== false && "), expr(sexp.dup), fragment(" !== nil &&"), expr(sexp.dup), fragment(" != null")]
+          [expr(sexp.dup), fragment(" !== false && "), expr(sexp.dup), fragment(" !== nil && "), expr(sexp.dup), fragment(" != null")]
         end
       end
     end
--- a/lib/opal/nodes/helpers.rb
+++ b/lib/opal/nodes/helpers.rb
@@ -95,7 +95,7 @@ module Opal
         end
 
         with_temp do |tmp|
-          [fragment("((#{tmp} = "), expr(sexp), fragment(") !== nil && (!#{tmp}.$$is_boolean || #{tmp} == true))")]
+          [fragment("((#{tmp} = "), expr(sexp), fragment(") !== nil && #{tmp} != null && (!#{tmp}.$$is_boolean || #{tmp} == true))")]
         end
       end
 
@@ -109,7 +109,7 @@ module Opal
         end
 
         with_temp do |tmp|
-          [fragment("((#{tmp} = "), expr(sexp), fragment(") === nil || (#{tmp}.$$is_boolean && #{tmp} == false))")]
+          [fragment("((#{tmp} = "), expr(sexp), fragment(") === nil || #{tmp} == null || (#{tmp}.$$is_boolean && #{tmp} == false))")]
         end
       end
 
@@ -131,7 +131,7 @@ module Opal
             expr(sexp)
           end
         elsif [:lvar, :self].include? sexp.type
-          [expr(sexp.dup), fragment(" !== false && "), expr(sexp.dup), fragment(" !== nil")]
+          [expr(sexp.dup), fragment(" !== false && "), expr(sexp.dup), fragment(" !== nil &&"), expr(sexp.dup), fragment(" != null")]
         end
       end
     end
--- a/lib/opal/nodes/logic.rb
+++ b/lib/opal/nodes/logic.rb
@@ -86,7 +86,7 @@ module Opal
       def compile
         with_temp do |tmp|
           push expr(value)
-          wrap "(#{tmp} = ", ", (#{tmp} === nil || #{tmp} === false))"
+          wrap "(#{tmp} = ", ", (#{tmp} === nil || #{tmp} === false || #{tmp} == null))"
         end
       end
     end
@@ -138,7 +138,7 @@ module Opal
         with_temp do |tmp|
           push "(((#{tmp} = "
           push expr(lhs)
-          push ") !== false && #{tmp} !== nil) ? #{tmp} : "
+          push ") !== false && #{tmp} !== nil && #{tmp} != null) ? #{tmp} : "
           push expr(rhs)
           push ")"
         end
@@ -146,7 +146,7 @@ module Opal
 
       def compile_if
         with_temp do |tmp|
-          push "if (#{tmp} = ", expr(lhs), ", #{tmp} !== false && #{tmp} !== nil) {"
+          push "if (#{tmp} = ", expr(lhs), ", #{tmp} !== false && #{tmp} !== nil && #{tmp} != null) {"
           indent do
             line tmp
           end
@@ -176,7 +176,7 @@ module Opal
           else
             push "(#{tmp} = "
             push expr(lhs)
-            push ", #{tmp} !== false && #{tmp} !== nil ?"
+            push ", #{tmp} !== false && #{tmp} !== nil && #{tmp} != null ?"
             push expr(rhs)
             push " : #{tmp})"
           end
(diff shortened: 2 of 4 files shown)


vais:
:+1:

Also, should `#{tmp} == true` be `#{tmp} === true` while we're at it?

adambeynon:
@vais No, javascript booleans are a pain to deal with. Some outside cases make it that `true === true` isn't always true. 

```javascript
Boolean.prototype.returnMe = function() {
  return this;
}

true.returnMe() == true // true
true.returnMe() === true // false
```

adambeynon:
This case happens a lot with things like `#tap()`, a lot of specs using `#should`, etc.

elia:
The original idea of not dealing with null and undefined in Ruby-land is to avoid propagating dangerous values that are not objects and will break things when used down the road. 

As a side note ` Native(null) ` should do the right thing returning nil. 

Anyway not taking sides as I see the value of considering null and undefined as falsy. I'm very interested in what @meh and @adambeynon have to say on the matter.

meh:
We had already tried this, and it didn't work out well, `Native(null)` and `Native(undefined)` should definitely return `nil` tho.

vais:
@adambeynon interesting, thanks for pointing it out, I did not think of that at all. Looks like `valueOf()` fixes it though:
```js
true.returnMe().valueOf() === true // true
```
I would prefer making this explicit instead of relying on javascript's bat-sh\*t crazy-a\*\* `==` operator. I've been [prepping](https://github.com/opal/opal/pull/865) for a full-scale war on bad JS with JSHint. We are locked and loaded, so this discussion is quite timely!

So, what do you think, should this be `#{tmp}.valueOf() === true` then?

jeremyevans:
Admittedly, I didn't know the backstory here.  I just ran into an issue converting some simple javascript to opal and this seemed like the easiest way to fix it.

If your ruby code shouldn't be dealing with plain javascript objects/primitives ever, then shouldn't opal wrap such objects/primitives before exposing them to ruby code?  I realize that may require a significant performance hit, though, and would probably require moving Native from stdlib to runtime.

I guess I was expecting opal to handle javascript objects similar to how jruby handles java objects.  Maybe that's an unrealistic expectation.

Currently, the way the Native handling works in opal, it's allocating new Native object for every method call on a native object, which can't be great for performance.  One of the future directions I was considering was adding syntax to call javascript methods directly from ruby code, so that you could use javascript objects in ruby land easily without wrapping them with Native.  Example:

```javascript
# In javascript code
obj.method() # call javascript method
obj.$method() # call ruby method

# In ruby code
obj.method() # call ruby method
obj.$method() # call javascript method
```

This would allow you to deal with javascript objects in ruby code without the Native overhead.  It does make the code no longer valid standard ruby code, though.  However, if the plan is move reject use to javascript objects in ruby land, it doesn't make sense to do that.

I do think the current behavior is suboptimal.  Either opal should handle the wrapping/unwrapping of javascript objects implicitly for you, or it should offer the ability to easily work with javascript objects in your ruby code.  Manually wrapping/unwrapping javascript objects in your ruby code really cuts down on the productivity benefits you get from using ruby instead of javascript.

meh:
We tried that too, every JavaScript object was seamlessly accessed, including `null` and `undefined`, it didn't work out well.

Also if I remember correctly `Native` is more performant than you might think, same goes for `method_missing`.

jeremyevans:
I've rebased this commit against master and added support for other conditionals and the and/or/not logic operators.  Especially for the logic operators I think this should be fixed, since those currently treat `null` and `undefined` as truthy and not falsy, they don't even raise an error when you try to use them.  I think for consistency it's best to handle `null` and `undefined` in conditionals as well, even if they aren't first class citizens in ruby land.

The benefit to doing this is that the principle of least surprise is followed, and the cost is adding a couple extra checks for `null` and `undefined` when a conditional or logic operator is used.  I think the benefit outweighs the cost, but I realize that is subjective.

vais:
> I think for consistency it's best to handle null and undefined in conditionals as well, even if they aren't first class citizens in ruby land.

For what it's worth, I agree, :+1: 

> ...and the cost is adding a couple extra checks for null and undefined

The cost of these extra checks could be somewhat mitigated by switching from using an *affirmation* of `!==` and `&&` to using a *negation* of `===` and `||`. I.e:
```
$ cat test.js 
var val, tmp, nil = 0,
    a = [false, nil, null, undefined, 3],
    i, len = a.length;

var t, j, n = 10000000;

t = Date.now();
for (j = 0; j < n; j++) {
    for (i = 0; i < len; i++) {
        tmp = a[i];
        val = (tmp !== false && tmp !== nil && tmp !== null && tmp !== undefined);
    }
}
console.log('affirm `&&` and `!==`  =>', Date.now() - t);

t = Date.now();
for (j = 0; j < n; j++) {
    for (i = 0; i < len; i++) {
        tmp = a[i];
        val = !(tmp === false || tmp === nil || tmp === null || tmp === undefined);
    }
}
console.log('negate `||` and `===`  =>', Date.now() - t);
```
```
$ node test.js 
affirm `&&` and `!==`  => 484
negate `||` and `===`  => 331
```
Super micro optimization, but hey :smile:

elia:
@adambeynon bump

elia:
@meh I think this should be merged (with a couple of fixes), anything against?

meh:
@elia I'd like to know if it gives any performance hits.

elia:
Priority should still go to false nil & objects and I'm assuming no perf hits. 

But I'll try to setup a jsperf.com experiment.

elia:
@meh http://jsperf.com/opal-if

meh:
I don't know, I still think it would be bad, you'd just be moving the error further down when you try to call a method on it, or pass it along, and it just makes your life much harder.

jeremyevans:
Monday will make a year that this issue has been open.  It seems that @meh is still against this due to the fact that accessing the value the if statement operates on may cause problems later.  However, the primary place I see this being useful is in code like:

```ruby
if foo.JS.bar
  # ...
end
```

or

```ruby
if foo.JS[:bar]
  # ...
end
```

or

```ruby
if `foo.bar`
  # ...
end
```

In all three of these cases, the objection raised by @meh does not apply, since the value the if statement operates on is not used again.  If this pull request is not accepted, I think we break the principle of least surprise in all three of these cases.  With that in mind, can this be reconsidered?

elia:
🎂 to the issue (almost) 😄 

@jeremyevans I'd be glad to merge, just a couple of thoughts on which I'd like to have some feedback before you rebase it:

1. Need to make sure there's no perf drawback and we still optimize for most common cases and give priority to ruby stuff
2. Was considering a compiler check on the node type for `if` conditions and enable this behavior it its an x-string, or a JS-call

jeremyevans:
For 1), it looks like we try null and undefined after other cases.  This is going to be a performance hit in the cases where we reach that code, but I doubt it has a significant effect.

For 2), the issue with that is that:

```ruby
  if a.JS.b
    # ...
  end
```

is not the same as

```ruby
  x = a.JS.b
  if x
    # ...
  end
```

That inconsistency is problematic, so I wouldn't recommend that approach.

elia:
@jeremyevans

1. js-perf is still down, so after the rebase I ask you if you can check there isn't any significant difference in running `rake mspec_ruby_nodejs` with and without the change just to be sure we're not screwing up common usage

2. ok sounds reasonable

In addition can you change `#{tmp} !== null && #{tmp} !== undefined` to `#{tmp} != null` and `#{tmp} === null || #{tmp} === undefined` to `#{tmp} == null` which are equivalent but shorter?

jeremyevans:
OK, I rebased and made the change you recommended, which helped performance quite a bit:
```
Original:      Finished in 73.195000 seconds
Previous Code: Finished in 74.866000 seconds
New Code:      Finished in 73.569000 seconds

All runs: 0 files, 11745 examples, 49057 expectations, 0 failures, 0 errors, 3695 tagged
```

With your suggestion, this new patch causes only a 0.5% slowdown when running `rake mspec_ruby_nodejs`.  I think and hope that's acceptable.

elia:
@jeremyevans yes, 0.5 should be fine

The build failed due to a bunch specs that should be easy to fix, they're brittle but atm still valueable. Let me know if you need any assistance for that.

_(BTW I have the hunch that if-conditionals perf can be further improved by optimizing the sequence of checks to favor the more common cases, e.g. Boolean values & nil over objects)_

jeremyevans:
I think I was able to fix the compiler specs.  Hopefully Travis can confirm that.  If we can further improve performance by reordering code, even better.

elia:
> If we can further improve performance by reordering code, even better.

Yeah, will wait for js-perf to come back and leave it for another PR/issue, was mostly just talking to myself 😁

PR #868, opened 2015-05-17
String#split, String#each_line, and String#lines fully compliant with rubyspec

vais:


vais:
That's strange: on both Chrome and Firefox, `"hi!".split(/()/)` => `[ "h", "", "i", "", "!" ]` So how is it that phantomjs returns `["", "h", "", "i", "", "!"]`? Any ideas?

meh:
No clue, it works here.

```
> phantomjs
phantomjs> "hi!".split(/()/)
{
   "0": "h",
   "1": "",
   "2": "i",
   "3": "",
   "4": "!"
}
```

```
> "hi!".split(/()/)
[ 'h', '', 'i', '', '!' ]
```

Maybe an old phantomjs version?

elia:
I think specs / Travis are running on 1.9 maybe it's that

vais:
@meh @elia thanks, so it looks like we're using an old phantomjs version on Travis. Is there a reason for that?

Basically I see two solutions: add `fail "String#split with Regexp includes all captures in the result array` or use latest phantomjs on Travis.

@elia if there's a reason we keep old phantomjs around, no problem, just let me know and I will add a `fail` back to bugs/string.

elia:
Yes, the api changed in 2.0 and our driver needs to be updated before going with the newer version. As a side note the same is true for Opal-rspec

vais:
@elia pardon my ignorance, is that a big project or something you can take care quickly? Cause, again, if that's a big deal, I'll just add the fail back into the bugs file for now to get this PR in. LMK...

elia:
@vais add the fail mark (maybe with a comment), I tried to update it "quickly" without success once and my next priority is on fixing opal-rails travis build (passes locally) so I don't think I'll try again for a while.

vais:
Sure, @elia, no problem. Thanks for the quick reply!

elia:
_quick reply kindly provided by [Trailer.app](https://github.com/ptsochantaris/trailer)_ :wink:

vais:
And that's why you're awsome, @elia, because it's a sacrifice to make yourself always available. Please know that it is appreciated.

elia:
errr, please stop or I'll end up believing it! :wink:

vais:
@elia you better believe it. Anyway, thanks! :smile:

Issue #816, opened 2015-04-28
Issue defining class method with the name 'name'


robfors:
The following code was run at opalrb.org/try  
```ruby
class User
  @name = 'rob'
  
  def self.name
    @name
  end

end

user = User.new
puts user.class #prints 'rob' but I was expecting 'User'
```
Using Opal: 0.7.0
It appears to be an issue with defining a class method with the name 'name'.

vais:
Good catch @robfors!

This can be [simplified further](http://opalrb.org/try/?code:class%20User%0A%20%20def%20self.name%0A%20%20%20%20%22hello%22%0A%20%20end%0Aend%0Aputs%20User.new.class). I.e. the class instance variable `@name` does not seem to have anything to do with the problem:
```ruby
class User
  def self.name
    "hello"
  end
end
puts User.new.class
```
outputs:
```
hello
```

vais:
The closest I came in trying to figure this out is https://github.com/opal/opal/blob/master/opal/corelib/module.rb#L538
```ruby
  def to_s
    name || "#<#{`self.$$is_mod ? 'Module' : 'Class'`}:0x#{__id__.to_s(16)}>"
  end
```
@adambeynon could you please weigh in on this?

Ajedi32:
Well, you could store the class name in some kind of hidden variable instead of depending on what the `name` method returns. That seems to be what MRI does, given that it can still get the correct name of the class even when the class's `name` method has been overridden.

elia:
@Ajedi32 there's the `.$$name` property defined on modules and classes that probably is the right thing to use in `Module#to_s`

vais:
@elia @Ajedi32 that works. @Ajedi32 wanna do a PR?

vais:
https://github.com/opal/opal/blob/master/opal/corelib/module.rb#L538
```ruby
  def to_s
    `self.$$name` || "#<#{`self.$$is_mod ? 'Module' : 'Class'`}:0x#{__id__.to_s(16)}>"
  end
```

Ajedi32:
I haven't made any significant changes to Opal so I'm not really set up for it, but this is a really small change so I wouldn't mind giving it a shot.

vais:
@Ajedi32 every journey starts with a single step :wink: If you have questions about getting started, take a look at https://github.com/opal/opal/blob/master/CONTRIBUTING.md and/or feel free to ask me any questions here or on https://gitter.im/opal/opal

Ajedi32:
[...]

PR #869, opened 2015-05-17
Fix issue #816

vais:
Defining a class-level self.name should not have any effect on class.to_s

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -542,7 +542,7 @@ class Module
   end
 
   def to_s
-    name || "#<#{`self.$$is_mod ? 'Module' : 'Class'`}:0x#{__id__.to_s(16)}>"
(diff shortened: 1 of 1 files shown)


vais:
@elia any doubts about this one?

elia:
I guess I just delayed and then forgot :)

but now that you're asking I wonder if there's any rubyspec that can be added to the bundle…

vais:
@elia unfortunately, this: https://github.com/opal/rubyspec/blob/master/core/module/to_s_spec.rb#L5

elia:
lol

elia:
I think merging MRI fork of rubyspec is top priority (along with all the other top priority stuff like: docs, releasing, improving the stdlib, etc.)

vais:
@elia indeed. I will give it another shot. A chance to practice my newly acquired mad git skilzz :stuck_out_tongue_winking_eye:

elia:
show off your git-💪🏼!

PR #872, opened 2015-05-18
Make sprockets a development dependency instead of runtime, allow tilt 2

jeremyevans:
Now that you can build complex opal projects using tilt instead of
sprockets, it doesn't make sense to require sprockets.

As sprockets 3+ doesn't require tilt, it doesn't really make sense
to keep tilt as a runtime dependency, since you should be able to
use sprockets without using tilt. However, Opal::Processor is
currently a subclass of Opal::TiltTemplate, so it doesn't matter
if you are using opal with tilt or sprockets, in either case tilt
is still required for using opal.

If Opal::Processor can be made to run without tilt, I think we
should make both tilt and sprockets development dependencies
instead of runtime dependencies.

The tilt/opal tests run fine with tilt 2, so I don't think we
should restrict people to using tilt 1.4, as then users of opal
can't use the new useful features in tilt 2, such as multiple
mappings, and they will miss out on useful bug fixes in tilt 2
that will not get backported to tilt 1.4, such as the sorting of
locals when caching templates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ matrix:
     - rvm: 2.2
       env: RUN=default TZ="/usr/share/zoneinfo/Pacific/Fiji"
 
+    - rvm: 2.2
+      env: RUN=rspec TILT_VERSION=2.0.1
+
     - rvm: 2.1
       env: RUN=rspec
 
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,9 @@ group :repl do
   gem 'therubyrhino', :platform => :jruby
 end
 
+tilt_version = ENV['TILT_VERSION']
+gem 'tilt', tilt_version if tilt_version
+
 unless ENV['CI']
   gem 'rb-fsevent'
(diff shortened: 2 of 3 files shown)


elia:
The plan is definitely to get both sprockets and tilt more an option but for now I'd keep sprockets in the runtime deps, many people (I think) use the default rack server, which is sprockets based, and would be a surprise to need to put sprockets in the Gemfile. 

The other reason is that before making one of them a dev dependency I'd like to reimplement the sprockets processor in the v3 format wich is just a `#call`-able.

Does that make sense?

jeremyevans:
That makes sense.  I'll push another commit that keeps sprockets as a runtime dependency, and just includes the tilt change.

elia:
If we keep support for both tilt v1 and v2 I think it's better to add an entry in the travis matrix for v2 (by default bundler uses the lowest version of a gems' runtime dependency, an example here (was used to test sprockets 3): c616d24a07 (Add different sprockets versions to the build)

PR #873, opened 2015-05-18
Fix issue #871

### LINKED ISSUE #871: Array#zip doesn't work if an enumerator is passed as an argument. (version 7.2)
Given the following code:

```ruby
e = [:one].each
[1].zip(e).to_a
```

Ruby produces the following: `[[1, :one]]`

Compiled Opal produces: `[[1, nil]]`

~~It looks like #zip tries to coerce the argument to an Array using to_ary, which Enumerator doesn't respond to, and so produces invalid output.~~

It looks like #zip just assumes that the others arguments will always be indexable arrays.

The work around is to explicitly coerce the enumerator to an array using to_a: 

```ruby
e = [:one].each
[1].zip(e.to_a).to_a
```

vais:
fixes #871

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -1780,12 +1780,22 @@ class Array
 
   def zip(*others, &block)
     %x{
-      var result = [], size = self.length, part, o;
+      var result = [], size = self.length, part, o, i, j, jj;
 
-      for (var i = 0; i < size; i++) {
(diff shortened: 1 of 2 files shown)


elia:
:+1: 

@vais be aware that issue title doesn't count for inter-issue mentions/autoclosing, you have to cite the issue number in the body or in a commit

vais:
Thanks, @elia, I was wondering about that. Now I know :)

PR #874, opened 2015-05-18
Support recv.$js_method syntax for calling javascript methods

jeremyevans:
This borrows the syntax of calling ruby methods from javascript
and uses it for calling javascript methods from ruby.

In javascript:

obj.method() # call javascript method
obj.$method() # call ruby method

In ruby:

obj.method() # call ruby method
obj.$method() # call javascript method

This will not break any working ruby/opal code, as the syntax is
not currently valid ruby/opal syntax.  Usage of this syntax will
mean that standard ruby will no longer be able to parse the code,
so any file that uses this syntax will be limited to executing
only with opal, but considering the syntax is used for calling
javascript methods, I think that should be acceptable.

This makes it much easier to use javascript objects inside ruby
code without wrapping them using Native(), and provides a work
around for edge cases where Native(obj) returns obj, such as
when using strings.  This syntax will handle many cases where
you currently need to drop down to raw javascript using backticks.


meh:
I'm strongly against this, it will break any tooling, making the point of using Ruby in the browser moot.

jeremyevans:
As I see it, there are two reasons to use opal:

1. To share ruby code between the client and the server
2. To use ruby instead of javascript on the client

This addresses a significant pain point for 2., and doesn't negatively affect 1., since if you want to share code between the client and server, you would just avoid using this syntax.

Could you mention briefly what tooling it will break?  In my current usage of opal, I'm just compiling the opal code to javascript and executing the javascript on the client, and it doesn't break that.

meh:
Any tooling that parses the language, code coverage, syntax highlighters, style checkers, anything.

jeremyevans:
Closing this, superseded by #879

Issue #877, opened 2015-05-19
Compiler generates accidental global variable if a local variable is also among the args to a local proc


vais:
E.g. `arg` never gets declared in compiled JS (opal/corelib/enumerable.rb):
```ruby
  def sort_by(&block)
    return enum_for :sort_by unless block_given?

    map {
      arg = Opal.destructure(`arguments`)

      [block.call(arg), arg]
    }.sort { |a, b| a[0] <=> b[0] }.map { |arg| `arg[1]` }
  end
```
E.g. `i` never gets declared in compiled JS (opal/corelib/array.rb):
```ruby
  def values_at(*args)
    out = [];

    args.each do |elem|
      if elem.kind_of? Range
        finish = Opal.coerce_to elem.last, Integer, :to_int
        start = Opal.coerce_to elem.first, Integer, :to_int
        
        %x{
          if (start < 0) {
            start = start + self.length;
            #{next};
          }
        }

        %x{
          if (finish < 0) {
            finish = finish + self.length;
          }
          if (#{elem.exclude_end?}) {
            finish--;
          }
          if (finish < start) {
            #{next};
          }
        }

        start.upto(finish) { |i| out << at(i) }
      else
        i = Opal.coerce_to elem, Integer, :to_int
        out << at(i)
      end
    end

    out
  end
```

Issue #876, opened 2015-05-19
Compiler generates accidental global variable compiling `yield` inside `class << ...`


vais:
@adambeynon I'm at the end of my rope with this one too, plz take a look when you have a chance...
```
$ bundle exec opal -cO opal/corelib/dir.rb
```
Note that `TMP_1` variable below is the only one that's never declared:
```javascript
/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $klass = Opal.klass;

  Opal.add_stubs(['$[]']);
  return (function($base, $super) {
    function $Dir(){};
    var self = $Dir = $klass($base, $super, 'Dir', $Dir);

    var def = self.$$proto, $scope = self.$$scope;

    return (function(self) {
      var $scope = self.$$scope, def = self.$$proto;

      self.$$proto.$chdir = TMP_1 = function(dir) {
        var $a, self = this, $iter = TMP_1.$$p, $yield = $iter || nil, prev_cwd = nil;

        TMP_1.$$p = null;
        try {
        prev_cwd = Opal.current_dir;
        Opal.current_dir = dir;
        return $a = Opal.yieldX($yield, []), $a === $breaker ? $a : $a;
        } finally {
        Opal.current_dir = prev_cwd;
        };
      };
      self.$$proto.$pwd = function() {
        var self = this;

        return Opal.current_dir || '.';
      };
      self.$$proto.$getwd = self.$$proto.$pwd;
      return (self.$$proto.$home = function() {
        var $a, self = this;

        return ((($a = $scope.get('ENV')['$[]']("HOME")) !== false && $a !== nil) ? $a : ".");
      }, nil) && 'home';
    })(self.$singleton_class())
  })(self, null)
})(Opal);

/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$exit']);
  return $scope.get('Kernel').$exit()
})(Opal);
```

Issue #875, opened 2015-05-19
Compiler generates accidental global variable compiling `super` inside `class << ...`


vais:
@adambeynon I'm at the end of my rope with this one, need your help to figure this out...
```
$ bundle exec opal -cO stdlib/singleton.rb
```
Please note `TMP_4` below (near the bottom) is the only variable that's never declared.
```javascript
/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $module = Opal.module;

  Opal.add_stubs(['$raise', '$class', '$__init__', '$instance_eval', '$new', '$extend']);
  return (function($base) {
    var self = $module($base, 'Singleton');

    var def = self.$$proto, $scope = self.$$scope;

    Opal.defn(self, '$clone', function() {
      var self = this;

      return self.$raise($scope.get('TypeError'), "can't clone instance of singleton " + (self.$class()));
    });

    Opal.defn(self, '$dup', function() {
      var self = this;

      return self.$raise($scope.get('TypeError'), "can't dup instance of singleton " + (self.$class()));
    });

    (function($base) {
      var self = $module($base, 'SingletonClassMethods');

      var def = self.$$proto, $scope = self.$$scope, TMP_1, TMP_2;

      Opal.defn(self, '$clone', TMP_1 = function() {var $zuper = $slice.call(arguments, 0);
        var self = this, $iter = TMP_1.$$p, $yield = $iter || nil;

        TMP_1.$$p = null;
        return $scope.get('Singleton').$__init__(Opal.find_super_dispatcher(self, 'clone', TMP_1, $iter).apply(self, $zuper));
      });

      Opal.defn(self, '$inherited', TMP_2 = function(sub_klass) {var $zuper = $slice.call(arguments, 0);
        var self = this, $iter = TMP_2.$$p, $yield = $iter || nil;

        TMP_2.$$p = null;
        Opal.find_super_dispatcher(self, 'inherited', TMP_2, $iter).apply(self, $zuper);
        return $scope.get('Singleton').$__init__(sub_klass);
      });
    })(self);

    (function(self) {
      var $scope = self.$$scope, def = self.$$proto;

      self.$$proto.$__init__ = function(klass) {
        var $a, $b, TMP_3, self = this;

[...]

PR #878, opened 2015-05-19
Fix all accidental global variables except the compiler-generated ones

vais:
Fix all accidental global variables except for those that are generated because of compiler bugs #875, #876, and #877.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -12,8 +12,14 @@
   "unused"        : true, // Warn when you define and never use your variables
 
   "globals": {
-    "Opal": true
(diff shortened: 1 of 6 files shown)


vais:
My secret hope is this will convince everyone of the value of having JSHint always on. These are trivial, easy-to-make human errors. JSHint will catch them for us, but we need to get the 3 above-mentioned compiler issues fixed before we can enable this check for good (I have it disabled in .jshintrc for now).

meh:
It convinced me for sure.

PR #879, opened 2015-05-19
Support recv.JS.js_method syntax for calling javascript methods

jeremyevans:
This adds a way to call native javascript methods from ruby using
valid ruby syntax, so it will work with existing tooling, in contrast
to #874.

This will only break working ruby code that defines a method named
"JS" that will accept no arguments, and uses that method in a method
chain.  I can't think of a way to reduce that breakage while still
making the resulting syntax valid ruby syntax. If someone does
actually have a method named "JS" that accepts no arguments and
that they want to use in a method chain, they could work around this
by adding a space (e.g. ".JS ."), as this handles ".JS." specially
in the lexer.

This makes it much easier to use javascript objects inside ruby
code without wrapping them using Native(), and provides a work
around for edge cases where Native(obj) returns obj, such as
when using strings.  This syntax will handle many cases where
you currently need to drop down to raw javascript using backticks.


elia:
looks good! way prettier than using a `$` prefix

spec/opal/core/runtime_spec.rb
```diff
@@ -5,3 +5,11 @@
     `Opal.hash({a: 1, b: 2})`.should == {a: 1, b: 2}
   end
 end
+
+describe 'javascript calls using recv.$meth' do
```
vais: @jeremyevans looks like the description need to be updated :wink:

Ajedi32:
If this gets merged, it would probably be a good idea to add some documentation so that this feature is discoverable by new users.

jeremyevans:
@Ajedi32 I definitely plan on updating the "Javascript from Ruby" section of http://opalrb.org/docs/compiled_ruby/ if this gets merged.

ylluminate:
BIG :+1: on this end.  Excellent call @jeremyevans.

meh:
This feels much better, now let's just hope nobody uses methods called `JS`.

elia:
> now let's just hope nobody uses methods called `JS`.

compiler option? on by default?

meh:
Compiler option, disabled by default would make it a definite :+1: from me.

But we really need to add header based compiler options sooner or later.

ylluminate:
On by default is preferred.  I think as a pre-1.0 release this would definitely be an allowed option to break some (minor) things such as this in favor of a much more Ruby way of doing it.  I don't really think many (if any) folks are using `JS` presently in the current context.

jeremyevans:
If we want to allow changing of the js prefix based on a per-compiler basis, we'll need to change the parser/lexer/compiler API slightly.  That's not difficult, and can probably be done in a backwards compatible manner.

I don't think this will be a big issue in practice.  Capitalized public methods are uncommon in ruby in general, especially capitalized methods that take no arguments, and all caps methods are basically never used.  The only common places for capitalized methods in ruby are conversion methods (e.g. `Kernel#Array`), which are generally private and take arguments, and class factories (e.g. Camping's `R`, Sequel's `Sequel::Model`), which generally take arguments and are either not called with an explicit receiver or are class methods called using `::` instead of `.`.  In 10+ years using ruby, I don't think I've ever seen a capitalized ruby method used in the middle of a method chain.

Considering the existing incompatibilities between opal and standard ruby, this incompatibility should be very small in comparison, especially since it can be easily worked around.

Disabled by default is definitely better than rejected, so I can certainly push additional commits that make this optional and off by default, if that is required to get feature accepted.

jaredcwhite:
Great work on this. I think this will be a huge deal. I haven't minded much doing backticks and native wrapping and whatnot, but then again I've been rather ruthless in trying to avoid getting into JS-world from Opal-world much. For people that want to jump back and forth pretty seamlessly, this kind of improvement will be very welcome.

meh:
We should carefully consider how the parameters and return values are handled, if they're converted to Native or what happens. The obvious `Hash` example being passed as is instead of being converted to its native representation will be the most common.

But another thing to consider is that passing `Proc`s to a method called with `.JS.` will result in the arguments of that `Proc` being native objects and not Opal objects.

I think this can be a good thing but we have to be careful about it, backticks have the positive side of clearly stating "I'm going down to JS, if I fuck up it's my fault", `Native` has the positive side of stating "I'm going down to JS carefully, if it breaks it's @meh's fault".

jeremyevans:
I think of:

```ruby
foo.JS.bar(baz)
```

as pure syntax sugar for:

```ruby
`#{foo}.bar(#{baz})`
``` 

I think of using `.JS.` the same way I think of backticks.  In the documentation I plan to write after this is merged, backticks will be mentioned first, and the `.JS.` syntax will be mentioned after as just a nicer alternative for the same thing. Just as backticks don't handle native wrapping and unwrapping, I don't think `.JS.` should either.  This is just a nicer syntax to write code that has to deal with javascript directly.  If you want conversions to/from javascript/ruby objects, you should use `Native`, not this.

jeremyevans:
@meh I pushed a commit that makes the `.JS.` syntax optional, configurable, and off by default.  Can you review and let me know if you are OK with it going in now?

ylluminate:
<sigh> I hope @meh changes his stance on the default option for this to be on by default.  Further, how will we enable this on CDN based usage?

lib/opal/parser/lexer.rb
```diff
@@ -73,6 +76,26 @@ def initialize(source, file)
...
+    #
+    # @param prefix [String|true|false|nil] prefix to use
+    def js_prefix=(prefix)
+      @js_prefix = case prefix
+                   when true
+                     DEFAULT_JS_PREFIX
+                   when false, nil
+                     nil
+                   else
+                     /\.#{Regexp.escape(prefix)}/
```
meh: Shouldn't this have a `\.` at the end like `DEFAULT_JS_PREFIX`?

jeremyevans: This is by design, to allow the user to choose a prefix that doesn't end in a `.`.  Would you like to force a `.` at the end?

meh: I think it would be better, @opal/owners?

elia: I'm ok with being less strict

adambeynon:
So, I have a couple of concerns/thoughts about this. Firstly, I like the idea. We had this right at the beginning of Opal, and it worked well, and was only removed when we went through a brief period of dynamic dispatching, and we could call native methods directly.

Anyway, some thoughts:

1. If this gets merged, I don't think it can be configurable or optional, it must be on at all times and use a set prefix. Otherwise, it can only safely be used by end user apps, and not shared libraries as an end user configuration could easier disable it, or change the prefix thus breaking the library functionality. I don't like that as a possibility. If you write some opal code, it should always work. It shouldn't break because of a configuration option. I like the `.JS` name, as specified.

2. I don't think the method calls should perform any conversion of ruby objects => javascript objects, or cast return values into ruby objects. This seems a general consensus.

3. This dispatching only works for function calls with a receiver. Is there any thoughts/needs for dispatching to global functions, e.g. jQuery? (A quick thought might be `JS.jQuery('.foo')`). Alternatively, the solution in part 4 might be more ideal.

4. The dispatching only works for function calls. Do we want to support property access on javascript objects. E.g., lets assume we have a `document` reference already as `doc`. Do we wish to support both getting `doc.JS[:title]`, and setting `doc.JS[:title] = "foo"`. Maybe, maybe not.

Of all those, the first part, not being configurable/optional is what I feel more strongly about. I don't want to write library code which is easily broken by accidental config options.

meh:
@adambeynon that's why we'd need an actually good gem/file compiler configuration thingy.

I don't feel that good about having it enabled by default, and the top-level function calling and property accessing do seem like important things or it'd just be an half-implemented feature in my eyes.

As things stand it's not really something I'd want.

ylluminate:
I personally think @adambeynon is spot on regarding point 1.  That is, too, why I'd like to see it all on and non-configurable.  I just don't see breakage being an issue here at this point in time.

jeremyevans:
Regarding 1) I agree that this should be on at all times and use a set prefix.  So first I'll need to revert the last commit.

Regarding 3), I deliberately chose to only handle calls on receivers.  For JQuery and other global functions in browsers, I think you can currently use window:

```ruby
`window`.JS.jQuery()
```

One of my future plans is to introduce a `JS` constant into the stdlib for handling javascript global functions (e.g. `parseFloat`) and javascript syntax that Opal doesn't natively support (e.g. `typeof`).  We could always make this `JS` constant define `method_missing` to call global javascript functions, so it would operate the same way as this syntax.  What are your thoughts on this?

Regarding 4), after this was accepted I planned to add support for both property access (e.g. `foo.JS.bar` => `$foo.bar`, `foo.JS.bar()` => `$foo.bar()` ) and array access (using `.JS[]` as you mentioned), not just function calls.  Having it work for property and array setters would be nice as well.  Just let me know which direction you'd like me to go and I'll implement it.

ylluminate:
Again, in our opinion on this end, your calls @jeremyevans are solid and desirable. :+1:

jeremyevans:
OK, I've pushed a revert of the commit to make this syntax optional, and 2 commits adding support for javascript properties (`foo.JS.prop`) and javascript arefs (`foo.JS[]`).  Javascript function calls without arguments now require parentheses (`foo.JS.func()`).    I plan on adding support for javascript property setters and aref setters next.

jeremyevans:
Now I've added support for javascript property and aref setters.  This does cause 8 reduce/reduce conflicts in the grammar, which I'm not yet sure how to eliminate.

adambeynon:
Nice! I will take a look at the reduce conflicts, its always delicate. One other thing I think we should include is support for command style calls, i.e. method calls without parens.

```ruby
JS.jQuery '.foo', '.bar'
# => jQuery('.foo', '.bar')
```

That would need to be added into the grammar here: https://github.com/opal/opal/blob/master/lib/opal/parser/grammar.y#L216-L238 .

Also, you have supported all the args being passed, but we should probably support blocks as well. A passed block, imo, should be appended to the list of normal arguments, and treated like the last argument passed to a function call. Again, this would work nicely with a  jQuery example:

```ruby
JS.jQuery do
  puts "document is ready!"
end

# => jQuery(function() { /* ... */ });
```

These require a few changes to the grammar, so it is something I can take care of later today, If you prefer.

jeremyevans:
@adambeynon If you could take care of the extra work in the grammar, that would be awesome.  This is my first time working with racc (or language parsers in general), so it would probably take me significantly longer.

jeremyevans:
@adambeynon I had some time today to work on this, so I've pushed a commit that makes the necessary grammar changes to support javascript method calls without parentheses and with blocks.  It eliminates the reduce/reduce conflicts as well.

cj:
@jeremyevans glad to see you're getting into Opal :smile:.  So would this pull request get rid of the need for me having this bit of code?  https://github.com/wedgeio/wedge/blob/master/lib/wedge/component.rb#L270-L294

jeremyevans:
@cj I don't think this would provide an exact replacement for `wedge_function`.  For one, `wedge_function` converts all javascript objects to ruby objects using `Native`, where this is explicitly designed to not do any conversions.  The purpose of this is to provide syntax for calling javascript methods and getting/setting javascript properties without having to use backticks everywhere.

ylluminate:
So this seems to be resolved as per @adambeynon 's note 12 days ago with his last commit. What's stopping this from being merged?

elia:
Ok finally merging on master, which is targeted at 0.9/1.0

Pouring this change into 0.8 would have meant to further delay the release. 

All this unless @adambeynon says otherwise

elia:
Sorry, closed by error from mobile

ylluminate:
Oh, I was thinking that the issues were resolved that would let it go on into 0.8.

vais:
:clap:

Issue #883, opened 2015-05-22
Support Non Top Level Scope for Bridged Classes Built From Functions


wied03:
If you build an Opal class from a function based superklass, this code is hit:
```
else if (typeof(superklass) === 'function') {
      // passed native constructor as superklass, so bridge it as ruby class
      return bridge_class(id, superklass);
    }
```

Would it be a good idea to allow passing base into bridge_class so that bridge class can use the (possibly not top level/Opal scope) when it calls create_scope?

PR #884, opened 2015-05-22
Support Non Top Level Scope for Bridged Classes Built From Functions

wied03:
This could have bigger implications then what I realize but rather than just complain (see issue https://github.com/opal/opal/issues/883), I thought I'd submit a PR.


wied03:
This issue was noticed when trying to use opal-jquery and react.rb in the same project and namespace collisions from the 'Element' class quickly surfaced.

elia:
Thanks!

I'll add just a couple of comments on styling, let me know if you can update the PR, otherwise I'll merge and then apply the fixes myself.

opal/corelib/runtime.js
```diff
@@ -560,14 +560,22 @@
...
    */
-  function bridge_class(name, constructor) {
+  function bridge_class(name, constructor, base) {
     var klass = boot_class_object(ObjectClass, constructor);
 
     klass.$$name = name;
 
-    create_scope(Opal, klass, name);
+	if (base === undefined) {
+		base = Opal;
```
elia: Indentation should be 2 spaces here

spec/opal/core/runtime/bridged_classes_spec.rb
```diff
@@ -62,3 +61,51 @@ def some_bridged_method
...
+  it "instances of class should be able to call native ruby methods" do
+    @instance.foo.should == "foobar"
+    @bridged.new.foo.should == "foobar"
+  end
+
+  it "allows new methods to be defined on the bridged prototype" do
+    @instance.some_bridged_method.should == [4, 5, 6]
+    @bridged.new.some_bridged_method.should == [4, 5, 6]
+  end
+end
```
elia: missing EOF newline

elia:
that's what I meant thanks!

Issue #862, opened 2015-05-15
Add Support of Bignum


jannishuebl:
I have a usecase where I need support of Bignums. I looked around a bit and found out that you had an Issue #746  with Bignums, too.

It would be great if opal would support Bignums like other Ruby-implementations do.

I realy would like to implement this feature for you. But I have some questions:

1. Should Bignum be in the corelib like in mri ruby or in stdlib?
2. Use a JS implementation and create a Wrapper around it or write a pure ruby-implementation that could be compiled by opal? (first would be much less work)

thanks 
Jannis

elia:
As a wrapper could live in stdlib or even in an external gem I think

jannishuebl:
When opal should support Bignum the way ruby does and the way it is needed for #746. There must be some changes in the lexer and parser so I think this could not be done in an external gem. 

I think it would be great when opal requires Bignums when the parser reads some. Then the Bignum implementation should be included and be load by the clinent. If the parser reads no Bignum in the Rubyprogramm, but the Opalprogramm require it. The Bignum implementation must be required manually.

elia:
@opal/owners seems reasonable to add bignum support to the compiler 

If that's not something we want to support it makes sense to raise an error when the parser encounters a number not supported by js. 

It could be a hook like Opal.bignum("34676556") that can be used by an external lib to add that functionality. 

This is anyway a solution I like.

hmdne:
There was a recent idea to map Integer to JavaScript BigInt. Ie. all Integer operations would become BigInt. This would hamper performance, but would make us have a correct class hierarchy.

PR #885, opened 2015-05-22
Preview of Bignum

jannishuebl:
Hey last days I started to implement bignums for Opal like mentioned in #862. I know there are still some test failures in the rubyspec tests, these failures are caused by not supporting Bigdecimal yet.

Now I would like to know if there is the possibility that my implementation will be integrated in opal. It would be nice if you comment on my implementation to give some input to improve the Bignum support. 

Some details on my implementation:
 1. I am using Bignum implementaion of the [forge project](https://github.com/digitalbazaar/forge/blob/master/js/jsbn.js) which comes original from [Tom Wu](http://www-cs-students.stanford.edu/~tjw/jsbn/), but the forge implementation is still maintained. Using this implementation would give the possibility to create a Cryptolib for opal with the RSA implementation from the forge project.

2. Bignums are compiled to "999999999999999999999".to_i which gives the possibility to add Bignum support by simply overwriting the  String#to_i method which must be done anyway. By this way Bignum can by required as a stdlib. 

For now my implementation includes Bignums in the corelib, but it can be changed easily. 

Thanks
Jannis


meh:
To maintain semantics overflow on `Integer` operations should make a `Bignum` as well, but that would break with the optimized operators.

jannishuebl:
Of course overflow must be handeld in Integer operations.  This is currently just handeld in the power function of Numeric (Numeric#**) to make the test of this method pass. The tests of the other methods do not cover the overflows. To test the automatic convertion I would take the rubyspec test suite.

Can you explain to me how it breaks with the optimized operators?

meh:
The optimized operators check if the two values are `Number` and inline the JavaScript operators.

It wasn't a critique, just a note to think about.

PR #890, opened 2015-05-24
Add Union + Options Methods to Regex

wied03:
Javascript doesn't support 'extended' exactly and Ruby doesn't support sticky exactly, but this should be a decent start and making Regex work a little better.


meh:
Could you squash the commits please?

wied03:
Squash complete

vais:
@wied03 :+1: for squashing the commits.

I see a couple other issues with this PR. If at all possible, corelib must:

1. use JS as the implementation language
2. use rubyspec for tests

Item one should be relatively easy for you to fix (if not, let me know if you have any questions).

Item two I have not had time to research, but if you point out specific problems that are stopping you from using rubyspec to test your changes, I will work with you on wrapping them in `not_supported_on :opal` blocks in our rubyspec fork. Having our own tests mirroring rubyspec adds maintenance burden the project *forever*, so there has to be *simply no other way* in order to justify going this route.

wied03:
@vais 

1. I see your logic here. The thought had occurred to me but I did submit an earlier PR (https://github.com/opal/opal/pull/853) that included Ruby in corelib. Should that one have been JS as well? I suppose it's not as good of a candidate as this regex one since it's mainly interrogating Ruby types anyways. I'll see if I can do this in JS.
2. I agree with not wanting separate tests and can see how the not_supported_on block was used in https://github.com/opal/rubyspec/pull/5. Below are my thoughts on that.

Here are some compatibility differences I've observed:
* Ruby supports the Regex 'x'/extended flag and Javascript, on the face of it, does not. As to what the extended flag does, http://doc.infosnel.nl/ruby_regular_expressions.html and a few other sites imply that it causes whitespace in the expression itself to be ignored.
* Javascript doesn't seem to support overriding those flags within a group, whereas Ruby does /(?-mix:dogs)|(?i-mx:cats)/ in Ruby is valid. Javascript does not like that.

As to how that impacts the Rubyspecs, the following files under corelib/core/regexp relate to these methods:
* options_spec.rb
* union_spec.rb
* new_spec.rb

In all 3 cases, they bundle examples that use the x flag and/or overrides described above inside the same it block with tests that don't. When the first test fails, the rest will not run (options_spec is particularly bad at this). I think there might be too many tests in each block there.

Here are the solutions I can think of:
1. Keep the same implementation I have (aside from JS vs. Ruby) and use the not_supported_on clause you mention in the tests.
2. Also keep the same implementation, try and separate out the Rubyspecs more upstream so that ignores can be more granular and we can allow the 'valid' tests to run.
3. Tweak the implementation to make the 'x' flag and per group overrides to work as if they do in Ruby and hide the magic behind the scenes.

To me 2 seems the ideal way but may not be easy to do (I don't know how amenable they are to that).
After that, 1 might be OK with 3 as a possible longer term goal (but that would be more complex and less performant).

Your thoughts?

wied03:
JS conversion is done (and squashed into single commit)

meh:
@wied03 to be clear about the Ruby/JavaScript conundrum, in certain cases, especially on methods that are commonly called many times, Ruby will be slower, and since it's in the corelib it's better to make it fast.

But it's also true that not all Ruby code will cause slowdowns, in this case, using `case` would slow down things.

wied03:
As an aside, assuming this pull request happens, the match method can probably made more efficient. There is no need to construct a new RegExp object because we already have one (it's self in this case) with the proper flags (see new initializer). Dropping the 'g' flag also shouldn't be a problem because Regexp#match in Ruby only returns the 1st match (as opposed to String#scan).

I'm willing to do another PR for that one unless I'm venturing too far here.

meh:
If the code can be faster, and still pass specs, do go for it :panda_face:

meh:
Also I think this can be merged, I just want more feedback on the specs issue from @opal/owners.

wied03:
@meh, I would work on a match PR, but it looks like @vais changed the code from using to self to building a new expression in 758297e64479f948d5b3da1ef21ae6070dced887. I wouldn't want to undo that if there was a good reason that was already changed.

vais:
@wied03 :+1: for option 2, making the rubyspecs more granular to allow us to filter out the failing cases as unsupported. In places where this may turn out to be not practical (too much entanglement), wrap the failing code in `not_supported_on :opal`. Submit a PR to opal/rubyspec.

Regarding creating a new regexp object versus reusing the one passed into the method. Two reasons to create a new one:

1. Global-matching regexps in JS may carry with them their own *state*, aka, *baggage* (e.g. see https://github.com/opal/opal/commit/758297e64479f948d5b3da1ef21ae6070dced887#diff-4279fbef29d580005d7ab9c433c6b4caL65). The easiest thing to do is to shed that incoming baggage (and prevent new outgoing baggage leaking out of the method) by simply creating a new regexp based on the one passed in. I have not measured the effect this has on performance. I prefer immutability and predictability. In any case, see next point.
2. I found that many rubyspecs simply assume multiline and don't even test for that. Granted, that should be fixed in rubyspec, and @meh is correct in saying that if it doesn't fail the specs, go for it, but this is assuming that you will fix the specs first, then, when the specs are correct, go for it :smile:. In the meantime, tested by rubyspec or by hand, the multiline flag needs to be set, and that means creating a new regexp.

Just a couple of side notes (no need to necessarily change anything here now, just an FYI):

1. It looks like you swung a little too far in the opposite direction in your conversion to JS. The point is not to make your life hell :smile: The basic guiding principle is to write mostly JS that calls out to Ruby, as opposed to mostly Ruby that calls down into JS. That's all. It means you can still call out to Ruby where that makes sense, for example, all your `raise` calls should be Ruby - there's little benefit to converting those calls to JS and the resulting JS is quite hideous.

2. I started using Ruby's snake case for local variable names and function names. Feel free to do the same if you'd like. I guess there's no agreement on stylistic issues.

wied03:
@vais 

Re: the match method, I added a comment so hopefully others see there was a good reason there

Re: your side notes, I made some changes to the PR to try and accommodate what you were saying. It's strange getting used to coding in 2 languages in the same method :)

Re: Rubyspecs, I can't see how is #2 really a desirable first option. The parent Rubyspec project has an uncertain future and I don't think anyone will be there to accept (or reject) the pull request given the front page says "RubySpec Has Been Closed." Maybe you guys know something I don't about that.

I'll go ahead and do what I can with a PR on Opal's Rubyspec fork.

vais:
@wied03 Opal's fork of rubyspec is exactly where you should send your PRs. Also, it's in sync now with https://github.com/ruby/rubyspec which is the canonical rubyspec now used by MRI, JRuby, et al. We have not submitted our changes to them yet, but our fork is up-to-date with ruby/rubyspec fork  (as of last week anyway).

wied03:
@vais, all of that is now in place. I didn't update the submodule version in the opal repo pointing to opal/rubyspecs yet. Assuming you're happy with everything, once https://github.com/opal/rubyspec/pull/7 is merged in, the submodule can be update to that newer revision and then these specs will pass

wied03:
@vais, FYI, the reason for the confusion was that with option 2 above, when I said upstream, I meant the original Rubyspec repo, not the Opal fork. option 1 was the Opal fork, which is what we ended up following.

vais:
@wied03 looking good :+1: I added some comments to the diff, please review. Also, comments on the diff here https://github.com/opal/rubyspec/pull/7 - there is one apparent bug to fix. Once that's fixed, I think it's good to be merged in - @meh any objections?

wied03:
@vais - made those white space tweaks. I'll look into why my editor isn't doing a new line at the EOF.

wied03:
@vais - I put in a TextMate bundle for that. It was a convention I was never really aware of.

elia:
> I put in a TextMate bundle for that. It was a convention I was never really aware of.

⚠️ shameless self promotion time ⚠️

@wied03 if you're on TM2 checkout my [avian-missing tmbundle](https://github.com/elia/avian-missing.tmbundle#readme)!

vais:
@wied03 :+1: https://github.com/opal/rubyspec/pull/7 accepted, please update the rubyspec submodule in this PR to make CI pass, and I will merge this one as well.

vais:
@elia LOL you're a man of many talents :smile:

wied03:
Cool. Travis looks good now.

wied03:
@vais - No love in the 0.8.0 CHANGELOG? I guess next time I'll need to slide in CHANGELOG updates on my next PR. ;)

vais:
Sorry, @wied03, please do update the CHANGELOG with your PRs in the future - it's too easy to miss/forget stuff then sifting through commits at a later date :sweat_smile:

elia:
:+1: we should start to encourage everyone to update the changelog (starting from me :yum:)

vais:
@elia I remember you did ask me to update the changelog with my early PRs, but I stopped doing it because it caused problems when I would rebase my branches to master. I guess because we always add to the head of the changelog. That's a bad reason not to maintain the changelog of course :stuck_out_tongue_winking_eye:

elia:
🐾 we'll find a solution for that eventually, could reversing it be of some help? (I think the problem would just be moved from the top to the bottom as changes concentrate on the latest version anyway, but could be wrong of course)

PR #891, opened 2015-05-24
Fix #877 (compiler generates accidental global var if proc arg used outside proc)

### LINKED ISSUE #877: Compiler generates accidental global variable if a local variable is also among the args to a local proc

vais:
Fix #877

Basically what was happening is that args to a proc inside of a method were incorrectly set as args on the parent scope. Then, since they supposedly already exist in the args array, they were excluded from being declared with `var`.

### DIFF
--- a/lib/opal/nodes/iter.rb
+++ b/lib/opal/nodes/iter.rb
@@ -115,7 +115,7 @@ module Opal
         sexp.each do |arg|
           if arg[0] == :lasgn
             ref = variable(arg[1])
-            scope.add_arg ref
+            self.add_arg ref
             result << ref
           elsif arg[0] == :array
             result << scope.next_temp


vais:
Example:
```ruby
def meth
  1.times do |xxx|
    puts xxx
  end
  xxx = 5
  puts xxx
end
```
BEFORE:
```javascript
/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {

  Opal.add_stubs(['$times', '$puts']);
  return (Opal.Object.$$proto.$meth = function() {
    var $a, $b, TMP_1, self = this;

    ($a = ($b = (1)).$times, $a.$$p = (TMP_1 = function(xxx){var self = TMP_1.$$s || this;
if (xxx == null) xxx = nil;
    return self.$puts(xxx)}, TMP_1.$$s = self, TMP_1), $a).call($b);
    xxx = 5;
    return self.$puts(xxx);
  }, nil) && 'meth'
})(Opal);

/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {

  Opal.add_stubs(['$exit']);
})(Opal);
```
AFTER:
```javascript
/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {

  Opal.add_stubs(['$times', '$puts']);
  return (Opal.Object.$$proto.$meth = function() {
    var $a, $b, TMP_1, self = this, xxx = nil;

    ($a = ($b = (1)).$times, $a.$$p = (TMP_1 = function(xxx){var self = TMP_1.$$s || this;
if (xxx == null) xxx = nil;
    return self.$puts(xxx)}, TMP_1.$$s = self, TMP_1), $a).call($b);
    xxx = 5;
    return self.$puts(xxx);
  }, nil) && 'meth'
})(Opal);

/* Generated by Opal 0.8.0.beta1 */
(function(Opal) {

  Opal.add_stubs(['$exit']);
})(Opal);
```
Note that BEFORE `xxx` is not declared, and AFTER `xxx` is declared as one of the `var`s.

PR #892, opened 2015-05-24
Fix issues #875 and #876 (compiler generates accidental globals compiling `class << ...`)

vais:
Fix #875 
Fix #876 

Basically the problem here is that the temporary variable is requested to be created in the parent scope while processing the body AFTER `to_vars` on the parent scope has already been called. I.e., it's too late for it to have any effect since `to_vars` has already been called. The solution is simple then: process the body first, then call `to_vars`.

PR #900, opened 2015-05-30
Enable JSHint's "'variable' is not defined" check (aka detect global leaks)

vais:
Now that https://github.com/opal/opal/pull/891 and https://github.com/opal/opal/pull/892 are merged in, we no longer have any accidental globals in JS, so it's time to enable JSHint's `'variable' is not defined` check for realz (FTW, for googness' sake, for the love of god, etc.)


elia:
🔛

vais:
:smile:

PR #901, opened 2015-05-30
Implement Array#to_h

vais:


elia:
Keys seem to be passed directly as keys of the map but I think they could be of any type. 

IIRC hash2 is just for hashes with string keys as an optimization for the compiler.

vais:
@elia oops, I did not think of that. What would you recommend? Is there any other recommended "fast" way to make a Ruby hash out of a JS one?

elia:
Try with Opal.hash but check because I'm not sure if it takes also an array of arrays

vais:
@elia I'm also looking into making Opal.hash pass all the `Hash.[]` specs BTW.

I think this one is good for now with using a ruby hash and calling `store` on it, I did not observe any significant performance difference of this vs the original using Opal.hash2.

Any objections to merging this updated PR at this point?

vais:
Also, @elia, @meh, in your expert opinion, from JS code, is it better to:

1. `hash = #{{}}` or `hash = Opal.hash()` to create a ruby hash? Does it matter?
2. `hash.$store(key, val)` or ```#{`hash`.store(`key`, `val`)}``` to store a value? Does it matter?

I mean in terms of what would be considered more idiomatic and maintainable in the long run.

elia:
I prefer {} and I see it as more maintainable 

But no strong feelings on this

vais:
Thanks, @elia

PR #902, opened 2015-05-30
Hash.[] fully compliant with rubyspec

vais:


vais:
@elia any objections? Self-merging feels dirty :)

PR #903, opened 2015-05-31
Run JSHint as part of Opal test suite

vais:


vais:
Gentlemen, could you please take a moment to clarify for me your usage of the `for in` loops in the following cases? The question is: did you deliberately want to dredge up any and all properties from the prototype chain, or is this an error of omitting the check for `hasOwnProperty()`? Once I collect your responses, I can enable this check in JSHint, depending on your response, either by correcting the error or by adding a JSHint comment to your code to exclude it from this check.

@elia
https://github.com/opal/opal/blob/master/opal/corelib/file.rb#L39
https://github.com/opal/opal/blob/master/opal/corelib/runtime.js#L928
https://github.com/opal/opal/blob/master/opal/corelib/runtime.js#L1203

@meh
https://github.com/opal/opal/blob/master/stdlib/native.rb#L172
https://github.com/opal/opal/blob/master/stdlib/native.rb#L210
https://github.com/opal/opal/blob/master/stdlib/native.rb#L456

elia:
Please consider those on my list worth checking, I may or may not have been aware of the actual behaviour. Will do the check myself eventually, but didn't get the time so far.

vais:
@elia no problem, take your time. The entire PR is on hold for now. The reason is that I want to move the jshint check from the dedicated rake task it is in now to the test task. Otherwise you get a nasty surprise only after you submit a PR and it fails on Travis. In the process I discovered new accidental global issues and other badness I need to take care of first. It did not come up before because there's much more getting linted in the test task (the JS generated for all the specs, imagine that). JSHint keeps on giving :wink: I will submit the new accidental global issue separately to make sure it does not fall through the cracks in the meantime.

vais:
@elia, need your Travis expertise here - does [this last commit](https://github.com/vais/opal/commit/4499ca770a7dd9e25a068c80797a7ddd89c8682c) make sense? The idea is that since jshint now runs as part of the test suite, we do not need a separate lint task, but we still need to make sure that jshint is installed on CI. Did I put that `npm install` in the right place? Thanks for taking a look for me!

elia:
@vais travis looks good

Took the time to read file.rb:39 and seems it's wrong and needs an additional hasown check

elia:
Just read the for in docs on MDN and if I'm not mistaken they should be fine in all cases where the object is known to be a literal. Prototype entries such as toString are not marked as enumerable so they shouldn't appear. 

Anyway the code just wants to walk "own" properties of an object literal. Being JavaScript I assume there could be pitfalls in any line of code so please tell me if you find out this code is broken. :P

vais:
Thanks, @elia! I think for now it's ok to leave the "The body of a for in should be wrapped in an if statement..." check disabled. I'm thinking of merging this PR today as is. Any objections?

elia:
👍🏼

PR #907, opened 2015-06-03
Benchmarks

vais:


benchmarks/prova.js.rb
```diff
@@ -1,13 +0,0 @@
...
-# p $1
-# p $2
-# `console.log(2, $opal.gvars['1'])`
-# b = $1
-# `console.log(1, b)`
-# nil
-
-puts [*1..3].to_s
-puts [*(1..3)].to_s
-puts [*Object.new].to_s
```
elia: 😖 ouch, didn't realized this had been committed

vais: Hehe, no problem @elia, it wasn't doing any harm and actually helped me get by bearings.

elia:
@vais what about a rake task that producing the nice output we saw on gitter?

vais:
@elia I cheated on that one - I just ran the script twice (once for Opal, once for MRI) to generate two output files, then used unix's `join` command to join them on the benchmark name, then used vim's `Tabularize` command to re-align the columns. Also, to run the benchmarks against MRI, I had to modify `run.rb` temporarily to use `load(file, true)` instead of `eval`, and to work around the fact that Opal's implementation of `Benchmark.measure` returns a scalar value, whereas MRI's returns an object, and you'd call e.g. `real` on it to get a scalar. Anyway, long story short, there's a bit of work still to be done before that output you saw on gitter can be generated automatically.

PR #908, opened 2015-06-03
Implements catch throw

catmando:


meh:
Some `Promise` changes snuck in, also I think those things should go in `kernel.rb` rather than their own file.

Issue #675, opened 2014-12-29
obj.methods(false) returns [] even when there are methods


fkchang:
``` ruby
class Foo
  def bar
    :bar
  end
end

f = Foo.new
f.methods(false) # returns []
```

Live code against 0.7.0.beta3 

http://fkchang.github.io/opal-irb/index-embeddable.html#code:class%20Foo%0A%20%20def%20bar%0A%20%20%20%20%3Abar%0A%20%20end%0Aend%0A%0Af%20%3D%20Foo.new%0Af.methods(false)

cj:
@meh @elia  I just ran into this.  `public_methods(false)` doesn't exist in Opal. 0.8.0.rc1

cj:
without it existing, how would you do something like this?  https://github.com/solnic/virtus/blob/2572ebe4c8346b51b372e1bf813940047db3f648/lib/virtus/instance_methods.rb#L207

elia:
"public_methods" should be added as an alias of "methods"

@cj can you do a PR for this?

PR #909, opened 2015-06-03
alias public_methods methods

cj:
@elia here's the PR as requested via #675.

I also saw @meh said "There's no concept of private or public in Opal...".  Shouldn't we add some way to track public and private methods?  There are a few ruby methods that rely on it, which would cause unexpected results if you are using the same class client/server side.

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -56,6 +56,8 @@ module Kernel
     }
   end
 
+  alias public_methods methods
+
   def Array(object)
     %x{
       var coerced;


elia:
@cj I'd like to have private/public support too, the problem is finding a way to implement it with decent performance. Until someone comes up with a good implementation it's "not supported".

cj:
@elia what solutions have you tried?

elia:
honestly I don't remember, maybe @meh has some historic memory of them, @adambeynon surely does…

Ajedi32:
Just throwing this idea out there (rough sketch):

```javascript
var some_object = {}

/***** Method definition *****/
some_object.some_method = function() {
  // Code here
}

// If method is public:
some_object.some_method.public = some_object.some_method;

// if Method is private
some_object.some_method.public = function() {
  throw new Exception("Calling a private method in a public context!");
}

/***********/

// Calling in private context
some_object.some_method();

// Calling in public context
some_object.some_method.public();
```

adambeynon:
We used to add two functions for every method definition. One for being called as a public method (with a receiver), and one for being called for a private method (without a receiver).

At runtime then, if we are defining a private method then the method body to be called in a private context was just replaced with a function that raised an error.

Was removed ~ 3 years ago.

elia:
maybe a visibility check can be added as an option for development only, like it's done with arity, having a simple boolean on the method body:

```rb
def.$foo = function TMP_def_foo(){…};
def.$foo.$$private = true
…
Opal.call_with_visibility(self, 'private', 'foo', args)
```

but probably would make generated code horrible…

Ajedi32:
@adambeynon Any idea why that was removed? It seems like a fairly efficient implementation to me, on the surface at least.

adambeynon:
@elia Yeap, agreed. Something that can be turned off (or off by default) would be the best approach.

@Ajedi32 I can't remember really. It was only ever used to be more compatible with ruby spec. I have nothing really against adding it (as long as its opt-in).

PR #910, opened 2015-06-03
Module class eval compilation

catmando:


adambeynon:
Hi @catmando. Not sure if this has already been picked up, but there seems to be some changes to promises in this commit as well. Was that intentional?

elia:
@catmando the eval stuff has been fixed in:

- 44b8cfca14 (BasicObject#instance_eval(str)` now compliant)
- 64d625301c (Fix context of evaluation for #eval and #instance_eval.)
- 23972813b1 (Add support for Module#module_eval with a string)

Thank you for your code that helped moving in the right direction!

PR #912, opened 2015-06-03
Add JS module for nicer calling of native javascript operators and global functions

jeremyevans:
This adds support for calling the following javascript operators:
delete, in, instanceof, new, typeof, and void.  It also supports
calling global javascript functions via call.  call is aliased
to method_missing, so that you can do JS.globalFunction() to
call global javascript functions.

### DIFF
--- a/spec/opal/stdlib/js_spec.rb
+++ b/spec/opal/stdlib/js_spec.rb
@@ -14,6 +14,11 @@ describe 'javascript operations using JS module' do
     JS.new(f).JS[:color].should == 'black'
   end
 
+  it 'JS.new handles blocks' do
+    f = `function(a){this.a = a}`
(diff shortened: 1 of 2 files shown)


elia:
Any reason Opal.global isn't ised in #call?

jeremyevans:
@elia Probably because I didn't know to use it.  I'll push a new commit later today for that uses `Opal.global`.

Looks like the Travis failure is unrelated to this code.

elia:
Yes, the failure was my fault 😖
But should be fixed on master

jaredcwhite:
@jeremyevans Really big :+1: on this -- I can hardly wait to begin using it. I love the idea of avoiding backticks for various JS one-liners throughout my Opal code. :)

PR #915, opened 2015-06-06
Make cruby benchmark.rb + tests run on Opal

vais:


vais:
If not for the one place where they used a series of `gsub!`s, we could have just required `benchmark.rb` directly from `test/cruby/lib`. That was the only reason I had to copy over the file. Oh, well :)

jgaskins:
Having a few troubles with this still:

```
➜  opal git:(538d7f1) ruby -v benchmark/run.rb test/cruby/benchmark/bm_app_answer.rb test/cruby/benchmark/bm_app_factorial.rb test/cruby/benchmark/bm_app_fib.rb
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
RUBY_GC_HEAP_INIT_SLOTS=1000000 (default value: 10000)
RUBY_GC_MALLOC_LIMIT=1000000000 (default value: 16777216)
test/cruby/benchmark/bm_app_answer.rb    benchmark/run.rb:27:in `read': No such file or directory @ rb_sysopen - test/cruby/benchmark/bm_app_answer.rb (Errno::ENOENT)
	from benchmark/run.rb:27:in `block in <main>'
	from benchmark/run.rb:20:in `each'
	from benchmark/run.rb:20:in `<main>'
```

Same with the Opal version:

```
➜  opal git:(538d7f1) bundle exec opal -v benchmark/run.rb test/cruby/benchmark/bm_app_answer.rb test/cruby/benchmark/bm_app_factorial.rb test/cruby/benchmark/bm_app_fib.rb
Opal v0.8.0.rc1
test/cruby/benchmark/bm_app_answer.rb    fs.js:500
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: ENOENT, no such file or directory 'test/cruby/benchmark/bm_app_answer.rb'
    at Error (native)
    at Object.fs.openSync (fs.js:500:18)
    at Object.fs.readFileSync (fs.js:352:15)
    at OpalClass.$read (/private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:26050:21)
    at $a.$$p.TMP_3 (/private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:26645:61)
    at Opal.yield1 (/private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:830:14)
    at Array.def.$each.TMP_9 [as $each] (/private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:7287:26)
    at /private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:26650:77
    at Object.<anonymous> (/private/var/folders/rm/rl4k78_n1bgg08fhgcptpjn40000gn/T/opal-nodejs-runner-20150606-13014-1nehdba:26658:3)
    at Module._compile (module.js:460:26)
```

Am I missing something?

vais:
@jgaskins looks like maybe you did not `git submodule update --init`?

elia:
The only thing about which I'm a bit doubtful is Kernel#sleep, but I'm aware of the fact that there aren't many solutions other than the tight loop. (Excluding the horror that npm's execAsync is last time I checked). 

Also the addition of process to the corelib highlights the fact that a mini-opal would be very useful.

vais:
@elia my thoughts on `Kernel#sleep` exactly. Consider this just a shim for compatibility. It does solve an immediate problem in Opal development, although it obviously should never be used in a production application.

Regarding adding process to the corelib, I just moved it from stdlib to corelib to match the way cruby works; i.e. process.c is in the corelib, so `Process` does not need to be `require`d in order to be used.

elia:
@vais yes, we talked many times with @adambeynon about a mini version of opal made by cherry-picking just the bare necessary from the corelib but it still to be done. e.g. those not using external gems will never need Process or `Kernel#sleep` so it makes sense having the option to load a stripped down version etc.

vais:
@elia a proper linker would be even better :wink: I.e. at compile time, Opal determines which parts of the corelib are used by the application and only includes those parts. Opal is already a step ahead of MRI in terms of deployment - the holy grail of compiling Ruby to a static executable with all the dependencies included (like, e.g. Go) - Opal kind of does that now! Except for the corelib :disappointed:

adambeynon:
mruby has an interesting approach to this as well to cut down file sizes.

vais:
@adambeynon in your expert opinion, how doable is something like this in Opal right now?

elia:
@adambeynon iirc they have parts of the corelib extracted as "gems" with progressive enhancement of functionality (e.g. basic gsub doesn't support regexps) is that right?

vais:
@elia interesting. Does that mean a developer would have to explicitly `require` pieces of corelib?

elia:
@vais that's what I remember but could be pre 1.0. wrt the linker and porting the precompilation back to mri the idea crossed my mind many times and I think it would be great to have

PR #917, opened 2015-06-06
Fixing RubySpec for Array#permutation

kaiwren:


vais:
Go @kaiwren! Merging as is for now, but in the future, please consider using mostly JS in the corelib for performance. Just consider a small example from your code as food for thought: Ruby `num == 1` compiles to JS `num['$=='](1)`. Ok, that's an extra function call already (not good), but let's see how `Number#==` is implemented:
```
  def ==(other)
    %x{
      if (other.$$is_number) {
        return self == Number(other);
      }
      else if (#{other.respond_to? :==}) {
        return #{other == self};
      }
      else {
        return false;
      }
    }
  end
```
So, that's all the code that needs to run only to compare `num == 1`. If you wrote `num === 1` in JS, then **that** would be the only code that would run. Does that make sense?

kaiwren:
@vais Agreed, and thanks for the feedback. I'll refactor `Array#permutation` and raise another pull request shortly.

vais:
@kaiwren no problem, thank you!

PR #918, opened 2015-06-06
Updating CHANGELOG

kaiwren:
Include compliance with RubySpec for Array#combination, Array#permutation, Array#values_at, Array#rotate and Array#rotate


vais:
Thanks, @kaiwren!

PR #919, opened 2015-06-07
Add a few passing numeric specs

eregon:
Float specs tend to be easy for Opal as JS is natively behaving quite similarly :grinning:


vais:
Thank you, @eregon!

PR #920, opened 2015-06-07
Using ruby/rubyspec for RubySpec

eregon:
Hello,

Here is a quick PR to discuss about using ruby/rubyspec directly, if that is what the Opal team wants.
This is not ready for merging as there are new failures due to new specs.

Using git submodule provides a very good isolation between specs and the rest, but also makes the process to change specs a bit rigid.
If changes to rubyspec are done mostly by a few people (@elia and @vais from what I saw), one solution is to allow these people to push directly to ruby/rubyspec for straightforward changes (and PR if not).

Another solution is to have a copy of rubyspec in the opal repository (with git subtree or manually) and integrate these changes upstream from time to time. This is more complex and messy but useful if many different people contribute directly to the specs.

What do you think?

P.S.: There are quite a few new failures due to new specs, do you have an automated way to tag them or is it done manually?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "spec/corelib"]
 	path = spec/corelib
-	url = https://github.com/opal/rubyspec
+	url = https://github.com/ruby/rubyspec.git
 [submodule "test/cruby"]
 	path = test/cruby
 	url = https://github.com/ruby/ruby.git


eregon:
Interestingly the Travis build passes, but that seems due to outdated cache.

vais:
First of all, :+1: for getting involved with Opal, @eregon!

Second, :+1: for this idea.

> If changes to rubyspec are done mostly by a few people (@elia and @vais from what I saw), one solution is to allow these people to push directly to ruby/rubyspec for straightforward changes (and PR if not).

This would be my first choice. If giving Opal direct access to `ruby/rubyspec` is an issue, my second choice would be for `opal/rubyspec` to be a fork of `ruby/rubyspec` so we can take advantage of the straight-forward, easy github workflow for keeping the two in sync as needed. To be perfectly honest, the current situation makes my primate brain *hurt*.

elia:
:+1: for commit access too

@eregon rebasing on master should properly break travis ([the cache was my fault](https://github.com/opal/opal/commit/ddd0b8bccb4eef8ef238e738dc03862b3dfef61a) :disappointed:)

eregon:
@elia @vais I have added both of you to the rubyspec team!

I think using ruby/rubyspec will help avoiding too large diffs between the version Opal uses and upstream. If it proves not flexible enough, it can of course always be reconsidered.

Feel free to commit straightforward changes directly to ruby/rubyspec (like guards, splitting examples, clear improvements, etc). For the rest, please push your changes to a branch (of ruby/rubyspec) and make a pull request so to get a chance to review changes before merging them in master.

In general, try to avoid using guards if it can be easily prevented, as it makes specs harder to read and potentially introduces more merge conflicts.
Guards to avoid parsing/compiling issues like the ones that were merged to ruby/rubyspec are OK, even though in the long term I would like those to migrate to filters/tags when that becomes possible.

The Travis build now fails as expected. Could one of you add the necessary filters?

vais:
Thanks, @eregon!

@elia, I can take care of this over the next couple of days (not feeling well tonight).

elia:
@eregon thanks! The ruby org looks great on the profile page!
@vais take care

vais:
@eregon, sorry, it was easier/quicker for me to get this done in a separate PR https://github.com/opal/opal/pull/926/files. Now Opal is using ruby/rubyspec directly! Thanks again for everything you've done to make this happen!

eregon:
@vais Awesome!
No problem, I just wanted to make clear the change I proposed with this PR.

eregon:
Just a little note on naming: both JRuby and Rubinius have RubySpec located in `spec/ruby` and MRI has it in `spec/rubyspec`.
`spec/corelib` is fine as well but it might be surprising to some people that this is actually a RubySpec submodule :wink:.

vais:
@eregon it was like that when I got here :smile: Maybe @elia knows the back story...

elia:
@eregon iirc we renamed it at the time of the great sl-* split in rubyspec/rubyspec & rubinius

eregon:
@elia Makes sense now!
Do you think it should be renamed again or not worth the trouble?

elia:
fine by me, but would prefer to get the green light from @opal/core too

vais:
Come to think of it, yes, it should be renamed, considering that now it actually contains stdlib specs as well under library. So, we have a folder called `corelib` that contains a folder called `library` that contains specs for stdlib.

I'd say let's call it `spec/rubyspec`

eregon:
@vais Exactly, RubySpec is larger than its `core/` part.

vais:
@eregon @elia https://github.com/opal/opal/pull/930

PR #922, opened 2015-06-08
Refactoring Array#permutation's implementation to use js over rb

kaiwren:
Cleaning up implementation based on feedback on #917


vais:
:+1:

PR #923, opened 2015-06-08
Updating Contributing.md to mention installation of jshint

kaiwren:


vais:
Good catch, thanks, @kaiwren

vais:
@elia do you think jshint could be vendored into stdlib/nodejs/node_modules?

elia:
stdlib node_modules are intended to be something used by stdlib/nodejs stuff, so I think it's ok to keep it external

vais:
Ok, thanks @elia

PR #927, opened 2015-06-09
Use ruby/mspec for MSpec

eregon:
So RubySpec runs on an expected version of MSpec.
Also, `ruby/mspec` does not depend on `fileutils` anymore.


elia:
Thanks! :+1:

PR #928, opened 2015-06-09
Define only the necessary constants for MSpec

eregon:
The first is needed for `ruby_exe` but I am not confident if this works yet.
The second is a standard `RUBY_` constant along VERSION, ENGINE, etc.


elia:
I wonder if any of those should go to `opal/corelib/variables.rb`

eregon:
@elia Yes, RUBY_PATCHLEVEL should.

`RUBY_EXE` is MSpec-specific.

PR #930, opened 2015-06-10
Rename spec/corelib to spec/rubyspec

vais:


elia:
:+1: 

Probably the rubyspec prefix can be dropped from `spec/rubyspecs` as they're all in the same place, but I'm ok with doing it at a later time

eregon:
Nice :+1:

PR #933, opened 2015-06-11
Fix a few of the new Rubyspec failures

vais:


vais:
@elia @meh there's this new thing in rubyspec regarding enumerators returned from a given method call in turn being able to correctly handle the `size` method call on themselves. Before I go off fixing these failures, I want to confirm with you if I'm on the right track. Here's what I did to fix the `String#each_char` [implementation](https://github.com/opal/opal/pull/933/files#diff-13a82ca8c6e6c08ecbc8c8001bc398e0R367):  `return enum_for(:each_char){self.size} unless block_given?` Note the block attached to the `enum_for` call that wasn't there before - it does the trick of making sure that per spec *returned enumerator size should return the size of the string*. I plan to use the same trick throughout. Any comments/objections/ideas? Thanks!

elia:
Makes sense to me but I'll leave the last word to @meh :)

vais:
@elia @meh this puppy's ready to merge. Any comments/objections/better ideas?

meh:
Sounds good to me.

PR #936, opened 2015-06-14
Rubyspecs list cleanup

vais:
This PR represents a switch from a "whitelist" to a "blacklist" way of thinking about RubySpec. In other words, until now, Opal was considered *guilty until proven innocent*, so we only ran a select few specs, fully expecting everything else to fail. With this change, Opal is considered *innocent until proven guilty* - we try to run all the specs, and those specs that do not pass are excluded with a `!` prefix.

This means a lot more `fails` lines in `filters/bugs` files in some cases, but I think it's worth it to have known problems be in plain sight to be dealt with. I consider `filters/bugs` to be my TO DO list, and it's comforting to know that it's not hiding anything from me.


meh:
:+1: from me.

vais:
Thanks, @meh!

PR #940, opened 2015-06-18
Run all of RubySpec (that is within Opal's scope)

vais:
The following is a list of 25 remaining RubySpecs we are *NOT* running as of this PR:

* argf
* bignum
* binding
* complex
* continuation
* dir
* encoding
* env
* fiber
* file
* filetest
* gc
* io
* main
* marshal
* mutex
* objectspace
* process
* random
* rational
* signal
* symbol
* systemexit
* thread
* threadgroup

As of this PR, the total number of RubySpecs Opal is passing is :sparkles:6227:sparkles: This PR marks the end of the free lunch - from now on getting another hundred passing specs will be hard work :smile: 

:beers:


opal/corelib/kernel.rb
```diff
@@ -1114,4 +1114,9 @@ def to_s
   end
 
   alias untaint taint
+
+  def Rational(*args)
+    #Just a stub to let unrelated rubyspecs run.
+    nil
```
elia: Maybe it's better to `raise NotImplementedError`, wdyt?

vais: @elia, it would be better, but then it fails with an exception, which I cannot use `fails` to filter out. I mean having it raise a NotImplementedError is no different in this case than the original NoMethodError that prompted me to add this stub. I agree, this is not pretty. The real solution is to go ahead and implement Rational, I guess :smile:

elia: ok, fair enough, Rational should be fun to implement after all :sweat_smile:

vais: @elia yes, as a PhD thesis :smile:

meh: I had some stuff in my numeric branch for that, I just have to find the will to live for doing all that bridge rework actually work.

Issue #882, opened 2015-05-21
`SyntaxError: Unexpected token !` if there's no white space around `!=`


adamcreekroad:
```ruby
"/desk_delegate/agent_online_check?v=1&#{force_update!=nil&&force_update ? 'force_update=true&' : ''}callback=?"
```

Works fine in IRB, comes up with syntax error in Opal 0.8

elia:
@adamcreekroad please paste here the error you get with its backtrace and the exact version (or git revision) you're using, thanks!

vais:
@elia 
```
$ cat ~/Desktop/adamcreekroad.rb 
puts "hello"
p "/desk_delegate/agent_online_check?v=1&#{force_update!=nil&&force_update ? 'force_update=true&' : ''}callback=?"
buts "bye"
```
```
$ bundle exec opal -v ~/Desktop/adamcreekroad.rb 
Opal v0.8.0.beta1

/Users/vais/GitHub/opal/opal-nodejs-runner.js:14453
 Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, force_update!$ = nil;
                                                                     ^
SyntaxError: Unexpected token !
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:929:3
```
@adamcreekroad as a temporary workaround, put spaces around `!=`. It's always a good idea to do that anyway.

Ajedi32:
Reduced even further:

```ruby
variable = 1
puts variable!=2
```

Result:

```text
SyntaxError: Unexpected token !
SyntaxError: Unexpected token !
    at $TryOpal.def.$eval_code (http://opalrb.org/javascripts/try.js:20:27290)
    at $TryOpal.def.$run_code (http://opalrb.org/javascripts/try.js:20:27179)
    at HTMLDocument.$a.$$p.TMP_3 (http://opalrb.org/javascripts/try.js:20:28057)
    at f (http://opalrb.org/javascripts/application.js:2:16766)
    at Object.p.fireWith [as resolveWith] (http://opalrb.org/javascripts/application.js:2:17547)
    at Function.K.extend.ready (http://opalrb.org/javascripts/application.js:2:12308)
    at HTMLDocument.pt (http://opalrb.org/javascripts/application.js:2:9736)
```

[...]

PR #951, opened 2015-06-19
Fixed regex for matching identifiers

seaneshbaugh:
This fixes https://github.com/opal/opal/issues/882.


meh:
Thanks :panda_face:

PR #955, opened 2015-06-20
Start working on Time bugs

vais:


vais:
Thanks, @elia. Not sure how to proceed with the rest of the `Time` failures until https://github.com/opal/opal/issues/957 is resolved. Seems to me like a high-level architectural decision needs to be made about what Opal wants to do with `Time`. Or maybe I'm just having a giant brainfart, and it is in fact possible to make time zones work with `Time` bridged to JS `Date` as it is now. In any case, I kinda hit a wall with it today. So, thanks for merging the little I had so far, cause I'm gonna take a break from it for a while.

PR #961, opened 2015-06-22
add wording to explicitly clarify which spaces the CoC applies in

shadowcat-mst:
This was originally the wording @strand and I worked out previously, which @adambeynon recently commented he thought at least the concept of was useful. It's been tweaked since though so please read the whole thread to understand where it's coming from.

I'm happy to (continue to :) bikeshed the phrasing until everybody's happy - I just think we need to draw *some* sort of explicit line here.

@adambeynon @sarciszewski @jaen @webmaven thoughts if you have any, please

PR #962, opened 2015-06-22
tweak sexual language wording

shadowcat-mst:
My thought here is that since harassment of any form is already specifically covered, it's better to only ban -overtly- sexual language so that a single genuine misunderstanding doesn't become a casus belli.

Bear in mind that once the misunderstanding has been explained, doing it again falls neatly under the dictionary definition of harassment and as such this tweak shouldn't allow anybody who -was- being a dick to weasel out of it.

@adambeynon @sarciszewski @jaen @webmaven thoughts if you have any, please

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -4,7 +4,7 @@ As contributors and maintainers of this project, we pledge to respect all people
 
 We are committed to making participation in this project a harassment-free experience for everyone, regardless of level of experience, gender, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, age, or religion.
 
-Examples of unacceptable behavior by participants include the use of sexual language or imagery, derogatory comments or personal attacks, trolling, public or private harassment, insults, or other unprofessional conduct.
+Examples of unacceptable behavior by participants include the use of overtly sexual language or imagery, derogatory comments or personal attacks, trolling, public or private harassment, insults, or other unprofessional conduct.
 
 Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct. Project maintainers who do not follow the Code of Conduct may be removed from the project team.
 


shadowcat-mst:
Note that I'm not referencing other tickets intentionally and instead @-ing people to pull them over here. Hopefully that'll give us a head start to get some discussing done.

Ajedi32:
Yeah, good idea. If you reference any of the other tickets where the drama was happening, chances are this PR will turn into another political debate.

Ajedi32:
As for the change itself; this looks good to me.

webmaven:
Meh. :neutral_face:

shadowcat-mst:
@webmaven I've seen this particular clause go south before, and I can't think of a single incident in the time I've been an op where this change would've stopped me banning somebody who deserved it.

adambeynon:
I agree with this change. Is this something that is being changed upstream in Contributor Covenant as well?

shadowcat-mst:
I've no idea what upstream would think of this change but would have no objection to somebody trying to do so. My first priority though is to get something in place here that people think will work for Opal.

webmaven:
I have no substantive objection.

I don't think it will help much, but fail to see how it would prevent non-overt sexualized language or imagery from falling under the harassment category in general, for example.

No matter what, you're going to have folks with differing *'I know it when I see it'* opinions as to what is *'sexualized'*, with this change you're also going to have folks with differing opinions as to what is *'overt'*.

domgetter:
> differing opinions as to what is 'overt'

@webmaven Easy, just look for the phrase "if you know what I mean...".

webmaven:
@domgetter:
> just look for the phrase "if you know what I mean...".

Yeah, but that's just the most obvious of the dog whistles.

shadowcat-mst:
@webmaven neither of the codes of conduct proposed in-thread included that phrase at all; @adambeynon chose to close those and accept this one instead, hence we're attempting to tweak this one.

adambeynon:
Please see my comment in https://github.com/opal/opal/pull/961#issuecomment-114533977 , as I think it applies to this pull request also.

meltheadorable:
The original code of conduct didn't specify an exact action to take for any specific code of conduct violation, so a genuine misunderstanding could already have been handled by just removing the post/comment/whatever and asking the person not to do it again.

Now it explicitly protects persistent innuendo and other sexual commentary. Since enforcement was always left with wiggle room, and didn't require drastic action in the event of a misunderstanding, I'm not sure the original concern (that a misunderstanding would tie everyone's hands and force a ban) is worth weakening the language this much, and I think you'd hear other women raise similar complaints if you were to raise the issue upstream.

strand:
> @adambeynon: I agree with this change. Is this something that is being changed upstream in Contributor Covenant as well?

**No.** I'm not a maintainer of the Contributor Covenant, but this change is clearly antithetical to the ethos of the original, and harmful to marginalized people.

sarciszewski:
> harmful to marginalized people

How does it harm marginalized people? 

Overt means, "open to view". 

If you're engaging in overtly sexual behavior, then it's obviously intentional, but I challenge you to enjoy a conversation with a group of strangers for a few minutes without one of them saying something that someone could in some sense take as sexual innuendo.

Ajedi32:
I'm hesitant to reply here because I don't want to dredge up all this crap again, but it seems to me that some people here are concerned about the COC excusing clearly inappropriate behavior, while others are concerned about it being used as a weapon against clearly innocent people.

The way I see it; unless you write a 100-page document in legalese laying out in excruciating detail precisely what is and isn't allowed, you're always going to have to rely on common sense anyway. (That's also why I've considered a CoC to be pointless from the beginning.)

Can we all just agree that the CoC in its current form *can* be interpreted in a reasonable way, and rely on common sense to ensure that it isn't used to protect the guilty or attack the innocent? (Which, by the way, is exactly how we'd deal with inappropriate behavior if we didn't have a CoC; just use common sense.)

PR #971, opened 2015-07-02
Implement BasicObject#!=

### LINKED ISSUE #970: != method not implemented on BasicObject
The `!=` method is not implemented on BasicObject. While the `!=` expression can be used in the normal way, it cannot be sent to an object using the `send` method.

Example:

```
$ opal-repl
>> s = ''
>> s == ''
>> true
>> s.send :==, ''
>> true
>> s != ''
>> false
>> s.send :!=, ''
undefined method `!=' for ""
```

I expect the last example to return false.

vais:
Fix-ish #970

### DIFF
--- a/opal/corelib/basic_object.rb
+++ b/opal/corelib/basic_object.rb
@@ -34,6 +34,10 @@ class BasicObject
     false
   end
 
+  def !=(other)
+    self != other
+  end
+
   alias eql? ==
   alias equal? ==
 
--- a/spec/filters/unsupported/basicobject.rb
+++ b/spec/filters/unsupported/basicobject.rb
@@ -1,6 +1,4 @@
 opal_filter "BasicObject" do
-  fails "BasicObject#! is a public instance method"
-  fails "BasicObject#!= is a public instance method"
   fails "BasicObject#initialize is a private instance method"
   fails "BasicObject#method_missing is a private method"
   fails "BasicObject#singleton_method_added is a private method"


elia:
I think even if this is good a real fix is needed at compiler level, one that makes JS call `#!=` instead of wrapping a `#==` call into `#!@`

vais:
@elia yeah, one would think the following would normally cause stack overflow, but in Opal it does not, because Opal has what in computer science I believe is called *stack overflow elimination* optimization built into it at the compiler level :stuck_out_tongue_winking_eye: 
```ruby
def !=(other)
  self != other
end
```
It works. Um... for now anyway :smile:

elia:
Merging of course, but I'll remove the fix so @mojavelinux's issue will remain open to remind us there's stuff to do in the compiler

vais:
Sounds good to me, thanks @elia!

elia:
btw I wonder how _stack overflow elimination_ is not also a Big Bang Theory episode title… 😹

vais:
Honestly, I'm not sure how this works at all, but I'm going on 4 hours of sleep, and it's been a looong day, so *stack overflow elimination* inside Opal sounds plausible at this point. I better shut down for the night before I do any real damage... :beers:

opal/corelib/basic_object.rb
```diff
@@ -34,6 +34,10 @@ def !
     false
   end
 
+  def !=(other)
+    self != other
```
Ajedi32: Haha, yeah it might be a good idea to make this more explicit and write `!(self == other)` instead. Otherwise this will break when the compiler is fixed.

mojavelinux: Why not implement it as follows?

```
def != other
  `self !== other`
end
```

That would avoid a stack overflow as it delegates directly to JavaScript. I could see, however, if this is a problem if Ruby and JavaScript don't agree on equality rules.

Ajedi32: Yeah, that would fail in the general case. For subclasses that redefine `==`, we want `!=` to behave correctly.

mojavelinux: Gotcha.

elia: @Ajedi32 feel free to send a PR to fix it ;)

vais:
https://github.com/opal/opal/pull/975

PR #977, opened 2015-07-04
Compile `!=` expression to a single method call instead of `==` + `!`

### LINKED ISSUE #970: != method not implemented on BasicObject

vais:
Fix #970 (for realz this time?)

### DIFF
--- a/lib/opal/nodes/if.rb
+++ b/lib/opal/nodes/if.rb
@@ -10,14 +10,14 @@ module Opal
       RUBY_ENGINE_CHECK = [:call, [:const, :RUBY_ENGINE],
                               :==, [:arglist, [:str, "opal"]]]
 
+      RUBY_ENGINE_CHECK_NOT = [:call, [:const, :RUBY_ENGINE],
+                              :!=, [:arglist, [:str, "opal"]]]
+
       RUBY_PLATFORM_CHECK = [:call, [:const, :RUBY_PLATFORM],
                               :==, [:arglist, [:str, "opal"]]]
 
-      RUBY_ENGINE_CHECK_NOT = [:call, [:call, [:const, :RUBY_ENGINE], :==,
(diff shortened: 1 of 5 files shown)


vais:
I was concerned there could be a performance hit since `!=` is defined higher up in the inheritance hierarchy than `==`, and there indeed seems to be one, albeit quite negligible I would say:
```
$ git checkout master
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.396
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.398
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.398

$ git checkout compile-not-equal-to-a-single-method-call
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.437
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.426
$ bundle exec opal -e "t = Time.now;1_000_000.times{'hello' != 'world'};puts Time.now - t"
0.414
```
Considering it's about 30 milliseconds spread over 1_000_000 iterations, I'd say it makes no difference for all practical purposes.

elia:
Great stuff, and benchmarks are always cool :sunglasses: 

Agree that the perf change is negligible.

PR #984, opened 2015-07-07
Fix Hash#has_value? when hash was created through initialize method.

iliabylich:
Basically, [here is the link to the demo of this issue](http://opalrb.org/try/?code:h%20%3D%20%7B%20a%3A%201%20%7D%0Aputs%20h.has_key%3F(%3Aa)%0Aputs%20h.has_value%3F(1))

It seems that `Hash#[]=` populates only one of `self.smap` / `self.map` for each key/value pair, but `Hash#has_value?` previously used just a `self.map`. I've already checked other methods - all of them use both `map` and `smap`, so this is the only affected method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ build/
 /node_modules
 /gh-pages
 /tmp
+.ruby-gemset
+.ruby-version
+.rvmrc
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -537,8 +537,22 @@ class Hash
 
   def has_value?(value)
     %x{
-      for (var khash in self.map) {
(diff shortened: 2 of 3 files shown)


vais:
Nice one, thank you @iliabylich :+1:

iliabylich:
@vais Do you have any ideas when it will be released? I understand, this fix is really minor, but I just don't like to have any gems from github in my Gemfile :smile:

elia:
I think this is a good candidate for backporting to 0-8-stable and be released in 0.8.0.rc3

iliabylich:
@elia Thanks!

vais:
:+1: @elia

elia:
done in 5bf2be5948335565a92a0dbb9529f682340b047c

PR #998, opened 2015-07-11
Array Frozen: The Resurrection

vais:
A bunch of Array specs pertaining to frozen status were mistakenly placed into the unsupported bucket. This is their resurrection...


elia:
I don't think it was by mistake …until you came along 😁

Just one question: I see #frozen? Is being used where for some reason I expected to see something like self.$$frozen is it something required by the spec or there's some other reason?

vais:
@elia not sure I understand your question, could you clarify with a specific example? I mean, `#frozen?` is used because that's the obvious way (for me) to check whether an object is frozen or not. Do you mean use `self.$$frozen` instead for performance?

elia:
Agree it's the obvious way, and yes, the concern was mainly on performance but also due to the fact that overriding #frozen? would "change" the frozen status of the object, if you see what I mean.

vais:
So, @elia, something like this in `array.rb`:
```javascript
function check_frozen() {
  if (self.$$frozen) {
    #{raise RuntimeError, "can't modify frozen Array"}
  }
}
```
then in every method where this check is needed:
```ruby
def []=(index, value, extra = undefined)
  `check_frozen`
  ...
end
```
Or should it be `Opal.check_frozen` defined in `runtime.js`? Hmm... then the message text needs to be parameterized in some way... WDYT?

elia:
Both look good, maybe we can start with something local to Array and scale it up to a global helper once we have a better picture of what we need in the other classes of the corelib and stdlib. 

What I mean is that full frozen support would need to do the check every time we set an instance variable.

meh:
Yeah, we cannot support freezing, even if it's simple in some classes.

vais:
@elia agree, local to array. @meh given how String works, I do not think we will ever have any "ideological purity" (all or nothing) vis-a-vis freezing, so I guess my take on it is "get what you can working". At this point it's fully supported for arrays.

vais:
Unless, @meh, you could expand on your comment a little - e.g. do you fee this is opening a "can of warms" we do not want to open? In any case, a little more explanation would make throwing away this PR a little easier to swallow :smile:

meh:
It's not really opening a can of worms, but having immutable `String` is a thing, having freezing only working as intended on `Array` or a couple other classes and having the rest just do nothing.

Freezing is on the same footing as tainting, it cannot be implemented without a huge performance hit, so it's either complete (which is not feasible) or none at all.

Now I don't remember why `#freeze`, `#frozen?` and friends are no-ops, but it's probably because some gem was using freezing or tainting for reasons that would make no sense in the browser.

vais:
@meh thanks for explaining your reasoning, it makes sense.

PR #1015, opened 2015-07-18
Remove console.log statements

vais:


elia:
😖 (embarrassed)

vais:
C'mon now, don't be silly @elia!

I am looking at `Hash` now, and my intentions are not good :smiling_imp: I was stuck trying to figure out how in Ruby two objects that have the same value for #hash can exist as two different keys in a Hash as long as they are not `eql?` to each other. In Opal that's just not possible, since #hash is the key we use for the internal JS map object, so two objects with the same #hash cannot be two keys in the map - first one wins. So, I was like, *how the hell does this work in Ruby??* There's a free chapter of Ruby Under the Microscope online that just happens to be about how Hash is implemented in MRI, very informative and totally explained it. Even linked to the Rubinius implementation that's a lot easier to understand than MRI's C. Now at least I understand how MRI/Rubinius/JRuby pull it off, but it's totally not applicable to Opal - they can take our JS map [out of our cold dead hands](https://en.wikipedia.org/wiki/From_my_cold,_dead_hands) - no way in hell we're implementing Hash in JS based on an array of buckets each containing an array of items, when we have a super-performant native JS object. So, the problem remains - I need to be able to have two objects that have the same #hash serving as keys in a Hash, pointing to different values. I got some ideas from the above reading though, so I'll play around with it, and if it's any good, I'll send a PR.

elia:
No worries jk. Now I can also confess that I started to realize how the eql? Was involved in Hash toward the end of [my journey](https://github.com/opal/opal/pull/628) but thought I had enough already. 😇

Don't know what you have in mind but a hybrid solution could be to have some kind of special array in the values only if a collision happened. With special I mean marked with something like .$$collision = true. Which of course would mean adding an additional check during access. But all things considered it shouldn't add much. Does it make sense?

Feel free to ignore and do things your way of course 😉 yet now I'm quite curious about your schemes for sneaking eql? inside Hash 😋

vais:
@elia yep, that's exactly what I had in mind as well. Keys are off limits - the key for the JS map must be the Ruby key's #hash. But the value we can do something with. My thought was to wrap the value in an object that has a `value` property and a `link` property to the next value object if there is one, and so on. This `value` wrapper can also be an array - I'm not sure about that yet. But we're thinking along the same lines - I guess there are only so many ways this can be done :smile: Performance will be the deciding factor.

elia:
🚀

PR #1022, opened 2015-07-23
some explanations added, using the index too

### LINKED ISSUE #1020: rack example broken
The rack example includes an index.html.erb, but does not in fact use it.

Since there is no readme and it states in other places that this index approach is the way to go, it seems natural for a beginner to assume the index.html.erb generates the loaded page.

It does not, as the title of the page gives away.

Some digging reveals that one has to set the index_path to the index file name and that this has been changed in 0.6!!

dancinglightning:
fixes #1020

### DIFF
--- a/examples/rack/config.ru
+++ b/examples/rack/config.ru
@@ -1,7 +1,20 @@
 require 'bundler'
 Bundler.require
 
+# Instructions: bundle in this directory
+# then run bundle exec rackup to start the server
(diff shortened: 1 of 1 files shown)


elia:
Great! :+1:

vais:
Excellent! Thank you, @dancinglightning! :sparkles: :beer: :sparkles:

PR #1032, opened 2015-07-30
raise server exception for unconfigured main asset path

artost:
Opal::Server won't work without at least configuring 'main'. This small change makes that clearer.


elia:
Thanks!

Can you add a hint on how to configure it in the exception message?

artost:
No problem, added a hint.

artost:
Added a couple fix commits, then squashed them.

elia:
thanks!

PR #1047, opened 2015-08-08
Proc#dup, #clone, #===, #yield, and #lambda?

### LINKED ISSUE #932: proc is empty when it shouldn't be
The proc is empty when it shouldn't be.  http://goo.gl/1FA7Q7

It seems to be something to do with `#deep_dup`

vais:
Fix #932


vais:
@meh :beers:

Issue #1045, opened 2015-08-05
Class#new cannot be optimized by v8


jgaskins:
I just noticed that, in Chrome/v8, instantiation of Ruby objects is unable to be optimized by the VM because it uses the `arguments` array (via `*args` in Opal): https://github.com/opal/opal/blob/247d921f43e2fcdec47405c1c2a38472fc89ca71/opal/corelib/class.rb#L42

Safari doesn't seem to have any trouble with it (it renders Clearwater apps 3x as fast as Chrome when I don't use render caching), but in rendering a pretty large virtual DOM several times in Clearwater just now, I spent 14.6% of my CPU time in `Class#new`.

I don't know if there's a way around it, but @vais asked me nicely to open an issue. :-) I've tried a few things, like explicitly setting positional arg names in `Class#new`, but that breaks when you _do_ want splatted args in class constructors.

jgaskins:
I did some more profiling and benchmarking and it looks like it's the reassignment of the splat variable that's causing the problem.

```ruby
require 'opal'

class Foo
  def bar(*args, &block)
    %x{}
  end

  %x{
    def.$baz = TMP_2 = function() {
      var $iter = TMP_2.$$p;
      var block = $iter || nil;

      var args = $slice.call(arguments, 0);
      TMP_2.$$p = null;
    };
  }
end
```

The two methods in this class compile to nearly identical JS. The only real difference is that `args` in `bar` is reassigned from the method signature. You can see I removed `args` from `baz`'s method signature and declared it as a variable within the function. Using this code that looks like it was written by a high-school student and not a veteran software engineer, I had some interesting results in both Chrome and Safari:

```ruby
foo = Foo.new

start = Time.now
count = 0
while Time.now - start < 10
  foo.baz
  count += 1
end

`document.write("<p>" + count + "</p>")`

start = Time.now
count = 0
while Time.now - start < 10
  foo.bar
  count += 1
end
```

In Chrome, the performance increase was roughly 18%:

```
19766704

16446747
```

In Safari, the results were even better (+24%):

```
31287542

25341646
```

[...]

Issue #934, opened 2015-06-12
Splat with nil should not pass nil as an argument to method


vais:
With arity check enabled, the following code fails with an `ArgumentError`, `wrong number of arguments(1 for 0)`
```ruby
def test
  puts 'ok'
end
a = nil
test(*a)
```
In MRI this works fine ~~because splat with nil is a special case - when splat is used with nil in a method call, the method does not receive any arguments~~ because Ruby calls `to_a` on the splat argument, and `nil.to_a` is an empty array, so the method receives no arguments.

vais:
Thank you, @meh!

PR #1048, opened 2015-08-08
Optimize Class#new

jgaskins:
This is the result of modifying `Class#new` with one of my experiments from #1045 and rendering an intentionally large VDOM in a Clearwater app went from 303ms down to 267ms in Chrome, a time savings of 11.8% — roughly what I saw in those experiments.

After the VM invokes the JIT/optimizer (about 2-3 renders in), rendering without this patch converged on 123ms. With this patch, those same renders hit 101ms. This is 17.9% better, so subsequent renders benefit even more.

### DIFF
--- a/opal/corelib/class.rb
+++ b/opal/corelib/class.rb
@@ -40,12 +40,13 @@ class Class
   end
 
   %x{
-    def.$new = TMP_2 = function() {
-      var $iter = TMP_2.$$p;
+    var TMP_NEW;
+    def.$new = TMP_NEW = function() {
+      var $iter = TMP_NEW.$$p;
       var block = $iter || nil;
 
       var args = $slice.call(arguments, 0);
-      TMP_2.$$p = null;
+      TMP_NEW.$$p = null;
 
       var obj = this.$allocate();
 
--- a/opal/corelib/class.rb
+++ b/opal/corelib/class.rb
@@ -39,15 +39,21 @@ class Class
   def inherited(cls)
   end
 
-  def new(*args, &block)
-    %x{
-      var obj = #{allocate};
+  %x{
+    def.$new = TMP_2 = function() {
+      var $iter = TMP_2.$$p;
+      var block = $iter || nil;
+
+      var args = $slice.call(arguments, 0);
+      TMP_2.$$p = null;
+
+      var obj = this.$allocate();
 
       obj.$initialize.$$p = block;
       obj.$initialize.apply(obj, args);
       return obj;
-    }
-  end
+    };
+  }
 
   def superclass
     `self.$$super || nil`


meh:
I don't think this is the right way, I'm pretty sure this optimization can be done directly in the compiler.

In short, not emitting the argument for the splat in the function body.

meh:
That is if I'm getting the optimization preventing issue correctly. Can you confirm please?

elia:
Yes, if we manage to fix this at compiler level all splat calls will benefit.

Here's where the `slice.call` happens:
https://github.com/opal/opal/blob/master/lib/opal/nodes/def.rb#L129-L134

And here's the var list (params):
https://github.com/opal/opal/blob/master/lib/opal/nodes/def.rb#L94

jgaskins:
I wasn't sure if it would actually benefit all calls because I only ran a few benchmarks in 2 browsers, so I figured I'd at least add a PR for code I already wrote that I know does help. :-)

meh:
I don't think it has any practical difference, it's still just a method call.

elia:
http://jsperf.com/opal-reassign-splat-args

I tried on the browsers I have available

meh:
Looks good on here, I'll patch the compiler if nobody wants to take it.

jgaskins:
@elia Nice, I added runs for Safari 9 and Chrome Canary. It's even more drastic than I thought on some browsers.

elia:
@meh it's all yours 😁
@jgaskins cool, I tried WebKit nightly but I think it's being reported as safari 8

vais:
@meh :+1: There is another splat issue that needs to be fixed in the compiler: https://github.com/opal/opal/issues/934 Basically, the code generated by the compiler should call `to_a` on the splat argument. Maybe while you are in there you can fix this one too :sweat_smile:

meh:
@jgaskins can you try the benchmarks again to make sure I didn't fuck the implementation up please?

jgaskins:
Nailed it!

meh:
Cool, thanks for spending time profiling, we really need more of it :panda_face:

ylluminate:
:+1:

meh:
@jgaskins also if you can test with the latest commit as well, I had to change splat **calls** to make them actually compliant with Ruby, I'm pretty sure the helper will be optimized out, but you never know.

jgaskins:
@meh Looks good:

```
Opal Default, no args, capture block: 27636548

Optimized JS, no args, capture block: 27814075

Opal Default, two args, no block: 28775224

Optimized JS, two args, no block: 29000361

Opal Default, two args, capture block: 26465777

Optimized JS, two args, capture block: 25358184
```

I ran it a few times. The `Opal Default` version was always a very slight bit lower for the first two pairs, which is probably due to `Opal.splat`, but the difference iss probably less than a GC pause would make.

This is really good work. :+1:

PR #1051, opened 2015-08-09
Optimize hash#clone

jgaskins:
We can just copy over the entire maps and keys wholesale, which saves anywhere from 60-80% of the cloning time, depending on the VM.

This patch dropped merging a 100,000-key hash from 89ms to 18ms in Safari — 189->73 in Chrome.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1163,21 +1163,12 @@
   };
 
   Opal.hash_clone = function (from_hash, to_hash) {
+    var key;
+    for(key in from_hash.map) { to_hash.map[key] = from_hash.map[key]; }
+    for(key in from_hash.smap) { to_hash.smap[key] = from_hash.smap[key]; }
+    to_hash.keys = from_hash.keys.slice(0);
     to_hash.none = from_hash.none;
     to_hash.proc = from_hash.proc;
-
-    for (var i = 0, keys = from_hash.keys, length = keys.length, key, value; i < length; i++) {
-      key = from_hash.keys[i];
-
-      if (key.$$is_string) {
-        value = from_hash.smap[key];
-      } else {
-        value = key.value;
-        key = key.key;
-      }
-
-      Opal.hash_put(to_hash, key, value);
-    }
   };
 
   Opal.hash_put = function (hash, key, value) {


meh:
@jgaskins looks like that breaks the key order.

jgaskins:
Indeed. I don't understand how, though, because it's copying the keys array (and everything else) as-is. It should be an identical object. I'll look into it a bit more and see if I can figure out what's up.

meh:
Most likely the `for .. in ..` loop does not maintain order since it's iterating over the object's keys in whatever order.

Replacing that with a C for would likely fix it.

jgaskins:
I tried iterating over `Object.keys` for both maps , but those two specs still fail. Looking at the spec output (from [this spec](https://github.com/ruby/rubyspec/blob/e27464cb8234e6d607f62f5a925486f05117046a/core/hash/shared/update.rb#L34), it looks like `merge_pairs` is what I expect, but `merge_bang_pairs` (unaffected by this patch) is way off. Am I reading that right?

vais:
@jgaskins the problem is that this is making a shallow copy of `keys` and `map`. This will seem to work, until it rightly breaks on a more complex case (such as the `merge!` spec). Object key representations that are stored in the `keys` array and in the `map` object cannot simply be copied to another hash, because then modifications in one hash will affect the other hash. If you would like to know more about how these object keys work, take a look here: https://github.com/opal/opal/blob/master/spec/opal/core/hash/internals_spec.rb#L47

jgaskins:
Ah, I see. I didn't look closely enough at how `map` was stored. I saw the stringified keys and thought "oh yeah, this will be super easy". I'll give it another shot and see if a deeper copy of `map` still yields any performance benefit.

vais:
@jgaskins that said, I still think you have a great idea there - `clone` can still be optimized for hashes that contain only string keys (which is, like, the 99.9% use case that you really want to optimize). The following *should* work:
1. Check if the `from_hash` `map` is empty
2. If it is, then do the optimization:
3. Iterate over the `from_hash` `keys` using a `for` loop and copy them over to the `to_hash` `keys`
4. Iterate over the `from_hash` `smap` using a `for in` loop (because order does not matter) and copy the props to the `to_hash` `smap`
5. Profit

Makes sense?

jgaskins:
If we only need to copy `smap`, we should just be able to use the `from_hash.keys.slice()` call, which is probably optimized already in the VM. There didn't seem to be any issue with those in this patch. Just the `map` keys.

In any case, I did want to optimize the object-key case but I think you're right that cloning hashes with object keys is likely to be so rare that it'll never be a performance bottleneck except for this exact case where I accidentally filled the hash with 100k numeric keys instead of strings. :-)

However, it looks like any performance gains we get from optimizing for that case are lost on small-medium hashes in checking to see whether the `from_hash.map` is empty (you can't just do `from_hash.map.length`, so you have to iterate over the keys and check `hasOwnProperty` on each and I've now flipped every table in my house), so we'd basically just be complicating it for marginal gains on cloning very large hashes. It's such a rare corner case (vast majority of hashes are quite small) that it just doesn't feel worth it. :-\

vais:
>  ...and I've now flipped every table in my house

@jgaskins LOL, yeah, realizing that there is no direct and backwards-compatible way in JS to check if an object is empty reminded me once again why I want to be doing my client-side programming in Opal instead :smile:

For my tests I actually just used `Object.keys`, but that's off-limits for core because, you know, IE... We could set (and maintain) an internal flag on the hash to know that it's string keys only, but yeah, I'm not sure either that going to great lengths to optimize the performance of huge hashes on the client side is really justified at this time.

PR #1052, opened 2015-08-11
Optimize rand with a range

jgaskins:
Creating an array out of the range and then choosing a random element is elegant but, as you might imagine, not very performant for wide ranges. :-)

This patch boosts performance of `rand(1...10)` by 8.5x in Safari 9 on my machine.

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -1003,9 +1003,10 @@ module Kernel
         return Math.random();
       }
       else if (max.$$is_range) {
-        var arr = #{max.to_a};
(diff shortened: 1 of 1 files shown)


elia:
Nice!

elia:
I tried in IRB and apparently we lack (ruby)specs and support for a couple of behaviors:
- returning nil if range.begin > range.end
- coercing passed range with `#to_int` unless begin and end are Numeric (e.g. `rand 'a'...'z')

It's just a reminder for anyone that wants to fix it :)

cc @vais

vais:
:+1:

PR #1057, opened 2015-08-14
Use official sprockets processor cache keys api

elia:
The old cache key hack has been removed.

Add `Opal::Processor.cache_key` and `Opal::Processor.reset_cache_key!` to 
reset it as it’s cached but should change whenever the `Opal::Config` changes.

- [ ] add a change log entry


elia:
_I'm dumb and this was intended to target 0-8-stable_

PR #1063, opened 2015-08-16
Refactor bridging and method definition.

### LINKED ISSUE #313: method_added support
I'm working on building a front-end framework, and I was hoping to use method_added to track when method's get added, then setup an alias method, that wraps the previous method, with additional tracking code.  (To make a really nice observer API)

So for example, in MRI, this prints out "ADDED blue"

```
class Something
  def self.method_added(method_name)
    puts "ADDED #{method_name}"
  end
  
  def blue
    puts "Blue"
  end
end
```

If you want, I can try to add the needed code.  I'm not sure if this would incur a significant performance penalty, since it only would run something when a method is added, but it might make since to make this another flag similar to method_missing.

Also, great work on opal.  I'm using it for a few projects and its super useful.  

Thanks

meh:
It's basically done par a not false-positive in the specs.

I want some eyes on it since it's a pretty big change, if I've done everything right it will allow proper `Numeric` semantics.

/cc @opal/owners 

fixes #313

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -8,7 +8,7 @@
   var Opal = this.Opal = {};
 
   // All bridged classes - keep track to donate methods from Object
-  var bridged_classes = Opal.bridged_classes = [];
+  var bridges = {};
 
   // TopScope is used for inheriting constants from the top scope
   var TopScope = function(){};
@@ -72,7 +72,7 @@
     var constant = this[name];
 
     if (constant == null) {
-      return this.base.$const_missing(name);
+      return this.base.$const_get(name);
     }
 
     return constant;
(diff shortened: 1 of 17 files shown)


meh:
Short version of what this does.

- `method_added` and `singleton_method_added` now work
- `alias` now works how it really should
- now you can bridge a class that already inherits from another (not `Object` or `BasicObject`) to a native class

meh:
Oh yeah, and now `Kernel` and `Object` aren't weird hacky things.

meh:
I confirm my `numeric-redux` branch is working with proper semantics.

```ruby
>> 2.class.ancestors
=> [Number, Numeric, Comparable, Object, Kernel, BasicObject]
>> `2` + 2
=> 4
>> `2`.conj
=> 2
>> class Numeric; def foo; 42; end; end
=> "foo"
>> 2.foo
=> 42
```

opal/corelib/runtime.js
```diff
@@ -433,21 +439,105 @@
...
+
+        if ($hasOwn.call(ancestor.$$proto, name) &&
+            ancestor.$$proto[name] &&
+            !ancestor.$$proto[name].$$donated &&
+            !ancestor.$$proto[name].$$stub &&
+            ancestor !== from) {
+          break;
+        }
+
+        if (ancestor == from) {
```
elia: Is the double eql intentional?

meh: Not really, but I don't think it changes anything in this case, does it?

elia: I don't think too, it's it just raises the above question in my brain when I see it :)
If there's not particular reason to do otherwise I'd change it to `===` just because it's the default.

elia:
Overall looks good, but it's also a big one (for the good). 
We can wait before merging and test it with something non-trivial such as opal-rspec or opal-browser.

The only other concern I have is performance on method definition but it's probably negligible given that the corelib `def`s are just 640.

> Oh yeah, and now Kernel and Object aren't weird hacky things.

<3

And really happy about the `iclass` stuff too!

vais:
@meh :heart:

PR #1064, opened 2015-08-17
Numeric redux

meh:
Obviously dependant on #1063.

/cc @vais


meh:
@vais also it would be nice if you could poke the rubyspec guys to stop using bignums in every single unrelated numeric spec.

vais:
:scream: :joy_cat:

PR #1066, opened 2015-08-17
Cherrypicking

meh:
Again dependant on #1063.

@elia try this with `opal-jquery` and `opal-rspec`, if it works I'm merging the bridging stuff.

/cc @adambeynon, @vais


meh:
Again, long story short.

We had stuff in stdlib that really belonged to corelib because of file size concerns. It just makes everything more complex for no good reason.

I remember me and @adambeynon talking about something like this a long time ago, so it's here, the common user shan't worry about it anyway.

Also added `unsupported.rb` so everything that is not supported is centralized for dummy methods and whatnot.

ryanstout:
Looks like a great idea :-)

elia:
I'd leave stdlib/{math,encoding}.rb with a require and a warning

eg. `stdlib/math.rb`

```rb
require 'corelib/math'
warn "DEPRECATION: math is now part of the core library, requiring it is deprecated"
```

elia:
Also I think we need a way to avoid raising `NotImplementedError` for stuff that can be easily faked like `.freeze` and `.taint`. We could either have warnings or noops, or let the user decide…

eg.

```rb
module Kernel
  `var ERROR = "Object freezing is not supported by Opal";`

  def freeze
    raise NotImplementedError, `ERROR` unless $OPAL_STUB_FREEZE
  end

  def frozen?
    raise NotImplementedError, `ERROR` unless $OPAL_STUB_FREEZE
  end
end
```

meh:
@elia I think it's a bad idea to have things working but not working, that's why I went that direction.

Ajedi32:
I think there's some benefit to both approaches. Seems like having a compiler option would be good. You could have the: "I'm writing/porting this code to specifically to work with Opal, show all errors" mode, and the "I want to see if this code works in Opal without any modification; try your best to compile it" mode.

meh:
@Ajedi32 yeah, I changed my mind on it, now you get stubs by default, but you can enable hard errors for them (which is done while running RubySpec).

PR #1072, opened 2015-08-23
Optimize Enumerable#sort_by

jgaskins:
I ran a benchmark that sorts a shuffled array of 50k objects. That
benchmark ran in 1040-1100ms under Safari.

Using JavaScript array indexing inside the sort function instead of Ruby array indexing dropped that number down to about 500-520ms. Mapping and sorting a single additional array (down from 3) reduced the range to 460-480ms.


vais:
:+1:

Issue #1077, opened 2015-08-26
Checking if class descendands from another using the less than operator returns true for the descendant class


wied03:
Code:

```Ruby
class Foobar
end

class Bar < Foobar
end

stuff = [Bar, Foobar]
stuff.each do |klass|
  puts "#{klass} inherit? #{klass < Foobar}"
end
```
MRI 2.2.2 output:
```
Bar inherit? true
Foobar inherit? false
```

Opal 0.8/0.9 output:
```
Bar inherit? true
Foobar inherit? true
```

PR #1078, opened 2015-08-26
Fix class descendant issue

wied03:
https://github.com/opal/opal/issues/1077

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -30,9 +30,14 @@ class Module
   end
 
   def <(other)
+    # class cannot be a descendant of itself
     %x{
       var working = self;
-
+      
+      if (working === other) {
+        return false;
+      }
+      
       while (working) {
         if (working === other) {
           return true;
--- a/spec/opal/core/language/class_spec.rb
+++ b/spec/opal/core/language/class_spec.rb
@@ -20,3 +20,21 @@ describe "Assigning Class.new to a constant" do
     ConstantWithAssignedClass.new.bar.should == :bar
   end
 end
+
+describe "Class descendant check using < operator" do
+  klass1 = Class.new
+  klass2 = Class.new(klass1)
+  klass3 = Class.new
+  
+  it "is a descendant" do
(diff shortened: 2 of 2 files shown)


Ajedi32:
Should this be added to RubySpec? Seems to me this isn't something specific to Opal.

wied03:
Probably. Can you find it anywhere in there before I submit a PR to rubyspec?

Ajedi32:
@vais or @eregon would probably know better than me on that point. I'm actually not very familiar with the RubySpec internals. I just assumed it didn't have such a test since you mentioned in your commit message that you "could not find any ruby specs for this".

eregon:
There is such a spec in https://github.com/ruby/rubyspec/blob/master/core/module/lt_spec.rb.

eregon:
Have a look at https://github.com/opal/opal/blob/47969e5500b1a30697a5992f621dac0923ecdf11/spec/filters/bugs/module.rb, essentially one of the first lines can be removed and now passes :wink: 
I think you can run RubySpec with `SUITE=rubyspec bundle exec rake`.

eregon:
(and part of the organization of the file structure is explained in http://ruby.github.io/rubyspec.github.io/organization/, but it should really make it to the README)

Ajedi32:
Ah, makes sense. Yeah, given this fix it's likely some of those filters can be removed.

Now I'm curious though about `spec/opal/core`. Given that we already have RubySpec, what purpose do those specs serve? It seems likely to me that a lot of those specs duplicate things already in RubySpcec. Or are they mostly just meant for behavior specific to Opal?

elia:
I think most of them are a fork of rubyspec with unsupported stuff removed and possibly (but I'm not sure) some Opal specific additions. E.g. in case of specs that had unsupported syntax but were still very useful.

wied03:
@eregon - I missed that. Argh! I looked at one lt_spec file but missed that one.

wied03:
I'll do another PR to use the Rubyspec route

PR #1081, opened 2015-08-27
Correct earlier duplication of RubySpec for class descendants issue

wied03:
Ruby specs were available for https://github.com/opal/opal/pull/1078 (commit 35b2a095dd6def54078f1ff57a141470ab62d566), so removing the custom specs that were added and using those


meh:
You don't need to add it to `rubyspecs`, it's already using the whole directory, that spec included, the one that start with a `!` are blacklisted.

wied03:
Ahh OK. Must have been the spec level filters that were excluding it then. Will fix and push.

wied03:
@meh - fixed and squashed

meh:
Cool, thanks.

eregon:
:+1:

PR #1083, opened 2015-08-28
Implemented nil#to_c, nil#to_r, nil#rationalize.

iliabylich:


iliabylich:
By the way, should we also remove the whole file `spec/filters/bugs/nil.rb`?

meh:
If it's empty, yes.

iliabylich:
@meh Done

meh:
Thank! :panda_face:

PR #1084, opened 2015-08-28
Fixed infinite recursion on calling Comparable#<=> when it was not defined.

iliabylich:
This solution creates some corner cases with `ancestors` like in the following example:
``` ruby
module ComparesByValue
  def <=>(other)
    self.value <=> other.value
  end
end

class A < Struct.new(:value)
  include ComparesByValue
  include Comparable
end

puts A.new(1) < A.new(2)
# true for MRI 2.2.3
# false for Opal
```

At the same time, rubyspecs for `Kernel#<=>` are [quite strange](https://github.com/ruby/rubyspec/blob/master/core/kernel/comparison_spec.rb). Because of specs implementation we have to call `==` in `Kernel#<=>`, while `Comparable#==` depends on `Kernel#<=>` (when it was not overwritten on subclass level) which calls infinite recursion.

I agree that this solution is not full and ideal, so I'm ok if you reject this PR, but does anyone have any other ideas?

Btw, running the following code in `opal-repl` causes MRI crash:
``` ruby
class A
  include Comparable
end
A.new <=> 1
```

### DIFF
--- a/opal/corelib/comparable.rb
+++ b/opal/corelib/comparable.rb
@@ -54,4 +54,8 @@ module Comparable
     return false if self > max
     return true
   end
+
+  def <=>(other)
+    nil
+  end
 end
--- a/spec/filters/bugs/comparable.rb
+++ b/spec/filters/bugs/comparable.rb
@@ -1,4 +1,0 @@
-opal_filter "Comparable" do
-  fails "Comparable#== when #<=> is not defined returns false and does not recurse infinitely"
-  fails "Comparable#== when #<=> calls super calls the defined #<=> only once for different objects"
-end


meh:
I have a better solution for the infinite recursion, but it doesn't fix the super spec.

meh:
Fixedi in ec8f400c658f0c81a1c8fae0b6d7c78b23969708.

iliabylich:
@meh wow, nice idea :smile: 

Actually I was thinking about [rubinius implementation](https://github.com/rubinius/rubinius/blob/dbd33a6dadd407041fcafa6651986ae02551f3ee/kernel/common/thread.rb#L45) - it uses `Thread.recursion_guard` for everything that may cause infinite recursion (mostly it's for inspecting/comparing recursive objects).

Opal has the same global issue, right? Just wondering do you have in mind any workaround for fixing it?

meh:
What do you mean with "global issue"?

iliabylich:
I mean every object is affected by this issie - you can not compare/inspect it if it is recursive, right?

meh:
No, `#inspect` already had recursive checks in place if I recall correctly, now the `#<=>` recursion is fixed as well.

PR #1092, opened 2015-09-02
Add validation of instance variable name to Kernel#instance_variable_defined?/get/set.

iliabylich:
Just a few questions:
1. Is it ok to move validations like this to a separated class/module? To be honest, I don't know which conventions are you using, so I've made it just like in rubinius (an equivalent to `Rubinius::Type`).
2. There was an issue with ivar name validation, initially I was going to make it using ruby regexes, but different rubyspecs related to regexes and `MatchData` started failing. All of them seems to be related to usage of `$'` which is a global variable containing last match data. So, the situation was the following: some spec calls `String#match` -> then it calls (somewhere inside) `Kernel#instance_variable_get` -> it calls opal's `String#match` -> overrides `$'` -> spec becomes broken, expected value returns new `MatchData`. So is it generally correct to test  global variables this way (I mean is it a RubySpec issue?).
3. I've added validation of instance variable names and some opal functionality appeared to be broken. `Struct` supports `-` in attribute names, and in opal it was using instance variables for its attributes (instance variables can't contain dash), so I had to change this behavior to store attrs using a simple hash inside. Is it ok? As an alternative, I can re-write it to call something like `attribute.underscore` to determine instance variable name and use in both reader- and writer-methods


opal/corelib/types.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    name = name.to_str
+
+    unless String === name
+      raise TypeError, "can't convert #{name.class} to String (#{name.class}#to_str gives #{name.class})"
+    end
+
+    `var properly_formatted = name.search(/^@[a-zA-Z_][a-zA-Z0-9_]*?$/) == 0;`
+
+    unless `properly_formatted`
```
meh: This can just be:

```ruby
unless `/^@[a-zA-Z_][a-zA-Z0-9_]*?$/.test(name)`
  ...
end
```

opal/corelib/types.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.validate_instance_variable(name)
+    unless name.respond_to?(:to_str)
+      raise TypeError, "#{name.inspect} is not a symbol nor a string"
+    end
+
+    name = name.to_str
+
+    unless String === name
+      raise TypeError, "can't convert #{name.class} to String (#{name.class}#to_str gives #{name.class})"
+    end
```
meh: From the start to here, just do:

```ruby
name = Opal.coerce_to!(name, String, :to_str)
```

meh:
Move the validation to `helpers.rb`, name it something like `instance_variable_name!`, bikeshedding welcome on this one.

meh:
Also it looks like some failures are happening because of this in `SUITE=opal`, dunno why.

iliabylich:
@meh all comments are fixed + rebased. But I still don't know what's wrong with specs, locally everything is fine, it seems to be broken only on CI. I've installed the same environment (ruby and node) and it also doesn't break anything for me. Do you have any ideas?

meh:
@iliabylich no clue, there's something finnicky going on with specs because I have some failures that on travis do not happen, @vais, @elia, had time to investigate?

meh:
Also @iliabylich were the changes to `Struct` intentional with this pull request?

If they were, it would be better they would be separated in another commit, and enable relevant failing specs.

iliabylich:
@meh done. Probably will make a fix for `Struct` later, for now broken spec is excluded from test suite.

meh:
Thanks! :panda_face:

PR #1093, opened 2015-09-03
[Docs] Adding instructions on how to use opal-parser.js

gabrielrios:
I saw that both on http://opalrb.org and on the README, we talk about `opal-parser.js` but there's no docs about how to use it

So I added some *very* basic  instructions on how to user opal-parser on the README just so the user can be aware of that possibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,38 @@ File.write 'app.js', builder.to_s
 ```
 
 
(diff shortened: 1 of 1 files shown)


elia:
Nice! Thanks!

(will add just a small tweak afterwards adding the meta/encoding tag to the HTML)

PR #1094, opened 2015-09-04
Store Struct attributes in hash instead of instance variables.

iliabylich:
Rubyspec says that it should be possible to override `Struct#initialize` and "try" to override attributes:
``` ruby
User = Struct.new(:name) do
  def initialize(*)
    self.name = 'Name'
    super
  end
end

# However, the username should be taken from arguments:
User.new.name
# => nil
```

To make method `user` to not raise error, variable `@data` need to be initialize before `initialize` is invoked.

That's why `Struct#@data` should be:
1. lazily initialized using method `data` (which makes impossible for `Struct` to have an attribute `data`)
2. explicitly initialized every time when we access it

Every object in MRI has some core methods, which makes impossible for `Struct` to have *any* attribute names:
``` ruby
S = Struct.new(:marshal_dump, :marshal_load)
Marshal.load(Marshal.dump(S.new))
# => ArgumentError: wrong number of arguments(1 for 0)
```
Can we just rename this 'virtual' method `data` (which actually doesn't exist yet) to something like `__data__` and lazily initialize variable `@data`?

### DIFF
--- a/opal/corelib/helpers.rb
+++ b/opal/corelib/helpers.rb
@@ -113,4 +113,10 @@ module Opal
 
     name
   end
+
+  def self.valid_method_name?(method_name)
+    method_name = Opal.coerce_to!(method_name, String, :to_str)
+
+    `/^[a-zA-Z_][a-zA-Z0-9_]*?$/.test(method_name)`
+  end
 end
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -15,6 +15,17 @@ class Struct
         args.each { |arg| define_struct_attribute arg }
 
         instance_eval(&block) if block
+
+        class << self
+          def new(*args)
+            instance = allocate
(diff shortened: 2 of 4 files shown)


iliabylich:
I've changed .travis.yml to run failed specs in separated suite and they are passing. Is it possible that some specs are responsible for breaking other specs?

meh:
Use a special field called `$$data` instead, and no method.

About `#initialize`, make `.new` do the initialization work instead.

Also you will need to rebase, and remove the travis changes, just add the failures even if they're spurious, we need to refactor the whole `spec/opal` business anyway.

iliabylich:
Finally specs are green. The was another place in the code with `Struct#initialize`. I suppose there was a kind of roulette of which code is invoked first. Also made a rebase, pull is ready for merging

meh:
Thanks! :panda_face:

PR #1096, opened 2015-09-05
More splat args optimization

jgaskins:
A lot of the optimizations we did before on splat args worked really well to speed things up, but I noticed some more places during profiling that said the VM couldn't optimize certain method calls because it used `Array.prototype.slice.call(arguments, n)` during arg splatting. v8's warning message for it was "bad value context for arguments value", presumably because we were calling an Array prototype function on something that's not technically an array.

This patch replaces the `slice` call with old-school iteration. Here are the benchmarks before and after the patch — they're iteration counts over 5 seconds, so larger is better:

```
Chrome 45

Before:
1 arg : 9,700,318
3 args: 7,923,521
5 args: 8,756,920

After:
1 arg : 18,278,855
3 args: 17,146,539
5 args: 16,672,049
```

```
Safari 9

Before:
1 arg: 13,294,087
3 args: 12,640,861
5 args: 11,544,130

After:
1 arg: 17,845,844
3 args: 17,696,489
5 args: 17,382,340
```

### DIFF
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -135,7 +135,12 @@ module Opal
       def compile_rest_arg
         if rest_arg and rest_arg[1]
           splat = variable(rest_arg[1].to_sym)
-          line "var #{splat} = $slice.call(arguments, #{argc});"
+          line "var array_size = arguments.length - #{argc};"
+          line "if(array_size < 0) array_size = 0;"
+          line "var #{splat} = new Array(array_size);"
+          line "for(var arg_index = 0; arg_index < array_size; arg_index++) {"
+          line "  #{splat}[arg_index] = arguments[arg_index + #{argc}];"
+          line "}"
         end
       end
 


jgaskins:
Well, this is causing problems I don't know how to solve at 4am. :-) I'll take another look at this over the weekend. It's working fine in a Clearwater app and in my benchmarks. I have no idea what it's doing that's causing it to fail tests.

elia:
Awesome, stuff!

(looks like Travis isn't able to find uglyfyjs, should be easy to fix)

meh:
Can you squash the commits please?

meh:
Also it looks like that exception is raised when you pass a negative number to `new Array`.

jgaskins:
I was gonna squash anyway, but I left it overnight hoping you all might know something about the craziness I was seeing. :-)

Updated benchmarks with the new changes since I flailed around with it last night:

```
Chrome 45:
1 arg: 19979570
3 args: 17279673
5 args: 17735849

Safari 9:
1 arg: 19104805
3 args: 18570975
5 args: 18290730
```

I'm surprised at how allowing Chrome's optimizer to work on this allows it to outrun Safari 9 in the single-arg case, considering it was trailing by 38% before. I also checked against Chrome Canary (though I don't have a base case for it) and the results were even more intense:

```
Chrome 47:
1 arg: 20328544
3 args: 18195190
5 args: 17787757
```

There are still a few places where the v8 optimizer can't run in the code (a few `try` / `catch` blocks, functions with too many different types of args passed in, etc), but I'm not sure they're worth it yet because it'd probably be too much work for too little gain. In a large Clearwater app, those spots only take up about 1% of the rendering run. This one was anywhere from 3-5% depending on the app (it was 13-15% before we did the previous splat optimizations) and with this patch it is gone from the first page of my profiling, which is great considering splat args are part of literally every Ruby object instantiation in Opal. :-)

meh:
Yeah, any optimizations to the compiled output and the corelib are going to affect everything in a really good way.

Is there any chance you could do the benchmarks on Firefox as well? I'd rather not optimize for v8 only and then have spidermonkey regress greatly (especially since I only use Firefox :panda_face:).

meh:
Anyway, merging, good work!

jgaskins:
Looks like Firefox is good. :-)

```
Firefox 40

Before:
1 arg: 5,416,028
3 args: 4,880,887
5 args: 4,351,953

After:
1 arg: 14,391,315
3 args: 12,675,545
5 args: 12,378,624
```

meh:
Awesome :+1:

vais:
Great stuff, rock on @jgaskins!

PR #1097, opened 2015-09-07
Optimize truthiness/falsiness checks

jgaskins:
I noticed we were first comparing to `nil`, then `!boolean || true`. Judging by the test that this patch fails, it seems to be that we're accounting for a special case where a monkeypatched `Boolean` method in Opal code could return `self`. I think it's great to cover all the bases, but it seems like an obscure case and truthiness checks would execute about 33% faster on average without it. (Benchmark code [here](https://gist.github.com/jgaskins/591474508055b2f8c591).)

Current master:

```
if true        : 19,636,814
if truthy value: 17,418,954
if false       : 15,485,197
if nil         : 17,246,398
```

With this patch:

```
if true        : 22,655,235
if truthy value: 23,646,471
if false       : 23,254,342
if nil         : 23,458,954
```

The difference in a real-world app isn't as high because most apps that care about performance and developer sanity aren't rapid-fire conditional statements, but it's about 5% when rendering an [obnoxiously huge Clearwater app](https://clearwater-roda-example.herokuapp.com/obnoxiously_huge_page):

```
Before:
[Log] Generated virtual DOM in 541ms

After:
[Log] Generated virtual DOM in 514ms
```

Admittedly, this part is a bit less scientific. This _is_ over several renders, but I'm too lazy to write the numbers down and figure out various stats over them, so I made a mental note of the numbers and just went with the ones for each version that seemed like they would be around the 20th percentile. They were all over the place (I had the devtools open to see the logs), so I could easily be wrong in one direction or another. One thing is for sure, though: it _is_ faster. :-)

I'm aware that this PR involves a tradeoff — we lose a piece of edge-case coverage for faster execution — so closing or merging this PR is really a matter of deciding which of the two is more important. I want to argue in favor of ditching the edge case because it seems so obscure to me but I realize I don't have all the context. I assume someone had a reason for it since there's a test. :-)

### DIFF
--- a/lib/opal/nodes/helpers.rb
+++ b/lib/opal/nodes/helpers.rb
@@ -92,7 +92,7 @@ module Opal
         end
 
         with_temp do |tmp|
-          [fragment("((#{tmp} = "), expr(sexp), fragment(") !== nil && (!#{tmp}.$$is_boolean || #{tmp} == true))")]
+          [fragment("((#{tmp} = "), expr(sexp), fragment(") !== nil && (#{tmp} !== false))")]
         end
       end
 
@@ -106,7 +106,7 @@ module Opal
         end
 
         with_temp do |tmp|
-          [fragment("((#{tmp} = "), expr(sexp), fragment(") === nil || (#{tmp}.$$is_boolean && #{tmp} == false))")]
+          [fragment("((#{tmp} = "), expr(sexp), fragment(") === nil || (#{tmp} === false))")]
         end
       end
 


vais:
@jgaskins a little off-topic, but Ruby's `benchmark` lib [is available](https://github.com/opal/opal/blob/master/stdlib/benchmark.rb) in Opal as well, so you can do `require 'benchmark'` and use it same as you would in regular Ruby - no need to roll your own :smile:

vais:
Oh, scratch that, you are going for `ips`-style benchmark, in which case nvm :sweat_smile:

meh:
That issue was raised from me, because I had hit it, but I honestly do not remember what the precise reason was.

A typical example would be something like `foo.tap {|v| if v; ...; end }` where `foo` is a function that could return an object or a boolean, then `false` would be truthy in that case.

It is very synthetic, but exactly because it could happen randomly, it would be very hard to debug and work-around.

Another solution is to make `Boolean` inherit from `BasicObject` and implement a `method_missing` that calls back into `Kernel` and checks if the returned value is a `Boolean` and returns the `.valueOf` on it.

meh:
Actually that wouldn't fix the case where someone implements methods on `Object`, `Kernel` or `Boolean` which return `self`.

But I'm getting a good idea, just need a little time.

meh:
@jgaskins while I figure this one out, can you benchmark with `.valueOf() !== false` to check?

If that works and keeps the performance it will be a lot less hacky.

jgaskins:
@vais Yeah, I've thought about adding `ips` to the Opal stdlib, I just haven't gotten around to it yet. :-)

@meh All good, it's not a bottleneck or anything. I just noticed it in the compiled JS and wanted to see if we could speed it up. Then I saw the failing test and figured there was a reason for it, but I just didn't know what it was.

jgaskins:
Sure, I'll check it against that when I get home today.

ryanstout:
If everyone votes to break compatibility, can it be done in a compiler flag? :-)

meh:
@ryanstout we won't break compatibility, just find a way to make it faster but keep it working.

ryanstout:
@meh even better :-)

Ajedi32:
I think the optimal strategy in all but the most extreme cases is to shoot for maximum compatibility, then add optional compiler flags which developers can use to increase performance at the cost of minor differences in behavior if they so choose. That's in keeping with the principle of least surprise, and makes debugging problems caused by compiler optimizations easier (since you can just compile without the optimizations to see if they're responsible for the bug). Any breaking optimization enabled by default should be prominently documented. (As is the case with symbols, etc in the docs now: http://opalrb.org/docs/compiled_ruby/)

I realize none of that applies in this case since it seems you're exploring optimizations that don't break anything, but I just thought I'd voice my opinion here for future reference.

Anyway great work here, keep it up! :+1:

meh:
@jgaskins any news about checking what effect `.valueOf()` has on the performance?

jgaskins:
@meh Two months later, I finally decided to follow up on this. With everything updated to the latest master:

```
current master:
if true: 18083365
if truthy value: 17269260
if false: 16704902
if nil: 19516664

this PR after rebasing onto master:
if true: 19502561
if truthy value: 20473591
if false: 18097225
if nil: 18476906

same as above using valueOf:
if true: 17663245
if truthy value: 17679874
if false: 15380077
if nil: 18528538
```

jgaskins:
If a couple numbers are close together between two separate runs, don't put too much stock in the difference. They varied by up to 1M iterations per run. My kingdom for `GC.disable` in JS.

elia:
An alternative could be to lengthen the check a bit giving priority to exact matches:

```
// IF
($a = foo) !== nil && (!$a.$$is_boolean || $a == true)
// vs.
($a = foo) === true || ($a !== nil && !$a.$$is_boolean) || $a == true)

// UNLESS
($a = foo) !== nil && (!$a.$$is_boolean || $a == true)
// vs.
($a = foo) === false || $a === nil || ($a.$$is_boolean && $a == false)
```

vais:
:+1: This and https://github.com/opal/opal/pull/867 kinda go together, no?

hmdne:
Please see #2293 for a continuation of this patch.

Issue #1019, opened 2015-07-22
JS Library Integration (Example?)


ashes999:
Hi,

Other than jQuery, which has its own `opal-jquery` gem, how do I go about integrating with other, pure Javascript libraries, with Opal?

For example, if I wanted to use CraftyJS, I want to call JS methods like `Crafty.c(...)` or `Crafty.e(...)`. It's not clear from the docs or the existing issues how I would go about doing that, or if I need to create wrappers or something else before this would work.

Is there an example I can look at where someone does this, even with just a simple JS function or class?

gabrielrios:
You can call JS from Ruby using backticks(`) http://opalrb.org/docs/compiled_ruby/#javascript-from-ruby

Or you can use the native module for simple wrapping http://opalrb.org/docs/compiled_ruby/#native-module

ashes999:
@gabrielrios since there are two options, can you suggest which to use, or what the differences are?

Is it sensible (or is it wasteful) to create Ruby objects with methods that just wrap backtick-calls?

Ajedi32:
@ashes999 That's what Native does. The way I see it, there are three levels of integration possible here:

1. Backticks
  - Quick and dirty.
  - Necessary in some situations (especially situations sensitive to performance).
  - A bit like writing inline assembly code in C.
2. `Native`
  - A quick way to wrap a JavaScript object and use it in Ruby.
  - Good for typical usage, or when a Ruby wrapper library isn't available.
3. Full wrapper libraries
  - Provide a Ruby-style interface to a JavaScript library
  - Preferable when available
  - If you're using a specific library a lot, it might be worth it to write your own wrapper library

gabrielrios:
I think using Native is better, as your code looks better that way.
This is exactly what libraries like opal-jquery does. 

So if you're using things on just one place you might want to just use Native module directly, but wrapping on ruby objects is a better long-term solution

ashes999:
If this can be documented somewhere, that would be fantastic, as it clearly identifies the trade-offs.

@Ajedi32 can you explain how backticks is a better option for performance-sensitive code? Roughly what kind of performance hit are we taking when we use native or wrapper libraries? (Also, please clarify if wrapper libraries means something other than what I described -- Ruby classes with methods/properties that use backticks internally.)

Ajedi32:
[...]

PR #1098, opened 2015-09-10
Clarify trade-off in calling JS, see #1019

ashes999:
- Added "wrapping javascript libraries" section with three options discussed in #1019, along with trade-offs.

### DIFF
--- a/docs/compiled_ruby.md
+++ b/docs/compiled_ruby.md
@@ -314,11 +314,11 @@ That’s all for now, bye!
 window.close!
 ```
 
-## Wrapping Javascript Libraries
+### Wrapping Javascript Libraries
 
 If you want to integrate a Javascript library with Opal, so that you can make Ruby calls, you can choose one of the following options:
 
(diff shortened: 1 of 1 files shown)


ashes999:
Code review please @elia

meh:
There's also another option, using the `.JS.` magic.

ashes999:
@meh I can include it in my change, if you can point me to an existing document that mentions/explains how to use it, and if you can articulate the trade-off (relative to the other three options mentioned).

Otherwise, I don't have enough experience with it to comment on it :)

meh:
I don't know where they're documented, the trade-off is they act like backticks, so they deal with native values without wrapping, while `Native` does automagical wrapping of native things when calling.

It's like `$document.JS.alert("whatever")` instead of ``` `document.alert("whatever")` ```.

elia:
There's https://github.com/opal/opalrb.org/pull/22 which I need to merge and port to `docs/`

Will review in a few hours (on the way to home)

Ajedi32:
Yeah, as far as I know, `.JS` is just syntactic sugar for backticks that you can use in a limited subset of circumstances.

PR #1099, opened 2015-09-10
Copy singleton methods on object cloning.

iliabylich:
1. I've tried to extract methods `dup` and `clone` from `nil`, `true` and `false` to separated modules like `Opal::Unclonable` and `Opal::Undupable` but it raises an error during compilation (in method `const_get`, `Object Undumpable has no method '$split'`). Any advice with it?
2. `Symbol#clone` should also raise an error, but as I see, Opal doesn't support it. Just wondering if you have any plans to implement it. If not - should I move related rubyspec to `/unsupported`?
3. `Array` and `String` classes have its own implementation of `clone` and `dup`. The only difference is that they don't call `allocate`. Was it made intentionally for optimization?
4. One of rubyspecs related to `String#clone` [calls](https://github.com/ruby/rubyspec/blob/master/core/string/clone_spec.rb#L53) `String#[]=`. I assume this rubyspec should be moved to unsupported group, because Opal doesn't support string modifications.


meh:
Symbols will never be supported.

meh:
Also I'd rather not use modules for this, it complicates things.

meh:
And yes, move them to unsupported.

iliabylich:
@meh Done

meh:
Thanks :panda_face:

PR #1101, opened 2015-09-10
Add two failing specs to /bugs.

iliabylich:
The first spec which is related to lambdas is failing because of its dynamic name. Sometimes it's called
```
Kernel.lambda returns from the kind_of? itself, not the creation site of the kind_of?
```
sometimes method name is different. When you run only one spec file containing this test, the method name in the message appears to be blank. This PR should be merged after [this pull](https://github.com/ruby/rubyspec/pull/129) is merged. And anyway, the spec if [failing](http://opalrb.org/try/?code:%24called%20%3D%20false%0A%0Adef%20test%0A%20%20lambda%20%7B%20return%20%7D.call%0A%20%20%24called%20%3D%20true%0Aend%0A%0Atest%0Aputs%20%24called%0A%0A%23%20----%0A%0A%24called%20%3D%20false%0A%0Adef%20test%0A%20%20proc%20%7B%20return%20%7D.call%0A%20%20%24called%20%3D%20true%0Aend%0A%0Atest%0Aputs%20%24called)

The second one is a bug in Opal. [Here](http://opalrb.org/try/?code:module%20M%0A%20%20eval%20%27def%20some_method%3B%20end%27%0Aend%0A%0Aputs%20Object.new.respond_to%3F(%3Asome_method)) is a quick demo. It seems that when Opal evaluates some code in context of class/module, the execution actually happens on `Kernel` and method becomes global. [This](https://github.com/ruby/rubyspec/blob/master/core/module/module_function_spec.rb#L232) is the test that generates a global method `test1`. And [this](https://github.com/ruby/rubyspec/blob/master/core/module/attr_writer_spec.rb#L11) is the test that fails because of the bug. So, one spec is actually broken, but it shows as green, the other spec becomes red because of the side effect.

### DIFF
--- a/spec/filters/bugs/basicobject.rb
+++ b/spec/filters/bugs/basicobject.rb
@@ -25,4 +25,9 @@ opal_filter "BasicObject" do
   fails "BasicObject#instance_exec raises a TypeError when defining methods on an immediate"
   fails "BasicObject#instance_exec raises a TypeError when defining methods on numerics"
   fails "BasicObject#instance_exec sets class variables in the receiver"
+  fails "BasicObject#__send__ has a negative arity"
+  fails "Kernel#send has a negative arity"
+  fails "Kernel#public_send has a negative arity"
+  fails "Module#prepend does not affect the superclass"
+  fails "Module#singleton_class? returns true for singleton classes"
 end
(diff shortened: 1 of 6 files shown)


iliabylich:
I've updated rubyspec to the current master, there are some new specs, could someone review their placement in bugs/unsupported?
@meh you mentioned that you also have two tests failing locally, could you check this PR please?

meh:
They were the same that were failing for you. The placements look fine for now. Good job tracking these down.

PR #1104, opened 2015-09-11
Improve PP stdlib defaults

wied03:
* Use $stdout by default, which is closer to what MRI does (they use $<, but that would require more extensive gvars changes)
* Remove the attempt at pretty printing by console.log'ing the object

Tests for this submitted as rubyspec PR https://github.com/ruby/rubyspec/pull/130

Once that makes its way into Opal's submodule, will need to add this to spec/rubyspecs

```
rubyspec/library/pp
```

### DIFF
--- a/stdlib/pp.rb
+++ b/stdlib/pp.rb
@@ -19,10 +19,8 @@ class PP
         p(*args)
       end
     else
-      def pp(obj, out=`console`, width=79)
-        if `#{out} === console`
-          `console.log(obj)`
-        elsif String === out
+      def pp(obj, out=$stdout, width=79)
+        if String === out
           out + obj.inspect + "\n"
         else
           out << obj.inspect + "\n"


elia:
@wied03 having `pp` forward everything to `console.log` is very useful in the browser as it allows to inspect object. I'm ok with removing this behavior only if we have a replacement. But of course the more compliant the better :smile:.

wied03:
I get that, but how often do people working w/ Opal use 'pp' for that as opposed to console.dir or something else? I use the latter but that doesn't necessarily mean anything.

meh:
I'm okay with this, we should probably merge the opal-browser `console.rb` in the stdlib instead.

wied03:
@meh - Merge instead or do both? pp (and RSpec's usage of pp) expect to be able to reassign $stdout and have pp use that by default, so that's why I'm pushing this.

meh:
@wied03 yeah sorry, I meant instead of having `console.log` in `pp` as is currently.

elia:
That's good for me too (merging "console. rb"). I use pp a lot for debugging but no strings attached, at least if I have a replacement. 

@meh I wonder how the usage of browser would change if it was part of the stdlib… did we ever discussed that with @adambeynon? Wdyt?

meh:
@elia I don't remember discussions, but I'd be against it, it's better outside the main tree, especially since it targets browsers and not any JS engine, it should just be publicized a little more.

wied03:
@meh @elia - Where does this leave us then? If the aim is to have a MRI equivalent pp implementation, then I'd say it's going to diverge even further from this console.log route as things like width, etc. are implemented.

If this PR goes through, then you can get the same effect as today by doing:

```ruby
PP.pp someobject, `console`
```

That might be a little verbose, so maybe defining a convenience method of 'ppo' on the PP class that uses the traditional behavior might work?

@elia - Do you usually use this by calling "PP.pp some_object" ?

elia:
@wied03 no, I just `pp some_object` after a `require 'pp'` in the main file usually.
@meh can you move the browser's `console.rb` to the stdlib, let me know if you can't and I'll take care

Merging.

@wied03 thanks for the patience :wink:, can you add an entry in the changelog too?

meh:
@elia done.

elia:
@meh thanks :+1:

wied03:
@elia - sorry, forgot to do the changelog

elia:
@wied03 no problem, really appreciate the work you're doing 😀

PR #1105, opened 2015-09-16
Raise/Fail Exception Rubyspec Compliance

wied03:
This passes all of the raise/fail rubyspecs (except private methods). It comes at a cost though. In order to reset $! after every rescue like MRI does, I had to alter the compiler to make it a stack (JS array).

I realize that may be too much since it adds a try {} finally {} to **every** rescue block. I thought I'd put it out here and see though.

If this can't be done cleanly using my approach or another approach, then I think Opal should explicitly say:
* $! is not a supported global variable
* As a result, you cannot re-raise exceptions by calling 'raise' without parameters

If this **is** acceptable, then:
* I could also use some feedback on how I did the compiler tweaks. I didn't spend much time trying to do it 100% right since the idea might get squashed anyways.
* I'll squash the commits

### DIFF
--- a/lib/opal/nodes/module.rb
+++ b/lib/opal/nodes/module.rb
@@ -31,7 +31,8 @@ module Opal
 
       def name_and_base
         if cid.type == :const
-          [cid[1].to_s, 'self']
+          # Using $scope.base, not self in case a class is created inside an anonymous class or block
+          [cid[1].to_s, '$scope.base']
         elsif cid.type == :colon2
           [cid[2].to_s, expr(cid[1])]
         elsif cid.type == :colon3


opal/corelib/kernel.rb
```diff
@@ -1006,21 +1006,38 @@ def warn(*strs)
...
       }
-      else if (exception.$$is_class) {
+      // using respond_to? and not an undefined check to avoid method_missing matching as true
+      else if (exception.$$is_class && #{exception.respond_to?(:exception)}) {
+        exception = #{exception.exception string};
+      }
+      else if (#{exception.kind_of?(Exception)}) {
+        // exception is fine
+      }
+      else if (exception.$$is_class && (exception === #{Exception} || #{exception < Exception})) {
```
iliabylich: I think `exception < Exception` comparison [is not](http://opalrb.org/try/?code:class%20ApiException%20%3C%20Exception%0Aend%0A%0Aclass%20Dog%0Aend%0A%0Aputs%20ApiException%20%3C%20Exception%0Aputs%20Dog%20%3C%20Exception%0Aputs%20Exception%20%3C%20Exception%0A) necessary

wied03: @iliabylich - opal's try site is still using 0.8.0 beta1 which was not compliant w/ RubySpecs for the < operator. That issue was fixed in https://github.com/opal/opal/pull/1105/files#r40012949.

If you run a very similar sample to your Opal try link in IRB, you'll get this output:
```ruby
true

NilClass
false
```

code:
```ruby
class ApiException < Exception
end

class Dog
end

puts ApiException < Exception
puts Dog < Exception
puts (Dog < Exception).class
puts Exception < Exception

```

wied03:
@adambeynon @meh @elia @vais - Any thoughts here?

wied03:
this PR was closed because I moved that work to a branch (like I should have done from the beginning), 1117 is a new PR with the same changes

PR #1106, opened 2015-09-16
Update to latest rubyspecs and include PP ruby spec for 7bd7884

wied03:
Promised rubyspecs for https://github.com/opal/opal/pull/1104


elia:
Thanks!

PR #1107, opened 2015-09-17
Add support for recursive Struct comparison using #== and #eql?.

iliabylich:
The [spec](https://github.com/ruby/rubyspec/blob/master/core/struct/eql_spec.rb#L8) `Struct#eql? returns false if any corresponding elements are not #eql?` just cannot be implemented because of JS limitations. It automatically typecasts 1.0 to 1 (so they become eql).

### DIFF
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -106,6 +106,78 @@ class Struct
     `self.$$data[name] = value`
   end
 
+  def ==(other)
+    return false unless other.instance_of?(self.class)
+
+    %x{
+      var recursed1 = {}, recursed2 = {};
+
+      function _eqeq(struct, other) {
+        var key, a, b;
+
+        recursed1[#{`struct`.__id__}] = true;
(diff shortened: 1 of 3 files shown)


meh:
Put the failures related to `Float == Integer` in `unsupported/float.rb` please.

The dream is one day they'll stop using that in unrelated specs, and that filter file will help.

Also use `Struct === a` instead of `a.is_a?(Struct)` please, it handles things that aren't Opal objects.

iliabylich:
@meh Done

meh:
Just noticed, also use `__id__` instead of `object_id` please.

iliabylich:
@meh Done. Could you explain what is the difference? As I see from the code, `object_id` is an alias to `__id__` for every opal object. Is it an optimization to avoid an extra call in delegation?

meh:
@iliabylich from what I recall it's valid to override `#object_id` in your class, but it isn't to override `__id__`.

PR #1109, opened 2015-09-21
Implement Enumerable#chunk.

iliabylich:
Three specs are `unsupported` - two of them because of symbols, [one](https://github.com/ruby/rubyspec/blob/master/core/enumerable/chunk_spec.rb#L63) because of opal's `String#equal?`


elia:
Great job! 😀

Merged in 3b96fe1a04e55b0e99c291de8507824172bb9178, rebased and added a changelog entry. 
Too bad it took so long! 😖

Issue #1110, opened 2015-09-23
Unable to define a new class inside an anonymous class definition


wied03:
code:

```ruby
Class.new do
  class Howdy
  end
  
  puts "right after, it's #{Howdy}"
end
```

MRI output:
```
right after, it's Howdy
```

Opal master output:
```
/tmp/opal-nodejs-runner-20150923-61-1ftz3za:4198
      throw exception;
            ^
NameError: uninitialized constant Object::Howdy
    at OpalClass.Opal.defn.self [as $new] (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:4412:15)
    at Opal.defn.self [as $raise] (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:4182:31)
    at Opal.defn.self (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:2362:19)
    at Opal.defn.self [as $const_get] (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:2344:19)
    at Opal.get (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:75:24)
    at OpalClass.$a.$$p.TMP_1 (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:18585:54)
    at Opal.defs.TMP_1 [as $new] (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:2751:15)
    at Opal.dynamic_require_severity (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:18585:102)
    at Object.<anonymous> (/tmp/opal-nodejs-runner-20150923-61-1ftz3za:18586:3)
    at Module._compile (module.js:456:26)
```

wied03:
https://github.com/ruby/rubyspec/pull/132

eregon:
See https://github.com/ruby/rubyspec/pull/132#issuecomment-142906014.
Howdy is defined on Object in this case, the fact we are in an anonymous class body does not change constant resolution and the lexical parent is still toplevel (Object).

wied03:
@eregon - Right, but in Opal, this only seems to surface with the anonymous class scenario. This example works properly in both MRI and Opal:
code:
```ruby
stuff = lambda do
  class Howdy
  end

  puts "right after, it's #{Howdy}"
end

stuff.call
```

output:
```
right after, it's Howdy
```

[...]

PR #1114, opened 2015-09-26
Fix classes defined within anonymous classes

wied03:
Attempt to fix https://github.com/opal/opal/issues/1110

Tests for this are in https://github.com/ruby/rubyspec/pull/135

### DIFF
--- a/lib/opal/nodes/module.rb
+++ b/lib/opal/nodes/module.rb
@@ -31,7 +31,8 @@ module Opal
 
       def name_and_base
         if cid.type == :const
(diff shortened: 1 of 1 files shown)


wied03:
@meh - Since I'm not an expert on the compiler, are you sure this is OK? It seemed like the simplest fix to the problem but wanted to make sure

meh:
From what I know that's the correct fix.

eregon:
This looks good, `self` is never used in constant resolution / definition.

PR #1117, opened 2015-09-27
Raise/Fail Exception Rubyspec Compliance

wied03:
(moved work in https://github.com/opal/opal/pull/1105 to another branch):

This passes all of the raise/fail rubyspecs (except private methods). It comes at a cost though. In order to reset $! after every rescue like MRI does, I had to alter the compiler to make it a stack (JS array).

I realize that may be too much since it adds a try {} finally {} to **every** rescue block. I thought I'd put it out here and see though.

If this can't be done cleanly using my approach or another approach, then I think Opal should explicitly say:
* $! is not a supported global variable
* As a result, you cannot re-raise exceptions by calling 'raise' without parameters

If this **is** acceptable, then:
* I could also use some feedback on how I did the compiler tweaks. I didn't spend much time trying to do it 100% right since the idea might get squashed anyways.
* I'll squash the commits

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -995,21 +995,34 @@ module Kernel
 
   def raise(exception = undefined, string = undefined)
     %x{
-      if (exception == null && #$!) {
+      if (exception == null && #$! !== nil) {
         throw #$!;
       }
-
       if (exception == null) {
         exception = #{RuntimeError.new};
       }
       else if (exception.$$is_string) {
         exception = #{RuntimeError.new exception};
       }
-      else if (exception.$$is_class) {
+      // using respond_to? and not an undefined check to avoid method_missing matching as true
+      else if (exception.$$is_class && #{exception.respond_to?(:exception)}) {
+        exception = #{exception.exception string};
+      }
+      else if (#{exception.kind_of?(Exception)}) {
+        // exception is fine
+      }
+      else if (exception.$$is_class && (exception === #{Exception} || #{exception < Exception})) {
         exception = #{exception.new string};
       }
-
-      #$! = exception;
+      else {
+        exception = #{TypeError.new 'exception class/object expected'};
+      }
+      
+      if (#$! !== nil) {
+        Opal.exceptions.push(#$!);
+      }
+            
+      #$! = exception;      
 
       throw exception;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
 
 * Fix issue where passing a block after a parameter and a hash was causing block to not be passed (e.g. `method1 some_param, 'a' => 1, &block`)
 
+* `Kernel#raise` now properly re-raises exceptions (regardless of how many levels deep you are) and works properly if supplied a class that has an exception method.
+
 
 ## 0.8.0 2015-07-16
 
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -115,7 +115,16 @@ module Opal
           push expr(variable), ';'
         end
 
-        line process(rescue_body, @level)
+        # Need to ensure we clear the current exception out after the rescue block ends
+        line "try {"
+        indent do
+          line process(rescue_body, @level)
+        end
+        line "} finally {"
+        indent do
+          line 'Opal.gvars["!"] = Opal.exceptions.length > 0 ? Opal.exceptions.pop() : Opal.nil;'
+        end
+        line "}"
         line "}"
       end
 
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -51,6 +51,9 @@
   // Exit function, this should be replaced by platform specific implementation
   // (See nodejs and phantom for examples)
   Opal.exit = function(status) { if (Opal.gvars.DEBUG) console.log('Exited with status '+status); };
+	
+  // keeps track of exceptions for $!
+  Opal.exceptions = [];
 
   /**
     Get a constant on the given scope. Every class and module in Opal has a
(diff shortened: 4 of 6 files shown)


elia:
_Quick note until I get the time to read the PR thoroughly._

> As a result, you cannot re-raise exceptions by calling 'raise' without parameters

I think this can be supported anyway as AFAIK `raise` will work inside a `rescue` where the current error is known.

wied03:
@elia - Yeah, but you'd have to deliberately do that, right? as in you'd have to rescue Exception => e and then "raise " as opposed to 'raise'.

Is there a more efficient way to still pass these specs than what I've done? It seems like you have to make $! a stack to make this work. I know this inflates code and slows things down, but if you have already have 1 try/catch, I'm not sure how much worse another is. It would help if JS had this concept natively like Ruby, Python, etc. do.

elia:
Another thing to consider is that `try/catch` blocks seem to be not-optimizable as pointed by @jgaskins [in the gitter channel](https://gitter.im/opal/opal?at=560740ebefd97f3e4669a564)

> try blocks can't be optimized, so I'm trying to figure out why it happens and see if there's a way we can work around it. For example, if the entire try block were contained within another function, that function could be optimized and we wouldn't need to worry about it, since the amount of time spent in the try block (in that stack frame, anyway) would be negligible.

wied03:
I say this to try and pick up some knowledge from a business app developer standpoint more than a systems programmer, but reading the info @ https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax, I think their "solution" (going back to C style return error codes) is pretty ugly. Although I suppose for Opal core stuff, the ugly way may be so much more performant it's worth doing.

A lot of this seems to reinforce the age old notion that exceptions aren't for flow control.

That said, outside of the code bloat, I would think this PR would make the already more expensive process of catching just a little more expensive. If they're in the catch block where this is, it's already going to be expensive, right?

For cases where that's unacceptable, it would seem like someone would try and go another route that doesn't involve begin/rescue/try/catch in the first place.

opal/corelib/kernel.rb
```diff
@@ -1006,21 +1006,38 @@ def warn(*strs)
 
   def raise(exception = undefined, string = undefined)
     %x{
-      if (exception == null && #$!) {
+      if (exception == null && #{$! != nil}) {
```
meh: Make this `#$! !== nil`, a little more performant.

opal/corelib/kernel.rb
```diff
@@ -1006,21 +1006,38 @@ def warn(*strs)
...
+      }
+      else if (exception.$$is_class && (exception === #{Exception} || #{exception < Exception})) {
         exception = #{exception.new string};
       }
+      else {
+        exception = #{TypeError.new 'exception class/object expected'};
+      }
 
-      #$! = exception;
+      if ($gvars["!stack"] === undefined) {
```
meh: Instead of having `$gvars["!stack"]` I'd add a special variable in `Opal`, `Opal.exceptions` or something.

meh:
Can someone figure out how to benchmark this change?

opal/corelib/kernel.rb
wied03: Yeah, I realize that was kind of a hack

wied03:
I'm no benchmarking guru, but I did this quick a dirty test.

```ruby
require 'benchmark'

def do_test
  begin
    raise 'some error'
  rescue
    a = 1
  end
end

n = 500000
Benchmark.bm do |x|
  x.report { n.times do; do_test; end }
end
```

Master:
```
       user     system      total        real
   7.380000   7.380000  29.520000 (  7.381000)
```

With this PR:
```
       user     system      total        real
   7.787000   7.787000  31.148001 (  7.791000)
```

That's about a 6% speed decrease on the real elapsed time

jgaskins:
Using an IPS benchmark, my results show about a [4-6% performance hit for the unwrapped case](https://gist.github.com/jgaskins/0a31974208a2d288b196)

A few notes:
- I realized after swapping between the PR and master that I accidentally left the duration at 1 sec. I usually do 5, but running it with 5 seconds didn't move the benchmarks any meaningful amount. Also, I forgot to check Firefox for @meh. :-)
- Actual difference may be a fair bit higher. My benchmark code for this uses blocks to try to reduce the amount of copy/paste I normally have to do. Unfortunately, using blocks adds overhead to the benchmark, which absorbs some of the performance ratio on blocks that execute very fast. My benchmarks usually use copy/pasted `while` loops for that reason. I tried to mitigate that by doing a lot of work inside the `begin` block before the `raise`, but I pointed it out because it's still a consideration.

wied03:
@jgaskins - Sorry for the miss, but I noticed 1 more spot (d30796f) that an optimization could be applied based on meh's feedback, might want to re-run your benchmark.

wied03:
@jgaskins - Do you have any thoughts about how tolerable (or not) this might be?

jgaskins:
@wied03 Gah, sorry, I forgot about this. I'll set a reminder to do it when I get home today. :-)

jgaskins:
@wied03 I updated the gist. It's now close to what it is on `master`. Not quite, but much, much closer.

The odd thing is that it improved the performance of wrapping the potentially exception-raising code within a closure in Safari (even over what's on `master`), but brought it down on Chrome.

wied03:
@meh - What are your thoughts? Mergeable? Personally I think the argument against merging should be based more on a) whether it hurts method optimization or b) code bloat. If you're inside a catch block, you've already gone down the path of (slower) exception code.

I do not think it would be good though if by adding a begin/rescue to your code, you prevent a bunch of optimization. I think that's a separate issue though. That would happen without this PR. The only thing this PR would affect is if a nested try/catch impacts optimization more than the outside try/catch.

wied03:
If https://github.com/opal/opal/pull/1151 is accepted, some of the raise if checks can be simplified

meh:
I'm fine with this, just squash it all and I'll merge.

wied03:
@meh - squashed

Issue #1090, opened 2015-09-02
&& operator does not evaluate the same as on MRI


wied03:
Opal doesn't handle && the same way when the LHS of the operator is nil (returning false rather than nil).

Code:

```ruby
var = nil
result1 = var && var[:junk]
var2 = {junk: 'foo'}
result2 = var2 && var2[:foo]
result3 = var2 && var2[:junk]
[result1, result2, result3].each do |res|
        puts "output is #{res}, type #{res.class}"
end
```

Output:

Opal:
```
result1 is false, type Boolean
result2 is , type NilClass
result3 is foo, type String
```

MRI:
```
result1 is , type NilClass
result2 is , type NilClass
result3 is foo, type String
```

wied03:
Compiler difference:

```javascript
result = (($a = var$ !== false && var$ !== nil) ? var$['$[]']("junk") : $a);
($a = result2 = var$, $a !== false && $a !== nil ?var$['$[]']("junk") : $a);
```

wied03:
@eregon - I can't find a rubyspec that covers this. language/and_spec.rb seems like the logical place to add it. agree?

The fix to lib/opal/nodes/logic.rb looks something like this around line 128:

```
         with_temp do |tmp|
           if truthy_opt = js_truthy_optimize(lhs)
             push "((#{tmp} = ", truthy_opt
             push ") ? "
             push expr(rhs)
-            push " : #{tmp})"
+            push " : ", expr(lhs), ")"
```

eregon:
@wied03 Yes, absolutely!
https://github.com/ruby/rubyspec/blob/5cab2a8cfd2793646ed68f1d0ef5a9c8e0bc59ee/language/and_spec.rb#L11-L14 should test it but it's rather complicated so it would be great to have an example showing just this behavior (for `and` as well).

wied03:
@eregon - I thought that one would test it, but opal handles a case with 3 booleans differently than 2.

wied03:
https://github.com/ruby/rubyspec/pull/136, will add the fix in once the rubyspec PR runs its course

PR #1119, opened 2015-09-28
Make return value from local_var &&/and something consistent

wied03:
Fixes https://github.com/opal/opal/issues/1090 with a compiler fix and update to latest rubyspecs


meh:
:+1: great work (and courage to touch the compiler).

Issue #1027, opened 2015-07-24
Missing API documentation for Native


Ajedi32:
Opal is currently missing API documentation for the `Native` class. This should be added.

----

This issue was created in response to the following conversation in issue #1024:

@ajedi32:
> My guess is you'd have to use Hash#to_native or #to_n or something? That doesn't seem to be documented anywhere though either, is it?

@ajedi32:
>Maybe some RDoc comments could be added to https://github.com/opal/opal/blob/master/stdlib/native.rb#L489-L520? Then we'd just need to use them to generate HTML and host them them somewhere. Better docs for some of the Opal-specific classes and methods / method implementations would be great.

@elia:
> yes, actually there's already a place for that kind of docs at http://opalrb.org/docs/ under "API documentation". It's unfortunate that for some reason Native isn't covered and we need to fix that.
>
> As of the array to_n should call to_n on all its elements, if it doesn't let me know and we'll ask @meh to fix it :blush:

ashes999:
To clarify the last comment cited by @elia above, I have code like:

```
hash = { 
  :images => ['background.png', 'player.png'],
  :audio => ['beep.ogg']
}
```

`hash.to_n` gives me back a hash object that doesn't include the Opal bridge properties (around ~300 properties, mostly starting with `$`).  However, the two arrays I get back still do have those same Opal properties.

This caused problems in #1024, because CraftyJS iterates via `for x in hash`. (They support a number of different cases for their preloader, including sub-hashes, but anyway, they worked around the problem on their side.)

Please fix this. It's simple enough to diagnose with something like this (using the hash above):

```
hash = ...
hash = hash.to_n

`
var count = 0;
for (x in hash['images']) {
  count++;
}
console.log("Found " + count + " properties in the images array.");
`
```

You should only get a count of `2`, not `300`.

Ajedi32:
This particular issue is about adding API *documentation* for the `Native` class. Fixing any incorrect behavior is a related but separate issue, IMO.

ashes999:
Okay, I've created #1028 .

elia:
Ok, so we now have API docs divided per version and per group (lib,corelib,stdlib)

http://opalrb.org/docs/api/v0.8.0/stdlib/index.html

Still lacks the actual documentation comments, I'd be happy to help out anyone who wants to takle this with a PR :smile: 

cc @meh

elia:
closed by 74f2a8a7c7 (docs: Add some API documentation for native.rb)

[...]

PR #1120, opened 2015-09-29
Make JS object->Hash conversion traverse arrays

### LINKED ISSUE #1095: Hash.new with JS object w/ arrays as values doesn't recurse into the array
I noticed it because we're using `gon` to transfer state to our client apps. When I ran this on the opal side to get my initial state:

```ruby
Hash.new(`gon`)
```

The top-level values were good, and so were any nested hashes, except ones contained within arrays. For example ([try link](http://opalrb.org/try/?code:gon%20%3D%20%25x%7B%0A%20%20%7B%0A%20%20%20%20a%3A%20%7B%20b%3A%202%20%7D%2C%0A%20%20%20%20c%3A%20%5B%7B%20d%3A%201%20%7D%5D%2C%0A%20%20%7D%0A%7D%0A%0Agon_hash%20%3D%20Hash.new(gon)%0Aa%20%3D%20gon_hash%5B%3Aa%5D%0Ac%20%3D%20gon_hash%5B%3Ac%5D%0A%0Ap%20a%0Ap%20c):

```ruby
gon = %x{
  {
    a: { b: 2 },
    c: [{ d: 1 }],
  }
}

gon_hash = Hash.new(gon)
a = gon_hash[:a]
c = gon_hash[:c]

p a
p c
```

The output is:

```
{"b"=>2}
TypeError: undefined is not a function (evaluating 'i.$__id__()')
$inspect@http://opalrb.org/javascripts/try.js:7:9276
i@http://opalrb.org/javascripts/try.js:5:49029
yield1@http://opalrb.org/javascripts/try.js:4:49598
g@http://opalrb.org/javascripts/try.js:7:4626
http://opalrb.org/javascripts/try.js:5:49052

eval code
eval@[native code]
$eval_code@http://opalrb.org/javascripts/try.js:20:52222
$run_code@http://opalrb.org/javascripts/try.js:20:52122
$@http://opalrb.org/javascripts/try.js:20:77095
o@http://opalrb.org/javascripts/try.js:20:2773
dispatch@http://opalrb.org/javascripts/application.js:3:4887
a@http://opalrb.org/javascripts/application.js:3:780
```

jgaskins:
Fixes #1095

### DIFF
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -470,8 +470,16 @@ class Hash
 
           if (value && value.constructor === Object) {
             smap[key] = #{Hash.new(`value`)};
-          }
-          else {
+          } else if (value && value.constructor === Array) {
+            value = value.map(function(item) {
+              if (item && item.constructor === Object) {
+                return #{Hash.new(`item`)};
+              }
+
+              return item;
+            });
+            smap[key] = value
+          } else {
             smap[key] = #{Native(`value`)};
           }
 
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -470,7 +470,7 @@ class Hash
 
           if (value && value.constructor === Object) {
             smap[key] = #{Hash.new(`value`)};
-          } else if (value && value.constructor === Array) {
+          } else if (value && value.$$is_array) {
             value = value.map(function(item) {
               if (item && item.constructor === Object) {
                 return #{Hash.new(`item`)};
--- a/spec/opal/stdlib/native/hash_spec.rb
+++ b/spec/opal/stdlib/native/hash_spec.rb
@@ -1,0 +1,36 @@
+require 'native'
+
+describe Hash do
+  it 'turns a native JS object into a hash' do
+    obj = %x{
+      {
+        a: 1,
+        b: "two",
+        c: {
+          d: 1,
+        },
+        e: [
+          {
+            f: 'g',
+          },
+        ],
+      }
+    }
+
+    h = Hash.new(obj)
+    expected_hash = {
+      a: 1,
+      b: "two",
+      c: {
+        d: 1,
+      },
+      e: [
+        {
+          f: 'g',
+        },
+      ],
+    }
+
+    expect(h).to eq(expected_hash)
+  end
+end


jgaskins:
Squashing the commits segfaulted the VM. Sweet!

elia:
Restarted the job, I ["gisted" the old log](https://gist.github.com/elia/7c159314ca5e096a30e2) in case anyone wants to submit the crash to the CRuby team :frog:

stdlib/native.rb
```diff
@@ -470,8 +470,16 @@ def initialize(defaults = undefined, &block)
 
           if (value && value.constructor === Object) {
             smap[key] = #{Hash.new(`value`)};
-          }
-          else {
+          } else if (value && value.constructor === Array) {
```
elia: can this be `(value && value.$$is_array)` instead? 

(see https://github.com/opal/opal/blob/master/opal/corelib/array.rb#L8)

elia:
Super happy about this fix, I feel though that it should be extended to arrays in general (vs. just inside `Hash.new`), maybe in `Array.new`. 

Could also make sense to define some kind of general api for classes that, something like `Hash.from_native` and `Array.from_native` which can still be used inside `.new` (eg. for compatibility) but would allow user defined classes to hook in:

```js
if (value && (klass = value.$$class) && klass.$from_native) {
  smap[key] = klass.$from_native(value);
}
```
cc @meh

vais:
@elia :+1: for a separate API for this instead of `native` monkey-patching core classes - I got bitten by this multiple times while working on Opal...

stdlib/native.rb
jgaskins: Oh, right, forgot we had those properties to work with. :-)

jgaskins:
I found myself wanting the same thing while writing this. Not being able to do this conversion recursively easily felt a little awkward, and I thought about adding something similar to `from_native`, but I wanted to make a small change before making larger structural changes. :+1:

meh:
A good idea might to be to expand the `Native(..)` functionality and make it use `T.from_native`.

elia:
@vais @jgaskins @meh here's what I'd like to do:

1. merge this PR (after the tiny `.$$is_array` fix)
2. add api docs for native.rb (#1027)
3. start the release cycle for 0.9
2. keep the native.rb changes on master (for the next release)

how does it sound?

jgaskins:
@elia :+1: I'll push the $$is_array thing after I get home.

vais:
Sounds great, @elia!

PR #1123, opened 2015-10-02
Struct Hash Fixes

wied03:
…ime.js so that Number#hash can stay how it is and not make a Ruby method call)

* Change kernel to hash the name, class ID, and object ID using the string hash method
* Since string's hash is not itself anymore, changed pathname to hash its path and keep its specs passing
* Rather than returning a string here, hash the overall result to be consistent with other changes
* Fix struct hash method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 
 * `Opal::Processor.load_asset_code(sprockets, name)` has been deprecated in favor of `Opal::Sprockets.load_asset(name, sprockets)`.
 
+* `Struct#hash` now works properly based on struct contents
+
 
 ## 0.8.0 2015-07-16
 
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -73,6 +73,10 @@ class Struct
   def members
     self.class.members
   end
+  
+  def hash
+    Hash.new(`self.$$data`).hash
+  end
 
   def [](name)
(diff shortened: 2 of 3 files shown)


meh:
I'm not entirely sure I'm fine with this. Performance is going to suffer for no really really really good reason.

wied03:
@elia - Build should now pass except some node stuff which broke on master before this PR
@meh - Which part concerns you most re: performance, the struct changes or the kernel/string/number changes?

wied03:
@meh - I ask because the struct behavior is what's more important (to me) and it can be fixed without the other changes. As is, structs break easily in Hash'es.

I just made the other changes to try and be consistent with MRI (and generate Fixnum hashes).

meh:
@wied03 the `#hash` changes, the `Struct` changes are fine.

If I recall correctly we decided the discrepancy with hashing from MRI was an acceptable compromise, so I'd rather keep it as is.

wied03:
@meh - got it, changes made

elia:
@wied03 can you fake a push in order to restart the ci process 
(amending the commit by just adding a space to the message should do the trick)

wied03:
@elia - Thanks, I believe Travis picked it up.

elia:
Good, yeah, I restarted the failing test but was unsure about the merging strategy of travis :)

elia:
Thanks! (Sorry for the delay)

PR #1128, opened 2015-10-03
Make dynamic symbols usable wherever symbols are

### LINKED ISSUE #1127: undef doesn't take a symbolized string
I was trying to use the `hamster` gem in a Clearwater app. It uses `undef` in a few of its classes to reduce API size, but it passes a symbol wrapped in quotes: `undef :"<=>"`. I checked and the fact that the string contains non-alphanumeric characters doesn't matter. This code also doesn't compile:

```ruby
class Foo
  undef :"object_id"
end
```

The thing is, `:"<=>"` is valid Ruby according to the compiler. It only breaks when passed to `undef` and other directives like `alias`.

If you've got ideas as to where this would be in the compiler, I can go ahead and try to fix it.

jgaskins:
Who has two thumbs and has no idea what he's doing with parsers? THIS GUY!

On the bright side, this PR fixes #1127. I don't know if we should enable any RubySpecs for it, though. Could use some guidance there from someone more knowledgeable about RubySpec than I am.


elia:
Nice! 👍🏼👍🏼

I checked the grammar and from what I can understand seems the right way, would be cool to add a test for this fix, to avoid regression.

elia:
@jgaskins I started writing a spec for it but something is broken, it probably needs some more work, here's a simple test that can be useful while developing:

```
$ bundle exec bin/opal -Oce 'class A; undef :"object_#{:id}"; end;' | grep "object" 
```

I uploaded the WIP test I added in this branch:

https://github.com/opal/opal/tree/elia/issue-1128-spec-dsym

the updated rake command is `rake mspec_opal_node`

jgaskins:
@elia It even outputs the array with the s-expressions into the JS code?

jgaskins:
Looks like I was chasing the wrong problem. `undef` doesn't work with strings, either:

```
parse error on value "\"" (tSTRING_BEG) :application:100
Source: application:100:8
 undef "object_#{:id}"
~~~~~~~^
```

jgaskins:
Nevermind. MRI doesn't allow that, either.

jgaskins:
I think I found [the problem](https://github.com/opal/opal/blob/acda5649755287b632e8dff7eddcb6d4d71ac0df/lib/opal/nodes/definitions.rb#L22). In that line, `mid` is `[(:dsym, "object_", (:evstr, (:sym, :id)))]`, so `mid[1]` is just `"object_"`.

What we need is for it to compile all of that node's children and, although I'm quickly learning more than I ever knew about compilers, I don't know how to do that yet.

jgaskins:
Notice it's the same in `AliasNode`.

elia:
cc @alexdowad

jgaskins:
@elia Played around with this a bit more tonight and managed to make your spec pass (included in this PR). It seemed to be [expecting a plain symbol](https://github.com/opal/opal/blob/386ae877b80f74f71d9f70705a06c771e210bd32/lib/opal/nodes/definitions.rb#L19-L34). I'm not sure if I went about it the right way, but what I was aiming for was "give me the output for this sexp" instead of assuming its type. I couldn't _quite_ get that, but I got reasonably close. Let me know if there's a better way to handle this.

elia:
Just did a check on master and this is still an issue:

```
opal:master ⤑ bundle exec bin/opal -cEO -e 'class A; undef :"object_#{:id}"; end;' 
(function(Opal) {
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $klass = Opal.klass;

  return (function($base, $super) {
    function $A(){};
    var self = $A = $klass($base, $super, 'A', $A);

    var def = self.$$proto, $scope = self.$$scope;

    return Opal.udef(self, '$(str "object_")');
  })($scope.base, null)
})(Opal);
```

→ `    return Opal.udef(self, '$(str "object_")');`

PR #1129, opened 2015-10-06
Regexp multiline white space, allocated fix, === fix, options optimization

wied03:
Could potentially store a private source in the initializer with the sub of . to [\S\s] already done so that each match doesn't have to do it, but thought I'd defer that until caching comes up.


wied03:
The original PR exposed an issue with calling #options on an empty regex (//). Since that is implemented in JS as /(?:)/, calling options on it was raising a TypeError. I restructured the 'allocate' approach to be more explicit so that #options (and #match) work correctly with an empty expression.

RubySpec PR for this is https://github.com/ruby/rubyspec/pull/137

wied03:
@meh or @elia. Any comments here? Would this be more palatable if the code that alters the actual expression to emulate ruby multiline behavior were moved to another PR?

meh:
Is `un_initialized` an instance variable on purpose?

Whether yes or no, could you rename it to `uninitialized`? (unless the `un_` there has another meaning)

wied03:
@meh - Yes, it's an ivar on purpose. It's kind of awkward but I didn't know any other way to create an uninitialized regexp on JS. I fixed the variable name nonetheless.

opal/corelib/regexp.rb
```diff
@@ -101,8 +107,8 @@ def ==(other)
     `other.constructor == RegExp && self.toString() === other.toString()`
   end
 
-  def ===(string)
-    `#{match(string)} !== nil`
+  def ===(string)
```
meh: Can you also remove the leftover spaces at the end please?

wied03:
@meh - done

meh:
Failure is my fault, just rebase again.

wied03:
closing/reopening PR works too :)

Issue #1131, opened 2015-10-07
Passing on blocks arguments something,hash, &block causes compiler to cut off block


wied03:
```ruby
def method1(stuff, &block)
  method2 stuff,:junk => ' nope', &block
end

def method2(stuff, options=, &block)
  block[stuff+options[:junk]]
end

method1 '123' do |result|
  puts "got #{result}"
end
```

When the compiler executes `method2 stuff,:junk => ' nope', &block`, it will cut off the &block parameter and the block never gets passed to method2.

Here is what the compiler ultimately generates for that call:
```
return self.$method2(stuff, $hash2(["junk"], {"junk": " nope"}));
```

This only happens when you have an argument (stuff in this case) followed by another argument that is a hash, then the block. If the passed argument is not a hash, then this bug doesn't happen.

I tried to identify the issue but am not compiler savvy enough yet.

Here is some of my debug output:
```
next token [:kDEF, ["def", [1, 0]]]
next token [:tIDENTIFIER, ["method1", [1, 4]]]
next token [:tLPAREN2, ["(", [1, 11]]]
next token [:tIDENTIFIER, ["stuff", [1, 12]]]
next token [:tCOMMA, [",", [1, 17]]]
next token [:tAMPER, ["&", [1, 19]]]
next token [:tIDENTIFIER, ["block", [1, 20]]]
new_args norm [:stuff] opt  rest  tail [nil, nil, :"&block"]
new_args result (:args, (:arg, :stuff), (:blockarg, :block))
next token [:tRPAREN, [")", [1, 25]]]
next token [:tIDENTIFIER, ["method2", [2, 2]]]
next token [:tIDENTIFIER, ["stuff", [2, 10]]]
next token [:tCOMMA, [",", [2, 15]]]
next token [:tSYMBEG, [":", [2, 16]]]
next token [:tIDENTIFIER, ["junk", [2, 17]]]
next token [:tASSOC, ["=>", [2, 22]]]
next token [:tSTRING_BEG, ["'", [2, 25]]]
next token [:tSTRING_CONTENT, [" nope", [2, 26]]]
next token [:tSTRING_END, ["'", [2, 31]]]
next token [:tCOMMA, [",", [2, 32]]]
next token [:tAMPER, ["&", [2, 34]]]
next token [:tIDENTIFIER, ["block", [2, 35]]]
next token [:tNL, ["\\n", [3, 0]]]
[...]

PR #1132, opened 2015-10-07
Fix block after hash compilation/parser issue

wied03:
My first foray into the parser world, fixes https://github.com/opal/opal/issues/1131

Rubyspecs that probably cover this are noted in the issue, but since we can't run those yet, addressed it in the parser specs


elia:
@wied03 can you rebase on master and add a test with the actual call (in `spec/opal/core/language/block_spec.rb`)? 

(If you need any help or can't do it for some reason let me know and I'll do it myself)

wied03:
@elia - done

elia:
@wied03 great! I'll merge just after the green :vertical_traffic_light: from travis

_(just a note on commit messages for the next PRs: would be good to avoid the leading `*` and more or less follow the [general rules](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) on how to format them)_

elia:
Thanks!

PR #1134, opened 2015-10-08
Tweak Exception#inspect to match the Rubyspecs

wied03:


wied03:
Additional rubyspec for bc57d26 is https://github.com/ruby/rubyspec/pull/139

PR #1135, opened 2015-10-08
OpenStruct now mostly compliant

wied03:
…ented `#new_ostruct_member`, `#delete_field`, `#marshal_load`, `#marshal_dump`. Fully compliant except for frozen behavior. Borrowed self referential logic from Opal's Hash#hash

* Update changelog

* Enable rubyspecs for ostruct

* Filter private/protected/frozen specs


meh:
Remove the marshal stuff, Opal will never be able to support marshalling.

Also move the freezing stuff into unsupported, there's a file for the freezing failures, Opal won't support freezing either.

wied03:
@meh - any idea why the phantom build keeps crashing ?

wied03:
I can't reproduce the error that the Phantom rubyspec run is having in my environment, even with the exact same ruby version. It doesn't appear to be Phantom itself, it's MRI when lexing/parsing. Sprockets was recently updated to 3.4.0, but another PR built with that OK. I'll try upping the MRI version to 2.2.3 since I've seen this error message (can't remove str from str_id) in some MRI commits.

wied03:
The problem "went away" when I changed the Phantom build to 2.2.3. I didn't parse in detail, but looks like that error, which was generated in symbol.c, might have been addressed in 2.2.1.

elia:
> The problem "went away" when I changed the Phantom build to 2.2.3. I didn't parse in detail, but looks like that error, which was generated in symbol.c, might have been addressed in 2.2.1.

:+1:

elia:
Now I wonder what kind of policy travis has for `2.2` not selecting `2.2.3`… 
Need to check their docs and possibly update all versions :unamused:

stdlib/ostruct.rb
```diff
@@ -60,10 +66,52 @@ def to_n
   def hash
     @table.hash
   end
+  
+  def table
```
meh: This is a problem if anyone defines a field named `:table`, use an internal special field called `$$table`.

stdlib/ostruct.rb
```diff
@@ -60,10 +66,52 @@ def to_n
...
+  def delete_field(name)
+    sym = name.to_sym
+    begin
+      singleton_class.__send__(:remove_method, sym, "#{sym}=")
+    rescue NameError
+    end
+    @table.delete sym
+  end
+  
+  def new_ostruct_member(name)
```
meh: This method doesn't seem to exist in Ruby, unless I'm worng, so again, make it a local helper instead of a method.

stdlib/ostruct.rb
```diff
@@ -60,10 +66,52 @@ def to_n
...
+    unless respond_to?(name)
+      define_singleton_method(name) { @table[name] }
+      define_singleton_method("#{name}=") { |x| @table[name] = x }
+    end
+    name
+  end
 
+  `var ostruct_ids;`
+  
   def inspect
```
meh: This method is likely better implemented in JavaScript, you kind of inverted what should be JS and what should be Ruby.

stdlib/ostruct.rb
```diff
@@ -60,10 +66,52 @@ def to_n
...
-    "#<#{self.class}: #{each_pair.map {|name, value|
-      "#{name}=#{self[name].inspect}"
-    }.join(" ")}>"
+    top = `(ostruct_ids === undefined)`
+    ostruct_id = `self.$object_id()`
```
meh: This should be Ruby, never call methods with their mangled name since method definitions could change any time.

Also never use `object_id`, always use `__id__`, just make it `self.__id__`.

spec/filters/unsupported/openstruct.rb
```diff
@@ -0,0 +1,7 @@
+opal_filter "openstruct" do
```
meh: There's an unsupported filter already present called freeze or frozen, I don't remember, move these there.

stdlib/ostruct.rb
wied03: That field was already there and I didn't rock the boat. That is a good point though.

stdlib/ostruct.rb
wied03: They are both shown on ruby-doc.org

stdlib/ostruct.rb
wied03: Yeah, I was just copying what Hash#hash has and then toggled it back and forth trying to figure out that source map crash issue.

stdlib/ostruct.rb
wied03: Yeah, table is a protected method in Ruby's impl (just now noticed this).

wied03: @meh - Do you think $$table must be a JS object literal or can it be a Ruby Hash?

spec/filters/unsupported/openstruct.rb
wied03: moved the 1 freeze related 1 over, kept the non freeze related ones here.

meh: You're right, I didn't notice that they weren't all related.

The private/protected ones should go to `privacy.rb`.

The marshal related ones should go to `marshal.rb`, create the file since it's not there, unless the marshal related specs are in their own spec files, then just exclude it from `rubyspecs`.

stdlib/ostruct.rb
meh: If it's a JS object, access will likely be faster, but if you need `Hash` methods feel free to use one.

stdlib/ostruct.rb
meh: I got that from empirical tests, don't trust the docs blindly, some methods may be coming in from the C side of things or god knows where.

```ruby
>> require 'ostruct'
=> true
>> OpenStruct.new.new_ostruct_member("test")
NoMethodError: undefined method `new_ostruct_member' for #<OpenStruct>
    (ripl):2:in `<main>'
>> 
```

wied03: That makes sense. That said, the rubyspecs have several tests for each of those (that expect to send to those methods since they are protected/private). It would be a shame to filter/lose those specs, which I think we would have to do if we turned them into local helpers (I assume you mean a lambda or something inside the class definition when you say local helper).

meh:
```ruby
>> require 'ostruct'
=> true
>> x = OpenStruct.new
=> #<OpenStruct>
>> x.new_ostruct_member = "lol"
=> "lol"
>> x.new_ostruct_member
=> "lol"
```

How can this work then?

wied03: MRI & and the Rubyspecs define method_missing, which i suppose is why that equals call succeeds.

meh: The equals is obvious, what I'm talking about is the final call that returns the proper value.

wied03: I should add that the new_ostruct_member rubyspecs all expect a @table instance variable as well. By implementing the self.$$table advice above, that will break that. ostruct seems to be designed with inheritance in mind given this and the new_ostruct, etc. being protected. One could make the argument that the rubyspecs are mucking with the internals of what they are testing in this case, but I could see it both ways.

wied03: When you call x.new_ostruct_member = 'lol', you're setting an attribute in its hash called new_ostruct_member and setting its value to lol. That method missing call defines a new reader/writer called new_ostruct_member.

meh: Which means you cannot call `new_ostruct_member` internally anymore.

wied03: It sure does. That said, I'm just trying to 'improve' the impl here such that it matches what MRI is doing

meh: Oh man, I just tested and the same happens in MRI, I thought it had some C-ext internal magic to avoid exploding when `new_ostruct_member` was redefined.

The `#new_ostruct_member` stuff is fine then, but `#table` doesn't seem to exist.

```ruby
>> x = OpenStruct.new
=> #<OpenStruct>
>> x.table
=> nil
>> x.y = 2
=> 2
>> x.table
=> nil
```

wied03: Yep. table is a protected method in the MRI impl since ostensibly they want people to be able to subclass openstruct.

The rubyspecs for new_ostruct... also expect to be able to set the `@table` instance variable, so are you OK with not using `self.$$table` ?

meh: It doesn't look to me like `#table` is a method at all, since it's returning `nil`, it looks like to me it just goes through the normal `method_missing` route.

Yeah, I'm fine with using `@table`.

wied03: It's a reader @ https://github.com/ruby/ruby/blob/trunk/lib/ostruct.rb:

```ruby
attr_reader :table # :nodoc:
protected :table
```

wied03:
@meh - If I've kept track of everything, this should be ready to merge

meh:
Looks good, just squash the commits and I'll merge.

wied03:
@meh - I would, but I tried to make this easier by merging master as it changed, which makes squashing hard (just tried). I probably should not merge w/ master until then end in the future.

wied03:
@meh - I'll see if I can cherry pick and start over

wied03:
@meh - Squashed all except for 1 master merge

meh:
That's the problem, you should not merge, you should rebase upon it :)

Issue #1121, opened 2015-09-30
Missing method_missing stubs for optimized operators


elia:
$ opal -Oce '(1+2).abs'

```js
/* Generated by Opal 0.8.0 */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  function $rb_plus(lhs, rhs) {
    return (typeof(lhs) === 'number' && typeof(rhs) === 'number') ? lhs + rhs : lhs['$+'](rhs);
  }
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$abs']);
  return ($rb_plus(1, 2)).$abs()
})(Opal);
[…]
```

`Opal.add_stubs(['$abs']);` should have been `Opal.add_stubs(['$abs', '$+']);` instead

wied03:
@elia - Yep, this is causing the be >, <, etc. operator to fail in RSpec since it handles these via method_missing (down to my last 30 tests or so)

PR #1136, opened 2015-10-08
Test and add support for operator method missing

wied03:


wied03:
As reported by @elia @ https://github.com/opal/opal/issues/1121

PR #967, opened 2015-06-27
`Time` cleanup

vais:
Big picture, Opal does not support `Time` methods that:

* Explicitly manipulate a `Time` object's time zone offset (limitation of JS Date object).
* Require sub-millisecond time resolution (limitation of JS runtime).
* Use Rationals (limitation of Opal at this time).

Hope that makes sense.

PR #1141, opened 2015-10-09
Fix Time#yday

elia:
overrides and closes #636

### DIFF
--- a/lib/mspec/opal/runner.rb
+++ b/lib/mspec/opal/runner.rb
@@ -44,6 +44,38 @@ class Object
   def opal_filter(description, &block)
     OSpecFilter.main.register_filters(description, block)
   end
+
+  # Copyed from MSpec, with changes.
+  def with_timezone(name, offset = nil, daylight_saving_zone = "")
+    zone = name.dup
+
+    if offset
+      # TZ convention is backwards
+      offset = -offset
+
+      zone += offset.to_s
+      zone += ":00:00"
+    end
+    zone += daylight_saving_zone
+
+    # WAS:
+    #
+    #   old = ENV["TZ"]
+    #   ENV["TZ"] = zone
+    #
+    #   begin
+    #     yield
+    #   ensure
+    #     ENV["TZ"] = old
+    #   end
+    #
+    if ENV["TZ"] == zone
+      yield
+    else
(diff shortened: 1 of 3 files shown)


iliabylich:
MSpec has a [utility helper](https://github.com/rubyspec/mspec/blob/a67d653fce45919d28a98da3574b0a569db084de/lib/mspec/helpers/datetime.rb#L29) `with_timezone` that internally changes timezone using `ENV['TZ']`. But it seems that node doesn't refresh it internally. Switching `ENV['TZ']` in runtime doesn't change anything at all :smile: . As a result, it uses system timezone (in this case, `Pacific/Fiji`) - check this [gist](https://gist.github.com/iliabylich/5752e2e1861d002e22d6), it really returns 118. @elia any thoughts? Can we change timezone to UTC in travis.yml? Or it should be fixed on rubyspecs' side?

elia:
@iliabylich the TZ entry in travis was added on purpose to ensure the rubyspec runs properly on time zones distant from UTC as there was a time in which we had random failures due to DST and all that :rainbow: fantastic stuff. If we could make node pick up timezones that would be great (but we still need a way to deal with TZs in phantoms), otherwise we can filter out that spec and add something custom in spec/opal.

vais:
@iliabylich hope https://github.com/opal/opal/pull/967#issuecomment-116166482 saves you the pain :stuck_out_tongue_winking_eye:

iliabylich:
@vais yes, I've found it after 2 minutes of googling, that's a very sad limitation. I was thinking about adding a [node-time](https://github.com/TooTallNate/node-time) to a node.js suite, patching a default `Date` js function and patching most of the `Time` methods (again, suite-specific patches). *Probably* it solves an issue.
But then we face two other problems:
+ node-time is a native extension, so I assume we can't just put into the `node_modules` dir, everyone has to setup it manually
+ I guess phantomjs doesn't work with native node modules

So, as @elia wrote before, it's much, much easier to exclude this particular test from the suite and write own tests.

vais:
@iliabylich I agree!

elia:
Thanks @iliabylich for pointing out that helper, I patched it to skip (actually to "noop") the spec if the requested TZ is not the current one. And I also fixed the `#yday` implementation (thanks moment.js!).

PR #1143, opened 2015-10-13
Load rubyspecs filters in testing tasks before loading actual specs

iliabylich:
Without this patch it's impossible to run rubyspecs with some pattern and use filters, e.g. task
```
env RUBYSPECS=true PATTERN="spec/rubyspec/core/array/initialize_spec.rb" bundle exec rake mspec_rubyspec_node
```
was invoking specs that are marked as `failed` or `unsupported`

### DIFF
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -41,9 +41,9 @@ module Testing
 
     if pattern
       custom &= rubyspecs if whitelist_pattern
-      specs = add_specs.(:custom, custom)
       specs = add_specs.(:filters, opalspec_filters)
       specs = add_specs.(:filters, rubyspec_filters)
+      specs = add_specs.(:custom, custom)
     elsif suite == 'opal'
       specs = add_specs.(:filters, opalspec_filters)
       specs = add_specs.(:shared, shared)


vais:
Yeah, I don't know at what point this whole running the tests thing went sideways... @meh?

elia:
@vais at some point we discovered that spec/opal conflicted with spec/rubyspec in terms of required filenames, we splitted them in two separate runs in the fastest way possible (`ENV['SUITE']`).

I recently did what (I think :trollface:) is a refactoring and produced different tasks for each node/phantom and opal/rubyspec combination.

vais:
@elia, right, that was after I took off on a two-week vacation that I never managed to come back from as far as my contributions to Opal :stuck_out_tongue_winking_eye: I kinda liked how the test suite "just worked" before - every test ran unless you chose to zero-in on a specific PATTERN. Now I never managed to make heads or tails of it. I guess I just need to spend some time with it, but there's just been none lately. Anyway, thanks to @iliabylich for the fixes!

elia:
@vais it's quite the same as it was (bugs apart) PATTERN is still supported

https://github.com/opal/opal/blob/master/tasks/testing.rake#L213-L228

PR #1144, opened 2015-10-13
Fix Array#initialize to match rubyspecs.

iliabylich:


meh:
Nice :panda_face:

vais:
:+1:

PR #1145, opened 2015-10-14
Return correct size of Array#combination enumerator.

iliabylich:


elia:
great :+1:!

vais:
@iliabylich :beers:

PR #1147, opened 2015-10-14
Fix Array#hash and Hash#hash in recursive cases.

iliabylich:


elia:
@iliabylich any reason for switching from the closure var to the global one (`hash_ids` → `Opal.hash_ids`)? 

#TMYK :smile:

iliabylich:
``` ruby
# before patch
h = {} ; rec = [h] ; h[:x] = rec; [{x: rec}].hash
=> "A,Hash,x,A,Hash,x,A,self"

# after patch
h = {} ; rec = [h] ; h[:x] = rec; [{x: rec}].hash
=> "A,Hash,x,self"
```

Current master doesn't detect cross-recursion when goes through the call chain
`Array#hash` -> `Hash#hash` -> `Array#hash`
(or)
`Hash#hash` -> `Array#hash` -> `Hash#hash`

So, `Opal.hash_ids` saves it across the calls sequence, that's the fix.

Are you ok with changing global state in local methods? Would it be better to move parts related to `hash_ids` to `runtime.js`? Currently `Array` and `Hash` are the only recursive structures with own `hash` method (things like `Struct` and `OpenStruct` are based on `Hash`), but probably some nested structures may need it in future (actually don't know, just wondering)

elia:
Merged, was good for now :smile: 

About having a helper on `runtime.js` for recursive checks I think it's a good idea, especially now that we rely on a property of `Opal`. I'm thinking about something like:

```rb
def hash
  %x{
    Opal.with_recursive_hash_check(hash_id, function(hash_ids){
      var result = ['A']
      for (i = 0; i < self.length; i++) {
        item = self[i];
        result.push(item.$hash());
      }
      return result.join(',');
    }
  }
end
```

That would deal with cleaning up the global and checking for top etc. is this what you were thinking about too?

vais:
I recall wondering if that `try {...} finally {...}` was really required, and if it has negative effect on performance (javascript engine's ability to optimize). Any thoughts?

elia:
I don't think it's possible but i don't know if it has any effect on perf putting it in a separate function, maybe @jgaskins has tried that and knows better

jgaskins:
Yeah, try/finally blocks are not optimizable, but as @elia mentioned extracting the contents of `try` to another function gets around that. I'm not sure why that helps (the stack still has to be unwound to the `try` in the event of an exception), but my benchmarks showed it works in Safari and Chrome.

But in all the Opal profiling I've done in Clearwater apps at least, I've never noticed `Hash#hash` as a bottleneck (otherwise I'd have tried to fix it already ;-)), but I almost never use hashes or arrays as hash keys. It's probably not a hotspot in most apps. I suppose if it's part of your code's hot path (if you have a `Set` of hashes/arrays that gets checked frequently, for example), it could *become* problematic, but I don't know if it's worth optimizing yet since it adds another layer of indirection.

vais:
Thanks @jgaskins. I guess my point is I do not understand why `try {} finally {}` is necessary *at all* in this case. I may be missing something, but it looks the code would be just fine without it. What are we really `try`-ing here? I think this was just a way to make **absolutely sure** that `delete Opal.hash_ids;` runs no matter what, but I don't see what could be expected to go wrong in the `try` block that necessitates wrapping it thusly.

IlyaBylich:
@vais `try {} finally {}` is required here. Some objects may have broken `#hash` method that raises an error. In that case we need to make sure that top most `#hash` flushes a global `Opal.hash_ids` value. Otherwise, calling `#hash` when `Opal.hash_ids` contains some data from previous call may cause a really unexpected behavior. Debugging that may become a huge pain.

vais:
@iliabylich Ah, I see what you mean... makes sense now!

IlyaBylich:
@vais Wait, `finally` is not really necessary here (`try` catches all bad scenarios), this is something like a default `ensure` in Ruby. But `try` is still required. Is `try {} finally {}` much faster then just `try {}`?

vais:
Nah, I don't think you can have a `try` without a `catch` or `finally` in JS.

jgaskins:
@IlyaBylich Not sure what you mean. How is it not necessary? The `try` won't handle resetting the global state.

But yeah, without a catch/finally, try isn't even necessary in JS.

IlyaBylich:
@jgaskins nvm, as @vais mentioned above, JS doesn't support `try {}` without catch/finally (just like Ruby).

PR #1150, opened 2015-10-15
throw exception when included object is a class, fixes #1146

### LINKED ISSUE #1146: Include Class instead of Module
If I try to include `Class` instead if `Module` I'll get weird error:

```
Uncaught TypeError: Cannot read property 'push' of undefined
```

example:

```ruby
class Test
end

class Test2
  include Test
end
```

fazibear:


vais:
@fazibear any reason not to lean on rubyspec for this instead of making this an Opal-specific test? E.g. https://github.com/ruby/rubyspec/blob/master/core/module/include_spec.rb#L44

@eregon there doesn't seem to be the equivalent to @fazibear's [kernel extend test](https://github.com/opal/opal/pull/1150/files#diff-a90cec31a62a55f4216ab162a7ac654aR22) in https://github.com/ruby/rubyspec/blob/master/core/kernel/extend_spec.rb Should there be? Looks like a rubyspec contribution waiting to happen :)

fazibear:
@vais All tests passes, even if I checkout submodules (rubyspec). So I write one that fails to test if it works correctly. BTW when I write a spec, noticed that I throws wrong exception and change it.

iliabylich:
@fazibear opal has filters for rubyspec tests, check out [this](https://github.com/opal/opal/blob/master/spec/filters/bugs/module.rb#L127) file, removing/commenting a line includes a spec into the test suite.

fazibear:
@iliabylich thanks, one changed, for other we need rubyspec contibution.

eregon:
@vais Yeah, it would be good to have this in the `Kernel#extend` spec too :smiley: 
That spec needs some love apparently.
Ideally `obj.extend(mod)` would be as simple `obj.singleton_class.include(mod)` but there are some hook differences.

opal/corelib/kernel.rb
```diff
@@ -205,8 +205,13 @@ def extend(*mods)
...
         var mod = mods[i];
 
-        #{`mod`.append_features `singleton`};
-        #{`mod`.extended self};
+        if (mod.$$is_module) {
+          #{`mod`.append_features `singleton`};
+          #{`mod`.extended self};
+        }
+        else {
+          throw Opal.TypeError.$new("wrong argument type Class (expected Module)");
```
iliabylich: 1. I think you can raise a Ruby-level exception here, this code depends on internal implementation of Opal
2. If an object is not module, it doesn't mean that it's a class.
``` ruby
raise TypeError, "wrong argument type #{`mod`.class}"
```

fazibear:
@iliabylich you're right, Class should not be hardcoded.

opal/corelib/module.rb
```diff
@@ -341,8 +341,13 @@ def include(*mods)
...
 
-        #{`mod`.append_features self};
-        #{`mod`.included self};
+        if (mod.$$is_module) {
+          #{`mod`.append_features self};
+          #{`mod`.included self};
+        }
+        else {
+          #{raise TypeError, "wrong argument type #{`mod`.class} (expected Module)"};
+        }
```
elia: can you change this to the code below to make the default flow more clear?

```
if (!mod.$$is_module) {
  #{raise TypeError, "wrong argument type #{`mod`.class} (expected Module)"};
}

#{`mod`.append_features self};
#{`mod`.included self};
```

fazibear:
flow fixed, also rubyspec updated

vais:
@fazibear @eregon :clap:

meh:
@fazibear can you rebase please?

fazibear:
Done.

spec/opal/core/kernel/extend_spec.rb
```diff
@@ -18,4 +18,4 @@ class A
     obj.extend KernelExtendSpecs::Mod
     obj.foo.should == 3.142
   end
-end
```
vais: @fazibear looks like you're missing a newline here

vais:
@fazibear could you please squash the commits and rebase so I can merge this?

fazibear:
@vais done.

vais:
Thank you @fazibear!

PR #1151, opened 2015-10-15
Exception#exception, Exception::exception, message/to_s fixes

wied03:


meh:
Let's get #1117 first so you can remove the last commit.

wied03:
@meh - Sounds good. once #1117 is merged, I think I can trim the last commit on this and cut out 1 of the if clauses in raise. Let me give that a go before merging this PR.

wied03:
@meh - "Reversed" that commit and also did some optimizations that can be done now that the other PR is in. If it looks good, I'll squash.

opal/corelib/error.rb
```diff
@@ -8,13 +9,17 @@ def self.new(message=nil)
...
   end
+  
+  def self.exception(*args)
+    new(*args)
+  end
 
-  def initialize(message)
-    `self.message = message`
+  def initialize(*args)
+    `self.message = (args.length > 0) ? args[0] : nil`
```
meh: This is probably better as `@message, = *args` although I don't know what it would compile to, if it compiles to shit make it ```@message = `...```` so it's a little clearer it's an ivar.

meh: Unless it's meant to be a special var, then call it `$$message`.

meh:
Added some nits, rest looks fine.

wied03:
@meh - Most of the 'junk' on initialize was from *args, not from the message set itself, so I change the instance var to $$message but left the same idea, then squashed the commits

elia:
I'll have limited internet connectivity for a few days but I'll review (and merge) all these great PRs when I'll be back to a more civilized place. Unless @meh or @vais will shoot first. 😸

Elia Schito

opal/corelib/error.rb
```diff
@@ -32,16 +37,31 @@ def backtrace
...
+        return self;
+      }
+      
+      var cloned = #{self.clone};
+      cloned.$$message = str;
+      return cloned;
+    }
+  end
+  
   def message
```
meh: Can't this just be `alias message to_s`?

meh:
Added one final nit, then I'll merge.

wied03:
@meh - I would use alias, except when you do that, you can't override `to_s` anymore (and have message reflect the change) and thus 1 of the rubyspecs fails. It's not just an Opal problem, that doesn't work on MRI either. I added a comment above `def message; to_s; end` explaining why I didn't do that for future reference.

meh:
:+1:

wied03:
@meh - had to go back to using @message, see #1155

PR #1153, opened 2015-10-16
Allow undef to take in multiple methods at the same time

wied03:
Rubyspecs for this are in https://github.com/ruby/rubyspec/pull/145

### DIFF
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -318,6 +318,11 @@ module Opal
       return returns s(:nil) unless sexp
 
       case sexp.type
+      # Undefs go from 1 ruby undef a,b,c to multiple JS Opal.udef() calls, so need to treat them as individual statements
+      # and put the return on the last one 
+      when :undef
+        last = sexp.pop
+        sexp << returns(last)
       when :break, :next, :redo
         sexp
       when :yield
--- a/lib/opal/nodes/definitions.rb
+++ b/lib/opal/nodes/definitions.rb
@@ -16,10 +16,21 @@ module Opal
     class UndefNode < Base
       handle :undef
 
-      children :mid
-
       def compile
-        push "Opal.udef(self, '$#{mid[1].to_s}');"
+        children.each do |child|
+          value = child[1]
+          statements = []
+          if child[0] == :js_return
+             value = value[1]
+             statements << expr(s(:js_return))
+          end
+          statements << "Opal.udef(self, '$#{value.to_s}');"
+          if children.length > 1 && child != children.first
+            line *statements
+          else
+            push *statements
+          end
+        end
       end
     end
 
--- a/spec/filters/bugs/language.rb
+++ b/spec/filters/bugs/language.rb
@@ -150,8 +150,6 @@ opal_filter "language" do
   fails "The super keyword without explicit arguments passes optional arguments that have a default value"
   fails "The super keyword without explicit arguments passes optional arguments that have a non-default value but were modified"
   fails "The super keyword without explicit arguments passes rest arguments including any modifications"
-  # can be removed when https://github.com/opal/opal/pull/1153 is merged
-  fails "The undef keyword allows undefining multiple methods at a time"
   fails "The unpacking splat operator (*) unpacks arguments as if they were listed statically"
   fails "The until expression restarts the current iteration without reevaluating condition with redo"
   fails "The until modifier restarts the current iteration without reevaluating condition with redo"
--- a/spec/lib/compiler_spec.rb
+++ b/spec/lib/compiler_spec.rb
@@ -51,6 +51,11 @@ describe Opal::Compiler do
     expect_compiled("Object").to include("Object")
     expect_compiled("Array").to include("Array")
   end
+  
+  it "should compile undef calls" do
+    expect_compiled("undef a").to include("Opal.udef(self, '$a')")
+    expect_compiled("undef a,b").to match(/Opal.udef\(self, '\$a'\);.*return Opal.udef\(self, '\$b'\);/m)
+  end
 
   describe "class names" do
     it "generates a named function for class using $ prefix" do
--- a/spec/lib/parser/undef_spec.rb
+++ b/spec/lib/parser/undef_spec.rb
@@ -12,4 +12,8 @@ describe "The undef keyword" do
   it "can take symbols or fitems" do
     parsed("undef :foo").should == [:undef, [:sym, :foo]]
   end
+  
+  it "can take multiple symbols" do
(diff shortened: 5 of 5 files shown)


meh:
I'd rather `Opal.udef` worked on one method at a time but `undef_method` expanded to multiple calls to `Opal.udef`.

meh:
Also does the same apply to `remove_method`?

wied03:
1) I can change this to put Opal.udef in the JS once for each call if that's what you'd prefer.
2) Yes remove_method does let you supply multiple methods in MRI.

wied03:
Rebased, pulled in the latest rubyspecs since https://github.com/ruby/rubyspec/pull/145 was merged

meh:
1) Yeah, I'd rather keep `Opal.udef` simple.
2) Could you do the same for `remove_method` then while you're there?

wied03:
@meh - Made those changes. It seemed like it was wrong to change the parser for undef to be like an arglist (since this is really a 'how do we represent this in JS' issue more than AST/Ruby issue). Therefore, in order to keep the return from messing things up when an undef is at the end, I wrapped multiple undefs in a function/closure.

Remove_method and undef_method already work properly w/ multiple args in Opal but I submitted another Rubyspec PR (https://github.com/ruby/rubyspec/pull/147/) to make sure it's tested.

meh:
Sounds good.

I only have a nit, the code there seems a little overcomplicated, basically `compile_udefs` is more complicated split out than inlined since you're passing whether or not there are more than one and you already know that because you're checking it in the caller.

wied03:
@meh - put a commit in w/ a lambda instead. These things can be made more concise but then I think it starts to be less obvious what is going on

meh:
It's still overcomplicated for no reason, I was talking about making it this.

```ruby
def compile
  push "(function() {"

  if children.length > 1
    indent do
      children.each {|c|
        line "Opal.udef(self, '$#{c[0].to_s}');"
      }
    end
  else
    push "Opal.udef(self, '$#{children[0][0].to_s}');"
  end

  line "})();"
end
```

It's much simpler to see what is going on this way.

wied03:
I understand. The main difference between what I have and what you just posted is that my code only puts the closure in there if there is more than 1 (trying to cut down verbosity). If that's not important, then your version is fine.

wied03:
The cleanest way I guess is to remove the closure/function and keep the
compiler from putting a return until the last undef call but I didn't know
how to do that without modifying the parser as I mentioned before. Is there
another way to do that within the node?

sent from my mobile device

meh:
Then move the closure inside the branch that needs it, which makes it clear why it's needed.

wied03:
Can this be done without the closure?

sent from my mobile device

wied03:
@meh - I see what you're saying. I was trying too hard to not duplictate the "udef" string. I incorporated your feedback here. It still might be nice to not have to have that closure if the return issue above can be dealt with.

wied03:
I'm going to take 1 more look in the next hour to see if this can be done without the closure

meh:
Sorry didn't see the notifications, I'm not entirely sure why the closure is required.

wied03:
I was only using it to prevent this from happening:
```js
return Opal.udef(self, '$a');Opal.udef(self, '$b');
```

elia:
@wied03 if possible this PR should make `undef` work with dynamic symbols too, eg.:

```rb
a = 'foo'
undef :"#{a}_bar"
```

wied03:
1. Re: dynamic symbols, can we crawl before we walk here? This PR is already taking quite a bit of time.
2. I've done a lot of work here trying to split this out into multiple Opal.udef calls. It seems like a lot of unnecessary parser/compiler complexity just to avoid refactoring Opal.udef to accept an array as a parameter. Is this really necessary?

meh:
@wied03 I'd rather keep the runtime as simple as possible, it's already a complexity nightmare.

On the other hand the compiler will only get worse with time, so I'd rather make the compiler slightly worse than the runtime.

elia:
1. @wied03 that's why I started with "if possible", for me this is good for a merge but needed to note the info that dynamic symbol support is still to be added and this PR seems the best place.
2. Not sure about it being necessary but I completely agree that's better

:smile:

wied03:
@elia - Understand if you want to capture that. Might an issue be a good place?

Made the changes. Now this:
```ruby
class T
  undef foo
end

class B
  undef foo, bar
  alias a b
end

class C
  undef foo, bar, ssss
end

class C
  undef foo, bar, ssss
  alias a b
end
```

compiles to this:
```js
/* Generated by Opal 0.9.0.dev */
(function(Opal) {
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $klass = Opal.klass;

  (function($base, $super) {
    function $T(){};
    var self = $T = $klass($base, $super, 'T', $T);

    var def = self.$$proto, $scope = self.$$scope;

    
    return Opal.udef(self, '$foo');
  })($scope.base, null);
  (function($base, $super) {
    function $B(){};
    var self = $B = $klass($base, $super, 'B', $B);

    var def = self.$$proto, $scope = self.$$scope;

    
    Opal.udef(self, '$foo');
    Opal.udef(self, '$bar');;

    return Opal.alias(self, 'a', 'b');
  })($scope.base, null);
  (function($base, $super) {
    function $C(){};
    var self = $C = $klass($base, $super, 'C', $C);

    var def = self.$$proto, $scope = self.$$scope;

    
    Opal.udef(self, '$foo');
    Opal.udef(self, '$bar');
    return Opal.udef(self, '$ssss');
  })($scope.base, null);
  return (function($base, $super) {
    function $C(){};
    var self = $C = $klass($base, $super, 'C', $C);

    var def = self.$$proto, $scope = self.$$scope;

    
    Opal.udef(self, '$foo');
    Opal.udef(self, '$bar');
    Opal.udef(self, '$ssss');;

    return Opal.alias(self, 'a', 'b');
  })($scope.base, null);
})(Opal);

/* Generated by Opal 0.9.0.dev */
(function(Opal) {

  Opal.add_stubs(['$exit']);
})(Opal);

```

meh:
Looks good to me.

meh:
Actually, I think we don't need a new token at all, there should already be an `:undef` node check there, you just have to verify the children are one or more, and dispatch to the proper compilation avoiding the `returns(..)`.

wied03:
I didn't exactly define a new token. I just passed through the undeflist concept, which the parser already knows about, into the compiler/node code.

meh:
Thing is, we already have the `:undef` node, and we're already able to know if it has one or more children, so it's simpler to just check that.

wied03:
done.

Issue #1154, opened 2015-10-16
Constants Behavior Degrades on 0.9


wied03:
```ruby
class Bar
  def self.get_klass
    k = Class.new(self)
    Foo.assign_const(k)
  end
end

module Foo
  def self.assign_const(group)
    scope = group.superclass
    scope = self if scope == Bar
    scope.const_set("MyStuff", group)    
  end
end

Bar.get_klass

puts Foo::MyStuff
puts Foo::MyStuff.name
```

MRI:
```
Foo::MyStuff
Foo::MyStuff
```

0.8 got 1/2 right:
```
MyStuff
Foo::MyStuff
```

0.9 is wrong on both:
```
Bar::MyStuff
Bar::MyStuff
```

wied03:
issue resolved by PR

PR #1156, opened 2015-10-16
Restore ability to call const_set on an anonymous class w/ non object super

wied03:
Issue that was breaking opal-rspec on 0.9

See https://github.com/opal/opal/issues/1154

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -705,7 +705,8 @@
     var scope = base_module.$$scope;
 
     if (value.$$is_class || value.$$is_module) {
-      if (value.$$base_module === ObjectClass) {
+      // only checking ObjectClass prevents setting a const on an anonymous class that has a superclass that's not Object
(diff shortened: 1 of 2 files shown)


wied03:
Re: #1154, the output now matches MRI exactly

meh:
I remember explicitly removing that check, but I don't remember why :rage4:

wied03:
@meh - I checked what you did before. 0.8 ignores whether it's a module or not and that's the difference. This brings back the 0.8 like logic but only for classes, not modules.

It's possible that the BasicObject piece isn't necessary, but i haven't tried that yet.

wied03:
@meh - I checked and the BasicObject check is needed for language/module_spec to pass, which is probably why you put it in there. I think this gets the best of both worlds. Keeps the module specs running like you did and also prevents regressing in this specific scenario.

meh:
Fine by me :+1:

PR #1159, opened 2015-10-17
Handle instance_eval/exec on class/modules

wied03:
Should allow defining class methods properly in blocks

Rubyspecs PR - https://github.com/ruby/rubyspec/pull/150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 
 * `Exception#exception`, `Exception::exception`, `Exception#message`, and `Exception#to_s` are fully implemented
 
+* Method defs issued inside `Module#instance_eval` and `Class#instance_eval`, and the respective `exec` now create class methods
 
 ## 0.8.0 2015-07-16
 
--- a/opal/corelib/basic_object.rb
+++ b/opal/corelib/basic_object.rb
@@ -52,7 +52,22 @@ class BasicObject
           result;
 
       block.$$s = null;
-      result = block.call(self, self);
+      
+      // need to pass $$eval so that method definitions know if this is being done on a class/module. Cannot be compiler driven since
+      // send(:instance_eval) needs to work
+      if (self.$$is_class || self.$$is_module) {
+        self.$$eval = true;
+        try {          
+          result = block.call(self, self);
+        }
+        finally {
+          self.$$eval = false;
+        }
+      }
+      else {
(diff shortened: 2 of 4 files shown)


meh:
Could you rename the special flag to `$$eval` and put the `try/finally` block only for the path that does require unsetting it please?

wied03:
@meh - Sure, wasn't sure about the best way to do that, was my first crack at it

wied03:
@meh - Complete, also covered the alias case and updated the rubyspec PR accordingly

meh:
@wied03 I'm not sure I understand the alias case, what is it for?

wied03:
If you do this:

```ruby
String.instance_eval do
 def stuff; end;
 alias stuff_alias stuff
end
```
Then String should have 2 class methods added (stuff and stuff_alias). It's something you can't do within a normal class definition to my knowledge without using send on the singleton class. E.g.

```ruby
String.singleton_class.send(:alias_method, :stuff_alias, :stuff)
```

All of that said, while the opal specs pass, I'm having a problem now running the rspec specs against this revision now (hangs at the end). Will post comment when I have updates.

meh:
If it helps `alias` should compile to `Opal.alias`.

wied03:
@meh - Yep, no compiler issue here. The source of problem in my last comment was caused by `instance_eval` being used in `Struct::new` instead of `class_eval`. `instance_eval` and `class_eval` were (incorrectly) used interchangeably since they were basically the same in Opal until now. I fixed that and submitted another rubyspec PR for struct to cover that case. I also referenced the Ruby source code for struct in that 2nd commit to this PR.

The only other usage of instance_eval in the opal code base I can see is the singleton stuff in stdlib but it looks like it's OK.

wied03:
and FYI, I will not be bombarding the project with PRs forever, just trying to get opal-rspec in a state where it works good with all the rspec specs on Opal 0.8 and Opal 0.9 :)

meh:
@wied03 aww too bad, I appreciate this kind of PR, I'm just a little pedantic with reviews :panda_face:

wied03:
@meh - I double checked singleton and ruby's https://github.com/ruby/ruby/blob/trunk/lib/singleton.rb uses instance_eval just like Opal master does so I suspect that's correct. I'm not sure why the specs didn't fail when this change was made but it probably has to do with class variables vs. class instance variables.

PR #1165, opened 2015-10-21
Remove flawed method check that was causing struct issues

wied03:
Remove flawed method check that was preventing something? from working with struct

Methods ending in ?, !, etc. are valid but this check did not like them. In addition, it would fail silently, which isn't good. Since it was only used by the struct and was incomplete, I think removing it is the best option. it's probably better to have a built-in check inside define_method anyways

### DIFF
--- a/opal/corelib/helpers.rb
+++ b/opal/corelib/helpers.rb
@@ -112,11 +112,5 @@ module Opal
     end
 
     name
-  end
-
-  def self.valid_method_name?(method_name)
-    method_name = Opal.coerce_to!(method_name, String, :to_str)
-
-    `/^[a-zA-Z_][a-zA-Z0-9_]*?$/.test(method_name)`
-  end
+  end  
 end
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -37,15 +37,13 @@ class Struct
 
     members << name
 
-    if Opal.valid_method_name?(name)
-      define_method name do
-        self[name]
-      end
-
-      define_method "#{name}=" do |value|
-        self[name] = value
-      end
+    define_method name do
+      self[name]
     end
+
+    define_method "#{name}=" do |value|
+      self[name] = value
+    end    
   end
 
   def self.members
--- a/spec/filters/bugs/struct.rb
+++ b/spec/filters/bugs/struct.rb
@@ -1,6 +1,5 @@
 opal_filter "Struct" do
   fails "Struct#hash returns the same fixnum for structs with the same content"
-  fails "Struct#initialize allows valid Ruby method names for members"
   fails "Struct#initialize can be overriden"
   fails "Struct#inspect returns a string representation of some kind"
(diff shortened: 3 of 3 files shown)


wied03:
This used to work fine on Opal 0.8 and has regressed.

IlyaBylich:
@wied03 This is how it works on MRI 2.2
``` ruby
S = Struct.new(:method1?, :method2!, :'method-3')
s = S.new(1,2,3)

puts s.method1?
# => 1
puts s.method2!
# => 2
puts s.send(:'method-3')
# => 3

s.send('method1?=', 4)
# => ok
s.send('method2!=', 5)
# => ok
s.send(:'method-3=', 6)
# => `method-3=' is not a struct member (NameError)

puts s
# => #<struct S :method1?=4, :method2!=5, :"method-3"=3>
```

`Struct` member that contains dashes should not have writer method (only reader is allowed).

You are write, current implementation of `valid_method_name?` is wrong and probably should be removed. Instead there should be some method like `valid_struct_member?` that uses a bit modified regex from `valid_method_name?` (like existing regex + `[!\?]?`). Don't know, probably `OpenStruct` may also need this validation, if not - it can be implemented as inline check directly in `Struct` class.

wied03:
Good points. I did some more research and MRI catches "static" calls using invalid names (of both `define_method` and `def`) with its interpreter (as does Opal).

For `define_method` using a variable, it lets you define 'invalid' method names and even call them with `send`. A rubyspec for `define_method` therefore does not seem to make sense.

I think you're right that a struct specific test needs to be done. I'll see if I can work on that.

wied03:
I actually cannot replicate the NameError on `s.send(:'method-3=', 6)` on 2.2.3 on Linux or OS X or on 2.0/system ruby on OS X.

1.9.3 doesn't even make it past `puts s.method1?`. Any specifics on how you reproduced this?

IlyaBylich:
@wied03 [This](https://gist.github.com/IlyaBylich/257a597b776a90d567bb) the exact code I was using for testing

wied03:
2.2.0? what platform?

IlyaBylich:
@wied03 linux

wied03:
Sorry, I made the last comment before I saw your gist included that, then deleted it.

wied03:
@IlyaBylich - Only 2.2.0-preview1 has that behavior in my quick tests, the actual 2.2.0 release (and 2.2.3) do not do that.

wied03:
I think that means we really don't need to do any checks here.

wied03:
I submitted a Rubyspec PR (https://github.com/ruby/rubyspec/pull/154) that ensures the valid method names work

wied03:
Failing due to https://github.com/ruby/rubyspec/pull/155

meh:
@wied03 can you rebase please?

wied03:
@meh - done, but since opal-browser specs are running, it will probably be tomorrow before it builds :)

wied03:
@meh - looks like this passed, master broke an unrelated spec I think

PR #1166, opened 2015-10-23
String#gsub optimization

wied03:
Convert Ruby return to JS return and avoid the returner logic which adds a try {} catch around the entire method


jgaskins:
Nice! :+1:

PR #1167, opened 2015-10-25
Optimize arguments slicing in more places

jgaskins:
We optimized splat args in the compiler before, but several places in the runtime still used the same slice idiom, which led to reduced performance in several code paths. This commit cleans up all the others I found.

### DIFF
--- a/opal/corelib/helpers.rb
+++ b/opal/corelib/helpers.rb
@@ -72,7 +72,9 @@ module Opal
         return args;
       }
       else {
-        return $slice.call(args);
+        var newArgs = new Array(args.length);
+        for(var i = 0; i < newArgs.length; i++) { newArgs[i] = args[i]; }
+        return newArgs;
       }
     }
   end
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -993,7 +993,9 @@
       method_missing_stub.$$p = null;
 
       // call method missing with correct args (remove '$' prefix on method name)
-      return this.$method_missing.apply(this, [method_name.slice(1)].concat($slice.call(arguments)));
+      var args = new Array(arguments.length);
+      for(var i = 0; i < args.length; i++) { args[i] = arguments[i]; }
+      return this.$method_missing.apply(this, [method_name.slice(1)].concat(args));
     }
 
     method_missing_stub.$$stub = true;
@@ -1184,7 +1186,14 @@
     }
 
     if (!args.$$is_array) {
-      args = $slice.call(args);
+      var newArgs = new Array(args.length);
+      var length = newArgs.length;
+
+      for(var i = 0; i < length; i++) {
+        newArgs[i] = args[i];
+      }
+
+      return block.apply(null, newArgs);
     }
 
     return block.apply(null, args);
@@ -1369,8 +1378,9 @@
   // @param mid  [String] ruby method to call
   // @return [Object] forwards the return value of the method (or of method_missing)
   Opal.send = function(recv, mid) {
-    var args = $slice.call(arguments, 2),
-        func = recv['$' + mid];
+    var args = new Array(Math.max(arguments.length - 2, 0));
+    var func = recv['$' + mid];
+    for(var i = 0; i < args.length; i++) { args[i] = arguments[i + 2]; }
 
     if (func) {
       return func.apply(recv, args);
@@ -1380,8 +1390,9 @@
   };
 
   Opal.block_send = function(recv, mid, block) {
-    var args = $slice.call(arguments, 3),
-        func = recv['$' + mid];
+    var args = new Array(Math.max(arguments.length - 3, 0));
+    var func = recv['$' + mid];
+    for(var i = 0; i < args.length; i++) { args[i] = arguments[i + 3]; }
 
     if (func) {
       func.$$p = block;


meh:
That gets ugly really fast.

Do you know if there's a way to maintain the performance but with a helper function? I was reading the link @wied03 posted on gitter about unoptimizable syntax and if I recall correctly there was a way to write a wrapper that would extract the arguments without leaking them.

Basically an `Opal.arguments` that does the conversion internally and returns an `Array` without leaking `arguments`.

jgaskins:
I believe [this](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments) is the section you're referring to. The problem is, abstracting this iteration has to be done at the compiler level or somewhere else in the build process, which [we are already doing](https://github.com/opal/opal/commit/7d4002fcfd686558c1aa99894d2784a26e55346a), but this diff involves code already written in JS that isn't twiddled by the compiler. :-\

You can't (well, shouldn't) pass the `arguments` object to another function. It's mentioned in that same section, and also I tried that already and it was equally slow. :-)

jgaskins:
I'm honestly surprised ES6 doesn't have something like an `arguments.toArray()`.

ryanstout:
Does it make since to add a compiler flag for "inline argument ...something". I know being able to read the generated code is a big selling point of Opal.

ryanstout:
Or maybe Opal needs the equivalent of -o2 -o3 :-)

jgaskins:
@ryanstout I'm not sure what you mean by inline argument.

meh:
@jgaskins welp, guess that's what we have to do then.

I'm just unsure about the ones that are with `args` instead of `arguments`, are we doing `args = arguments` or something?

jgaskins:
@meh [This](https://github.com/opal/opal/pull/1167/files#diff-aa9d58fa0fe0369084f2356dd3b9a5eeL1023) is the only one of those I found. Are there any others?

meh:
@jgaskins [this](https://github.com/opal/opal/pull/1167/files#diff-782785d42637cc1075a5823353c3d031L75) too.

ryanstout:
@jgaskins sorry, I meant inline the argument rebuild into an array.

meh:
@jgaskins can you confirm if those two instances of `args` instead of `arguments` are proper?

jgaskins:
Yeah, they're good. One's in [`Opal.destructure`](https://github.com/jgaskins/opal/blob/07d79d347ba85a088f616019114af13230b699d9/opal/corelib/helpers.rb#L66-L80) and the other is in [`Opal.yieldX`](https://github.com/jgaskins/opal/blob/07d79d347ba85a088f616019114af13230b699d9/opal/corelib/runtime.js#L1013-L1036)

meh:
@jgaskins what I meant is that `args` there is already an `Array` and not an `arguments`, so they don't need the loop expansion.

jgaskins:
@meh Ah, okay. I just searched the repo for `$slice.call(`. I figured if it was an array, it would just be `args.slice()`. :-)

meh:
@jgaskins wait, I think you're right, gimme a bit.

meh:
@jgaskins yeah I think those two could receive an `arguments` from outside, but in that case we're already leaking an arguments, and I think it won't happen in compiled code.

ryanstout:
@jgaskins you probably already looked at this, but is ```Array.prototype.slice.apply(arguments)``` slow also?  Thanks!

jgaskins:
@meh Ah, right, I don't know why I didn't notice that. Since we're already leaking `arguments`, it's probably going to be slow anyway. I'll see if I can trace that down. I should probably just search for any time `arguments` appears inside of parentheses. :-)

@ryanstout Yep, that's how we got here. ;-) It's how `Opal.slice` (and therefore `$slice` inside `runtime.js`) is implemented. The problem is that the VM can't JIT functions that receive an `arguments` object. The optimizer warning for this in v8 is "Bad context value for arguments object" or something similar.

Removing those calls is a huge part of how sped up Opal 0.9 — `Class#new` used the `$slice.call` method because of splat args. To get an idea of the impact of those optimizations, `Class#new` went from 13-15% of the time spent rendering a Clearwater app on Opal 0.8 to below 1% on Opal 0.9. :-)

ryanstout:
@jgaskins thanks for the info.  Can you inline the ```Array.prototype.slice.apply(arguments)``` call?  ```slice``` is implemented in C in most of the browsers, or is the cost just calling any function?

ryanstout:
@jgaskins also, great work on the perf stuff :-)

jgaskins:
@ryanstout Thanks!

I see what you mean about slice. What I said was actually misleading. It wasn't the fact that it _receives_ `arguments` (or maybe it is, I can't remember off the top of my head), but that we're sending it from the calling function. So the it isn't the array copying that is slow, but that it makes the entire calling function unoptimizable.

ryanstout:
@jgaskins cool, good to know.  Keep up the good work.

elia:
@jgaskins please update (and ping), the merge is on it's way :)

jgaskins:
@elia Ah, right, I completely forgot about this branch. :-) Rebased and CI seems to be passing still.

jgaskins:
@elia Ping :-)

elia:
Pong!   ➰⚪️

PR #1171, opened 2015-10-28
Implement Enumerable#minmax.

iliabylich:


opal/corelib/enumerable.rb
```diff
@@ -177,6 +177,10 @@ def count(object = undefined, &block)
   end
 
   def cycle(n = nil, &block)
+    if `arguments.length > 1`
```
meh: I don't think this is needed when arity checks are enabled.

opal/corelib/enumerable.rb
```diff
@@ -945,7 +949,41 @@ def min_by(&block)
   end
 
   def minmax(&block)
-    raise NotImplementedError
+    unless block_given?
```
meh: And here I'd rather have a `block ||= proc { |a, b| a <=> b }`.

opal/corelib/enumerable.rb
IlyaBylich: @meh But without this `if` an associated spec fails. Is it a problem with running build on CI (that we do not pass some extra options to compiler) or am I missing something?

meh: That's weird, I'm pretty sure the arity checker should catch this.

IlyaBylich: Opal master:
``` ruby
>> [].cycle(2,3)
=> #<Enumerator: []:cycle(2)>
```

meh: You have to explicitly enable arity checking, but I'm unsure how it's done with `opal-repl` or if it's doable at all.

@elia @wied03 is it possible something changed in the rake tasks that disabled arity checking? Or is this case not accounted by it?

wied03: re: arity checking being turned off in the Rake task, not that I know of, i still see '-A' in the args when the mspec_opal_node and mspec_rubyspec_node Rake tasks are run.

IlyaBylich: @meh Even with arity check enabled:
``` ruby
opal --arity-check -e 'puts [].cycle(2,3)'
#<Enumerator:0x102>
```

wied03: The arity check, when compiled in for the cycle method, looks weird:

```js
var $arity = arguments.length;if ($arity < 0) { Opal.ac($arity, -1, this, "cycle"); }
```

How could the args array ever have a length < 0?

wied03: Seems to impact a lot of cases where there is a default parameter value (not just blocks):

```ruby
def stuff(foo=123)
end
```

```js
/* Generated by Opal 0.9.0.beta1 */
(function(Opal) {
  var OPAL_CONFIG = { method_missing: true, arity_check: true, freezing: true, tainting: true };

  return (Opal.defn(Opal.Object, '$stuff', function(foo) {var $arity = arguments.length;if ($arity < 0) { Opal.ac($arity, -1, this, "stuff"); }
    var self = this;

    if (foo == null) {
      foo = 123
    }
    return nil;
  }), nil) && 'stuff'
})(Opal);

/* Generated by Opal 0.9.0.beta1 */
(function(Opal) {
  var OPAL_CONFIG = { method_missing: true, arity_check: true, freezing: true, tainting: true };

  Opal.add_stubs(['$exit']);
})(Opal);

```

elia: in cRuby it's < 0 as well:

    ruby -e "p method(def a(b=1) end).arity" # => -1

elia: @wied03 The arity check for less than zero because the minimum arity is zero, the real problem is that a max-arity check is missing. Fixing it on master.

PR #1172, opened 2015-10-28
`NameError` and `NoMethodError` - add `#name` and `#args` attributes

wied03:
Wasn't sure whether to leave the filled out `NameError` and `NoMethodError` classes in the same spot, but did so and preserved the indentation


meh:
I'd rather you leave them in the definitions as they were before, and monkey patch them below.

Also I'd rather they generated the error message themselves upon taking the parameter, and instead of `raise T.new(blabla, x)` it was `raise T, x`.

wied03:
1. MRI (and thus rubyspecs) require you to supply a message, doesn't generate by itself
2. They also don't let you supply supplementary args to `raise`.  see below.
code:
```ruby
begin
  raise NoMethodError, 'no method', 'food'
rescue NoMethodError => e
  puts e.name
end

begin
  raise NoMethodError.new('no method', 'food')
rescue NoMethodError => e
  puts e.name
end
```

output:
```

food
```

meh:
Ah, I was sure that worked, welp.

PR #1179, opened 2015-11-04
Add max-arity checks

elia:
ref #1171 

Probably `Opal.ac` still needs an update to support error messages with min and max arity


vais:
:+1: :+1:

PR #1180, opened 2015-11-05
Fix Kernel#instance_variables for Hash and nil

iliabylich:
Before:
``` ruby
>> nil.instance_variables
=> ["@apply", "@call", "@constructor", "@toString"]
>> {}.instance_variables
=> ["@map", "@keys", "@smap", "@constructor", "@none", "@proc", "@toString"]
>> Class.new.instance_variables
=> ["@constructor", "@toString"]
>> Module.new.instance_variables
=> ["@constructor", "@toString"]
```
After:
``` ruby
>> nil.instance_variables
=> []
>> {}.instance_variables
=> []
>> Class.new.instance_variables
=> []
>> Module.new.instance_variables
=> []
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 0.9.0 (edge)
 
+* Renamed:
+    * `Hash.keys` => `Hash.$$keys`
+    * `Hash.map` => `Hash.$$map`
+    * `Hash.smap` => `Hash.$$smap`
+
 * A `console` wrapper has been added to the stdlib, requiring it will make available the `$console` global variable.
 
 * `Kernel#pp` no longer forwards arguments directly to `console.log`, this behavior has been replaced by stdlib's own `console.rb` (see above).
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -199,14 +199,14 @@ module Opal
             arg_name = kwarg[1]
             var_name = variable(arg_name.to_s)
             add_local var_name
-            line "if ((#{var_name} = $kwargs.smap['#{arg_name}']) == null) {"
+            line "if ((#{var_name} = $kwargs.$$smap['#{arg_name}']) == null) {"
             line "  #{var_name} = ", expr(kwarg[2])
             line "}"
           when :kwarg
             arg_name = kwarg[1]
             var_name = variable(arg_name.to_s)
             add_local var_name
-            line "if ((#{var_name} = $kwargs.smap['#{arg_name}']) == null) {"
+            line "if ((#{var_name} = $kwargs.$$smap['#{arg_name}']) == null) {"
             line "  throw new Error('expecting keyword arg: #{arg_name}')"
             line "}"
           when :kwrestarg
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -93,16 +93,16 @@ class Hash
         return false;
       }
 
-      if (self.keys.length !== other.keys.length) {
+      if (self.$$keys.length !== other.$$keys.length) {
         return false;
       }
 
-      for (var i = 0, keys = self.keys, length = keys.length, key, value, other_value; i < length; i++) {
+      for (var i = 0, keys = self.$$keys, length = keys.length, key, value, other_value; i < length; i++) {
         key = keys[i];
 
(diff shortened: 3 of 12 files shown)


elia:
I'm a bit unsure about the change of `constructor` to `$$constructor`, I fear it can have side effects

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor

vais:
@iliabylich have you tried also removing these RubySpec failures? https://github.com/opal/opal/blob/master/spec/filters/bugs/kernel.rb#L44-L46

iliabylich:
@vais Will try, thank you

vais:
@iliabylich I'm worried about `constructor` as well - technically it's not *ours* to mess with (as opposed to hash.keys and such other, which are fair game AFAIK)

iliabylich:
What do you think about adding `if ivar_name != 'constructor'` to `Kernel#instance_variables`? Or even better, to `Opal.valid_instance_variable_name?` (setting an instance variable `@constructor` breaks an object). For `Hash` and `nil` it's fixed now, renamed methods starting with `$$` do not conflict with instance variables anymore.
@vais You were totally right, these specs are passing, I've removed them from `/bugs`

vais:
@iliabylich :+1: go with the *...even better...* option :wink:

elia:
@iliabylich there's a list of reserved words somewhere in `lib/opal/nodes` that will be checked against for local and/or instance vars, eg. `class` is turned to `class$`.

iliabylich:
@elia all changes related to `$$constructor` have been reverted

iliabylich:
@elia Any updates on this? Is there something that still should be fixed? I really need this fix for another functionality :smile:

meh:
There are other internal private properties, check `Opal.ivar`.

iliabylich:
@meh Yes, but all of them are prefixed with `$`, so `instance_variables` ignores all of them

meh:
`nil.call` and `nil.apply` where like that on purpose, because then you have duck typing for a `Function`.

meh:
Also I'd rather go with a check in `#instance_variables` rather than adding a helper.

iliabylich:
@meh Is it for passing `nil` as a block and calling it (and therefore raising an exception)? If so, it doesn't break anything. Does Opal use it at all?

iliabylich:
@meh Updated according to your comments.

meh:
I also think it's just better to inline the check instead of having an external function, also I wouldn't check the validity of the instance variable name, but would rather just check it's not one of the special ones, like `constructor`, `__proto__` and friends.

iliabylich:
I've updated the code so now it includes only changes related to `Hash` and `nil`. I'd like to finish the rest of the work in a separated PR, it's hard to track everything in a single commit :smile:

elia:
Looks good, will wait for @meh to see this and then I think we can merge.

_A note on commits (please @meh correct me):_ I don't think there's a preference for having a single commit in a PR, rather it's better to have focused commits with specific changes and just avoid the sequence of changes and reverts that tell an interesting story but make the use of `git blame` quite problematic. 

That said those are quite loose requirements in the spirit of keeping contributing enjoyable 🤓

vais:
@elia I think you put it so well that this could as well be copied and pasted into CONTRIBUTING.md

elia:
<3

opal/corelib/number.rb
```diff
@@ -710,6 +710,10 @@ def positive?
   def negative?
     `self == -Infinity || 1 / self < 0`
   end
+
+  def instance_variable_set(*)
+    raise RuntimeError, "can't modify frozen #{self.class.name}"
```
meh: Freezing is unsupported, I'd rather not have this.

meh:
Last nitpick then it's good to merge.

opal/corelib/number.rb
elia: @meh to me looks like it should fail anyway, looks odd to store ivars on a numeric object (even if I can confirm it works on latest safari) no strong feelings tho

meh: I'd just leave it as undefined behavior.

iliabylich:
Done. @elia could you merge it please? :smile:

elia:
Sure nice work!

PR #1185, opened 2015-11-09
doc: On the number of commits per PR

elia:
In response to @vais (https://github.com/opal/opal/pull/1180#issuecomment-155060845) :)


vais:
:+1:

PR #1186, opened 2015-11-10
Fix Kernel#instance_variables, Kernel#instance_variable_get and Kernel#instance_variable_set for reserved JS keywords as ivar names.

iliabylich:


opal/corelib/kernel.rb
```diff
@@ -779,16 +779,33 @@ def instance_variable_get(name)
...
+        name === "hasOwnProperty$" ||
+        name === "valueOf$")
+      {
+        return name.slice(0, name.length - 1);
+      }
+
+      return name;
+    }
+  }
+
```
elia: I think this can be simplified as `name.replace(/\$/g, '')` or alternatively `name.replace(/\$$/, '')`.

vais: @elia you mean remove all the `===`'s or just instead of `name.slice(...)`? In case it's the latter, wouldn't `name.slice` be more efficient than doing a regexp?

IlyaBylich: Initially I thought that we should not just blindly return all own properties that match some pattern. I mean
``` ruby
obj = Object.new
`#{obj}.attribute$ = 'some value'`
obj.instance_variables
```
should return a blank array, right? There are only a few reserved words in JS that must be explicitly escaped. I can revert this changes if you think that we can ignore such cases.

IlyaBylich: Check out the last spec in this PR, it shows exactly what I mean

elia: @iliabylich I see what you mean, I think we can ignore that check as we already look for all own properties that match a valid ivar name. Also MRI doesn't allow for `$` to be in an ivar name:

```
irb(main):001:0> o = Object.new
=> #<Object:0x007fad74c7c530>
irb(main):002:0> o.instance_variable_set '@asdf$', 123
NameError: `@asdf$' is not allowed as an instance variable name
	from (irb):2:in `instance_variable_set'
	from (irb):2
	from /Users/elia/.rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
irb(main):003:0> 
```

@vais good point, this should be better: `return name.substr(-1) === '$' ? name.slice(0, name.length-1) : name`

meh:
I don't really like the helper function.

elia:
> I don't really like the helper function.

Me too, maybe it makes sense to move it to `Opal.js2ivar` or something like that because I don't like this name either, as we all now _naming is hard!_

iliabylich:
Agree, changed to inline js

elia:
:+1: Looks good

elia:
@iliabylich would you mind adding a changelog entry for this?

iliabylich:
@elia Done. Didn't realize that this fix is important :smile:

PR #1188, opened 2015-11-12
fix JS.new to work with DOM objects

### LINKED ISSUE #1183: JS.new with native objects
Trying to make JS.new work with native objects, for example:

```ruby
# `new Websocket("ws://sdfsdf")
x = JS.global.JS['WebSocket']
JS.new(x, "ws://localhost")
#<Exception: Failed to construct 'WebSocket': Please use the 'new' operator, this DOM object constructor cannot be called as a function.>
```

I fugured it out that this implementation work in browser:

```ruby
  # Use new to create a new instance of the prototype of the function.
  def new(func, *args, &block)
    args.insert(0, `this`)
    args << block if block
    `new (#{func}.bind.apply(#{func}, #{args}))`
  end
```

But tests fails, because phantomjs not support `bind` (as well as older browsers I guess).
This is a `bind` implementation that works with phantomJS. 

```js
if (typeof Function.prototype.bind != 'function') {
    Function.prototype.bind = function bind(obj) {
        var args = Array.prototype.slice.call(arguments, 1),
            self = this,
            nop = function() {
            },
            bound = function() {
                return self.apply(
                    this instanceof nop ? this : (obj || {}), args.concat(
                        Array.prototype.slice.call(arguments)
                    )
                );
            };
        nop.prototype = this.prototype || {};
        bound.prototype = new nop();
        return bound;
    };
}
```

I need a little help to merge this `bind` implementation and `JS.new` to one method. What do you think about this solution ?

fazibear:
Fix #1183

### DIFF
--- a/stdlib/js.rb
+++ b/stdlib/js.rb
@@ -27,7 +27,7 @@ module JS
     def new(func, *args, &block)
       args.insert(0, `this`)
       args << block if block
-      `new (#{func}.bind.apply(#{func}, #{args}))`
+      `new (#{func}.bind.apply(#{func}, #{args}))()`
     end
   else
     def new(func, *args, &block)
--- a/stdlib/js.rb
+++ b/stdlib/js.rb
@@ -23,11 +23,19 @@ module JS
   end
 
   # Use new to create a new instance of the prototype of the function.
-  def new(func, *args, &block)
-    args << block if block
-    f = `function(){return func.apply(this, args)}`
-    f.JS[:prototype] = func.JS[:prototype]
-    `new f()`
+  if `typeof Function.prototype.bind == 'function'`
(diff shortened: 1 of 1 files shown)


meh:
Shouldn't the `if` check be inside the `def`?

That way it's only gonna be compiled once with whatever matches.

fazibear:
This `if` is outside because it check if environment have supports `bind`, so it can be executed once, while module is loaded. It's not necessary to check this condition every time `JS.new` is executed.

meh:
Ah yeah didn't read carefully, just drop the RubySpec update or put it in a separate commit and I'll merge.

fazibear:
Rubyspec changes reverted.

PR #1190, opened 2015-11-16
Clean up RubySpec bug filters

vais:
Unlocks :sparkles: 62 :sparkles: more passing specs FTW, brings the current total to 6408 :tada:


elia:
well done!

vais:
@elia sorry, I'm still confused about the existence of this file: https://github.com/opal/opal/blob/master/spec/filters/bugs/language_opal.rb

elia:
Those filters are for `spec/opal` specs only, all the other filters are for `spec/rubyspec`, that is because we now run the two suites separately, and to keep track of unused filters.

Since we are in control of `spec/opal` I think the best option would be to skip the filtered examples in `spec/opal` (by removing, commenting or marking them as pending) and remove `filters/bugs/language_opal.rb` entirely.

vais:
@elia, ok, but where do these examples come from? Are they duplicates of what's in RubySpec or are they "original works" in which case why are they not just contributed upstream? That's what I'm confused about...

elia:
@vais most of them are "forks" from rubyspec, that just avoids the parts we still not support (e.g. some splat syntaxes that still can't be parsed). Others are very opal-specific. Some others can surely be contributed upstream.

I'm quite sure there are some that can be deleted entirely as their original rubyspec version is now being run, and also many that are out of sync with rubyspec that changed since they were forked back in the days.

vais:
Oh boy, it's quite a project, but I think it needs to be done sooner or later. The only opal-specific specs we should have in the end are those that are... well... opal-specific! Language specs are a prime example of what's *not* in that category :wink:

PR #1191, opened 2015-11-17
Basic implementation of Marshal.load and Marshal.dump.

iliabylich:
Things that has been changed to make `Marshal` working:
+ `Array#instance_variables` doesn't return indexes
+ `Module#instance_variables` doesn't return constants
+ `String` subclasses do not return `@literal` in ivars list.

Limitations:
+ `Marshal.dump(:symbol)` - no Symbols support
+ String encoding - dumping/loading a string with custom encoding always dumps/loads utf-16 encoded string
+ Everything about String mutating - extended strings, ivars on strings and so on.
+ `Bignum`'s can't be encoded/decoded

218 examples from rubyspec are passing (`Marshal#restore` is an alias to `Marshal#load`, so probably even 300), ~50 are failing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,7 @@ Whitespace conventions:
 - Fixed an issue with `"-"` inside the second arg of `String#tr`
 - Fixed Base64 and enabled specs
 - Fixed method definition in method body.
+- Partially implemented `Marshal.load`/`Marshal.dump`. In order to use it require `opal/full`.
 
 
 ### Removed
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -2227,4 +2227,8 @@ class Array < `Array`
       }
     }
   end
+
+  def instance_variables
+    super.reject { |ivar| `/^@\d+$/.test(#{ivar})` || ivar == '@length' }
+  end
 end
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -175,7 +175,7 @@ class Hash
   end
 
   def default(key = undefined)
-     %x{
+    %x{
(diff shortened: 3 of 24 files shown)


iliabylich:
@elia I know in MRI `Marshal` is a part of corelib. What do you think about moving it to `stdlib` for Opal? Then it will not change the size of `opal/base` or `opal/mini`

elia:
@iliabylich this what I was thinking, let me know your considerations, I was pondering whether to introduce an `opal/full` and slim down the default opal including in it the things that are useful in the browser for most of the time, that way we can both 1. keep its size in check and 2. freely implement cool stuff like marshalling without fearing for the size (it's just a `require "corelib/THING"` away for those ho need it).

Just as a note I'll add that MRuby goes even further including in its basic bundle a dumbed down version of many methods (e.g. `gsub` can't accept a block and stuff like that). I think in our case it would make the core lib more fragile, but it's still a solution to keep in mind.

cc @meh

eregon:
FWIW, I fixed mspec upstream in https://github.com/ruby/mspec/commit/203ec67db1855bf650bda4069cce5d39eac656e8.

iliabylich:
@eregon Thank you so much! I was going to open a PR to mspec but for some reason I thought that it will not be be accepted :smile:

eregon:
@iliabylich @elia @vais Anything that makes `mspec` simpler is good and a welcome change!

iliabylich:
@eregon Thanks, pulling udpates!@elia I'll try to rebase it to the latest master (if it's possible :smile:). Just a final question, are you ok with excluding marshal stuff from `opal/opal` and forcing everyine to require 'corelib/marshal' manually?

elia:
@iliabylich I am, let's also add an `opal/full`

iliabylich:
Done

elia:
w00t!

PR #1193, opened 2015-11-18
Implement Kernel#remove_instance_variable

vais:


opal/corelib/kernel.rb
```diff
@@ -787,6 +787,20 @@ def instance_variable_set(name, value)
     `self[Opal.ivar(name.substr(1))] = value`
   end
 
+  def remove_instance_variable(name)
+    name = Opal.instance_variable_name!(name)
+
+    %x{
+      var key = Opal.ivar(name.substr(1)),
+          val = self[key];
+      if (val === undefined) {
```
elia: Any reason for not checking with `hasOwnProperty` here?

vais: Good point!

IlyaBylich: @elia Does `Kernel#instance_variable_get` also need this check?

elia: @IlyaBylich I think it makes sense, unless we discover some weird behavior of ivars makes sense to ignore any possible property coming from the prototype.

PR #1194, opened 2015-11-18
Implement shellwords (mostly)

elia:
There are a couple of failure in the few specs/tests.
Not sure if it's worth spending more time on this tho, long term target here is `optparse`.


hmdne:
We have shellwords now in 1.4! See: #2326

PR #1195, opened 2015-11-18
Adds Nashorn runner

Mogztter:
@elia You can add `-doe` option to see the stack trace.
`jjs` binary must be in your `PATH`.

With jdk1.8.72:
```
array.js:3611:6 Error: [object Object]
array.js:3611:6 Error: [object Object]
	at jdk.nashorn.internal.objects.NativeError.initException(NativeError.java:137)
	at jdk.nashorn.internal.objects.NativeError.captureStackTrace(NativeError.java:150)
	at jdk.nashorn.internal.scripts.Script$Recompilation$192$114426A$array.L:3917$L:3922$L:3931(array.js:3941)
	at jdk.nashorn.internal.scripts.Script$Recompilation$191$106310AA$array.L:2469$L:2480$L:3590(array.js:3606)
	at jdk.nashorn.internal.scripts.Script$Recompilation$187$array.L:14708$L:14713$TMP_1(array.js:14741)
	at jdk.nashorn.internal.scripts.Script$Recompilation$162$66464A$array.L:2266$L:2272$TMP_2(array.js:2336)
	at jdk.nashorn.internal.scripts.Script$Recompilation$184$417462AA$array.L:14708$L:14977$TMP_16(array.js:14990)
	at jdk.nashorn.internal.scripts.Script$Recompilation$180$417842A$array.L:14996(array.js:15002)
	at jdk.nashorn.internal.scripts.Script$array.:program(array.js:14996)
	at jdk.nashorn.internal.runtime.ScriptFunctionData.invoke(ScriptFunctionData.java:623)
	at jdk.nashorn.internal.runtime.ScriptFunction.invoke(ScriptFunction.java:494)
	at jdk.nashorn.internal.runtime.ScriptRuntime.apply(ScriptRuntime.java:393)
	at jdk.nashorn.tools.Shell.apply(Shell.java:393)
	at jdk.nashorn.tools.Shell.runScripts(Shell.java:322)
	at jdk.nashorn.tools.Shell.run(Shell.java:171)
	at jdk.nashorn.tools.Shell.main(Shell.java:135)
	at jdk.nashorn.tools.Shell.main(Shell.java:111)
```

.array.rb
```ruby
require 'set'

['', 'DEFAULT'].to_set
```


Mogztter:
NOTE: `puts` is not working so the following will also crash:

.hello.rb
```ruby
puts "Hello"
```

elia:
> NOTE: puts is not working so the following will also crash:

For that I think it's a good idea to add `stdlib/nashorn.rb` that implements as much Kernel/IO stuff as it makes sense, see `stdlib/phantomjs.rb` and `stdlib/nodejs.rb` as a ref.

Mogztter:
@elia How do you test the CLI locally ? I had to modify the code to use relative `require`, ex:
```ruby
#require 'opal/cli_options'
require File.join(File.dirname(__FILE__), '../lib/opal/cli_options')
```
But I think there's a better way to do it ? 
I also added `stdlib/nashorn.rb` to implement IO stdout/stderr write.

Did you manage to run Nashorn on your machine ? Do you have an idea why `to_set` is not working ?
Thanks!

elia:
Bundle exec should be enough to make the require work

    bundle exec bin/opal -R nashorn hello.rb

Didn't have a chance to try it yet, but once I get the time first thing is to try to run rubyspec with it (likely a variation of mspec_rubyspec_node)

elia:
Also don't forget to add a changelog entry! 😀

Mogztter:
> For that I think it's a good idea to add stdlib/nashorn.rb that implements as much Kernel/IO stuff as it makes sense, see stdlib/phantomjs.rb and stdlib/nodejs.rb as a ref.

I think `stdlib/nashorn.rb` is ignored because the JavaScript code is still:

```javascript
typeof(process) === 'object' ? function(s){process.stdout.write(s)} : function(s){console.log(s)}
```

As defined in [corelib/io.rb#68](https://github.com/opal/opal/blob/master/opal/corelib/io.rb#L68). If I replace `corelib/io` code with:

```ruby
$stdout.write_proc = `function(s){print(s)}`
$stderr.write_proc = `function(s){print(s)}`
```

Then the following is working:

.hello.rb
```ruby
puts "Hello"
```
```
$ bundle exec bin/opal -R nashorn hello.rb
Hello
```

Mogztter:
Wow and with this change, the following code is now working:

.array.rb
```ruby
require 'set'

puts "#{['', 'DEFAULT'].to_set}"
```

```
bundle exec bin/opal -R nashorn array.rb
#<Set:0x104>
```

Voodoo magic :exclamation:

elia:
Should work with 

     bundle exec bin/opal -R nashorn -r nashorn array.rb

As the runner (-R) and the require from stdlib (-r) are two different things

Mogztter:
@elia thanks for your help :+1: 

How this is gonna work if I want to build a version of Asciidoctor.js compatible with Node.js / Nashorn / Browser ?
Why Node.js implementation is defined in [corelib/io](https://github.com/opal/opal/blob/master/opal/corelib/io.rb#L68) ?
Should I also add Nashorn implementation in `corelib/io` ?
```ruby
STDOUT.write_proc = %x{
if (typeof(process) === 'object') {
  function(s){process.stdout.write(s)}
else if (typeof(console) === 'object') {
  function(s){console.log(s)}
} else {
  function(s){print(s)}
}
}
```

elia:
That's something I was thinking about too and was leaning toward something like stdlib/detect_platform.rb but still not sure about the name. Ideally it would detect the platform and require the right file. 

The reason for having the node implementation in corelib/io was a way - I think - to make life easier on the most popular platforms (browser and nodejs).

elia:
As of building asciidoctor if suggest doing a simple detection that will require the correct file, eg. :

    require "nodejs/io" if nodejs?
    require "nashorn/io" if nashorn?

Pseudo code of course but that's the idea.

elia:
Started trying to do something similar in opal-rspec 

https://github.com/opal/opal-rspec/blob/7ea91f869ba0858000af849f62bdfcea65a00bc8/opal/opal/rspec.rb

elia:
@Mogztter Just added [`opal/platform`](https://github.com/opal/opal/blob/eb4575b70e1b24c1a6e034b0e273a2542948ed8b/stdlib/opal/platform.rb) to stdlib, that is the file to update with the proper platform detection code for nashorn.

I think the next step is to update the stock write_proc to always use console.log (discarding the `process` part)

Mogztter:
@elia Cool! We have done something similar in Asciidoctor.js in the [opal_ext](https://github.com/asciidoctor/asciidoctor.js/tree/master/lib/asciidoctor) directory.
I wonder if you could review the opal_ext directory in Asciidoctor.js and tell me if I can port some changes in Opal ? We have mostly overridden I/O methods to make Asciidoctor works in different environment. I believe that some of the override are generic enough to improve the "out-of-the-box" experience in Opal. What do you think ?

elia:
@Mogztter looks like we converged to the same solution, everything in there is worth porting over to Opal. Maybe at this point it's worth moving everything under the `Platform` namespace or maybe try to mimic `rbconfig`… I'm open to suggestions on this.

Anyway that's outside the scope of this PR. 
In order to merge this the only thing missing is a changelog entry.

Mogztter:
> looks like we converged to the same solution

Indeed!

>  everything in there is worth porting over to Opal. 

:+1: 

> Anyway that's outside the scope of this PR. 

Yes, of course.

> In order to merge this the only thing missing is a changelog entry.

Done :smile:

elia:
Thanks!

You think there's a way to test it on Travis ?

Mogztter:
Apparently JDK 8 is available on Travis `oraclejdk8` but I don't know if `jjs` binary is available in the PATH. If this is the case then you can run the Nashorn engine on Travis !

PR #1198, opened 2015-11-19
Pass "Kernel#instance_variables immediate values returns the correct …

vais:
…array if an instance variable is added"

### DIFF
--- a/opal/corelib/numeric.rb
+++ b/opal/corelib/numeric.rb
@@ -186,4 +186,8 @@ class Numeric
   def clone
     raise TypeError, "can't clone #{self.class}"
   end
+
+  def instance_variable_set(name, value)
+    raise RuntimeError, "can't modify frozen #{self.class}"
+  end
 end
--- a/spec/filters/bugs/kernel.rb
+++ b/spec/filters/bugs/kernel.rb
@@ -44,7 +44,6 @@ opal_filter "Kernel" do
   fails "Kernel#instance_variable_set on frozen objects keeps stored object after any exceptions"
   fails "Kernel#instance_variable_set on frozen objects raises a RuntimeError when passed replacement is different from stored object"
   fails "Kernel#instance_variable_set on frozen objects raises a RuntimeError when passed replacement is identical to stored object"
-  fails "Kernel#instance_variables immediate values returns the correct array if an instance variable is added"
   fails "Kernel#is_a? returns true if given a Module that object has been extended with"
   fails "Kernel#iterator? is a private method"
   fails "Kernel#itself returns the receiver itself"


elia:
I think this makes sense, but since @meh [said otherwise][1] I'd like to discuss it a bit before merging.

As I see it Opal doesn't support _freezing_ but should acknowledge already "frozen" objects like numbers and strings, so that `#freeze` is not defined or is just a _noop_ but `#frozen?` works fine.

_Disney's Frozen :snowflake: pun goes here_

[1]: https://github.com/opal/opal/pull/1180#discussion-diff-44304223

vais:
@elia I think this case is different because numbers are "born" frozen - they are not frozen explicitly, they're just always frozen by nature.
```
$ ruby -ve "42.instance_variable_set(:@ultimate_question, 'the answer')"
ruby 2.0.0p645 (2015-04-13 revision 50299) [universal.x86_64-darwin15]
-e:1:in `instance_variable_set': can't modify frozen Fixnum (RuntimeError)
	from -e:1:in `<main>'
```

meh:
My opinion is every frozen related filter should just be moved to unsupported.

vais:
@meh, I'm confused... Setting aside RubySpec (bug filters, unsupported filters, all that) for a moment, the situation right now is that MRI raises a runtime error if you try to set an instance variable on a number, but Opal happily obliges. Is this part of a broader vision for a more mutable variant of Ruby? :trollface: Seriously though, the question is: would you like to see this issue addressed somehow, or simply move this spec into unsupported and leave the current behavior as is?

meh:
@vais yes, I have plans to support mutable numerics, they're very important to future projects I have in mind.

If I understand it correctly, the spec is not passing because RubySpec is checking something unrelated to what it should actually be checking for.

Which means the spec should be fixed and not the frozen support semi-added. Everyone using Opal should know from the start what is not unsupported, mutable strings, freezing, tainting; having it half-there is bad in my opinion.

vais:
> Everyone using Opal should know from the start what is not unsupported, mutable strings, freezing, tainting; having it half-there is bad in my opinion.

:+1: 

but:
> I have plans to support mutable numerics

:confused: 

You mean "upstream" in MRI as well?

meh:
@vais I was joking, but it was really tongue-in-cheek.

vais:
Oh, ok :sweat_smile:

@meh, I understand your point about half-supporting freezing et al, and I am 100% on board with that. However, I just don't think this particular case has anything to do with freezing other than the mention of *...frozen...* in the runtime error message text. The word *frozen* is not even mentioned anywhere in this spec.

I believe we should raise a runtime error in this case (RubySpec or no RubySpec), and perhaps the text of the message could be edited to not mention the word *frozen*?

meh:
I'd be fine with that, if not for the fact there are other core things that are frozen, `true`, `false`, `nil`, and possibly others, and they're frozen because freezing is supported.

So it's not just a matter of `#instance_variable_set` working or not.

vais:
> and they're frozen because freezing is supported

The way I see it, there is the *idea* of these objects being immutable, and then there's the specific *implementation* of this idea, which in MRI is *freezing*. I think these two things are orthogonal - we can support immutability of these objects without supporting freezing. In other words, I don't see these as in conflict with each other.

elia:
I don't care about the message either, but would like to see the error when I'll mistakenly try to assign ivars to a number 🤓

PR #1199, opened 2015-11-19
fix "backslash must appear before newline" when Opal ERB code block includes a function call with doublequotes

jakutis:


elia:
Thanks!

Issue #1200, opened 2015-11-20
Opal 0.9 Beta1 1->2 regression with cloning instance variables


wied03:
Broke opal-rspec.

Now that the `instance_variables` behavior is fixed, `#clone` is inconsistent with that and breaks opal-rspec in places.

example:
```ruby
class Foobar
  attr_reader :hello
  attr_accessor :bar
end

object = Foobar.new
puts "ivars #{object.instance_variables}"
cloned = object.clone
puts "cloned ivars #{cloned.instance_variables}"
```

MRI:
```
ivars []
cloned ivars []
```

Opal 0.9 beta 2:
```
ivars 
cloned ivars @constructor,@hello,@bar,@toString
```

PR #1201, opened 2015-11-20
Ensure cloned objects do not copy ivars that are not returned by #instance_variables

wied03:
Makes https://github.com/opal/opal/pull/1186 behavior consistent

Fixes https://github.com/opal/opal/issues/1200


wied03:
Wasn't sure if this spec belonged in rubyspecs but since it seems quasi related to https://github.com/opal/opal/pull/1186, put it in opal specs.

wied03:
Might be a valid debate here about cloning bridged objects but went for consistency for now since this broke an rspec matcher.

Issue #842, opened 2015-05-11
Can't redefine self.[] in a subclass of array


catmando:
```
class S < Array
  def self.[](*args)
    puts "i never get to this line"
  end
end

S[1, 2, 3]
```
removing subclass of array and all works well

changing name from self.[] to self.build and all works well

meh:
Subclasses of natively bridged classes are very very hacky right now, and they're hard to fix.

vais:
@catmando like @meh said, `String` and `Array` inheritance only kinda sorta works in Opal.

I was pretty bummed out about it at first, but then I remembered that I have never actually subclassed `String` or `Array` in my "normal" Ruby usage, and that **it is generally not recommended in Ruby to subclass core classes** (see [this excellent explanation by Steve Klabnik](http://words.steveklabnik.com/beware-subclassing-ruby-core-classes)).

So, subclassing core Ruby classes is a bad idea even in MRI where it obviously *does* work. In the end, the way I coped with this presently *not* working in Opal is I decided that I never liked it *anyway* :smile:

vais:
The more I think about it, @meh, http://words.steveklabnik.com/beware-subclassing-ruby-core-classes gives some very good reasons for pulling native inheritance out of Opal altogether. This way you can forget the native bridging rewrite as a nightmare. Or is that going too far?

cc @adambeynon @elia

Ajedi32:
I don't think it's a good idea to break compatibility with MRI except in places where we absolutely have to. Ideally we would be able to pass any pure-Ruby library or program into Opal and get working JavaScript code out the other end.

It's fine if you want to prioritize other features over subclassing of core classes (as I agree that that kind  of subclassing isn't something people should be doing under most circumstances), but I think for compatibility with existing programs the ability to subclass core classes is definitely useful.

meh:
Yeah, we're not going to disable it completely, but it is a low priority, especially since it's quite hard to do right.

elia:
fixed by #1206

Issue #1187, opened 2015-11-12
Class#method_name overrides Module#method_name


iliabylich:
``` ruby
class Class
  def my_method
    puts "Called from Class"
  end
end

class Module
  def my_method
    puts "Called from Module"
  end
end

Module.new.my_method
Class.new.my_method
```
Prints
```
Called from Class
Called from Class
```

Tested on Opal master

iliabylich:
Same issue:
``` ruby
class Array
  def method1
    puts 'Array called'
  end
end

class UserArray < Array
end

class Object
  def method1
    puts 'Object called'
  end
end

UserArray.new.method1
# => Object called
```

iliabylich:
Can the first example be somehow related to the fact that `Module.new.singleton_class` is a `Class`?

The second example is pretty weird, `UserArray` is actually an `Array::Wrapper` that relies on `method_missing` to delegate everything to its `@literal`. But there's a method defined on `Object`, so `method_missing` is never fired, and delegation doesn't work in that case. Don't even know how to fix that.

/cc @elia

elia:
> Can the first example be somehow related to the fact that Module.new.singleton_class is a Class?

That would be interesting and somehow makes sense…

> The second example…

I'd say that maybe having the wrapper inheriting from BasicObject could help, but I didn't check the code and it could be easily be already the case.

/cc @meh added to the 0.9 milestone, if it's not fixable in the current status feel free to revert

iliabylich:
It seems that the issue is with a `singleton_class` of  a`Module` instance
``` ruby
class Class
  def is_a_class?
    true
  end
end

[Module.new, Class.new].each do |mod|
  puts "For #{mod}"
  puts mod.singleton_class.methods.grep(/is_a_class/).inspect
  puts mod.singleton_class.instance_methods.grep(/is_a_class/).inspect
end
```

Output for opal master:
```
For #<Module:0x100>
["is_a_class?"]
["is_a_class?"]
For #<Class:0x102>
["is_a_class?"]
["is_a_class?"]
```

Output for ruby 2.2.3:
```
For #<Module:0x000000018723b8>
[:is_a_class?]
[]
For #<Class:0x00000001872390>
[:is_a_class?]
[:is_a_class?]
```

[...]

PR #1206, opened 2015-11-26
Make Array subclasses < Array and stop relying on method missing to avoid conflicts with ancestors.

iliabylich:
See #1187 for details.

I don't like how `Array#new` checks for a break in a passed block, but I don't see any other solution.

Also there's a performance regression:

+ `master`:
```
test/cruby/benchmark/bm_so_array.rb     34.48100018501282
test/cruby/benchmark/bm_vm2_array.rb    0.04900002479553223
test/cruby/benchmark/bm_vm2_bigarray.rb 0.03899979591369629
===============================================
Executed 3 benchmarks in 34.569000005722046 sec
```
+ `patch`:
```
test/cruby/benchmark/bm_so_array.rb     35.486000061035156
test/cruby/benchmark/bm_vm2_array.rb    0.046000003814697266
test/cruby/benchmark/bm_vm2_bigarray.rb 0.039999961853027344
==============================================
Executed 3 benchmarks in 35.57200002670288 sec
```


meh:
It was made like that for a reason, you can't change the class of a native `Array` because you can't touch the prototype chain.

Basically this patch shouldn't work at all, or did you somehow solve that problem and I missed it?

iliabylich:
@meh The idea was to make the following code working:
``` ruby
class Array
  def method1
    puts 'Array called'
  end
end

class UserArray < Array
end

class Object
  def method1
    puts 'Object called'
  end
end

UserArray.new.method1
# => 'Array called'
# before the patch it was 'Object called'
```
There's no need to change the prototype chain of `Array`, this PR is about broken `Array::Wrapper` (well, not broken, just not fully functional). This commit makes inheritance from `Array` real (previously it was resolved through delegation with fake inheritance).

meh:
It looks like the inheritance is actually working, there's just that failure to fix.

Could you also do the same for `String`?

vais:
:rocket: :sparkles:

vais:
@iliabylich that performance regression is truly negligible, keep going! :+1:

Side note: this change is not really about fixing the problem that *causes* https://github.com/opal/opal/issues/1187. Rather it fixes another big problem which, once fixed, *bypasses* the https://github.com/opal/opal/issues/1187 problem for subclasses of Array.

https://github.com/opal/opal/issues/1187 still remains to be solved.

iliabylich:
@meh I would like to split `Array`/`String` fixes to two separate PRs, it's hard to track changes in big PRs. Are you ok with it?
@vais You are right, these changes fix only the second item from #1187

iliabylich:
Does it make sense to add some utility method that converts a js array to a ruby array like:
``` javascript
function toArraySubClass(obj, klass) {
  if (obj.$$klass.$$name == 'Array') {
    return obj;
  } else {
    return klass.$allocate().$replace(obj.$to_a());
  }
}
```
and use it everywhere instead of `self.class.allocate`? This may give a serious performance improvement to `Array`-owned methods that take care about inheritance/subclasses stuff. Patched version of `Array.[]` is 100x times slower than original.

elia:
I wonder if this solves #842 too

elia:
> Does it make sense to add some utility method that converts a js array to a ruby array like:

Makes sense to me.

@meh there's anything else you want to be addressed before merging?

iliabylich:
@elia Yes, seems it fixes #842.
@vais 
``` ruby
class A < Array
end

puts `#{A.new}.length`
# => 0
```

iliabylich:
@elia Added `toArraySubclass` function to `Array` class, it should make `Array#` methods faster.

opal/corelib/array.rb
```diff
@@ -63,7 +76,15 @@ def initialize(size = nil, obj = nil, &block)
   end
 
   def self.new(*args, &block)
-    [].initialize(*args, &block)
+    obj = allocate
+    breaker = obj.initialize(*args, &block)
+    %x{
+      if (obj.$$has_breaker !== undefined) {
```
iliabylich: @vais In https://github.com/opal/opal/pull/1246 you had a similar problem with `break`. Do you think it's better to make the same thing here?
``` ruby
def initialize(*)
  # ...
  %x{
    value = block(i)
    if (value === $breaker) {
      return $breaker.$v
    }
  }
end

def self.new(*)
  obj = allocate
  breaker = obj.initialize
  breaker == obj ? obj : breaker
end
```

vais: @iliabylich LOL, that was [my question there as well](https://github.com/opal/opal/pull/1246/files#r47440335). If I understand correctly, you are checking the return value of `initialize`, and since it normally should be the object itself, you can *assume* that if it's *anything else*, then it *must* be, can *only ever be*, `$breaker.$v`. Right? That's basically what I'm doing with the return value of `String#upto` called from `Range#each`, but I have no idea if that's "the right way". Actually, there may not be a right way, given that the semantics of `break` appear to be [broken](https://github.com/opal/opal/issues/1247) at this time. ¯\_(ツ)_/¯

IlyaBylich: Yes, that's my assumption here, `Array#initialize` **usually** returns self. And I'm not sure if it's a good idea to rely on it :smile:

iliabylich: Yes, that's my assumption here, `Array#initialize` **usually** returns self. And I'm not sure if it's a good idea to rely on it :smile:

vais: It's probably not, but at least in my case that seemed to be the least bad of the available options, until the way `break` works is fixed.

elia: see #1250 ;)

iliabylich:
Rebased to the latest master, ready for merging (I guess). Thanks @elia for fixing an issue with `break`!

opal/corelib/array.rb
```diff
@@ -7,8 +7,18 @@ class Array < `Array`
   # Mark all javascript arrays as being valid ruby arrays
   `def.$$is_array = true`
 
+  %x{
+    function toArraySubclass(obj, klass) {
+      if (klass.$$name === 'Array') {
```
elia: this test looks a bit weak, any way to make it more reliable?

I'm thinking of stuff like:

```rb
module Mylib
  class Array < ::Array
  end
end
```

opal/corelib/array.rb
```diff
@@ -58,7 +70,9 @@ def initialize(size = nil, obj = nil, &block)
   end
 
   def self.new(*args, &block)
-    [].initialize(*args, &block)
+    obj = allocate
+    obj.initialize(*args, &block)
+    obj
   end
```
elia: If it relies on `#allocate` + `#initialize` it can be removed entirely

opal/corelib/array.rb
```diff
@@ -544,14 +561,14 @@ def clear
...
     copy.initialize_clone(self)
     copy
   end
 
   def dup
-    copy = []
+    copy = self.class.allocate
     copy.initialize_dup(self)
     copy
   end
```
elia: `#clone` and `#dup` can probably be removed to:

https://github.com/opal/opal/blob/a8669f9fd1433f192564ee4a2a43b04b39dd814e/opal/corelib/kernel.rb#L138-L163

iliabylich:
@elia Fixed. Any other comments?

opal/corelib/array.rb
```diff
@@ -7,8 +7,18 @@ class Array < `Array`
   # Mark all javascript arrays as being valid ruby arrays
   `def.$$is_array = true`
 
+  %x{
+    function toArraySubclass(obj, klass) {
+      if (klass.$name() === 'Array') {
```
elia: maybe `klass === Opal.Array`?

iliabylich:
@elia Agree, fixed

elia:
@iliabylich looks good to me, just add a changelog entry under `0.10.0 (edge)`

@meh you did the initial work for array inheritance, anything else before it's merged?

meh:
Looks good, the only reason is I have absolutely no idea why it's working :rage4:

iliabylich:
@elia Added a changelog entry

vais:
@elia @meh so... who's gonna press the `Merge` button? :stuck_out_tongue_winking_eye:

elia:
gun1 gun2

tumbleweed

meh:
I'll merge it, can you change the `whatever.$to_a()` to `#{`whatever`.to_a}` first?

elia:
@iliabylich ping :)

iliabylich:
@elia Sorry for delay :smile: Reverted all `$to_a()` back to `#{to_a}`, but after rebase one spec fails for me (and it also fails on the latest master) see #1271

iliabylich:
@elia nvm, can be merged

meh:
:rage4:

PR #1208, opened 2015-11-29
Optimize super as a bareword

jgaskins:
Currently, it's hardcoded in the compiler to use `$slice.call(arguments, 0)`. This PR just does the usual arguments->array copy via iteration while also making the JS code less terrible than I've been doing by declaring that target variable with the rest of the locals. :-)

Results from `rake bench` (from left to right: master, this PR, MRI):

```
test/cruby/benchmark/bm_vm2_zsuper.rb              13.864  1.780   0.547
```

As a bonus, I renamed the index variable for splat args and added a local variable for it. I hadn't intended to commit it but I apparently wasn't paying attention. :-) The only reason I changed it at all was because I originally used `arg_index` in both places, but `Struct.new` uses both splat args and bareword `super` so `var arg_index` in two places within the same scope caused JSHint to barf.

### DIFF
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -81,11 +81,11 @@ module Opal
 
           if scope.uses_zuper
             add_local '$zuper'
-            add_local 'zuper_index'
+            add_local '$zuper_index'
 
             line "$zuper = [];"
-            line "for(zuper_index = 0; zuper_index < arguments.length; zuper_index++) {"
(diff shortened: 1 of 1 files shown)


elia:
I'd use "$zuper_index" and "$splat_index" to avoid potential conflicts with local vars

jgaskins:
Ah, good idea.

jgaskins:
Alrighty, preceding dollar signs added, ready to go.

meh:
:+1:

PR #1209, opened 2015-11-29
Optimize Struct access

jgaskins:
When using methods on a struct, we don't need to check whether the struct has that as a member. If it doesn't we'll be invoking method_missing, not the subscript operator, so we can bypass all that and head straight to the good part.

Additionally, traversing the struct's members as an array is O(n), but checking whether the struct's data object has that key is O(1), so this commit speeds that up, as well.

Results from `rake bench` (from left to right: master, this PR, MRI):

```
test/cruby/benchmark/bm_vm2_struct_big_aref_hi.rb  12.814   0.796   0.261
test/cruby/benchmark/bm_vm2_struct_big_aref_lo.rb   9.565   0.794   0.266
test/cruby/benchmark/bm_vm2_struct_big_aset.rb      9.404   1.224   0.285
test/cruby/benchmark/bm_vm2_struct_small_aref.rb    5.989   0.049   0.190
test/cruby/benchmark/bm_vm2_struct_small_aset.rb    7.255   0.275   0.281
```

### DIFF
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -85,11 +85,7 @@ class Struct
     elsif String === name
       %x{
         if(!self.$$data.hasOwnProperty(name)) {
-          Opal.Kernel.$raise(
-            Opal.NameError.$new(
-              "no member '" + name + "' in struct", name
-            )
-          )
+          #{raise NameError.new("no member '#{name}' in struct", name)}
         }
       }
     else
--- a/opal/corelib/struct.rb
+++ b/opal/corelib/struct.rb
@@ -38,7 +38,7 @@ class Struct
     members << name
 
     define_method name do
-      self[name]
+      `self.$$data[name]`
     end
 
     define_method "#{name}=" do |value|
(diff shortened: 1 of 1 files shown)


jgaskins:
Updated benchmarks with setter methods

opal/corelib/struct.rb
```diff
@@ -83,7 +83,15 @@ def [](name)
 
       name = members[name]
     elsif String === name
-      raise NameError.new("no member '#{name}' in struct", name) unless members.include?(name.to_sym)
+      %x{
+        if(!self.$$data.hasOwnProperty(name)) {
+          Opal.Kernel.$raise(
+            Opal.NameError.$new(
+              "no member '" + name + "' in struct", name
+            )
```
elia: Unless it's a perfect bottleneck it's better to keep the raise in an interpolation for better consistency with the rest of the codebase

jgaskins: I had some trouble with the interpolation. I can't remember what it was, but it was really late last night and I don't think I paid attention to the failure so I just ended up doing the thing I knew would work. Maybe doing it when I'm more awake would help. :-)

elia: s/perfect/performance 

Yeah, "night time is the right time" :)
Just a stylistic note anyway, the code looks correct. 👍🏼

jgaskins:
Well, I'm not sure what I was doing last night that didn't work but tests are passing using the interpolation this time. :-)

PR #1210, opened 2015-12-01
BasicObject#instance_eval(str) now compliant

### LINKED ISSUE #480: BasicObject#instance_eval(string, [, filename [, lineno]] ) is not implemented

### LINKED ISSUE #499: instance_eval for Class doesn't define method
```
class Base
end
obj = Object.new
Base.instance_eval do 
  @a=3
  def foo
    puts @a
    puts self
  end
  
end

Base.foo #undefined method `foo' for Base
#which in ruby is ok.
```

### LINKED ISSUE #621: #instance_eval string-argument version half implemented or not correctly implemented ?
`#instance_eval` only support the block argument version and not the string argument version.

```
instance_eval {puts "ok"}`
=> "ok"

instance_eval("puts 'ok' ")`
ArgumentError: no block given
@ opalrb.org/javascripts/try.js:5:19601
...
```

if I do
```
instance_eval("puts 'ok' ") {}`
=> # nothing gets printed
```

and if I do:

```
instance_eval(puts "ok") {}`
=> "ok"
#while it should have said "no implicit conversion of nil into String" (at least on Ruby 2.0.0)
```

Tested in Opal online REPL

elia:
Needs `require "opal-parser"`.

fixes #621
fixes #480
fixes #499


vais:
:tada: :sparkles:

PR #1212, opened 2015-12-02
fix Native to work with Array

### LINKED ISSUE #1211: Array of Objects is not wrapped by Native
```ruby
x = `[{"a": "b"},{"a":"c"},{"a":"d"}]`
y = Native(x)

# this is work
puts Native(y[0])["a"]

# this is not
puts y[0]["a"]
```

cc @meh

fazibear:
fix #1211

### DIFF
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -217,6 +217,10 @@ module Kernel
       obj.map do |o|
         Native(o)
       end
+    elsif obj.is_a?(Proc)
+      proc do |*args|
+        Native(obj.call(*args))
+      end
     else
       obj
     end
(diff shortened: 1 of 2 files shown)


stdlib/native.rb
```diff
@@ -213,6 +213,10 @@ def Native(obj)
       nil
     elsif native?(obj)
       Native::Object.new(obj)
+    elsif obj.is_a?(Array)
```
elia: This is good but should be probably generalized to any bridged class (Array, Numeric, Proc, String, …)

I also have a slight concern about perf.

fazibear: `Array` and `Proc` not work properly, but `String`, `Float`, `Integer` work good.

meh: As a note, each bridged class has a `$$bridge` special property.

elia:
@fazibear thanks! 

Forgot to ask for a changelog entry before merging… (no worries, I'll add it :smile:)

PR #1213, opened 2015-12-02
Update to latest RubySpec

vais:
@opal/core adding `spec/a_file.rb` is a hack around this new problem we have with RubySpec: https://github.com/ruby/rubyspec/blob/master/fixtures/constants.rb#L287 (added in this commit https://github.com/ruby/rubyspec/commit/e78454c7467e89f44b3e81b55e46b8882eed8a10) Any ideas about what would be the "right" way for us to handle this?


elia:
@vais should be stubbed, probably in https://github.com/opal/opal/blob/master/tasks/testing.rake#L11-L18

that way the require should be ignored by the parser

elia:
eg.

    opal -e 'module M;autoload :A, "a_file"; end'
    # vs.
    opal -s a_file -e 'module M;autoload :A, "a_file"; end; p 4'

vais:
Thanks, @elia, that did it!

elia:
Now wondering what the MRI behavior is for using autoload on toplevel, bc noticed that it's kinda ignored by opal…

vais:
@elia I must confess that I have *never* used autoload :stuck_out_tongue_winking_eye: You are always told to avoid it, so I did. Kinda like the `@@` class vars, an evolutionary dead end.

PR #1214, opened 2015-12-03
Introduce opal/platform (stdlib) & cleanup testing.rake

elia:
**About opal/platform**

This file will take care of detecting the current platform and loading the
proper IO & Kernel adapters. It's supposed to be expanded to other
platforms too.

**About the cleanup**

- use CLI runners instead of dumping to a file to better test runners*
- use opal/platform to have a single task implementation for both node &
phantom
- end the platform names with "js" so that they match the CLI runners
- use a stock MSpec formatter (DottedFormatter)

This still leaves the need for separately testing the Sprockets stack.

* this will make debugging slightly harder by not writing the resulting
JS to disk, but it will still be possible to copy the command
outputted by rake and manually add "-c > output.js".

@opal/core heads up: the `mspec_*` task names have been changed


elia:
(as usual it's better to [ignore the whitespace](https://github.com/opal/opal/pull/1214/files?w=1) in order to see the actual changes in the diff)

vais:
Good idea, @elia! testing.rake was getting kinda out of control little by little :wink:

Issue #462, opened 2013-12-23
Instance methods have precedence over prefix operators


aost:
```ruby
# Code
puts -5
puts (-5).abs
puts -(5.abs)
puts -5.abs

# Opal output
-5
5
-5
-5

# MRI output
-5
5
-5
5
```

alexdowad:
The author of this grammar put an explicit precedence declaration on the rule which parses unary +/- followed by a number. It was given very, very low precedence to avoid a conflict with a rule which parses unary +/-, followed by a number, followed by the exponentiation operator.

alexdowad:
...Which is understandable, because `**` actually has higher precedence than unary +/-. So `-1 ** 2` is `-(1 ** 2)`, not `(-1) ** 1`.

alexdowad:
It looks like this was fixed somewhere since 2013, but nobody noticed!

alexdowad:
This should be closed.

elia:
Probably the issue is still there for method calls:

```
$ opal -e 'puts -5.abs' # => -5
$ ruby -e 'puts -5.abs' # => 5
```

elia:
No, wait, it's wrong only if it's preceded by another call:

```
$ opal --sexp -e '-5.abs'                                                     ~
(:call, (:int, -5), :abs, (:arglist))
$ opal --sexp -e 'puts -5.abs'                                                ~
(:call, nil, :puts, (:argils, (:call, (:call, (:int, 5), :abs, (:arglist)), :-@, (:arglist))))
```

alexdowad:
I've found that even MRI parses `-a.method` as `-(a.method)`. It seems to special-case numeric literals.

alexdowad:
(Good find about method calls, by the way.)

alexdowad:
My updated PR fixes this. I'm not versed in all the intricacies of your lexer/parser, so hopefully there won't be any adverse side effects! The fix *seems* right, but it would be nice if the original author also had a look.

elia:
Looks good, and the test suite should a good protection against side effects 👍🏼

(Unfortunately the original author isn't active on the project anymore)

alexdowad:
This is fixed and should be closed.

elia:
Fixed by #1215

PR #1215, opened 2015-12-03
Accept { 'symbol literal': val } syntax

### LINKED ISSUE #1202: Symbol from String as Hash key
```ruby
hash = { 'test': :test }
```

throws exception

```
parse error on value ":" (tCOLON) :application:24
Source: application:24:14
sh = { 'test': :test }
~~~~~~~~~~~~~^
```

alexdowad:
This is intended to fix #1202.

BUT! But, test coverage is still needed! You can see I tried to add a test, but it seems that it's not even running. Can you point me to the best place to add tests for this syntax?

Rather than just testing that the parser returns the right AST, I'd like to actually test the semantics of expressions which use this syntax. I'd also like to add a test ensuring that the semantics of `cond ? a : b` are not broken.


alexdowad:
Ooh, I forgot to regenerate the parser too. But tests come first.

alexdowad:
I just slipped a fix for #462 in here as well. But likewise, it still needs tests!

alexdowad:
Tried adding a test, but I am still not sure at all that these tests are even running. Some help, please!

alexdowad:
BTW, I love the way you use a bit vector as a stack of `true` and `false` values. Bitshift to push a `false`, bitshift and bitwise AND to push a `true`, another bitshift to pop. `parser` could gain some performance by doing the same thing.

elia:
The specs look good, a bit in a hurry now (7.50am) but you can look at CONTRIBUTING.md to see how to deal with and add rubyspec specs. 

All the cool stuff in the gramma/parser/lexer/compiler is almost entirely @adambeynon's work (props to him, once again)

vais:
@alexdowad in order to run [the RubySpec's symbol spec](https://github.com/ruby/rubyspec/blob/master/language/symbol_spec.rb), just remove the following line from our `specs/rubyspecs` file: https://github.com/opal/opal/blob/master/spec/rubyspecs#L129

The `!` at the beginning of that line means that this specific file is currently *excluded* from spec runs even though the entire `rubyspec/language` directory is *included*. I tried explaining how this works [here](https://github.com/opal/opal/blob/master/CONTRIBUTING.md#down-the-rabbit-hole) :wink:

spec/lib/parser/unary_spec.rb
```diff
@@ -5,6 +5,10 @@
     context 'with an integer' do
       it "parses unary op. with the right precedence" do
         parsed("-1.hello").should == [:call, [:int, -1], :hello, [:arglist]]
+        parsed("-1 ** 2").should == [:call,
+                                      [:call, [:int, 1], :**, [:arglist, [:int, 2]]],
+                                     :-@,
+                                     [:arglist]]
```
elia: It's probably better to keep this in a single line (not super important tho)

alexdowad:
@vais, thanks! But that spec also contains the new `%i{}` syntax.

spec/lib/parser/unary_spec.rb
alexdowad: Updated.

alexdowad:
I found what looks like a good test at `spec/opal/core/language/versions/hash_1.9.rb`, but RSpec doesn't run it because it isn't suffixed with `_spec.rb`. Is this intentional? I have given it a `_spec.rb` suffix here, just to make it run, but I'm sure there is a better way to do this.

elia:
That file should either match the spec pattern or be required by another file. 

If it adds good specs I'm ok with renaming it.

elia:
Btw I thought the %i syntax was supported …

alexdowad:
> I thought the %i syntax was supported …

It's implemented incorrectly:

```
20:30 ~prog/Ruby/opal % pry                                  [str_sym_literals_in_hash]
[1] pry(main)> $:.unshift Dir.pwd+'/lib'
=> ["/home/alex/Programming/Ruby/opal/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/coderay-1.1.0/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/slop-3.6.0/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/method_source-0.8.2/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/pry-0.10.3/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/yard-0.8.7.6/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/pry-doc-0.8.0/lib",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/site_ruby/2.2.0",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/site_ruby/2.2.0/x86_64-linux",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/site_ruby",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/vendor_ruby/2.2.0",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/vendor_ruby/2.2.0/x86_64-linux",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/vendor_ruby",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/2.2.0",
 "/home/alex/.rbenv/versions/2.2.2/lib/ruby/2.2.0/x86_64-linux"]
[2] pry(main)> require 'opal'
=> true
[3] pry(main)> Opal::Parser.new.parse("%i{a b c}")
=> (:array, (:str, "a"), (:str, "b"), (:str, "c"))
[4] pry(main)> Opal::Parser.new.parse("[:a, :b, :c]")                                   
=> (:array, (:sym, :a), (:sym, :b), (:sym, :c))
[5] pry(main)> %i{a b c}
=> [:a, :b, :c]
```

alexdowad:
> That file should either match the spec pattern or be required by another file

Grepping for the original name of the file doesn't find anything...

And yes, it does add some good specs.

elia:
%i has been probably implemented by extending the %q case, and since in opal strings and symbols are the same it shouldn't be a problem, but of course the implementation is not strictly "correct"

alexdowad:
Thank you! If any further problems come up with this, please let me know so I can work on it more!

elia:
Thank you for you help, you make changing the grammar look easy! 😸

PR #1222, opened 2015-12-05
Resolves #1220, Implement File.read in Nashorn

### LINKED ISSUE #1220: Implement File.read in Nashorn (Java8+ JavaScript engine)
Here is the implementation we are using:

```javascript
var Paths = Java.type('java.nio.file.Paths');
var Files = Java.type('java.nio.file.Files');
var lines = Files.readAllLines(Paths.get(path), Java.type('java.nio.charset.StandardCharsets').UTF_8);
var data = [];
lines.forEach(function(line) { data.push(line); });
return data.join("\n");
```

We could maybe do something even simpler: http://www.adam-bien.com/roller/abien/entry/java_8_reading_a_file but I don't know if this is working in Nashorn.

Mogztter:


stdlib/nashorn/file.rb
```diff
@@ -0,0 +1,13 @@
+class File < IO
```
elia: Can you remove the `< IO` part just to avoid conflicts if the superclass of `File` ever changes?

elia:
Thanks, I think should tweak the pattern a bit and maybe move the platform detection to `corelib/constants` and have `opal/platform` just `case OPAL_PLATFORM`. Also I was unsure about having the implementation for `File` required automatically but I think it makes sense for most use cases.

Mogztter:
You are welcome :smile: 

> Also I was unsure about having the implementation for File required automatically but I think it makes sense for most use cases.

What is the drawback ? A bigger Javascript file ?

PR #1223, opened 2015-12-05
Correct defaults for Date.new

vais:


elia:
TIL Date::Infinity 🤓

vais:
@elia that makes two of us :)

PR #1224, opened 2015-12-05
Fix parsing of `method do ... end` inside lambda body

### LINKED ISSUE #544: kDO_BLOCK inside ->{} produces SyntaxError
```ruby
a -> { b do |c|
end }
```

```
/Users/elia/Code/opal/lib/opal/parser.rb:81:in `on_error': An error occurred while compiling: -e (RuntimeError)

parse error on value "do" (kDO_BLOCK) :-e:1
Source: -e:1:11
a -> { b do |c|
~~~~~~~~~~^
	from /Users/elia/.rvm/gems/ruby-2.2.3/gems/racc-1.4.12/lib/racc/parser.rb:271:in `_racc_do_parse_c'
	from /Users/elia/.rvm/gems/ruby-2.2.3/gems/racc-1.4.12/lib/racc/parser.rb:271:in `do_parse'
	from /Users/elia/Code/opal/lib/opal/parser.rb:53:in `parse_to_sexp'
	from /Users/elia/Code/opal/lib/opal/parser.rb:38:in `parse'
	from /Users/elia/Code/opal/lib/opal/compiler.rb:149:in `compile'
	from /Users/elia/Code/opal/lib/opal/builder_processors.rb:91:in `compiled'
	from /Users/elia/Code/opal/lib/opal/builder_processors.rb:101:in `requires'
	from /Users/elia/Code/opal/lib/opal/builder.rb:41:in `build_str'
	from /Users/elia/Code/opal/lib/opal/cli.rb:110:in `block in build'
	from /Users/elia/Code/opal/lib/opal/cli.rb:154:in `evals_or_file'
	from /Users/elia/Code/opal/lib/opal/cli.rb:109:in `build'
	from /Users/elia/Code/opal/lib/opal/cli.rb:119:in `compiled_source'
	from /Users/elia/Code/opal/lib/opal/cli.rb:85:in `run_code'
	from /Users/elia/Code/opal/lib/opal/cli.rb:69:in `run'
	from bin/opal:22:in `<main>'
```

**update** updated the error output using master

alexdowad:
The lexer has a special state in which it lexes "do" as kDO_BLOCK rather than
kDO. It enters this state when parsing the arguments for a "command". (I am
not 100% sure, but I believe a "command" is a method call without explicit
receiver and without parentheses for arguments.)

Although this 'cmdarg' state was entered when parsing the argument list
(even an empty argument list) for a command, it wasn't cleared when entering
the body of a subsequent { } block. This meant that some constructs could not
be parsed correctly inside a { } block... unless there happened to be an
intervening construct which cleared the 'cmdarg' lexer state.

So, clear the 'cmdarg' state when entering a { } body, and restore it when
exiting.

fixes #544


elia:
@alexdowad added the `fixes` ref to the description, thanks a lot!

I'm also quite interested in learning more about your debug process, like the actual commands you use to parse with debug flag, then check the lexer etc. Documenting it a bit would serve the double purpose of allowing me to learn and improve our CONTRIBUTING.md guide (soon to be renamed to "Opal Hacking Guide" :wink:).

alexdowad:
Here is what I did:

1. In `tasks/building.rake`, added `-t` flag to invocation of racc.
2. `rake racc`
3. `pry`
4. Try parsing the sample code:

```
$:.unshift Dir.pwd + '/lib'
require 'opal'
p = Opal::Parser.new
p.instance_eval { @yydebug = true }
p.parse("sample code")

5. The Racc-generated parser will print a bunch of stuff at the console as it works -- each token it reads, when it decides to shift, when it decides to reduce, etc.
6. Read through the debug trace, noted that as soon as the parser hit `kDO_BLOCK`, it reduced the preceding `tIDENTIFIER` down as far as it could, then threw an error. LALR parsers tend to do this -- when they hit an illegal token, they will often perform one or more reduces before getting really stuck and then erroring out. LR parsers, on the other hand, know exactly what tokens are valid at each and every point, and will error out the moment they see an invalid token. But they have a far greater number of parser states. (I digress.)
7. At this point, I had to start examining the grammar, looking back and forth between the grammar and the debug trace, and trying to figure out what "path" the parser was trying to follow. What rules was it "expecting" to use later on? Why didn't it "think" that the rule which includes `kDO_BLOCK` was a possibility?
8. For comparison, I also tried using `parser` in debug mode and looked over its parser debug trace as well.

vais:
@alexdowad :bow:

PR #1225, opened 2015-12-05
Fix parsing of `return %{string}`

alexdowad:
The (ludicrously convoluted) Ruby lexer has a number of states, which affect
how various constructs are lexed. There is a state called 'expr_mid', which is
entered after lexing 'return', 'rescue', and a couple of other keywords.

'expr_mid' state is very, very, very, very much like 'expr_beg' state. I still
haven't figured out exactly why it is needed. Anyways, when in expr_mid state,
we should lex %{ just like expr_beg state.

### DIFF
--- a/lib/opal/parser/lexer.rb
+++ b/lib/opal/parser/lexer.rb
@@ -899,7 +899,9 @@ module Opal
             @lex_state = :expr_beg
             return new_op_asgn('%')
           elsif check(/[^\s]/)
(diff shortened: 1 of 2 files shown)


elia:
Thanks! I'm updating the contributing file with your instructions and also made easier to show Racc debug output in ce9aa248a96024a27971a5dbe96de0457e2984dd.

(tip: https://help.github.com/articles/closing-issues-via-commit-messages ← this works for PR descriptions too)

PR #1231, opened 2015-12-08
Fix #1230 (interpolating an empty expr into string)

### LINKED ISSUE #1230: SyntaxError: Unexpected token ')'
Invalid javascript generated when trying to interpolate an empty expression `"#{}"`

http://opalrb.org/try/?code:%22%23%7B%7D%22

vais:
¯\_(ツ)_/¯

### DIFF
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -129,7 +129,7 @@ module Opal
             push part.inspect
           elsif part.type == :evstr
             push "("
-            push expr(part[1])
+            push part[1] ? expr(part[1]) : '""'
             push ")"
           elsif part.type == :str
             push part[1].inspect


elia:
@vais i suggest to add a regression test for this, the one in #1230 looks good enough

elia:
aaand there's no spec proposed in #1230, sorry 😖

(I thought I saw it in one of the email notifications, will get some sleep and try again)

vais:
@elia this definitely needs a test, just wasn't sure where to put it :blush: LMK...

elia:
I'd say in rubyspec under "language", even if it could be argued that it can be seen as a buggish behavior…

Cc @eregon

eregon:
Yeah, this is mostly a parser thing, so language seems right.
An empty expression like that actually evaluates to `nil`:
```ruby
[1] pry(main)> def nil.to_s; "ahah"; end
=> :to_s
[2] pry(main)ahah> "#{}"
=> "ahah"
```
Feel free to implement or not that behavior though :wink:

PR #1232, opened 2015-12-08
Fix method lookup on a Module when the same method was defined on a Class.

### LINKED ISSUE #1187: Class#method_name overrides Module#method_name

iliabylich:
Closes #1187

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -1,8 +1,7 @@
 class Module
   def self.new(&block)
     %x{
-      function AnonModule(){}
-      var klass         = Opal.boot_class(Opal.Module, AnonModule);
+      var klass         = Opal.boot_module_object();
       klass.$$name      = nil;
       klass.$$class     = Opal.Module;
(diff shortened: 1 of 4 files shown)


iliabylich:
I"m not sure, but it seems that `Module.new` was inherited from a `Class` (which is inherited from a `Module`). That's why it had both `Class`/`Module` methods. At least chrome inspector shows this prototype inheritance chain.

elia:
Looks good, although I wonder if there's a way to test this without touching Module and Class directly. 

@meh any other comments?

elia:
@iliabylich sorry, my last PR introduced some conflicts, can you rebase? I'll merge afterwards :)

iliabylich:
@elia Done!

PR #1233, opened 2015-12-08
Fix parsing of do..end block after lambda argument

### LINKED ISSUE #1228: Stabby lambda without parantheses breaks the parser
Hello.

The following code breaks Opal (master @ 9d3d2ec):

```ruby
def within(some_proc, &block)
end

# This breaks: (Giving error: parse error on value "do" (kDO))
within ->{ whatever } do
end

# This works:
within(->{ whatever }) do
end
```

This code worked fine with Opal 0.8.1, so looks like a regression to me.

alexdowad:
In 7b5e479, a `lexer.cmdarg_push` and `lexer.cmdarg_pop` pair was added
around the parsing of `lambda_body`. This ensured that when parsing the lambda
body, the lexer started out in the correct state.

The problem was that when lexing the terminating RCURLY of the lambda body,
the lexer would *also* perform a `lexer.cmdarg_pop`. So we had push + pop + pop.
Enough to get the lexer in the right state when starting the lambda body, but
not enough to get it back into the right state after finishing the lambda body.

So, remove the previous fix, and instead, make sure the lexer performs a
`cmdarg_push` when lexing LAMBEG, as it already does for LCURLY. This way,
the pushes and pops are balanced.

EXCEPT... the grammar also allows a STRING_DBEG token to be matched with
RCURLY. And STRING_DBEG does not perform a `cmdarg_push`. But RCURLY still
does the `cmdarg_pop`.

There are bound to be problems in some cases which use dstrs... we just
haven't found them yet. Anyways, this does fix one problem at least.

Fixes #1228.


elia:
👍🏼

PR #1246, opened 2015-12-13
Range#to_a fully compliant with RubySpec

vais:


opal/corelib/range.rb
```diff
@@ -38,6 +38,36 @@ def cover?(value)
...
+      if (#@begin.$$is_string && #@end.$$is_string) {
+        value = #{@begin.upto(@end, @exclude, &block)};
+
+        // The following is a bit hackish: we know that
+        // String#upto normally returns self, but may
+        // return a different value if there's a `break`
+        // statement in the supplied block. We need to
+        // propagate this `break` value here, so we
+        // test for equality with `@begin` string to
+        // determine the return value:
```
vais: I would love to hear your thoughts on this. What's the current "best practice" for propagating the `break` value, e.g. in this case where we delegate to another method which takes a block, so we are passing the block down the line, but we still need to know if `break` was used somewhere (anywhere, arbitrarily deeply) down the line... :confused:

opal/corelib/string.rb
```diff
@@ -1575,7 +1575,7 @@ def upto(stop, excl = false, &block)
     return enum_for :upto, stop, excl unless block_given?
     stop = Opal.coerce_to(stop, String, :to_str)
     %x{
-      var a, b, s = self.toString();
+      var a, b, s = self.toString(), value;
```
vais: Here I realized that there's probably a bunch of methods that currently do not respect/support `break` values. This fixes `String#upto`, so that:
```ruby
'a'.upto('z'){break 'hello'} #=> "hello"
```

vais: @eregon I tried to look for how RubySpec currently handles testing methods that take a block which can have a `break` with a value called inside of it. I don't see much... e.g. https://github.com/ruby/rubyspec/blob/master/core/integer/times_spec.rb#L35 Am I missing something? Is this handled in some generic fashion elsewhere? Otherwise it looks like this case is very under-represented.

eregon: Yes, `break` is a language construct, it breaks out of the call site with the block lexically containing the `break`
(like an invisible exception implicitly caught by the call site with the block):
```ruby
ret = foo.bla(1,2,3) { break 42 if rand < 0.5 }
# execution continues here
```
would break of the full call to `#bla` and `ret` would be assigned the value passed to break.
This should be in language specs, and maybe it makes sense to add a comment to specs using `break` in core to clarify it's not method-specific (or alternatively remove them since it's not method specific).
Sorry for the confusion, we had the same problem in JRuby+Truffle as well.

vais: Thank you, @eregon, I understand. We have a problem with this in Opal, I have opened issue https://github.com/opal/opal/issues/1247. No idea how to go about resolving this... @elia, @meh?

eregon: It's quite similar to a return, but provided by the caller, so maybe it could be implemented similarly.
(I would guess an exception + some ID to avoid returning/breaking not far enough)

meh: @eregon the problem with that is that `try/catch` prevents optimizations from the engine.

eregon: Yeah, I thought that might be an issue :disappointed: , how do you implement `return`? What about non-local `return` or `return` in blocks?

meh: @eregon it uses `$returner` instead of `$breaker`, so it's likely broken as well.

eregon: I meant in general, not for this particular code. Or do you have any pointer to the code maybe?

meh:
```ruby
def test; lol { return 2 }; end
```

```javascript
/* Generated by Opal 0.9.0.dev */
(function(Opal) {

  Opal.add_stubs(['$lol']);
  return (Opal.defn(Opal.Object, '$test', function() {try {

    var $a, $b, TMP_1, self = this;

    return ($a = ($b = self).$lol, $a.$$p = (TMP_1 = function(){var self = TMP_1.$$s || this;

    Opal.ret(2)}, TMP_1.$$s = self, TMP_1), $a).call($b);
    } catch ($returner) { if ($returner === Opal.returner) { return $returner.$v } throw $returner; }
  }), nil) && 'test'
})(Opal);
```

vais: Thanks for your help, @eregon! You can follow this issue if you're interested: https://github.com/opal/opal/issues/1247#issuecomment-164503043

alexdowad:
Thanks @vais!

PR #1250, opened 2015-12-16
Treat `break` with try/catch from the call place

### LINKED ISSUE #1247: `break` does not break out of more than one level deep
```ruby
def m1(&block)
  yield
  'm1'
end

p(m1{break 42})
```
Result in Opal: `42`, same as MRI.

Let's add more levels:
```ruby
def m2
  yield
  'm2'
end

def m1(&block)
  m2 &block
  'm1'
end

p(m1{break 42})
```
Result in Opal: `"m1"`
Result in MRI: `42`

elia:
fixes #1247


meh:
I think all the `Enumerable` and everywhere else magic that does manual checking of break should be reverted to not doing it.

That's likely why it's failing.

iliabylich:
@elia The following example is very strange (tested on your branch)
``` ruby
def self.it(description, &block)
  block.call
end

result = it 'does something' do
  value = [1,2,3].all? { break 10 }
  puts value
  value
end

puts result
# Opal output:
10
# MRI output
10
10
```

Looks like `$breaker` gets caught too early:
``` js
Opal.defs(self, '$it', TMP_1 = function(description) {
    var self = this,
        $iter = TMP_1.$$p,
        block = $iter || nil;

    TMP_1.$$p = null;
    return block.$call();
});
result = (function() {
    try {
        ($a = ($b = self).$it, $a.$$p = (TMP_2 = function() {
            var self = TMP_2.$$s || this,
                $a, $b, TMP_3, value = nil;

            value = ($a = ($b = [1, 2, 3])['$all?'], $a.$$p = (TMP_3 = function() {
                var self = TMP_3.$$s || this;

                Opal.brk(10)
            }, TMP_3.$$s = self, TMP_3), $a).call($b);
            self.$puts(value);
            return value;
        }, TMP_2.$$s = self, TMP_2), $a).call($b, "does something")
    } catch ($breaker) {
        if ($breaker === Opal.breaker) {
            return $breaker.$v
        }
        throw $breaker;
    }
})();
return (function() {
    try {
        self.$puts(result)
    } catch ($breaker) {
        if ($breaker === Opal.breaker) {
            return $breaker.$v
        }
        throw $breaker;
    }
})();
```

It also happens on CI with some tests that invoke `break`.
`self.$puts(value)` definitely should be somewhere in the `catch` section

elia:
@iliabylich yes I think I mis-interpreted the semantics, I thought the `break` worked up to the topmost block, but instead should just stop at the first one (see also [this comment][1] by @eregon)

[1]: https://github.com/opal/opal/issues/1247#issuecomment-165059594

elia:
@iliabylich the latest commit fixes your example, thanks!

iliabylich:
@elia I think the issue is still there, check out the latest build. `break` stops the whole test suite.

elia:
@iliabylich yeah, need to check, the CI is expected to fail anyway, as @meh pointed out all previous uses of `$breaker` need to be fixed (e.g. in `Array` and `Enumerable`)

vais:
@elia :bow: 

Shutting down for the night now... if you think all that's left to do is fix the usage of `$breaker` throughout, I may be able to carve out a couple hours tomorrow to get this done. LMK...

elia:
@vais I have a first (half?) pass done locally, I'll see how far I can get before night and then be happy to pass the ball :rugby_football: :smile: (thanks!)

elia:
_status update_

### the good news

only one spec failing

### the bad news

only one spec failing

---

this code

```rb
lambda { break }.call
```

as of now compiles more or less to this:

```js
(function(){
  var $brk = Opal.new_brk();
  try {
    self.$lambda.$$p = (TMP_1 = function(){var self = TMP_1.$$s || this;
      Opal.brk(nil, $brk); // <~~~~ break
    }, TMP_1.$$s = self, TMP_1)
    return self.$lambda();
  } catch (brk) {
    if (brk === $brk) { return brk.$v } else { throw brk } 
  }
})().$call() // <~~~~ Proc#call
```

because the break is catched in the context of the method call owning the block.

### Solution?

What I think could be the solution is that basically the `$brk` object that's being thrown needs to be registered on the function/block itself, so that `Proc#call` can catch it and return the associated value.

Probably something link this:

```js
… TMP_1.$$s = self, TMP_1.$$brk = $brk …
```

I'd like to hear a second opinion though, maybe there's a better way :)

cc @opal/core @eregon @iliabylich

elia:
Ok as usual explaining to someone else (or a rubber duck) helps. It's implemented in 6a661d6 and seems to work fine! 🐳

elia:
so, need to take a _break_ now :stuck_out_tongue_winking_eye: 

(@vais if you have some time there's still to do a pass to start removing the legacy `$breaker` checks from the corelib… fell free to push to this branch, I'll let you know if I need to force push)

eregon:
@elia About the Proc#call case: I forgot about that one, it behaves as if block-calling methods catch it, so Proc#call but also `yield`, {instance,class}_{eval,exec}:
```ruby
[1] pry(main)> l=lambda {break 42}
=> #<Proc:0x005615f3878ac0@(pry):1 (lambda)>
[2] pry(main)> def foo; p yield; :after; end
=> :foo
[3] pry(main)> foo(&l)
42
=> :after
```
An alternative is of course to catch it at the top of the lambda compiled body, but then it must be done eagerly.

elia:
@eregon thanks for the heads up, I think I know how to do it…

PR #1256, opened 2015-12-24
Return original value if it cannot be nativized

jgaskins:
Prior to #1249, `{ foo: MyObject.new }.to_n` returned the object value, but now it returns `nil` if `MyObject#to_n` is not defined. This occurred because of the move to `Native.try_convert` for the conversion (which just lets native JS objects pass through).

I think using that method was a great move, but I don't think `nil` as a return value for arbitrary Ruby objects makes sense. If I understand the intent correctly, the idea is to always return a native JS value. The original Ruby object may be closer to the JS object you want than `nil`. That is, I can think of a lot of ways to use the original Ruby object from JS, but if I get `nil` in JS, I probably actually wanted `null`.

### DIFF
--- a/spec/opal/stdlib/native/hash_spec.rb
+++ b/spec/opal/stdlib/native/hash_spec.rb
@@ -40,5 +40,12 @@ describe Hash do
       native = obj.to_n
       `#{native}.a_key.key`.should == 1
     end
+
+    it 'passes Ruby objects that cannot be converted' do
+      object = Object.new
+      hash = { foo: object }
+      native = hash.to_n
+      expect(`#{native}.foo`).to eq object
+    end
   end
 end
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -24,7 +24,7 @@ module Native
     }
   end
 
-  def self.try_convert(value)
+  def self.try_convert(value, default=nil)
     %x{
       if (#{native?(value)}) {
         return #{value};
@@ -33,7 +33,7 @@ module Native
         return #{value.to_n};
       }
       else {
-        return value;
+        return #{default};
       }
     }
   end
@@ -472,7 +472,7 @@ class Struct
     result = `{}`
 
     each_pair {|name, value|
-      `#{result}[#{name}] = #{Native.try_convert(value)}`
+      `#{result}[#{name}] = #{Native.try_convert(value, value)}`
     }
 
     result
@@ -488,7 +488,7 @@ class Array
       for (var i = 0, length = self.length; i < length; i++) {
         var obj = self[i];
 
-        result.push(#{Native.try_convert(`obj`)});
+        result.push(#{Native.try_convert(`obj`, `obj`)});
       }
 
       return result;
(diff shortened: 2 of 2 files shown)


iliabylich:
Yeah, returning back `nil` is not a good idea. But usually (I mean, in MRI) when you have an object that can't be converted to some type, you get back an error, right?
``` ruby
[1,2,3].join({})
TypeError: no implicit conversion of Hash into String
```
In Opal `Helpers` module [is responsible](https://github.com/opal/opal/blob/master/opal/corelib/helpers.rb#L14-L32) for coercing and it always raises an error in such cases.

Probably raising something like `TypeError` looks more natural here, because actually Ruby objects are not so close to JS objects :smile: .

jgaskins:
@iliabylich The `try_convert` method is the "safe" version of the [`convert` method](https://github.com/opal/opal/blob/d1c442664a200dfaf46919133fae38c7c5b116b9/stdlib/native.rb#L41-L53), which does raise an error if it can't convert. :-)

elia:
Maybe a #to_n method can be defined on Object that internally just calls self.instance_variables.to_n

jgaskins:
Admittedly, in an ideal scenario where I intend to allow an object to be passed unmodified to JS libs with `to_n`, I'd implement the method on my Ruby object to return `self`. This PR is based purely on the fact that something that worked prior to #1249 changed in what looks to be an unintentional way and just restores that behavior.

I'm totally cool with it being rejected if the new behavior is closer to what is desired, but I'd recommend adding a line in the changelog and a test covering that scenario. I spent a couple hours trying to figure out why suddenly one of my objects (a [`virtual-dom` hook](https://github.com/Matt-Esch/virtual-dom/blob/master/docs/hooks.md#hooks) implemented in Ruby) suddenly became `nil` when I upgraded Opal. :-)

jgaskins:
> Maybe a #to_n method can be defined on Object that internally just calls self.instance_variables.to_n

I thought about something like this, too, but just returning the original value accomplishes pretty much the same thing. :-)

meh:
The behavior of `Native.try_convert` is correct as is, what's wrong is the `#to_n` for `Hash` and friends, the call should be `Native.try_convert(value) || value`.

jgaskins:
@meh I was going to go for that, actually, but I was concerned with the semantics of something like `{ foo: nil }.to_n`. Currently, JS `null` is treated as truthy in Opal, but if/when that changes in the future, `Native.try_convert(nil) || nil` would return `nil` instead of `null`. :-\

meh:
Yeah right, I didn't consider that, maybe add an additional parameter to `Native.try_convert`?

So it could become `Native.try_convert(whatever, whatever)` and it will return whatever if it cannot be converted to native, since it would default to `nil` the behavior would remain the same for the other cases.

Basically `Native.try_convert` was trying to follow the `Array.try_convert` API, but an extension is fine by me.

jgaskins:
Good idea. I went ahead with that.

meh:
:+1:

PR #1260, opened 2015-12-26
Rack2

JelF:
Adds rack '>=2' support, which is required by rails 5

closes https://github.com/opal/opal-rails/issues/76

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,21 +1,24 @@
 source 'https://rubygems.org'
 gemspec
 
+tilt_version = ENV['TILT_VERSION']
+rack_version = ENV['RACK_VERSION']
+
 # Stick with older racc until
 # https://github.com/tenderlove/racc/issues/22
 # is solved.
 gem 'racc', '< 1.4.10' if RUBY_ENGINE == 'jruby'
 gem 'json', '< 1.8.1'  if RUBY_VERSION.to_f == 2.1 and RUBY_ENGINE == 'ruby'
 gem 'rubysl', :platform => :rbx
-gem 'thin', platform: :mri
+gem 'thin', platform: :mri unless rack_version
+
+gem 'rack', rack_version if rack_version
+gem 'tilt', tilt_version if tilt_version
 
 group :repl do
   gem 'therubyracer', :platform => :mri, :require => 'v8'
   gem 'therubyrhino', :platform => :jruby
 end
-
-tilt_version = ENV['TILT_VERSION']
-gem 'tilt', tilt_version if tilt_version
 
 unless ENV['CI']
   gem 'rb-fsevent'
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,9 @@ rack_version = ENV['RACK_VERSION']
 gem 'racc', '< 1.4.10' if RUBY_ENGINE == 'jruby'
 gem 'json', '< 1.8.1'  if RUBY_VERSION.to_f == 2.1 and RUBY_ENGINE == 'ruby'
 gem 'rubysl', :platform => :rbx
-gem 'thin', platform: :mri unless rack_version
+
+# thin requires rack < 2
+gem 'thin', platform: :mri if !rack_version || (rack_version < '2')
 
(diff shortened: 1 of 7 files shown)


lib/opal/sprockets/server.rb
```diff
@@ -4,7 +4,14 @@
...
 require 'rack/directory'
-require 'rack/showexceptions'
+
+# rack changed some paths in 2.0.0alpha
+begin
+  require 'rack/showexceptions'
+rescue LoadError
+  require 'rack/show_exceptions'
+end
+
```
elia: Would it feasible to use a version check here?

```
if rack_version < 2
  require …
```

lib/opal/paths.rb
```diff
@@ -42,6 +42,8 @@ def use_gem(gem_name, include_dependencies = true)
     def require_paths_for_gem(gem_name, include_dependencies)
       paths = []
       spec = Gem::Specification.find_by_name(gem_name)
+      # If spec is nil, something should be raised instead of NoMethodError
+      # in next line
```
elia: Agree, `Opal::GemNotFound` should do it

lib/opal/sprockets/server.rb
JelF: rack defines `RELEASE` constant inside 'rack.rb', so it could be accessed only if full rack would be required

Gemfile
```diff
@@ -1,22 +1,25 @@
...
+rack_version = ENV['RACK_VERSION']
+
 # Stick with older racc until
 # https://github.com/tenderlove/racc/issues/22
 # is solved.
 gem 'racc', '< 1.4.10' if RUBY_ENGINE == 'jruby'
 gem 'json', '< 1.8.1'  if RUBY_VERSION.to_f == 2.1 and RUBY_ENGINE == 'ruby'
 gem 'rubysl', :platform => :rbx
-gem 'thin', platform: :mri
+gem 'thin', platform: :mri unless rack_version
```
elia: Why is this needed?

If it can't be avoided an explanatory comment will probably help future readers (e.g. me in 5min :stuck_out_tongue_winking_eye:)

lib/opal/sprockets/server.rb
elia: I see now. I think it's fine to `require 'rack'` since it uses autoload and doesn't require anything else… 

thanks to this I just discovered that since it uses autoload the requires can be skipped altogether and the `ShowExceptions` and `ConditionalGet` by virtue of autoload :D

lib/opal/paths.rb
JelF: Did you mean `Gem::GemNotFoundException`?

spec/lib/paths_spec.rb
```diff
@@ -2,8 +2,12 @@
 
 describe 'Opal.use_gem' do
   it 'adds require_paths of a gem to Opal paths' do
-    Opal.use_gem 'sprockets'
-    # Will fail if GEM_HOME has "sprockets" in the path, that's ok.
+    begin
+      Opal.use_gem 'sprockets'
+    rescue 
+      skip %(Will fail if GEM_HOME has "sprockets" in the path, that's ok)
```
JelF: Actualy, it raised because `rack` gem not found, basicaly because it was redefined inside Gemfile. Should i just change pending message?

lib/opal/paths.rb
elia: No, though about using it, but the meaning is different, so if anyone wants to rescue it they can differentiate between the two.

spec/lib/paths_spec.rb
elia: Ok, let's pick another gem, like `rake` which will likely never be redefined in the Gemfile.

elia:
@JelF ok, one last thing, can you squash the commits? (see the [note on commits here](https://github.com/opal/opal/blob/master/CONTRIBUTING.md#contributing-tldr))

if you need any git help or don't have time it's fine, I can do it myself, but if possible I prefer to merge the PR from GH rather than manually :)

PR #1263, opened 2015-12-27
Allow promises to have multiple then/rescue/always

jgaskins:
According to the A+ promise specification, ["`then` may be called multiple times on the same promise"](https://promisesaplus.com/#point-36). Since `fail` and `always` are just `then` with different success/failure semantics, I added the same functionality to them, as well.

@wied03, @meh, and I spoke about supporting this in the past. Last I heard, @meh was still contemplating the change. I probably should've put this in before the 0.9 release, but this PR is backwards-compatible with the current release.

### DIFF
--- a/spec/opal/stdlib/promise/always_spec.rb
+++ b/spec/opal/stdlib/promise/always_spec.rb
@@ -19,12 +19,23 @@ describe 'Promise#always' do
     x.should == 3
   end
 
-  it 'raises an exception when the promise has already been chained' do
+  it 'can be called multiple times on resolved promises' do
     p = Promise.value(2)
-    p.then {}
+    x = 1
+    p.then { x += 1 }
+    p.fail { x += 2 }
+    p.always { x += 3 }
 
-    proc {
-      p.always {}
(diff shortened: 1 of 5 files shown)


elia:
:+1:

Master is already targeting 0.10 but if someone needs it individual commits/PRs can be backported to 0.9 and released in 0.9.1

meh:
If it works I'm fine with it, I just have one request, can you add `then!` and friends that check for presence of an existing next?

jgaskins:
I'm not sure what you mean. You want one that does raise if called multiple times on the same promise?

meh:
Yes, alternative bang methods that raise.

jgaskins:
What was the reason behind the single-use method again? I've asked this before but I don't remember.

meh:
I just personally prefer it that way to avoid bugs in the code when you only want to chain and not add up.

jgaskins:
Alrighty, done, but it looks like GitHub diffs are broken at the moment?

meh:
I'll merge after they're back.

PR #1269, opened 2016-01-02
Fix using more than two "rescue" in sequence

elia:
Previously only the first two rescues were being compiled.

ref: https://github.com/opal/opal/pull/687#issuecomment-69344736

The following code was ignoring the `IOError` rescue, now works with any number of `rescue`s.

```ruby
begin
  raise IOError, 'foo'
rescue RangeError              # this one is correct
rescue TypeError               # miss a return
rescue IOError                 # following two lines disappear in js
  puts "I got #{$!.message}"
end
```


elia:
@alexdowad can you review and merge if it looks ok?

vais:
:scream:

elia:
@vais you think it's serious enough to justify a bug fix release?

vais:
@elia well, it definitely goes against *the principle of least surprise* to say the least. Not sure what you mean by a bug fix release (i.e. back-porting this to which/how many versions would be involved), but I think this should be fixed in 0.9 as opposed to 0.10 if that makes sense.

elia:
@vais was thinking about 0.9.1 only too.

Maybe I'll ask on gitter if anyone have some bug fix that would like to be backported to 0.9.

vais:
:+1:

huochunpeng:
@elia 
Glad to see the fix, but sadly there is one more bug. In my original comment `# miss a return` is not solved. It looks like separate issue, see my comments below.

```
(function(Opal) {
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $gvars = Opal.gvars;
  if ($gvars["!"] == null) $gvars["!"] = nil;

  Opal.add_stubs(['$raise', '$puts', '$message']);
  try {
  return self.$raise($scope.get('IOError'), "foo")
  } catch ($err) {if (Opal.rescue($err, [$scope.get('RangeError')])) {
    try {
      return nil
    } finally {
      Opal.gvars["!"] = Opal.exceptions.pop() || Opal.nil;
    }
    }else if (Opal.rescue($err, [$scope.get('TypeError')])) {
    try {
      nil //***** HERE should be return nil
    } finally {
      Opal.gvars["!"] = Opal.exceptions.pop() || Opal.nil;
    }
    }else if (Opal.rescue($err, [$scope.get('IOError')])) {
    try {
      self.$puts("I got " + ($gvars["!"].$message())) //***** HERE should be return self.$puts...
    } finally {
      Opal.gvars["!"] = Opal.exceptions.pop() || Opal.nil;
    }
    }else { throw $err; }
  }
})(Opal);
```

elia:
@huochunpeng yeah, apparently the return is missing from all the `rescue`s but the first, no need for a separate issue this time, I'll keep it in this PR.

@vais will wait for the complete solution before backporting to 0.9.1

elia:
@huochunpeng I may be wrong but the return isn't needed there, unless the whole begin block is an expression, in fact if an assignment is added taking the whole block the `return`s appear as expected:

```rb
r = begin
  raise IOError, 'foo'
rescue RangeError              # this one is correct
rescue TypeError               # miss a return
rescue IOError                 # following two lines disappear in js
  puts "I got #{$!.message}"
end
```

```
opal:elia/pull-687/fix-multiple-rescue ⤑ pbpaste | be bin/opal -cEO
(function(Opal) {
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $gvars = Opal.gvars, r = nil;
  if ($gvars["!"] == null) $gvars["!"] = nil;

  Opal.add_stubs(['$raise', '$puts', '$message']);
  return r = (function() { try {
  return self.$raise($scope.get('IOError'), "foo")
  } catch ($err) {
    if (Opal.rescue($err, [$scope.get('RangeError')])) {
      try {
        return nil
      } finally { Opal.gvars["!"] = Opal.exceptions.pop() || nil; }
    } else 
    if (Opal.rescue($err, [$scope.get('TypeError')])) {
      try {
        return nil
      } finally { Opal.gvars["!"] = Opal.exceptions.pop() || nil; }
    } else 
    if (Opal.rescue($err, [$scope.get('IOError')])) {
      try {
        return self.$puts("I got " + ($gvars["!"].$message()))
      } finally { Opal.gvars["!"] = Opal.exceptions.pop() || nil; }
    }else { throw $err; }
  }})()
})(Opal);
```

huochunpeng:
Even without assignment, it's still an expression, it still needs to return. In Ruby, almost (maybe don't need to say "almost" in Ruby 2, but I am not sure) everything is expression which needs to return a value, and it is implicitly returned if this is the last one evaluated in a method.

Consider return value of this method
```
def foo
  begin
    //...
  rescue ErrorOne
    //...
  rescue ErrorTwo
    //...
  end
end
```

elia:
Good point, luckily seems to be working properly, wrapping the whole begin block within a function and adding returns where, needed, I'll do more tests tomorrow tho.

As of the `statement` status, please consider this example in which the returns would give the wrong result:

```rb
def bar
  begin
    raise IOError, 'foo'
  rescue RangeError              # this one is correct
  rescue TypeError               # miss a return
  rescue IOError                 # following two lines disappear in js
    puts "I got #{$!.message}"
  end
  :baz
end
p bar # should be :baz
```

huochunpeng:
Adding `return` statement is not my point... treat it as expression is my point.

In the sense of expression, the body of your method may be something like this:
```
var last_expression;
var expression1 = (function() { /* needs return for everything */ })();
last_expression = expression1;
var expression2 = "barz";
last_expression = expression2;
return last_expression;
```

Wrapping any complex Ruby expression in `(function() {})()` maybe a good idea.

One thing surprised me in Opal is that the compiled JavaScript doesn't have anything to track `last_expression`, it's all depending on the `return` statement at right point, which I imagine it's very difficult way to mimic Ruby's "return the last expression". I hope this makes sense.

elia:
Ok merging this, #1270 is still open as the fix is trickier thank I though.

@vais the backport to `0-9-stable` should consist in cherry-picking 2672fe3f3308b75c0d5b1ba7d4ea2c68402d2cfd only.

huochunpeng:
Tried Opal again after 9 months regarding implicit return of last expression, following test is still negative in 0.10.1.

```
def foo
  begin
    raise IOError, 'foo'
  rescue RangeError # this one is correct
  rescue TypeError  # miss a return
  rescue IOError    # miss a return
    "bar"
  end
end

puts foo
```

This should print "bar" but it doesn't.

Issue #1271, opened 2016-01-05
Exception instead of NoMethodError when method is missing


iliabylich:
``` ruby
begin
  1 <=> []
rescue Exception => e
  puts e.class
end
# MRI:
# blank output (comparison returns nil)
# Opal master
Exception
```
Don't even know why CI is still green, I'm getting [this spec](https://github.com/ruby/rubyspec/blob/master/core/array/comparison_spec.rb#L76) failing locally (it invokes internally pretty much the same code, but it expects to catch a `StandardError`, that's why the spec fails for me).

And it seems that that code was [ok](http://opalrb.org/try/?code:begin%0A%20%201%20%3C%3D%3E%20%5B%5D%0Arescue%20Exception%20%3D%3E%20e%0A%20%20puts%20e.class%0Aend) before (on 0.8), so it looks like a regression.

Looks like in this case Opal doesn't generate method missing stubs correctly.

elia:
Closed by #1273

PR #1273, opened 2016-01-05
Always populate all stub_subscribers with all method stubs.

### LINKED ISSUE #1272: Array comparison spec failing when ran alone
```
rake mspec_rubyspec_nodejs 
# vs
rake mspec_rubyspec_nodejs PATTERN="spec/rubyspec/core/array/comparison_spec.rb"  
```

cc @vais

iliabylich:
Closes #1272 and #1271

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -738,6 +738,10 @@
     Opal.donate_constants(module, klass);
   };
 
+  // Table that holds all methods that have been defined on all objects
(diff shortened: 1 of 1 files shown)


elia:
Reading the diff I was wondering if caching results of stub_for into Opal.stubs would save any work…

elia:
Also can you try to add a spec that fails without this patch?

elia:
Never mind, just couldn't wait :D

Issue #1274, opened 2016-01-07
0.9 RC1->Final Nil Operator Expression Regression


wied03:
Appears to have been introduced with https://github.com/opal/opal/commit/cbb05e53de526cd234e18a4ce5f095b9e27d129d but that is probably not the root cause of the problem.

Problem appears to be related to operator precedence and usage of parentheses with the `nil` global in JS. I've observed the problem with ternary expressions and guards.

Test Source:

```ruby
def return_nil
  %x{
    return nil;
  }
end

class Foo
  def <(other)
    %x{
      return nil;
    }
  end
end

puts "no class #{return_nil ? 'yes' : 'no'}"
puts "class w/o parens #{Foo.new < 2 ? 'yes' : 'no'}"
puts "class w/ parens #{(Foo.new < 2) ? 'yes' : 'no'}"
puts "guard w/ parens" unless (Foo.new < 2)
puts "guard w/o parens" unless Foo.new < 2
```

Can't run this MRI due to the JS, but here is the actual results from opal-0-9-stable:
```
no class no
class w/o parens yes
class w/ parens no
guard w/ parens
```

The result should be:
```
no class no
class w/o parens no
class w/ parens no
guard w/ parens
guard w/o parens
```

This problem could manifest itself anywhere the `nil` JS value is used in expressions. The first case I encountered was that it breaks module comparison unless parentheses are used.

wied03:
Causing https://github.com/opal/opal-rspec/issues/51

wied03:
Simplifying this has the same result:

Test Source:

```ruby
module Mod2
end

class Bar
end

def report(klass)
  puts "#{klass} < mod2 ? #{klass < Mod2 ? 'yes' : 'no'}"
end

report Bar
```

MRI:
```
Bar < mod2 ? no
```

Opal 0.9 final:
```
Bar < mod2 ? yes
```

wied03:
Rubyspecs should be catching this with existing tests but the behavior of this code when I add it into ```rubyspec/core/module/lt_spec.rb``` is different than when I run the standalone code above.

iliabylich:
It doesn't look like https://github.com/opal/opal/commit/cbb05e53de526cd234e18a4ce5f095b9e27d129d causes it. And it also happens for a regular `if`, not just with a ternary operator.
``` ruby
value = Foo.new

if value < 2
  puts 'yes'
else
  puts 'no'
end
```
Compiles to:
``` js
value = $scope.get('Foo').$new();
[...]

PR #1276, opened 2016-01-08
Fix JS Truthy Optimization

wied03:
Rather than using just an operator comparison to decide whether optimization can happen, look at the call type and only optimize operator based calls to ValueNode/NumericNode based types.

Should address https://github.com/opal/opal/issues/1274 in a way that preserves operator optimization (rg_gt, etc.) when possible.

Since no tests broke when I changed this, added compiler specs and also added a spot check at a higher level (opal mspec).

Can squash once approach is reviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.1 (unreleased)
 
 * Backport rack2 compatibility from #1260
+* Fixed issue with JS nil return paths being treated as true (#1274)
 
 
 ## 0.9.0 2015-12-20
--- a/lib/opal/nodes/base.rb
+++ b/lib/opal/nodes/base.rb
@@ -22,6 +22,10 @@ module Opal
           end
         end
       end
+      
+      def self.truthy_optimize?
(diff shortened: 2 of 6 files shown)


wied03:
@elia - If this looks good/gets merged, I can also help port the fix to master (or alternatively, I can redo this against master and it can be backported to 0.9 stable). Let me know.

meh:
I like the approach, could you do the thing you put in the TODO?

I think it would be cleaner, and easier to extend in the future.

wied03:
@meh @elia - OK to squash?

meh:
Looks good to me.

elia:
👍🏼 can send a PR for master too?

wied03:
@elia - You didn't want to squash first before that was merged?

elia:
Ouch, did it from mobile and confused the last commit with the result of a squash 😞

wied03:
@elia - submitted another PR for master. I don't think I can squash this one now since I can't rebase and push directly into 0-9-stable

PR #1283, opened 2016-01-10
Resolves #1281, do not generate source block when module option is specified

### LINKED ISSUE #1281: Avoid to generate an empty block if the source is empty from the command line
I'm using the following command to generate JavaScript code:

```
opal --compile -I lib -I asciidoctor -g asciidoctor -r asciidoctor/converter/docbook5 --dynamic-require ignore --no-opal --no-exit > build/asciidoctor-docbook5.js < /dev/null
```

It would be nice if I could tell Opal that there's no code to generate (apart from the `docbook5` module).
A safe solution is to add an option `--no-source` to skip the generation.

Mogztter:
@elia I didn't find out how to put the code of the module inside an Opal module... in Asciidoctor.js we're using `Opal::Environment` but this is now deprecated. Maybe it's better to leave as it is (no need to manually require the module) ? Wdyt ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Whitespace conventions:
 - Add Rack v2 ccompatibility (#1260)
 - Newly compliant with RubySpec:
     * `Array#slice!`
+- Add -M --module option to compile only the code of the module (#1281)
 
 
 ### Fixed
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -7,7 +7,7 @@ module Opal
   class CLI
     attr_reader :options, :file, :compiler_options, :evals, :load_paths, :argv,
                 :output, :requires, :gems, :stubs, :verbose, :port, :preload,
-                :filename, :debug, :no_exit
+                :filename, :debug, :no_exit, :module_only
 
     def compile?
       @compile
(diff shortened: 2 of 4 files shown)


elia:
> how to put the code of the module inside an Opal module

Not sure what do you mean by that, the code seems fine as it is.  Only have a couple of requests:

- can you add an exception if module_only is provided along with file/filename/eval?
- I think lowercase -m is better unless it's used in any other option or is used by MRI for something (even if it's not yet implemented in Opal 
- a changelog entry

:)

Mogztter:
> Not sure what do you mean by that, the code seems fine as it is

The generated code begins with `(function(Opal) {` (automatically loaded). Previously in the Asciidoctor.js build we had `Opal.modules["my_module"] = function(Opal) {` (require an explicit `Opal.require('my_module');`)

> can you add an exception if module_only is provided along with file/filename/eval?

Sure, I should raise an `ArgumentError` right ?

> I think lowercase -m is better unless it's used in any other option or is used by MRI for something 

Lowercase `m` is used by "Show sourcemap": `on('-m', '--map', 'Show sourcemap')`

> a changelog entry

Oups always forget about this one, sorry :smile:

Mogztter:
@elia How can I check if the block `evals_or_file` is defined ? This is beyond my Ruby skills :disappointed: 

```ruby
def evals_or_file
  if evals.any?
    yield evals.join("\n"), '-e'
  else
    if file and (filename != '-' or evals.empty?)
      yield file.read, filename
    # else the block is undefined :'(
    end
  end
end
```
I tried the following but this is not working:

```ruby
evals_or_file ||= nil

if !evals_or_file.nil? and module_only
 raise ArgumentError
end
```

Thanks!

elia:
No worries, I'll add the check and merge manually

Mogztter:
@elia Ok great, could you tell me once this is done I'm curious about the solution :wink:

elia:
Sure, nothing fancy, just an added check [here](https://github.com/opal/opal/blob/ec6ad0b5711ee03b5d078f7df281bc49502bc68c/lib/opal/cli.rb#L61) and relative specs, but I'll ping you when it's up :)

elia:
@Mogztter done: https://github.com/opal/opal/commit/57947157aab8f15a4eb05e151a2e7e2711fa5f40#diff-fc68deb5a23846daa1db3a62f9da5f24L61 🙂

Mogztter:
Nice, thanks!

PR #1284, opened 2016-01-11
Fixed Opal.is_a to use Opal.ancestors of passed object and object's metaclass.

### LINKED ISSUE #1279: Module Case Operator
Code:

```ruby
module ModuleCase
end

module ModuleCase2
  include ModuleCase
end

class ModuleCase3
  include ModuleCase2
end

def module_case_works_right?
  instance = ModuleCase3.new
  ModuleCase === instance && instance.kind_of?(ModuleCase)
end

puts module_case_works_right?
```

MRI:
```
=== result true, kind of true
true
```

Opal master:
```
=== result false, kind of false
false
```

iliabylich:
Fixes #1279.


iliabylich:
Rebased on the latest master

PR #1294, opened 2016-01-17
Resolves #1218, Implement Kernel.open method

### LINKED ISSUE #1218: Implement Kernel.open() method
In Asciidoctor.js we have a basic implementation of `Kernel.open()` that allow to read files:

```ruby
module Kernel
  # basic implementation of open, enough to work
  # with reading files over XmlHttpRequest
  def open(path, *rest)
    file = File.new(path, *rest)
    if block_given?
      yield file
    else
      file
    end
  end
end
```

With this implementation, the following code in Ruby is working fine with Opal:

```ruby
open(include_file, 'r') {|f| f.read }
```

Mogztter:


elia:
Should be

      def open(*args, &block) File.open(*args,&block) end

Mogztter:
@elia Done and rebased on master (and with a changelog entry :wink:)

PR #1304, opened 2016-01-26
Generate meaningful names for functions representing Ruby methods.

iliabylich:


vais:
@iliabylich please rebase with master and try pushing again.

iliabylich:
@vais Done

elia:
```
$ node
> foo = 123
123
> function foo(){}
undefined
> foo
[Function: foo]
> foo = 123
123
> bar = function foo(){}
[Function: foo]
> foo
123
```

So, just for the record, seems that a named function will override a local var only if it is not an expression (vs. statement). I don't think we ever compile to function in that way so it should be fine.

iliabylich:
@elia I think opal generates a separate context here:
``` ruby
# test.rb
Class.new {
  `var value = 123;`
  def value
    456
  end
  `value`.should == 123
}
```
Which compiles to:
``` js
// opal -c --no-opal test.rb
var value = 123;
Opal.def(self, '$value', function value() {
  var self = this;
  return 456;
});
```

Here's a simplified example:
```js
var a = 123;
function wrapper() {}

wrapper(function a() {})

console.log(typeof(a))
console.log(a)

// node test.js
// => number
// => 123
```

I'll add a test for this case to `language_spec.rb`.

iliabylich:
By the way, here's a trace that I'm getting by raising an exception from instance method `A#a`:
```
RuntimeError: a
    at Opal.defs.self [as $new] (/tmp/opal-nodejs-runner-20160127-23250-98rcx1:4586:15)
    at $A.raise (/tmp/opal-nodejs-runner-20160127-23250-98rcx1:4338:48)
    at $A.a [as $a] (/tmp/opal-nodejs-runner-20160127-23250-98rcx1:18682:19)
    at Opal.dynamic_require_severity (/tmp/opal-nodejs-runner-20160127-23250-98rcx1:18685:33)
    at Object.<anonymous> (/tmp/opal-nodejs-runner-20160127-23250-98rcx1:18686:3)
    ...
```

Is it possible to make something with the last anonymous function (which is, I guess, is a function that wraps a compiled ruby file)? I just can't imagine a function name that has a valid identifier and represents a file name (i.e. `lib/something/whatever.rb`)

elia:
Most JS should pick up the name from an assignment I think node does that for example, for the other browsers the only option seems to be `displayName`.

elia:
Can't wait to see those `TMP_2` go away from the backtraces!

Issue #1305, opened 2016-01-26
SyntaxError: `return try {`


vais:
The following Ruby:
```ruby
  it "can be used to handle a specific exception" do
    begin
      raise SpecificExampleException, "Raising this to be handled below"
    rescue SpecificExampleException
      :caught
    end.should == :caught
  end
```
compiles to the following JavaScript:
```javascript
($a = ($c = self).$it, $a.$$p = (TMP_3 = function(){var self = TMP_3.$$s || this;
    return try {
        return self.$raise($scope.get('SpecificExampleException'), "Raising this to be handled below")
      } catch ($err) {
        if (Opal.rescue($err, [$scope.get('SpecificExampleException')])) {
          try {
            return "caught"
          } finally { Opal.pop_exception() }
        } else { throw $err; }
      }.$should()['$==']("caught")}, TMP_3.$$s = self, TMP_3), $a).call($c, "can be used to handle a specific exception");
```
which, when executed, produces a SyntaxError:
```
    return try {
           ^^^

SyntaxError: Unexpected token try
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3
```
Due to this issue, we can no longer execute https://github.com/ruby/spec/blob/master/language/rescue_spec.rb#L17

iliabylich:
I guess for now we can just put `nil` to the end of every test like this (and that's probably what Opal compiler should do).

iliabylich:
Ahh, this spec comes from the rubyspec :smile:

vais:
@iliabylich lol, yes :)

elia:
it fails when treated as an expression, starting a 0.10 milestone to keep track of this stuff
in the meanwhile I think it's better to pin down a working mspec version in a submodule and restore the green badge

vais:
@elia
> restore the green badge

??

https://github.com/opal/opal/pull/1306 is green, but `language/rescue` spec had to be commented out because of this syntax error. Not sure why you're referring to MSpec though. I guess @eregon made a change so some of the evaled specs are easier to filter out. We could have done that without updating to latest spec suite, it involves just changing some existing filters to remove the embedded `\n`s.

elia:
[...]

PR #1306, opened 2016-01-26
Update to latest Ruby Spec Suite

vais:


spec/rubyspecs
```diff
@@ -112,6 +112,7 @@ rubyspec/language
 !rubyspec/language/next_spec
 !rubyspec/language/predefined_spec
 !rubyspec/language/proc_spec
+!rubyspec/language/rescue_spec
```
vais: Unfortunately we lost the ability to run language/rescue spec due to https://github.com/opal/opal/issues/1305 :cry:

PR #1308, opened 2016-01-29
Wrap begin;rescue;end into a function when there's a method call on returning value.

### LINKED ISSUE #1305: SyntaxError: `return try {`

iliabylich:
Prepend rescue statement with js_return when it has ensure.
Fixes #1305.

### DIFF
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -44,7 +44,14 @@ module Opal
       end
 
       def body_sexp
-        wrap_in_closure? ? compiler.returns(begn) : begn
+        if wrap_in_closure?
+          sexp = compiler.returns(begn)
+          # 'rescue' is an edge case that should be compiled to
+          # try { return function(){ ..rescue through try/catch.. }() }
+          sexp.type == :rescue ? s(:js_return, sexp) : sexp
+        else
+          sexp = begn
+        end
       end
 
       def ensr_sexp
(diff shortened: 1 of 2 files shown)


elia:
:-1:

vais:
:bow:

eregon:
Nice! Rescue specs should load-able again with this :smile:

vais:
@eregon https://github.com/opal/opal/pull/1309 :wink:

PR #1309, opened 2016-01-29
Update to latest Ruby Spec Suite

vais:
With special thanks to @iliabylich for https://github.com/opal/opal/pull/1308


eregon:
:+1:

PR #1310, opened 2016-01-29
Fix context of evaluation for #eval and #instance_eval.

### LINKED ISSUE #1245: Regression "Reopening a class adds new methods to subclasses"
[This PR](https://github.com/opal/opal/pull/1243) added a few more specs to run, and (as a result?) the ["Reopening a class adds new methods to subclasses"](https://github.com/ruby/rubyspec/blob/master/language/class_spec.rb#L233-L242) spec started failing and had to be [filtered out](https://github.com/opal/opal/blob/master/spec/filters/bugs/language.rb#L125) from the spec suite. I suspect there's some global interference at play here, since there was no code changed, just more code being executed.

iliabylich:
1. Not sure about `<script type='text/ruby'>` - it can be broken, but locally specs are green. Are there any tests for this functionality?
2. Probably there are some specs from ruby/spec that are green now
3. I guess it closes #1245

For `#eval`:
``` ruby
require 'opal-parser'

class A
  eval <<-SOURCE
    def a_method
      puts "From " + self.class.name + "#a_method"
    end
  SOURCE
end
A.new.a_method

class B < A
end
B.new.a_method

class C
end
C.new.a_method
```

Before:
```
From A#a_method
From B#a_method
From C#a_method
```
Now:
```
From A#a_method
From B#a_method
a_method: undefined method `a_method' for #<C:0x352>
```

For `#instance_eval`:
``` ruby
o = Object.new
o.instance_eval "def m; :m; end"
puts o.m

puts Object.new.m
```
Before:
```
m
m
```

After:
```
m
m: undefined method `m' for #<Object:0x2d4>
```


iliabylich:
Replaced `with` with `(function(self) {})(self)`. I'm not sure it's a good idea to make jshint ignoring `with`. I suppose this PR is ready for merging.

elia:
Looks good

elia:
Thank you, sorry for the wait :)

PR #1324, opened 2016-02-02
Fix nodejs implementation of Dir.pwd and Dir.[].

### LINKED ISSUE #1323: Delegation and scope
Recently I've implemented `Pathname.expand_path` (#1296).

`Pathname.expand_path` method (defined in `stdlib`) delegates to `File.expand_path` method (also defined in `stdlib`). 
`File.expand_path` method uses `Dir.pwd` method to find out the current directory. We have defined a method in `stdlib/nodejs` to use `process.cwd()`. We also have an implementation of `Dir.pwd` method in `corelib`.

Here's a "flow chart" of the method call:

```
Pathname.expand_path (stdlib) ---> File.expand_path (stdlib) ---> Dir.pwd (stdlib/nodejs)
                                                             ---> Dir.pwd (corelib)
```

The problem is that the method `Dir.pwd` defined in `stdlib/nodejs` is not called.
I think the scope (Opal) defined in `File.expand_path` is not "updated/overloaded" with the methods defined in the module `nodejs`.
Does that make any sense ? :8ball:

I tried to debug the `Opal.defn` function but my knowledge of Opal internals are very limited...

iliabylich:
Closes #1323.

### DIFF
--- a/stdlib/nodejs/dir.rb
+++ b/stdlib/nodejs/dir.rb
@@ -2,12 +2,14 @@ class Dir
   @__glob__ = node_require :glob
   `var __glob__ = #{@__glob__}`
 
-  def self.[] glob
-    `__glob__.sync(#{glob})`
-  end
+  class << self
+    def [] glob
+      `__glob__.sync(#{glob})`
+    end
 
-  def pwd
-    `process.cwd()`
+    def pwd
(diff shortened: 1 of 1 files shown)


elia:
😁 ugh, thanks

PR #1330, opened 2016-02-05
Fix issue #1320 (Unable to create string literal using `%&string&`)

vais:
Fix issue #1320


elia:
:+1: (sorry for the wait)

PR #1333, opened 2016-02-06
Resolves #1302, Remove encoding flag before calling Node.js openSync method

### LINKED ISSUE #1302: Node.js File.open method should ignore invalid flags
Currently `flags` are sent directly to Node.js `fs.open`: 

```ruby
@fd = `__fs__.openSync(path, flags)`
```
Source: https://github.com/opal/opal/blob/master/stdlib/nodejs/file.rb#L86

Unfortunately Node.js does not support the same flags as Ruby does: https://nodejs.org/api/fs.html#fs_fs_open_path_flags_mode_callback

We should remove "invalid" flags before calling Node.js `fs.open` (and log a warning ?). Otherwise Node.js throws an exception.

Exemple:

```
Error: Unknown file open flag: r:UTF-8
    at stringToFlags (fs.js:451:9)
    at Object.fs.openSync (fs.js:502:51)
```
Source: 
https://github.com/opal/opal/blob/master/lib/opal/path_reader.rb#L12

Mogztter:
Resolves #1302

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -80,7 +80,8 @@ class File < IO
   # Instance Methods
 
   def initialize(path, flags)
-    flags = flags.gsub(/b/, '')
+    flags = flags.gsub(/b/, '') # binary flag is unsupported
+    flags = flags.gsub(/:(.*)/, '') # encoding flag is unsupported
     @path = path
     @flags = flags
     @fd = `__fs__.openSync(path, flags)`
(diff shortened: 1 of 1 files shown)


elia:
Looks good, can you add the warning?

Mogztter:
I extracted the regexp in a variable and use `match` to check if `flags` contains binary and/or encoding flag. Not sure if we can do better with Ruby ?

elia:
Thanks!

PR #1334, opened 2016-02-07
inline block

elia:
replacing this:

```js
var array = [1,2,3];
array.$$p = function(element){ console.log(element) };
array.each();
```

```js
var array = [1,2,3];
array.each(Opal.block, function(element){ console.log(element) });
```

---

### Current status:

this works:

```
$ bundle exec bin/opal -O -ropal/mini -rcorelib/io -e 'p :ciao'
"ciao"
```

### DIFF
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -228,16 +228,25 @@ module Opal
       self.helpers << name
     end
 
+    def indent!
+      previous = @indent
+      @indent += INDENT
+      @space = "\n#@indent"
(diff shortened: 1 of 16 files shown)


wied03:
Is this mainly cosmetic/JS readability? I wouldn't think it would make much of a difference from an optimization standpoint

elia:
It is, any possible improvement in performance is of secondary importance in this context (eg for not passing a TMP var for the method to retrieve the block)

PR #1339, opened 2016-02-11
Fix Enumerator#with_index to return the result of the previously called method.

### LINKED ISSUE #1325: map.with_index returns the original values
Running against the `master` branch:

```ruby
p [1,2,3].map.with_index { |n, index| 4 }
```

Expected: `[4, 4, 4]`
Actual: `[1, 2, 3]`

[Example on opalrb.org/try](http://opalrb.org/try/?code:p%20%5B1%2C2%2C3%5D.map.with_index%20%7B%20%7Cn%7C%204%20%7D)

iliabylich:
Closes #1325.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Whitespace conventions:
 - Fixed context of evaluation for `Kernel#eval` and `BasicObject#instance_eval`
 - Fix `Module#===` to use all ancestors of the passed object (#1284)
 - Fix `Struct.new` to be almost compatible with Rubyspec (#1251)
+- Fix `Enumerator#with_index` to return the result of the previously called method.
 
 
 
--- a/opal/corelib/enumerator.rb
+++ b/opal/corelib/enumerator.rb
@@ -72,13 +72,7 @@ class Enumerator
         return value;
       }
 
-      self.$each();
-
-      if (result !== undefined) {
-        return result;
-      }
-
-      return self.object;
+      return self.$each();
     }
   end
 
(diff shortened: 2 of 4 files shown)


vais:
@iliabylich you're a machine! @meh is this the `Enumerator#with_index` fix you were looking for?

> fkchang, yeah, there's something broken with all the with_index stuff it seems

iliabylich:
@vais The fix is trivial, but I'm not 100% sure that it's correct. I can't find any specs for this functionality in `ruby/spec`

vais:
@iliabylich pass the existing specs is the best we can do, and it looks they're passing. Let's see what @meh has to say, he is *the* authority on this subject :wink:

meh:
Yeah I'm not sure it'sw 100% correct, but I can't remember why I did it like that :panda_face:

vais:
Ok, thanks, @meh :smile:

@iliabylich, we are already [running all of ruby's enumerator specs] (https://github.com/opal/opal/blob/master/spec/rubyspecs#L11). Please try removing [these bug filters](https://github.com/opal/opal/blob/master/spec/filters/bugs/enumerator.rb#L3-L5), and if the spec suite still passes, I think we should just merge your fix :rocket:

iliabylich:
@vais Done. Two other specs that you mentioned are still broken.

vais:
Looks good to me. @elia, @meh, any objections to merging?

elia:
:shipit:

vais:
@iliabylich could you please rebase with master?

iliabylich:
@vais Done

vais:
:tada: thank you, @iliabylich!

PR #1342, opened 2016-02-12
First pass at BigDecimal implementation

vais:
Passing ~50% of the BigDecimal spec suite. At this point one should be able to use this to do basic money arithmetic: addition, subtraction, multiplication, and division are all implemented :sweat_smile: The rest I hope will get fleshed out as needed - I am out of :hourglass: 

My hope is this could also be used as an example for wrapping an existing JS lib. I made generous use of the new `.JS.` syntax, hehe. As such, any feedback would be appreciated :wink: cc @elia @meh


meh:
Looks good to me.

elia:
:+1:

vais:
:tada:

PR #1345, opened 2016-02-14
Resolves #1253, Add an option to suppress "not supported" warning message

### LINKED ISSUE #1253: Add an option to suppress "not supported" warning message
It would be great if we could suppress "not supported" warning message. In Asciidoctor we are using `freeze` method but we don't want to display a warning message.

We could add a config parameter `OPAL_CONFIG.suppress_warning_freezing` ?

Mogztter:
Resolves #1253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,9 @@ Whitespace conventions:
 - Removed the previously deprecated `Opal::Fragment#to_code`
 - Removed the previously deprecated `Opal::Processor.load_asset_code`
 - Removed the previously deprecated acceptance of a boolean as single argument to `Opal::Server.new`
+- Add an option to suppress 'not supported' warning message (#1253)
+    * `Opal::Config.suppress_warning_freezing`
+    * `Opal::Config.suppress_warning_tainting`
 
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


Mogztter:
Rebased on master

Mogztter:
Rebased on master

elia:
Looking again at the PR I'd move the warning stuff to the runtime:

```
%x{
  var warnings = {};
  Opal.config = {};
  Opal.warn = function warn(config_name, error) {
    if (Opal.config[config_name] && !warnings[string]) {
      warnings[string] = true;
      #{Kernel.warn(`string`)};
    }
    else {
      #{raise NotImplementedError, `ERROR`}
    }
  }
}

module Kernel
  `var ERROR = "Object freezing is not supported by Opal";`

  def freeze
    `Opal.warn(freezing, ERROR)`
    self
  end

end
```

which should also save a few bytes in the compiled output.

@meh is that ok for you?

Mogztter:
@elia OK
Waiting for @meh's approval :smile:

elia:
I'm going to apply [this patch](https://gist.github.com/elia/b6d1d132897e29118c12) which should solve the issue in a simpler way, I'll close this when it's pushed to master

elia:
fixed by:
- 1219c201be (Manually manage Opal runtime config)
- 0815306245 (Fix the name of `Opal.dynamic_require_severity`)
- b418bdd494 (Decouple runtime/compiletime require errors config)

Mogztter:
Nice thanks @elia

PR #1347, opened 2016-02-14
Resolves #1219, Implement File.extname method

### LINKED ISSUE #1219: Implement File.extname method
In Asciidoctor.js we are using this implementation:

```ruby
  def self.extname(path)
    return '' if path.nil_or_empty?
    last_dot_idx = path[1..-1].rindex('.')
    last_dot_idx.nil? ? '' : path[(last_dot_idx + 1)..-1]
  end
```

Mogztter:
Resolves #1219

### DIFF
--- a/opal/corelib/file.rb
+++ b/opal/corelib/file.rb
@@ -30,6 +30,17 @@ class File < IO
       split(path)[-1]
     end
 
+    def extname(path)
+      raise TypeError, 'no implicit conversion of nil into String' if path.nil?
+      path = path.to_path if path.respond_to?(:to_path)
+      raise TypeError, "no implicit conversion of #{path.class} into String" unless path.is_a?(String)
+      filename = basename(path)
+      return '' if filename.nil?
+      last_dot_idx = filename[1..-1].rindex('.')
+      # extension name must contains at least one character .(something)
+      (last_dot_idx.nil? || last_dot_idx + 1 == filename.length - 1) ? '' : filename[(last_dot_idx + 1)..-1]
+    end
+
     def exist? path
       `Opal.modules[#{path}] != null`
     end
--- a/spec/rubyspecs
+++ b/spec/rubyspecs
@@ -15,6 +15,7 @@ rubyspec/core/exception
 
 rubyspec/core/false
 rubyspec/core/fixnum
+rubyspec/core/file/extname_spec
 rubyspec/core/float
 rubyspec/core/hash
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Whitespace conventions:
 - Generate meaningful names for functions representing Ruby methods
 - Implement Pathname.join and Pathname.+ methods (#1301)
 - Added support for `begin;rescue;else;end`.
+- Implement File.extname method (#1219)
 
 
 ### Fixed


elia:
@Mogztter can you add a spec for this? (either from ruby spec or in spec/opal)

opal/corelib/file.rb
```diff
@@ -30,6 +30,12 @@ def basename(path)
       split(path)[-1]
     end
 
+    def extname(path)
+      return '' if path.nil_or_empty?
```
elia: Is `#nil_or_empty?` defined somewhere?

Mogztter:
@elia Sure. I've never done that before, I just need to add a new line in here https://github.com/opal/opal/blob/master/spec/rubyspecs#L129 ? or write a spec file here: https://github.com/opal/opal/tree/master/spec/opal/core ?

opal/corelib/file.rb
Mogztter: Good catch, I though this was a standard Ruby method but in fact this is an Asciidoctor core extension [nil_or_empty.rb](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/core_ext/object/nil_or_empty.rb).
Do you want me to add `nil_or_empty` extension or replace the line with a standard Ruby code ?

vais: My :+1: for *standard Ruby code* :smile:

elia: Indeed only stuff from corelib is allowed in corelib 😊

elia:
@mogztter see HACKING.md for an explanation on how to deal with specs

(excuse brevity, AFK, I'll get into more details as soon as I can 🙂)

Mogztter:
@elia I'm now 100% compatible with Ruby spec !

opal/corelib/file.rb
```diff
@@ -30,6 +30,18 @@ def basename(path)
       split(path)[-1]
     end
 
+    def extname(path)
+      raise TypeError, 'no implicit conversion of nil into String' if path.nil?
+      path = path.to_path if path.respond_to?(:to_path)
+      raise TypeError, "no implicit conversion of #{path.class} into String" unless path.is_a?(String)
+      filename = basename(path)
```
Mogztter: `File.basename` method is not fully compliant with the Ruby specs but all tests passed nevertheless :wink:

opal/corelib/file.rb
```diff
@@ -30,6 +30,18 @@ def basename(path)
...
+      return '' if filename.nil?
+      last_dot_idx = filename[1..-1].rindex('.')
+      extname = last_dot_idx.nil? ? '' : filename[(last_dot_idx + 1)..-1]
+      return '' if extname.tr('.', '').empty? # extname contains only dots
```
Mogztter: I tried to use a Regexp `extname.match(/^.+$/)` but didn't make it work...

elia: Have you tried escaping the dot? `\.`

spec/rubyspecs
```diff
@@ -127,3 +127,5 @@ rubyspec/library/pp
 rubyspec/library/set
 rubyspec/library/singleton
 rubyspec/library/stringscanner
+
+rubyspec/core/file/extname_spec
```
elia: Can you move this up to keep the loose alphabetical order?

elia:
> @elia I'm now 100% compatible with Ruby spec !

Great!

opal/corelib/file.rb
Mogztter: Yes I did. 
I've took a different approach, now I'm just checking the last dot index with the filename length.

spec/rubyspecs
Mogztter: Ok

elia:
:+1:

Mogztter:
Cool, thanks :tada:

PR #1350, opened 2016-02-17
`A, B = 1, 2` does not work

mame:
Hello,

Opal fails to compile multiple assignment to constants:

~~~~
 opal -e 'A, B = 1, 2'
/home/mame/local/lib/ruby/gems/2.3.0/gems/opal-0.9.2/lib/opal/nodes/masgn.rb:103:in `compile_assignment': An error occurred while compiling: -e (RuntimeError)
Bad child node in masgn LHS: (:cdecl, :A). LHS: (:array, (:cdecl, :A), (:cdecl, :B))
*snip*
~~~~

Though I'm unsure if my patch is right, it fixes this error anyway.  Could you please review it?

### DIFF
--- a/lib/opal/nodes/masgn.rb
+++ b/lib/opal/nodes/masgn.rb
@@ -3,7 +3,6 @@ require 'opal/nodes/base'
 module Opal
   module Nodes
     class MassAssignNode < Base
-      # TODO: does this work for cvars?? constants??
       SIMPLE_ASSIGNMENT = [:lasgn, :iasgn, :lvar, :gasgn, :cdecl]
 
       handle :masgn
--- a/lib/opal/nodes/masgn.rb
+++ b/lib/opal/nodes/masgn.rb
@@ -4,7 +4,7 @@ module Opal
   module Nodes
     class MassAssignNode < Base
       # TODO: does this work for cvars?? constants??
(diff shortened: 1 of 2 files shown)


vais:
Is this another case of a missing spec? Perhaps it should be here: https://github.com/ruby/spec/blob/master/language/variables_spec.rb
cc @eregon

eregon:
Indeed, I'd guess there is no spec for this given it's quite an edge case and there are so many possible cases for masgn. But please add one :smile:

lib/opal/nodes/masgn.rb
```diff
@@ -4,7 +4,7 @@ module Opal
   module Nodes
     class MassAssignNode < Base
       # TODO: does this work for cvars?? constants??
```
eregon: Maybe the TODO can be updated since it should work with constants now?

mame:
> Maybe the TODO can be updated since it should work with constants now?

I think so if the fix is appropriate.

I encountered the error, found the TODO, tried the fix, and just worked.  I'm not sure if this fix is appropriate because I don't understand the source code at all.

elia:
Thanks @mame, the fix looks appropriate, the comment about constants can be removed. 
For me it's ok to merge if any spec is added (either to rubyspec or to spec/opal).

vais:
@elia merge at will, a spec is in the works ^

elia:
Thanks @mame!

PR #1351, opened 2016-02-17
`(-1)[1]` should return 1 instead of 0

mame:
Hello,

`Number#[]` returns a wrong result when it picks a high bit of negative number.

~~~~
$ opal -ve 'p (-1)[1]'
Opal v0.9.2
0
~~~~

It should return 1 because -1 is represented by an integer whose bits are all 1.  Other Ruby implementations including MRI do so.

Thank you,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ Whitespace conventions:
 - Fixed `Module#const_get` for dynamically created constants.
 - Fixed `File.dirname` to return joined String instead of Array.
 - Fixed multiple assignment for constants, i.e., allowing `A, B = 1, 2`.
+- Fixed `Number#[]` with negative number. Now `(-1)[1]` returns 1.
 
 
 
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -214,7 +214,7 @@ class Number < Numeric
       }
 
       if (self < 0) {
-        return (((~self) + 1) >> #{bit}) % 2;
+        return 1 - ((~self) >> #{bit}) % 2;
       }
       else {
         return (self >> #{bit}) % 2;
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -209,16 +209,13 @@ class Number < Numeric
     bit = Opal.coerce_to! bit, Integer, :to_int
 
     %x{
-      if (#{bit} < #{Integer::MIN} || #{bit} > #{Integer::MAX}) {
+      if (#{bit} < 0) {
         return 0;
       }
-
-      if (self < 0) {
-        return 1 - ((~self) >> #{bit}) % 2;
+      if (#{bit} >= 32) {
+        return #{ self } < 0 ? 1 : 0;
       }
-      else {
-        return (self >> #{bit}) % 2;
-      }
+      return (self >> #{bit}) & 1;
     }
   end
 


vais:
Wow, @mame, do you know if `Number#[]` is documented anywhere? I'm trying to find a ruby spec for it...
cc @eregon

mame:
Now that you mention it... I think this behavior is natural extension of two's-complement representation of integers, but I don't know it is explicitly documented anywhere.  [Rdoc](http://ruby-doc.org/core-2.3.0/Fixnum.html#method-i-5B-5D) says just:

> Bit Reference—Returns the +n+th bit in the binary representation of fix, where fix[0] is the least significant bit.

[RubySpec](https://github.com/ruby/spec/blob/master/core/fixnum/element_reference_spec.rb) does not cover `Fixnum#[]` with a negative integer at all.

Ajedi32:
This is kinda tricky for Opal, since Opal doesn't make a distinction between Fixnums and Floats for performance reasons. Maybe there is a reasonable behavior we could use here, but it's hard to say for sure.

mame:
I think `n[b]` is a short-hand to `(n>>b)&1` regardless of whether `n` is positive or negative.  `((-1)>>1)&1` evalates to 1 in JavaScript.  So, this is not so special in JavaScript.

I know it is impossible to do exactly the same behavior as MRI when `n` is so large (`>= 2**31` or `< -(2**31)`), but IMO it should do so when `n` can be exactly represented as a signed 32-bit integer.

eregon:
@mame @vais RubySpec is not perfect, please help to make it more complete and accurate.
There is [one case](https://github.com/ruby/spec/blob/master/core/fixnum/element_reference_spec.rb#L16-L18) with `-1` but the description is completely wrong unfortunately.

On another note, I wish one could derive the behavior just from the spec descriptions, but that really depends on which spec and most do not allow this because they were written by people not fully understanding the functionality of the method (which is hard, the only true way is by studying MRI's code and abstracting from it but that is time consuming), or were derived from the RDoc which is not always very accurate.
In this case, mentioning it behaves like `(n >> b) & 1` would go a long way to explain it better (and even suggest one possible implementation).

elia:
@mame thanks!

did a quick test comparing with MRI and the result are a bit different, not sure although if it falls under the impossibility to have the same behavior:

```
opal:pr/1351 ⤑ be bin/opal -e '[-21, -1, -10, 10, 1, 21].each {|n| 100.downto(0) do |i| print n[i] end; puts }'
01011111111111111111111111111111010111111111111111111111111111110101111111111111111111111111111101011
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
10110111111111111111111111111111101101111111111111111111111111111011011111111111111111111111111110110
01010000000000000000000000000000010100000000000000000000000000000101000000000000000000000000000001010
00001000000000000000000000000000000010000000000000000000000000000000100000000000000000000000000000001
10101000000000000000000000000000101010000000000000000000000000001010100000000000000000000000000010101
opal:pr/1351 ⤑ ruby -e '[-21, -1, -10, 10, 1, 21].each {|n| 100.downto(0) do |i| print n[i] end; puts }'
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111101011
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111110110
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001010
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010101
```

this the C impl from Ruby 2.2:

```c
static VALUE
fix_aref(VALUE fix, VALUE idx)
{
    long val = FIX2LONG(fix);
    long i;

    idx = rb_to_int(idx);
    if (!FIXNUM_P(idx)) {
        idx = rb_big_norm(idx);
        if (!FIXNUM_P(idx)) {
            if (!BIGNUM_SIGN(idx) || val >= 0)
                return INT2FIX(0);
            return INT2FIX(1);
        }
    }
    i = FIX2LONG(idx);

    if (i < 0) return INT2FIX(0);
    if (SIZEOF_LONG*CHAR_BIT-1 <= i) {
        if (val < 0) return INT2FIX(1);
        return INT2FIX(0);
    }
    if (val & (1L<<i))
        return INT2FIX(1);
    return INT2FIX(0);
}
```

Didn't had the time to convert this to a ruby spec, @eregon is it ok to open issues on ruby/spec for missing specs?

mame:
Fixed :-)

~~~
$ bundle exec bin/opal -e '[-21, -1, -10, 10, 1, 21].each {|n| 100.downto(0) do |i| print n[i] end; puts }'
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111101011
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111110110
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001010
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010101
~~~

elia:
@mame great!

can you rebase on master? (if you can't let me know and I'll merge manually later)

mame:
I tried `git rebase master` and `git push --force`.  Does this work for you?

elia:
@mame that's perfect, thank you

eregon:
> @elia: Didn't had the time to convert this to a ruby spec, @eregon is it ok to open issues on ruby/spec for missing specs?

Yes, definitely! Although a PR (or the start of it) is even better :smiley:

PR #1352, opened 2016-02-18
`File.dirname` should return a String instead of Array

mame:
Hello,

`File.dirname` should return a joined path instead of an array of each directory name.

Before:

~~~~
$ opal -e 'p File.dirname("foo/bar/baz")'
["foo", "bar"]
~~~~

After:

~~~~
$ bundle exec opal -e 'p File.dirname("foo/bar/baz")'
"foo/bar"
~~~~

Thank you,


elia:
Thanks! good catch :smiley:

PR #1353, opened 2016-02-18
Initialize $ScratchPad = []

mame:
On my machine, `bundle exec rake` fails.

~~~
1)
The rescue keyword Fix using more than two "rescue" in sequence #1269 ERROR
NoMethodError: undefined method `<<' for nil
<<: undefined method `<<' for nil
*snip*
~~~

Looks the initialization of $ScratchPad was forgotten.  (I'm unsure why this was overlooked.)


iliabylich:
Looks like it depends on the order of execution - https://github.com/opal/opal/blob/master/spec/opal/core/kernel/require_tree_spec.rb#L3

elia:
:+1:

PR #1355, opened 2016-02-20
Limit rack version to 1.6.x

epaulet:
Rack file names were changed in >= 2.0.0.alpha. https://github.com/rack/rack/commit/857641dab255dbec490fead1d3a0f1ff999b2137. Resulted in the following error:

```
opal-0.8.0/lib/opal/sprockets/server.rb:7:in `require': cannot load such file -- rack/showexceptions (LoadError)
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 gemspec
 
 tilt_version = ENV['TILT_VERSION']
-rack_version = ENV['RACK_VERSION']
+rack_version = ENV['RACK_VERSION'] || '~> 1.6.4'
 
 # Stick with older racc until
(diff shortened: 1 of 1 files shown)


elia:
Great! thanks for the PR!

I got caught in this kind of error too, just to discover it's ok to just require "rack" and rely on autoload. 
Should work correctly on 0.9.2: https://github.com/opal/opal/commit/ccef813dae1f42dfac1d562c12ca20bdb9325eb4#diff-9ff187537248de757981a87cae9cbd24

Let me know if there's anything preventing you from upgrading and we'll try to sort it out :)

epaulet:
:+1:

I was using 0.8.x but it works on 0.9.2. Closing since this has already been fixed.

PR #1357, opened 2016-02-22
RFC/experiment: Skipping defs for unused methods

elia:
```ruby
require "opal"
puts Opal::Compiler.new("def foo; bar; end", calls: [:baz]).compile
```

    (function(Opal) {

      return
    })(Opal);

cc @meh @vais

### DIFF
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -118,6 +118,8 @@ module Opal
 
     compiler_option :eval, false, as: :eval?
 
+    compiler_option :calls, nil
+
     # @return [String] The compiled ruby code
     attr_reader :result
 
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -38,6 +38,8 @@ module Opal
       end
 
       def compile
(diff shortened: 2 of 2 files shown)


elia:
update:

```
opal:elia/bare-necessities ⤑ be ruby -e 'require "opal"; p size: Opal::Builder.new(optimize_calls: true).build("opal").to_s.size'                                         ~/C/opal
DCE.37.96.163.195.209.214.215.215
{:size=>419772}
opal:elia/bare-necessities ⤑ be ruby -e 'require "opal"; p size: Opal::Builder.new(optimize_calls: false).build("opal").to_s.size'                                        ~/C/opal
{:size=>579802}
```

160k saved, I think it's not bad

still doesn't work for multiple builder calls

meh:
What about `#send`?

elia:
It's intended as heavy optimization, so not supported, same goes for `method_missing`.

An option to mitigate this limitation could be to restrict the DCE to the core/std lib, assuming you're not going to use `send` and `method_missing` to call `String#gsub` or `Enumerable#select`.

PR #1358, opened 2016-02-22
Add support for keywords arguments in lambdas.

iliabylich:
So, most of the issues related to parsing method/block parameters are fixed in this PR (but not all of them are supported by compiler yet). Sexp for lambda parameters were different from method parameters, so I've changed it to avoid code duplication.
1. optional arguments for lambdas: `->(a=1){} -> (:optarg, :a, (:int, 1))` (was an `:arg` with a `default_block`)
2. rest arguments for lambdas: `->(*a){} -> (:restarg, :a)` (was `:splat :a`)
3. The only difference between lambda/method parameters are `mlhs_args` (`lamda { |(a,b)| }`), all other code for parameters is shared now.
4. There are still a few tests in `ruby/spec` that can't be compiled (but that's not related directly to block/method parameters)


iliabylich:
Closes #1313

iliabylich:
Ready for review

lib/opal/nodes/iter.rb
```diff
@@ -50,88 +36,120 @@ def compile
...
+            line "  #{var} = #{source};"
+            line "} else {"
+            line "  #{var} = #{source}[0];"
+            line "}"
+          else
+            # decompressing |(a, b)| argument
+            line "if (#{source} == null || !#{source}.$$is_array) {"
+            indent do
+              line "if (self.$$is_lambda) {"
+              line "  throw Opal.ArgumentError.$new('wrong number of arguments');"
```
elia: I think this can be skipped if arity check is disabled

lib/opal/nodes/iter.rb
```diff
@@ -50,88 +36,120 @@ def compile
...
       end
 
-      # does this iter define a block_pass
       def extract_block_arg
-        if args.last.is_a?(Sexp) and args.last.type == :block_pass
-          block_arg = args.pop
-          block_arg = block_arg[1][1].to_sym
+        if args.is_a?(Sexp) && args.last.is_a?(Sexp) and args.last.type == :block_pass
+          self.block_arg = args.pop
+          self.block_arg = block_arg[1][1].to_sym
```
elia: Any reason for not doing `self.block_arg = args.pop[1][1].to_sym`?

lib/opal/nodes/iter.rb
```diff
@@ -50,88 +36,120 @@ def compile
...
-          s(:array, args_sexp)
+        sexp = if Fixnum === args_sexp or args_sexp.nil?
+          s(:args)
+        elsif args_sexp.is_a?(Sexp) && args_sexp.type == :lasgn
+          s(:args, args_sexp)
         else
           args_sexp[1]
         end
+
+        # compacting _ arguments into a single one (only the first one leaves in the sexp)
```
elia: Just wanted to say that this kind of comments are super valueable! :sparkles:

lib/opal/nodes/node_with_args.rb
```diff
@@ -0,0 +1,148 @@
...
+            add_local var_name
+            line "if ((#{var_name} = $kwargs.$$smap['#{arg_name}']) == null) {"
+            line "  #{var_name} = ", expr(kwarg[2])
+            line "}"
+          when :kwarg
+            arg_name = kwarg[1]
+            var_name = variable(arg_name.to_s)
+            add_local var_name
+            line "if ((#{var_name} = $kwargs.$$smap['#{arg_name}']) == null) {"
+            line "  throw Opal.ArgumentError.$new('expecting keyword arg: #{arg_name}')"
```
elia: Can be skipped unless arity check is enabled

lib/opal/nodes/node_with_args.rb
```diff
@@ -0,0 +1,148 @@
...
+          line "  $kwargs = $hash2([], {});"
+          line "}"
+        else
+          line "if ($kwargs == null) {"
+          line "  $kwargs = $hash2([], {});"
+          line "}"
+        end
+
+        line "if (!$kwargs.$$is_hash) {"
+        line "  throw Opal.ArgumentError.$new('expecting keyword args');"
```
elia: Can be skipped unless arity check is enabled

elia:
Hey, big :clap:! 

Added a few minor notes but otherwise looks good, uniforming the args is great, especially in the perspective of moving to the `parser` gem. Btw did you get the sexp names from there by any chance?

@alexdowad can you have a check too?

iliabylich:
@elia Moving to parser is really necessary, opal's grammar misses a looot of syntax cases. At least, compare the size of [this](https://github.com/whitequark/parser/blob/master/lib/parser/ruby23.y) and [this](https://github.com/opal/opal/blob/master/lib/opal/parser/grammar.y), still wondering why there are just a few known parsing issues.

PR #1366, opened 2016-02-26
Fix nodejs/yaml

psparrow:
nodejs/yaml was throwing errors upon usage as it relied on OpalNode.node_require instead of Kernel#node_require like the nodejs/file class.


elia:
Thanks! That part really needs some <3 

:)

psparrow:
No problem. I’m really interested in using opal to build node and node-webkit apps. So, maybe I’ll try to contribute more.

PR #1367, opened 2016-02-28
Fix #1237 and refactor super

### LINKED ISSUE #477: Wrong super calling with block and parenthesis
```ruby
class Foo
	def foo(*args, &block)
    puts [*args, block].compact.length
	end

  def bar(*args, &block)
    puts [*args, block].compact.length
  end
end

class Bar < Foo
	def foo(a)
		super do end
	end

  def bar(a)
    super() do end
  end
end

Bar.new.foo(1)
Bar.new.bar(1)
```

This prints `2 1` in Ruby and `1 1` in Opal.

### LINKED ISSUE #1237: Passing block explicitly to superclass method fails
At one point this worked (or maybe was not fully working) in Opal 0.9 but now it complains:

```ruby
class SuperClass
  def add_stub(&block)
    puts 'super method, calling block'
    block.call
  end
end

class InheritingClass < SuperClass
  def add_stub(*args, &block)
    super(*args, &block)
  end
end

InheritingClass.new.add_stub do
  puts 'block ran'
end
```

Runs OK in MRI. It does not happen with a standalone method nor does it happen when 1 method calls another in the same fashion, just when calling the superclass' method. This also causes arity checking to break.

MRI:
```
super method, calling block
block ran
```

Opal:
```
super method, calling block

/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:4249
      throw exception;
            ^
call: undefined method `call' for nil
    at OpalClass.Opal.defs.self [as $new] (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:4466:15)
    at NilClass.Opal.defn.TMP_1 (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:3028:54)
    at NilClass.method_missing_stub [as $call] (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:888:35)
    at $InheritingClass.Opal.defn.TMP_1 (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:19037:20)
    at $InheritingClass.Opal.defn.TMP_2 [as $add_stub] (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:19056:72)
    at Opal.dynamic_require_severity (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:19061:66)
    at Object.<anonymous> (/tmp/opal-nodejs-runner-20151209-19233-13hhbbd:19062:3)
```

wied03:
The main idea here besides fixing #1237 is to focus the runtime method solely on finding the function, then alter the compiler to cover missing functionality use the existing call node to avoid duplicating logic.

Things it would be nice to get an opinion on before I squash:
* Embedding errors - I did this for a few cases where the compiler can detect an error (using super in a block where you can't for example). I bake in raise into the code to make it a runtime error and be MRI/rubyspec compliant, but at least this way the compiler does the heavy lifting of checking. Is this an OK way to go?
* Call/Super - Started out trying to reuse stuff and I think it largely remained that way. On the contrary, maybe the SuperNode customizes enough that it might be worth it to not inherit from CallNode? I think overall, inheriting is still slightly preferable.
* Call refactoring - Tried to break up the compile method a little better and make if statements/methods more intuitive as to what's going on
* Compiler specs - Made it a lot easier to track down everything going on here
* Renamed a bunch of child expressions in the CallNode - This made it easier for me to follow when I was refactoring but I can back these out

Fixes #1237 
Fixes #477


meh:
Looks good to me and I'm fine with all points, also agree that it's better to inherit from `CallNode`.

lib/opal/nodes/super.rb
```diff
@@ -52,45 +151,51 @@ class DefinedSuperNode < BaseSuperNode
...
-        # insert method body to find super method
-        self.compile_dispatcher
-
-        wrap '((', ') != null ? "super" : nil)'
+        add_method(nil)
+        # will never come back null with method missing on
+        if compiler.method_missing?
+          wrap '(!(', '.$$stub) ? "super" : nil)'
+        else
+          # TODO: With method_missing support off, something breaks in runtime.js's chain
```
elia: This is interesting, we should either deprecate disabling MM or add some specs to ensure it "kinda" works, because I'm quite sure rubyspec relies on it and so does probably mspec (could be wrong tho).

@opal/core thoughts? (Personally I'm for deprecating it telling ppl to tell us if they actually disable it)

elia:
> Renamed a bunch of child expressions in the CallNode - This made it easier for me to follow when I was refactoring but I can back these out

but now it feels much less like a real™ compiler!

jk, nice job!

TBH I'm a bit conflicted about the new call specs because I "predict" they'll need an update every time someone touches the compiler, but after all it's probably a good thing and will force to think about what one's doing.

:vertical_traffic_light: green light for the squash 😀

meh:
@elia I'm pretty sure we already have that kind of test in other places, at least I remember getting beaten by it.

wied03:
The main reason I added them is I was getting lost in the complexity of all the permutations.

wied03:
Commits have been squashed

wied03:
I would like to have fixed more but there are a few other problems with yields within blocks, etc.

elia:
> I would like to have fixed more but there are a few other problems with yields within blocks, etc.

This is great stuff already!

wied03:
@meh @elia - I might be repeating myself, but I found it near impossible to debug the compiler issues without the compiler specs. Running the mspecs generated too much noise when trying to analyze the node structure. etc.

What *might* help the brittleness of the compiler specs is writing some RSpec custom matchers to DRY up things a bit. then there's just 1 matcher that translates "should pass X along in a block" into string testing

elia:
@wied03 no worries, overall (as said) I think it's good to have those specs :)

PR #1368, opened 2016-02-29
Updated ruby specs for super PR

wied03:
Exclude encoding related string spec introduced in rubyspec commit ruby/spec@2e3116d41adf1e56c060e0c158a7e48831609291


elia:
:+1:

PR #1372, opened 2016-03-01
More super fixes

wied03:
Transition error handling to runtime (too difficult and error prone to do it on the compiler side), fix method_missing error handling , case, added missing test cases, DRY find_iter_super_dispatcher

This gets opal-rspec in the neighborhood again

### DIFF
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -71,17 +71,8 @@ module Opal
             add_local '$zuper_length'
 
             line "$zuper = [];"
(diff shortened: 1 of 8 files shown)


wied03:
I should add that some error checks (like trying to do the implicit params check (super() vs super) in define_method vs. a block were going to be impossible without moving to checking them at runtime. The compiler just doesn't know where a block will be passed and how it will be used.

elia:
I thought I had merged this already… thanks!

PR #1373, opened 2016-03-01
singleton method_added fixes

wied03:
Ruby specs - https://github.com/ruby/spec/pull/200

Another RSpec related issue


opal/corelib/runtime.js
```diff
@@ -1350,7 +1350,7 @@
       }
     }
 
-    if (obj.$method_added && !obj.$method_added.$$stub) {
+    if (obj.$method_added && !obj.$method_added.$$stub && (obj.$$singleton_of !== Opal.BasicObject)) {
```
elia: In light of what pointed out by @eregon in https://github.com/ruby/spec/pull/200#issuecomment-190636009 this probably should look like 
```rb
if (obj.$method_added && !obj.$method_added.$$stub && !obj.$$singleton_of) {
  …
}
else if (obj.$singleton_method_added && !obj.$singleton_method_added.$$stub && obj.$$singleton_of) {
  …
}
```

and that would probably fix some more ruby specs too… :)

wied03:
@elia - Implemented your feedback. I've always found it a little odd to grasp the singleton class stuff, but as the code in this PR stands now. Should we be calling $singleton_method_added inside the else on obj instead? Or alternatively, should the else look like this?

```js
else if (singleton_of && singleton_of.$singleton_method_added && !singleton_of.$singleton_method_added.$$stub)
```

elia:
Honestly I'm not sure without delving into it again, in doubt skip the else :)

wied03:
It passes the rubyspecs either way, which makes me uncomfortable. Maybe I can dig in and find out why.

eregon:
@wied03 I think your condition above in the commend is good.
`obj` is a Module or a Class here, which might be confusing naming, and `singleton_of` is the singleton class' instance AFAIK. `singleton_method_added` is called on the singleton instance, not a module.
The tests pass because modules/classes respond to `singleton_method_added` since they are BasicObject.

wied03:
OK. That makes sense. I pushed that change as well and squashed the commits.

PR #1385, opened 2016-03-06
Super/Aliasing Fix

### LINKED ISSUE #1384: Super/Alias Problem
Weird edge case. I might be able to fix this soon but wanted to document it. Has something to do with alias_method & super

code:
```ruby
module Mod
  attr_reader :foobar
end

class Foo
  include Mod
    
  def initialize
    @foobar = :yes
  end
    
  def foobar
    super()
  end

  alias_method "foobar?", :foobar
    
  def foobar
    @foobar
  end
end

puts 'call regular'
puts Foo.new.foobar
puts 'call alias'
puts Foo.new.foobar?
```

ran `ruby test.rb`:
```
call regular
yes
call alias
yes
```

ran `bundle exec opal -A test.rb`:
```
call regular
yes
call alias

/tmp/opal-nodejs-runner-20160306-1265-182v9i:1073
      throw new Error("could not find current class for super()");
            ^
Error: could not find current class for super()
    at Opal.find_obj_super_dispatcher (/tmp/opal-nodejs-runner-20160306-1265-182v9i:1073:13)
    at Opal.find_super_dispatcher (/tmp/opal-nodejs-runner-20160306-1265-182v9i:1025:27)
    at Opal.find_iter_super_dispatcher (/tmp/opal-nodejs-runner-20160306-1265-182v9i:1055:17)
    at $Foo.$a.$$p.TMP_2 [as $foobar?] (/tmp/opal-nodejs-runner-20160306-1265-182v9i:19891:37)
    at Opal.dynamic_require_severity (/tmp/opal-nodejs-runner-20160306-1265-182v9i:19907:57)
    at Object.<anonymous> (/tmp/opal-nodejs-runner-20160306-1265-182v9i:19908:3)
    at Function.Module._load (module.js:312:12)
```

wied03:
Fixes #1384. Another weird quirk of the Ruby language (if it's a weird corner, RSpec will find it). It still needs the following:
- Rubyspec built from the anon_mod.rb file *done*
- This is probably incomplete as it really needs to be something that also fixes the Rubyspec `Module#alias_method can call a method with super aliased twice` *save this for another PR*

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1070,9 +1070,10 @@
     var klass = obj.$$meta || obj.$$class;
     jsid = '$' + jsid;
 
+    // first we need to find the class/module current_func is defined/donated on
     while (klass) {
-      if (klass.$$proto[jsid] === current_func) {
-        // ok
+      // when a module has been included in a class, look for that
+      if ((klass.$$iclass && klass.$$module === current_func.$$owner) || klass === current_func.$$owner) {
         break;
       }
 
@@ -1086,7 +1087,7 @@
 
     klass = klass.$$parent;
 
-    // else, let's find the next one
+    // now we can find the super
     while (klass) {
       var working = klass.$$proto[jsid];
 
@@ -1358,6 +1359,8 @@
   //
   Opal.defn = function(obj, jsid, body) {
     obj.$$proto[jsid] = body;
+    // for super dispatcher, etc.
+    body.$$owner = obj;
 
     if (obj.$$is_module) {
       Opal.donate(obj, jsid);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ Whitespace conventions:
 
 ### Fixed
 
+- `super` works properly with overwritten alias methods (#1384)
 - `NoMethodError` does not need a name to be instantiated
 - `method_added` fix for singleton class cases
 - Super now works properly with blocks (#1237)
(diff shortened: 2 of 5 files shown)


wied03:
@meh - any thoughts here?

opal/corelib/runtime.js
```diff
@@ -1347,6 +1356,8 @@
   //
   Opal.defn = function(obj, jsid, body) {
     obj.$$proto[jsid] = body;
+    // super dispatcher find, etc. might need to lookup by function body
+    obj.$$funcbodies[body] = jsid;
```
elia: This rings an alarm for me, `[body]` will convert the function to a string and use that as the key, which means that will use the actual source code of the function to identify it. That, to me, is a problem because:

- two methods in the inheritance chain can have the same implementation this defeating this technique (even if I'm aware it's surely not common)
- I'm not sure how standard is to have `Function.prototype.toString()` converto the function to its source code, what if the implementation is changed?
- looks like a hack
- could mislead a future me reading this code

Even if I recognize it's practical, I'd rather prefer to have a more robust system, that actually checks a prototype chain for the particular method or alternatively assign an `Opal.uid()` to each function and use that instead of the function body.

wied03: What about the first way/first commit?

sent from my mobile device

wied03:
rubyspecs - https://github.com/ruby/spec/pull/212

wied03:
@elia - Thanks for the earlier feedback. I think with this 2nd commit, I got the nested loop eliminated and have something that can work OK.

elia:
:+1: :)

PR #1387, opened 2016-03-09
Add ruby/spec/language/block spec to the test suite

iliabylich:


lib/opal/nodes/def.rb
```diff
@@ -199,7 +199,10 @@ def compile
               scope.mlhs_mapping[child] = tmp
 
               child.children.each do |child|
-                scope.add_temp variable(child.last)
+                # No support for nested mlhs yet
+                if child.type != :mlhs
+                  scope.add_temp variable(child.last)
+                end
```
elia: shouldn't this blow up when it founds unspported args?

iliabylich: What do you mean by unsupported arguments? From what I see in grammar.y `mlhs` can include only `norm_arg`, `restarg` or another `mlhs`. All other cases are rejected on the stage of translating to sexp.

elia: what I meant is this:

```
raise NotImplementedError, "No support for nested mlhs yet" if child.type == :mlhs
scope.add_temp variable(child.last)
```

but maybe it doesn't make sense

IlyaBylich: This code raises an error on the compilation stage and makes `ruby/spec/language/block_spec` unparsable. Could we just ignore nested `mlhs` arguments for now?

iliabylich: This code raises an error on the compilation stage and makes ruby/spec/language/block_spec unparsable. Could we just ignore nested mlhs arguments for now?

elia: :+1:

elia:
Awesome

iliabylich:
Thanks!

Issue #821, opened 2015-04-30
Backslash for line continuation does not work


vais:
```
$ cat test.rb 
a = "a"
b = "b"
puts a == \
  b
```
```
$ bundle exec opal test.rb 

/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20150430-1056-wi8a05:14300
  return self.$puts(a['$=='](self.$\
                                  ^^
SyntaxError: Unexpected token ILLEGAL
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:929:3
```

vais:
But the following works:
```
$ cat test.rb 
e = 8 + 9 \
    + 10
puts e
```
```
$ bundle exec opal test.rb 
27
```
Looks like the issue is specifically with a backslash following `==`

wied03:
@vais - Same error happens in rspec's spec code:
```ruby
expect(group.examples.first.exception.message).to \
          match(/may not be used outside of examples/)
```

vais:
@wied03 I couldn't figure it out, need @adambeynon's help on this one.

wied03:
Another case that doesn't work, I noticed while foolishly trying to make the entire activesupport lib work in Opal:

```ruby
def advance(options)
    unless options[:weeks].nil?
      options[:weeks], partial_weeks = options[:weeks].divmod(1)
      options[:days] = options.fetch(:days, 0) + 7 * partial_weeks
    end

    unless options[:days].nil?
      options[:days], partial_days = options[:days].divmod(1)
      options[:hours] = options.fetch(:hours, 0) + 24 * partial_days
    end

    d = to_date.advance(options)
    datetime_advanced_by_date = change(:year => d.year, :month => d.month, :day => d.day)
    seconds_to_advance = \
      options.fetch(:seconds, 0) +
      options.fetch(:minutes, 0) * 60 +
      options.fetch(:hours, 0) * 3600

    if seconds_to_advance.zero?
      datetime_advanced_by_date
    else
      datetime_advanced_by_date.since(seconds_to_advance)
    end
  end
```

vais:
@iliabylich I believe this one can be closed now, could you please confirm?

iliabylich:
[...]

PR #1390, opened 2016-03-10
Fix parsing of method call starting after "\" on the next line.

### LINKED ISSUE #1388: Line continuation mishandling
There appears to be at least one case where a line continuation isn't handled correctly by Opal. Attempting to execute the compiled output of [this method](https://github.com/nofxx/georuby/blob/v2.5.2/lib/geo_ruby/simple_features/point.rb#L422-L432) causes an error to be thrown.

The compiled method:
```javascript
Opal.defs(self, '$from_latlong', function(lat, lon, srid) {
  var $a, $b, TMP_3, self = this, p = nil, point = nil;

  if (srid == null) {
    srid = $scope.get('DEFAULT_SRID')
  }
  p = ($a = ($b = [lat, lon]).$map, $a.$$p = (TMP_3 = function(l){var self = TMP_3.$$s || this, $a, $b, sig = nil, deg = nil, min = nil, sec = nil, cen = nil, dec = nil;
 null) l = nil;
  $b = self.$\
n(/(-)?(\d{1,2})\D*(\d{2})\D*(\d{2})(\D*(\d{1,3}))?/).$flatten(), $a = Opal.to_ary($b), sig = ($a[0] == null ? nil : $a[0]), deg = ($a[1] == null ? nil : $a[1]), min = ($a[2] == null ? nil : $a[2]), sec = ($a[3] == null ? nil : $a[3]), cen = ($a[4] == null ? nil : $a[4]), $b;
    if ((($a = l['$=~'](/W|S/)) !== nil && (!$a.$$is_boolean || $a == true))) {
      sig = true};
    dec = $rb_plus(deg.$to_i(), $rb_divide(($rb_plus($rb_times(min.$to_i(), 60), ((("") + (sec)) + (cen)).$to_f())), 3600));
    if (sig !== false && sig !== nil) {
      return $rb_times(dec, -1)
      } else {
      return dec
    };}, TMP_3.$$s = self, TMP_3), $a).call($b);
  point = self.$new(srid);
  return point.$set_x_y(p['$[]'](0), p['$[]'](1));
});
```

Specifically, this line: `$b = self.$\`

iliabylich:
Closes #1388

### DIFF
--- a/lib/opal/parser/lexer.rb
+++ b/lib/opal/parser/lexer.rb
@@ -674,6 +674,8 @@ module Opal
         @lex_state = :expr_end
       end
 
+      self.yylval = matched
+
       return matched =~ /#{REGEXP_START}[A-Z]/ ? :tCONSTANT : :tIDENTIFIER
     end
 
--- a/spec/lib/parser/call_spec.rb
+++ b/spec/lib/parser/call_spec.rb
@@ -23,6 +23,10 @@ describe "Method calls" do
     parsed("foo\n.bar").should == [:call, [:call, nil, :foo, [:arglist]], :bar, [:arglist]]
     lambda { parsed("foo\n..bar") }.should raise_error(Exception)
   end
+
+  it "parses method starting on the next line after \\" do
+    parsed("\\\nfoo").should == [:call, nil, :foo, [:arglist]]
+  end
 end
 
 describe "Operator calls" do


vais:
I think this should close https://github.com/opal/opal/issues/821 as well :bow:

PR #1391, opened 2016-03-10
Add ability to run inverted test suite

iliabylich:
Not 100% sure that feature is really needed, but basically it allows to run only those specs which are marked as "bugs" in the `/filters/bugs` directory.
1. By setting env var `INVERT_RUNNING_MODE` opal's rake task will switch the logic of excluding examples from the test suite.
2. New formatter doesn't print failed examples. Instead, it prints specs that are marked as "bugs" but in fact are passing.

Here's the output that I've got from running this task on `PATTERN=spec/ruby/core/*/*_spec.rb rake mspec_ruby_nodejs`:
```
Expected to fail:
  1) "Module#module_function as a toggle (no arguments) in a Module body affects evaled method definitions also even when outside the eval itself"
  2) "Module#module_function as a toggle (no arguments) in a Module body functions normally if both toggle and definitions inside a eval"
  3) "Module#class_eval passes the module as the first argument of the block"
  4) "Module.allocate returns a fully-formed instance of Module"
  5) "Module#class_variable_set sets the value of a class variable with the given name defined in an included module"
  6) "Module#constants includes names of constants defined after a module is included"
  7) "Module#include recursively includes new mixins"
  8) "Module#include preserves ancestor order"
  9) "Module#include ignores modules it has already included via module mutual inclusion"
  10) "Module#ancestors returns only modules and classes"
  11) "Module#ancestors when called on a singleton class includes the singleton classes of ancestors"
  12) "Proc#arity for instances created with -> () { } returns positive values for definition \n    @a = -> (a:) { }\n    @b = -> (a:, b:) { }\n    @c = -> (a: 1, b:, c:, d: 2) { }"
  13) "Proc#arity for instances created with -> () { } returns positive values for definition \n    @a = -> (a, b:) { }\n    @b = -> (a, b:, &l) { }"
  14) "Proc#arity for instances created with -> () { } returns positive values for definition \n    @a = -> (a, b, c:, d: 1) { }\n    @b = -> (a, b, c:, d: 1, **k, &l) { }"
  15) "Proc#arity for instances created with -> () { } returns positive values for definition \n    @a = -> ((a, (*b, c))) { }\n    @b = -> (a, (*b, c), d, (*e), (*)) { }"
  16) "Proc#arity for instances created with lambda { || } returns positive values for definition \n    @a = lambda { |a:| }\n    @b = lambda { |a:, b:| }\n    @c = lambda { |a: 1, b:, c:, d: 2| }"
  17) "Proc#arity for instances created with lambda { || } returns positive values for definition \n    @a = lambda { |a, b:| }\n    @b = lambda { |a, b:, &l| }"
  18) "Proc#arity for instances created with lambda { || } returns positive values for definition \n    @a = lambda { |a, b, c:, d: 1| }\n    @b = lambda { |a, b, c:, d: 1, **k, &l| }"
  19) "Proc#arity for instances created with lambda { || } returns positive values for definition \n    @a = lambda { |a, b, c:, d: 1| }\n    @b = lambda { |a, b, c:, d: 1, **k, &l| }"
  20) "Range#hash generates a Fixnum for the hash value"
  21) "Class#allocate returns a fully-formed instance of Module"
  22) "Class#superclass for a singleton class of a class returns the singleton class of its superclass"
  23) "Class#new passes the block to #initialize"
  24) "Class#dup sets the name from the class to nil if not assigned to a constant"
  25) "Class#dup stores the new name if assigned to a constant"
  26) "Struct#inspect returns a string representation of some kind"
  27) "Time#strftime should be able to show the number of seconds since the unix epoch"
  28) "String#split with Regexp includes all captures in the result array"
  29) "Kernel#extend makes the class a kind_of? the argument"
  30) "Kernel.Array does not call #to_a on an Array"
  31) "Kernel#Array does not call #to_a on an Array"
  32) "Kernel#is_a? returns true if given a Module that object has been extended with"
  33) "Kernel#kind_of? returns true if given a Module that object has been extended with"
  34) "Kernel#raise raises RuntimeError if no exception class is given"
  35) "BasicObject metaclass has Class as superclass"
  36) "BasicObject metaclass contains methods for the BasicObject class"
```

And from running `spec/language/*_spec.rb`:
```
Expected to fail:
  1) "A lambda literal -> () { } assigns variables from parameters for definition '@a = -> (*) { }'"
  2) "A lambda literal -> () { } assigns variables from parameters for definition '@a = -> (*, &b) { b }'"
  3) "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |*| }'"
  4) "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |*, &b| b }'"
  5) "Predefined global $! remains nil after a failed core class \"checked\" coercion against a class that defines method_missing"
  6) "Predefined global $! should be set to the value of $! before the begin after a successful rescue"
  7) "Predefined global $! should be set to the value of $! before the begin after a successful rescue within an ensure"
  8) "Predefined global $! should be set to the new exception after a throwing rescue"
  9) "Predefined global $! in bodies without ensure should be cleared when an exception is rescued"
  10) "Predefined global $! in bodies without ensure should be cleared when an exception is rescued even when a non-local return is present"
  11) "Predefined global $! in bodies without ensure should not be cleared when an exception is not rescued"
  12) "Predefined global $! in bodies without ensure should not be cleared when an exception is rescued and rethrown"
  13) "Predefined global $! in ensure-protected bodies should be cleared when an exception is rescued"
  14) "A singleton class inherits from Class for classes"
  15) "A singleton class is a subclass of a superclass's singleton class"
  16) "A singleton class is a subclass of the same level of superclass's singleton class"
  17) "A singleton class for BasicObject has Class as it's superclass"
  18) "A singleton class for BasicObject has the proper level of superclass for Class"
  19) "Instance methods of a singleton class for a singleton class includes instance methods of the singleton class of Class"
  20) "Class methods of a singleton class for a class include class methods of Class"
  21) "Class methods of a singleton class for a class include instance methods of the singleton class of Class"
  22) "Class methods of a singleton class for a singleton class include class methods of the singleton class of Class"
```

(Funny moment here: there are some test in `spec/ruby/language/while_spec` that are compiled as a JS with infinite loop, I had to comment it temporary, need to investigate later)

Feel free to ping me about the code issues, I don't really like what I've got.
Will add some documentation after approval, not sure that Opal needs it.


spec/mspec-opal/runner.rb
```diff
@@ -14,7 +24,7 @@ def register
 
   def ===(description)
     @seen << description
-    @filters.include? description
+    @filters.filtered?(description)
```
elia: Why don't just put an "if ENV" here and stick to Set?

Same goes for the formatter, I'd just wrap the formatter assignment with "if ENV" and assign the inverted formatter instead. 

A $global instead of ENV would be ok too. 

As non-DRY as it is, it's always easy to do a project-wide search for the ENV var and find all the places in which it's used. Especially since this is basically internal configuration code (v. library code).

elia:
This is absolutely needed and I'm very happy you tackled this finally :)

I'm quite sure that there's some standard MSpec mechanism to do the same, but given the nature of the code I'm ok to merge this as it is (despite I'd prefer it to be simpler as per the other comment). 

The only important part to have is some brief documentation in hacking.md with a red to the ENV var (more or less the description of this PR). 

😊

iliabylich:
@elia Finally green. Made fixes according to your comments. I was 100% sure that it's impossible to use `ENV` in opal (even if node runs it). I've tried putting `p ENV` to the beginning of `spec_helper` and didn't notice that platform-specific code gets required after `spec_helper`. Added docs, fill free to correct any mistakes :smile:

elia:
👍🏼😸 great to see we were already passing al that stuff

PR #1393, opened 2016-03-13
Optimize number spaceship

jgaskins:
A function containing a try/catch construct cannot be optimized. This commit extracts the work to a function that can be optimized. The actual `<=>` operator just wraps it and potentially rescues the exception.

I made a [jsperf](http://jsperf.com/deoptimizations-with-try-catch-finally) for a talk I gave a couple weeks ago that demonstrates the kind of performance benefit this sort of extraction has.

### DIFF
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -176,18 +176,32 @@ class Number < Numeric
     }
   end
 
-  def <=>(other)
-    %x{
+  # Compute the result of the spaceship operator inside its own function so it
+  # can be optimized despite a try/finally construct.
+  %x{
+    var spaceship_operator = function(self, other) {
       if (other.$$is_number) {
         if (isNaN(self) || isNaN(other)) {
           return nil;
         }
 
-        return self > other ? 1 : (self < other ? -1 : 0);
+        if (self > other) {
+          return 1;
+        } else if (self < other) {
+          return -1;
+        } else {
+          return 0;
+        }
       }
       else {
-        return #{__coerced__ :<=>, other};
+        return #{__coerced__ :<=>, `other`};
       }
+    }
+  }
+
+  def <=>(other)
+    %x{
+      return spaceship_operator(self, other);
     }
   rescue ArgumentError
     nil


meh:
Maybe this could actually be implemented on the compiler side?

jgaskins:
@meh Extracting the body from a `def` / `rescue` / `end` method would be great in the compiler because it would speed up every one of these calls. I haven't seen many places where this has been a big deal, but I would assume all `<=>` calls have this — or at least all standard Ruby ones.

That does bring up the question of what to name the extracted functions so two of them in the same scope won't collide.

elia:
The compiler has built in support for progressively named TMP vars which should work ok in this case

jgaskins:
@elia That's true, and that would definitely help with naming collision, but after the work done to add method names so stack traces/profiling/timelines weren't a mess of `(anonymous function)`, I thought it might be nice to include a descriptive name along with it.

Another possibility would be to have something that works as if it were written like this:

```ruby
class Number
  %x{
    var rescued_functions = {};
  }

  def <=> other
    `rescued_functions['$<=>'].apply(self, arguments)`
  rescue
    nil
  end

  %x{
    rescued_functions['$<=>'] = function(other) {
      // Insert actual implementation here
    };
  }
end
```

This way, there are no naming collisions because the JS functions are named the same as the wrapped version and the names are just as descriptive as they are currently. It also doesn't expose the unwrapped version as a callable method. Maybe the wrapping could be something other than a POJO, but I'm just throwing out ideas.

iliabylich:
@jgaskins :+1: for generating meaningful function names. But there's a problem: you may have 2 begin/rescue statements in a single method. So we can't compile a method with `rescue` to something like
``` js
function method_a_rescue_body() {
  // ...
}
Opal.defn('method_a', function() {
  try { method_a_rescue_body.apply ... } catch ()
})
```
These `*_rescue_body` functions may depend on some local variables defined outside of the rescue. What I'm talking about is:

``` ruby
def a_method
  %x{var a = 1, b = 2}
  # first rescue
  begin
    %x{a = 3}
  rescue
  end

  # second rescue
  begin
    %x{b = 4}
  rescue
  end
end
```

We can't detect cases like that on the compilation stage. Don't even know how to fix it.

(Probably we can make an optimizatios for methods containing **only** a begin/rescue/end statement, it doesn't look complex)

elia:
Personally I'd fix the TMP vars generator to accept a name so that other places can benefit from that and get better names too. Eg. Methods with blocks, lambdas etc.

jgaskins:
> you may have 2 begin/rescue statements in a single method
I wasn't too concerned about this. I think if you're going to wrap only part of the method, it's up to you to extract the boundaries of that begin/rescue/end. I was thinking more along the lines of def/rescue/end.

The rescue block having variables in scope from the wrapped version is indeed a blocker for this solution, however we may be able to dynamically define (and cache) the function within the Ruby method. I'll write an example of what I mean when I'm not on a mobile device. :-)

iliabylich:
V8 optimizes functions defined in `try/catch` so we can simply wrap
`s(:resbody)` with a `(function() { ... })()`
https://gist.github.com/iliabylich/cfd3ec68b2c1b5d13393

jgaskins:
@iliabylich It optimizes at the function level, so you're correct that functions called in and around the `try` / `catch` / `finally` construct can indeed be optimized. However, for a function to qualify for optimizer passes, it has to execute more than a certain number of times. A function created within the try/catch and immediately executed is executed only once and thrown away because it's not being stored, so it won't realistically be optimized. The reason your example does get optimized is because you _are_ storing it in the `a` variable.

iliabylich:
@jgaskins Yes, you are definitely right. So, the code
``` ruby
def m
  1
rescue
end
```
should be compiled to
``` js
function m_res_body() {
  return 1;
}
Opal.defn(self, 'm', function() {
  try {
    return m_res_body();
  } catch(e) {}
})
```
And the only case when it can be optimized is when our method's body has only `:rescue` node, right?

And on more thing to mention: `m_res_body` **must** take all the arguments that were passed to the method. It can be challenging :smile:

Btw, what happens if `rescue` block uses `arguments` directly? In this case we have to pass method arguments to `m_res_body` and then we should duplicate the logic for extracting `arguments`... Don't know if it makes sense, ping me if you have any questions.

elia:
Maybe they can be stored on the method itself:

```js
Opal.defn(self, 'm', TMP_1 = function() {
  TMP_1.$$rescue_1 = TMP_1.$$rescue_1 || function m_res_body_1() {
    return 1;
  };
  try {
    return TMP_1.$$rescue_1();
  } catch(e) {}
})
```
that way it should also be easier to implement in the compiler…
didn't think through it tho, does it make sense?

iliabylich:
@elia V8 says it's not optimized. @jgaskins is right, assigning `rescue_body` to an external variable (outside of the function with try/catch) makes it optimizable. (Just in case - https://gist.github.com/iliabylich/0971d5ed60e3f32cdfba , creating local variable like `outer = test.$rescue_body` and using `outer` directly makes it optimizable)

elia:
@iliabylich shouldn't it be `test.$rescue_body = ` **`test.$rescue_body ||`** `function(){ … }`?

jgaskins:
That may not be a bad idea. I keep forgetting JS lets you do store properties on functions, even though that's how half the language is implemented here. :-)

The fact that you've enumerated this also points out that this can be done multiple times. I'd only been concerned with the performance impact of the def/rescue/end construct because so many core methods use it specifically for gracefully, but this might be a decent way to improve performance for all of them.

> @iliabylich shouldn't it be `test.$rescue_body = test.$rescue_body || function(){ … }?`

Yep, and that results in `Function is optimized`. :-)

iliabylich:
haha, you are totally right! :smile:

iliabylich:
@elia what do you think about extracting all the code for `begin/rescue/else/ensure/end` to a runtime method that takes 4 callbacks?
``` js
// well, the name is discussable :)
Opal.begin_rescue_else_ensure_end(begin_body, rescue_body, else_body, ensure_body) {
  try {
    // invoking begin_body
    // and somehow else_body
  } catch(e) {
    // rescue_body
  } finally {
    // ensure_body
  }
}
```

iliabylich:
I've almost opened a PR that closes it :smile: . https://github.com/iliabylich/opal/commit/173e1df8fead296fc0e729e3584b859e0f7f5e82 solves an issue, and the `$$rescue_body` function becomes optimized. However, simple benchmark shows that it becomes 2-3 times slower.
``` ruby
def m
  1 + 1
rescue
end

t = Time.now
10_000_000.times do
  m
end
p Time.now - t
```
master: 0.8
master with patch: 2.1

I'm pretty sure we can close this PR (and I can remove my branch too :smile: ), this optimization brings more overhead that we expected

PR #1395, opened 2016-03-15
Split arguments logic to separate files

iliabylich:


lib/opal/nodes/def.rb
```diff
@@ -52,9 +52,14 @@ def compile
 
           add_temp 'self = this'
 
-          compile_rest_arg
-          compile_opt_args
-          compile_keyword_args
+          if has_complex_args?
+            compile_complex_args(false)
```
elia: maybe it's better to use `"false"` instead of `false` since it's gonna be JavaScript… in addition it might be better to clarify what's that using kwargs: `compile_complex_args(expand_code: 'false')` or something like that

opal/corelib/runtime.js
```diff
@@ -996,6 +996,215 @@
...
+  // @example
+  //   parameters = [1, 2, 3, 4, { d: 1, e: 2 }]
+  //   scheme = [
+  //     ['arg', 'a'],
+  //     ['restarg', 'b'],
+  //     ['arg', 'c'],
+  //     ['kwarg', 'd'],
+  //     ['kwrestarg', 'kwrest']
+  //   ]
+  //   (which is equivalent to (a, *b, c))
```
elia: shouldn't this be `(which is equivalent to (a, *b, c, d:, **kwrest))`?

spec/filters/bugs/language.rb
```diff
@@ -31,26 +31,14 @@
...
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |(*a, b)| [a, b] }'"
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |*, **k| k }'"
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |a, | a }'"
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition \n    @a = lambda do |(a, b, *c, d), (*e, f, g), (*h)|\n      [a, b, c, d, e, f, g, h]\n    end"
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition \n    @a = lambda do |a, (b, (c, *d, (e, (*f)), g), (h, (i, j)))|\n      [a, b, c, d, e, f, g, h, i, j]\n    end"
-  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition \n    @a = lambda do |a, b=1, *c, (*d, (e)), f: 2, g:, h:, **k, &l|\n      [a, b, c, d, e, f, g, h, k, l]\n    end"
+  fails "A lambda expression 'lambda { ... }' assigns variables from parameters for definition '@a = lambda { |a, | a }'" # requires arity checking in
```
elia: typo?

> requires arity checking ~~in~~ _on_

elia:
looks good, while you're there can you add some short doc before `node_with_args.rb` methods hinting the kind of args that methods is dealing with?

eg:

```rb
# kwopt is `k` in `def foo(a, b, j:, k: nil)
def kwopt_args
```

iliabylich:
@elia Fixed. Is it a good idea to use kwargs in the compiler's code? I thought it's better to prefer simple code constructions. We may get a situation when breaking kwargs breaks methods compilations in `opal-parser`.

Also a quick question. Do you think it's ok to pass method schema directly to the JS world? (That's what I'm really worrying about). And do you see any parts of `Opal.scan_args` that can/should be moved to the `NodeWithArgs`?

elia:
> Do you think it's ok to pass method schema directly to the JS world? (That's what I'm really worrying about). 

Yeah, it's probably a valid concern, wasn't sure either, if you come up with some more descriptive method name that would be great, otherwise it's ok as it is I guess :+1: 

> And do you see any parts of Opal.scan_args that can/should be moved to the NodeWithArgs?

Still need to properly peruse and understand scan_args(), not a small bite to swallow :yum:

opal/corelib/runtime.js
```diff
@@ -996,6 +996,223 @@
...
+  // :kwoptarg  is :g     in "def m(g: 1) end"
+  // :kwrestarg is :f     in "def m(**f) end"
+  //
+  // @param parameters [Array] arguments that were passed to the method
+  // @param scheme [Array] array in format [ [<arg type>, <arg name>] ]
+  // @param expand [true, false] expands +parameters+ if it contains a single Array element and expand is +true+
+  Opal.scan_args = function(parameters, scheme, expand) {
+    if (expand && scheme.length > 1 && parameters.length == 1 && parameters[0].$$is_array) {
+      parameters = parameters[0];
+    }
```
elia: This check seems a candidate for moving to `node_with_args`

opal/corelib/runtime.js
```diff
@@ -996,6 +996,223 @@
...
+        kwarg_name = argument_scheme[1];
+
+        if (!kwargs.$$is_hash || !kwargs.$$smap.hasOwnProperty(kwarg_name)) {
+          throw Opal.ArgumentError.$new("missing keyword: " + kwarg_name);
+        }
+        scheme_right--;
+        kwarg_value = kwargs.$$smap[kwarg_name];
+        used_kwargs[kwarg_name] = true;
+        result[kwarg_name] = kwarg_value;
+      } else if (argument_type === 'kwoptarg') {
```
elia: can you change this to:

```js
  …
}
else if (…
```

I'm more of a `} else` guy too, but it's for consistency with the rest of the code :)

iliabylich:
Any other comments before rebasing?

iliabylich:
Splitted args compilation to separate files. Removed `Opal.scan_args` (that was initially a bad idea :smile: ). All the logic is now in the `/nodes/`. There are ~20 failing examples in `spec/lib` which is expected, there are a lot of changes to the compiled code (I'll fix them a bit later). @opal/core Can I get a quick review please?

iliabylich:
ok, finally green :smile: . I've added few optimizations, so compiled corelib doesn't use post args at all.

lib/opal/nodes/args/initialize_kwargs.rb
```diff
@@ -0,0 +1,32 @@
...
+      def initialize_kw_args
+        helper :hash2
+
+        line "if ($kwargs && !$kwargs.$$is_hash) {"
+        line "  throw Opal.ArgumentError.$new('expected kwargs');"
+        line "}"
+
+        line "if (!$kwargs || !$kwargs.$$is_hash) {"
+        line "  $kwargs = $hash2([], {});"
+        line "}"
```
elia: I think this check should use `== null` that includes `null` and `undefined`.

```js
if ($kwargs != null && !$kwargs.$$is_hash) {
  throw Opal.ArgumentError.$new('expected kwargs');
}
if ($kwargs == null || !$kwargs.$$is_hash) {
  $kwargs = $hash2([], {});
}
```

or 


```js
if ($kwargs == null || !$kwargs.$$is_hash) {
  if ($kwargs == null) {
    $kwargs = $hash2([], {});
  } else {
    throw Opal.ArgumentError.$new('expected kwargs');
  }
}
```

not sure which is better, the last one saves a check for the happy path of actually passing kwargs.

lib/opal/nodes/args/initialize_kwargs.rb
```diff
@@ -0,0 +1,32 @@
...
+    # @see PostArgsNode
+    #
+    class InitializeKwargsNode < Base
+      def compile
+        unless scope.kwargs_initialized
+          initialize_kw_args
+        end
+      end
+
+      def initialize_kw_args
```
elia: I'm ok with this, but here's an alternative way I'd like to propose:

```rb
def compile
  return if scope.kwargs_initialized

  # … the current initialize_kw_args body
```

elia: or maybe removing `#compile` altogether and call `#initialize_kw_args` explicitly in subclasses instead of `super`, this I think would make more clear what's going on.

lib/opal/nodes/args/kwrestarg.rb
```diff
@@ -0,0 +1,40 @@
...
+        extract_code = "Opal.kwrestargs($kwargs, #{used_kwargs});"
+
+        if kwrestarg_name
+          var_name = variable(kwrestarg_name.to_sym)
+          add_temp var_name
+
+          line "#{var_name} = #{extract_code}"
+        else
+          # kwrestarg can be blank def m(**) end
+          line extract_code
```
elia: Is this really necessary? couldn't it just be skipped? (sorry if it's a dumb question…)

elia:
:clap: really impressive, great job :D

.travis.yml
```diff
@@ -44,9 +44,6 @@ matrix:
     - rvm: rbx
       env: RUN=rspec
 
-    - rvm: jruby
-      env: RUN=rspec
-
```
elia: this is no longer necessary: 23d73a544d (Use the latest jRuby stable version on Travis) :)

iliabylich:
@elia Fixed! Anything else?

elia:
and boom :boom: 

:)

PR #1405, opened 2016-03-21
Resolves #1404, Implement File.stat

### LINKED ISSUE #1404: Implement File.stat method
http://ruby-doc.org/core-2.2.0/File.html#method-c-stat

Mogztter:
`File::Stat` object has a *lot* of methods but for now I've just implemented `file?`.

Resolves #1404

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -74,8 +74,10 @@ class File < IO
     end
   end
 
-
-
+  def self.stat path
+    path = path.path if path.respond_to? :path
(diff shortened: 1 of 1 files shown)


Mogztter:
@coveralls Not sure how the coverage can increase by 0.09% without test :confused: (rounding bug?)

elia:
@Mogztter I'm not even sure coveralls works at all with our setup… :weary: :D

elia:
@Mogztter a set of specs targeting node and stdlib/nodejs explicitly is more needed every day

PR #1292, opened 2016-01-17
Resolves #1290, generate opal-builder.js

Mogztter:
I made some progress toward a 100% JavaScript to compile Ruby code to JavaScript :smile: 

Remaining things to do:
- [x] Implement `Pathname.expand_path` method (used in [Hike](https://github.com/sstephenson/hike/blob/v1.2.3/lib/hike/trail.rb#L49)) #1291
- [x] Implement `Kernel.open()` method (used in [Opal.PathReader](https://github.com/opal/opal/blob/master/lib/opal/path_reader.rb#L12)) #1218
- [x] Implement  `Pathname.join` method (used in [Hike](https://github.com/sstephenson/hike/blob/v1.2.3/lib/hike/paths.rb#L23)) #1301  
- [x] Incorrect type with splat operator in HikePathFinder with Opal.paths #1331

PR #1406, opened 2016-03-21
Replace gsub! with gsub

Mogztter:
`gsub!` doesn't work in a JavaScript environment. To be able to generate `opal-builder.js`, we need to replace `gsub!` with `gsub`.

Related to #1292

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -119,7 +119,7 @@ module Opal
     end
 
     def process_require(filename, options)
-      filename.gsub!(/\.(rb|js|opal)#{REGEXP_END}/, '')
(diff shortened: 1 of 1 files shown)


wied03:
@iliabylich - Isn't this the kind of thing those tests should catch?

iliabylich:
@wied03 `gsub!` explicitly raises an error in Opal, I think `opal-parser` doesn't invoke this code (I mean, it's running only on the MRI side). However I'm not sure, you should better ask @elia .

Mogztter:
gsub! raises an exception at runtime when using opal-compiler.js.
Currently this code is not invoked but it will when #1292 will be merged.

elia:
Looking forward to when #1292 will be completed :+1: :smile:

PR #1408, opened 2016-03-21
Resolves #1407, Implement Pathname.entries (Node.js)

### LINKED ISSUE #1407: Implement Pathname.entries (Node.js)
http://ruby-doc.org/stdlib-2.1.0/libdoc/pathname/rdoc/Pathname.html#method-i-entries

Mogztter:
Resolves #1407

### DIFF
--- a/stdlib/nodejs/dir.rb
+++ b/stdlib/nodejs/dir.rb
@@ -1,6 +1,8 @@
 class Dir
   @__glob__ = node_require :glob
+  @__fs__ = node_require :fs
   `var __glob__ = #{@__glob__}`
+  `var __fs__ = #{@__fs__}`
 
   class << self
     def [] glob
@@ -10,6 +12,22 @@ class Dir
     def pwd
       `process.cwd()`
     end
+
+    def mkdir(path)
+      `__fs__.mkdirSync(#{path})`
+    end
+
+    def entries(dirname)
+      %x{
+        var result = [];
+        var entries = __fs__.readdirSync(#{dirname});
+        for (var i = 0; i < entries.length; i++) {
+          result.push(entries[i]);
+        }
+        return result;
+      }
+    end
+
     alias getwd pwd
   end
 end
--- a/stdlib/pathname.rb
+++ b/stdlib/pathname.rb
@@ -209,6 +209,10 @@ class Pathname
       Pathname.new(File.join(*relpath_names))
     end
   end
+
+  def entries
+    Dir.entries(@path).map {|f| self.class.new(f) }
+  end
 end
 
 module Kernel
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -255,6 +255,7 @@ task :test_nodejs do
     %w[
       opal-parser.rb
       test_file.rb
+      test_dir.rb
     ],
     includes: %w[test/nodejs],
     js_filename: js_filename,
--- a/test/nodejs/test_dir.rb
+++ b/test/nodejs/test_dir.rb
@@ -1,0 +1,17 @@
+# Copied from cruby and modified to skip unsupported syntaxes
+require 'test/unit'
+require 'nodejs'
+require 'nodejs/dir'
+
+class TestNodejsDir < Test::Unit::TestCase
+  def test_dir_entries
+    path = "/tmp/testing_nodejs_dir_implementation_#{Time.now.to_i}"
+    Dir.mkdir(path)
+    result = Dir.entries(path)
+    assert_equal(result.length, 0)
+    File.open(path + '/bar', "w") {}
+    File.open(path + '/baz', "w") {}
+    result = Dir.entries(path)
+    assert_equal(result.length, 2)
+  end
+end
--- a/test/nodejs/test_file.rb
+++ b/test/nodejs/test_file.rb
@@ -5,7 +5,7 @@ require 'nodejs/file'
 
 class TestNodejsFile < Test::Unit::TestCase
   def test_write_read
-    path = '/tmp/testing_nodejs_file_implementation'
+    path = "/tmp/testing_nodejs_file_implementation_#{Time.now.to_i}"
     contents = 'foobar'
     assert !File.exist?(path)
     File.write path, contents


stdlib/nodejs/pathname.rb
```diff
@@ -0,0 +1,30 @@
...
+  # Instance Methods
+
+  def initialize(path)
+    #raise ArgumentError if path == "\0"
+    if Pathname === path
+      @path = path.path
+    else
+      @path = path
+    end
+  end
```
elia: shouldn't this go to stdlib/pathname? or it's specific to nodejs?

Mogztter: I think you're right this can go to `stdlib/pathname`

Mogztter: @elia but `stdlib/nodejs/pathname` still need a constructor no ? I can safely remove `initialize` from `stdlib/nodejs/pathname` and Opal will use `initialize` method from `stdlib/pathname` instead ?

elia: @Mogztter exactly, probably also fixing the implementation to accept `Pathname` instances too, which I don't think currently works:

https://github.com/opal/opal/blob/0cc85df99943f720124f01a030e23db8f8449ac9/stdlib/pathname.rb#L2-L5

Mogztter:
@elia I've also modified Pathname constructor in https://github.com/opal/opal/pull/1409/, I will do a dedicated pull request for this change.

wied03:
@Mogztter - want to write some rubyspecs for this? :)

Mogztter:
Sure, I still need to figure out the best way to create temporary files and directories in my test without FileUtils. Any advice ?

wied03:
They only need to be read, correct? It would be good to do these in the Rubyspecs. I would say we create a `fixtures/entries` directory underneath `spec/ruby/library/pathname` and seed that with as many directories as needed to cover the permutations we'd want to test. @eregon - You agree?

eregon:
Pathname was origianlyl implemented in Ruby in MRI since it was a simple stdlib.
The C extension in MRI only optimizes Pathname#<=>, so why not reusing the existing Ruby code instead? https://github.com/ruby/ruby/blob/4bf3cb5ba9c0242bd5a6d0d55b7db9f837c09edf/ext/pathname/lib/pathname.rb

As you can see, Pathname#entries is really just Dir#entries, and there are already quite a few specs for it: https://github.com/ruby/spec/blob/master/core/dir/entries_spec.rb
So I would recommend implementing Pathname#entries in terms of Dir#entries, and then add a small spec for Pathname#entries that just checks it is what Dir#entries returns as Pathname instances.

wied03:
@eregon - didn't think about that. doh! @Mogztter - Now I know what you mean. It looks like FileUtils (at least this version, https://github.com/ruby/ruby/blob/trunk/lib/fileutils.rb, don't know if the Rake gem has one too) is mainly implemented in terms of `File` or `Dir` methods. I suppose if those were implemented, then FileUtils could run and then the `Dir::entries` specs could run.

wied03:
It's cool if the time isn't there to do that now. That said, if we can't have tests, my preference is to keep `entries` out of pathname and keep this PR around until it's testable. I think no implementation is better than one that we can't keep working in a sustainable way (with tests). It's not really my say though. :)

elia:
@Mogztter confirm that the implementation for `.entries` should go to `Dir` and have the standard Pathname rely on it. Up to you if you want to port the original MRi pathname over (that would be great btw). :+1:

Mogztter:
@eregon Thanks for pointing that, I didn't know either.
@elia OK I will open a new pull request to port the original Pathname implementation.

wied03:
@elia @Mogztter - any precedent for running the FileUtils test_unit tests (e.g. https://github.com/ruby/ruby/blob/trunk/test/fileutils/test_fileutils.rb) on Opal? It would make it a lot easier to implement FileUtils and then rely on the FileUtils calls in the rubyspecs. I see a test/cruby directory but not sure if anything uses that.

wied03:
@elia @Mogztter To add to my last comment, there are no rubyspecs for FileUtils but a LOT of test unit tests. It would be a shame to not do anything until all of those FileUtils tests are ported over to mspec.

Mogztter:
Ok so now `Pathname.entries` delegates to `Dir.entries`.

@wied03 Not sure to understand the implication, I'm a bit lost between `mspec`, `rspec`, `cruby` and Opal tests...
As far as I understand `mspec` is a "minimal" spec with the simplest Ruby language features (ie. some sort of smoke tests for implementation) and `rspec` is the complete Ruby spec suite but some spec are disabled in Opal via `opal/spec/ruby_specs` file. 
> there are no rubyspecs for FileUtils but a LOT of test unit tests

But [test_fileutils.rb](https://github.com/ruby/ruby/blob/trunk/test/fileutils/test_fileutils.rb) is a rubyspecs of FileUtils no ? What do you mean by "a LOT test unit tests" ?

wied03:
You sort of have it right.

RSpec is being used to test MRI side stuff in Opal (has nothing to do with `/spec/ruby_specs`). `/spec/ruby_specs` is a file level filter for the rubyspecs/mspec (see below)

MSpec is the testing framework used to run all of the rubyspecs (as in github.com/ruby/spec).

The MRI team had originally written a bunch of test_unit stuff before the rubyspecs came to be that's a git submodule under test/cruby.

test_fileutils.rb is one of those. The main point I'm making here is we can probably advance Opal further by getting the fileutils cruby tests (of which test_fileutils.rb is one) to pass on Opal instead of porting all of them over to rubyspecs first. Today I submitted a PR to start running a subset (see testing.rake) of those test_unit tests again with Travis.

Mogztter:
Thanks @wied03 for the detailed explanation :smile:

Mogztter:
@elia I've added a test for `Dir.entries` and I've implemented `Dir.mkdir` to be able to test `Dir.entries` :confused:

stdlib/nodejs/pathname.rb
Mogztter: Done

Mogztter:
Awesome thanks @elia

PR #1409, opened 2016-03-21
Implement missing Pathname methods

Mogztter:
- split
 - dirname
 - basename
 - directory?
 - extname
 - comparaison

### DIFF
--- a/stdlib/pathname.rb
+++ b/stdlib/pathname.rb
@@ -94,7 +94,7 @@ class Pathname
   end
 
   def <=>(other)
-    `self.path > other.path ? 1 : (self.path < other.path ? -1 : 0)`
+    self.path <=> other.path
   end
 
   alias eql? ==
--- a/stdlib/pathname.rb
+++ b/stdlib/pathname.rb
@@ -1,7 +1,14 @@
+require 'corelib/comparable'
+
 class Pathname
+  include Comparable
+
   def initialize(path)
-    raise ArgumentError if path == "\0"
-    @path = path
+    if Pathname === path
+      @path = path.path
+    else
+      @path = path
+    end
   end
 
   attr_reader :path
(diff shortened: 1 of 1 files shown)


stdlib/pathname.rb
```diff
@@ -66,6 +73,33 @@ def join(*args)
...
+  def directory?
+    File.directory?(@path)
+  end
+
+  def extname
+    File.extname(@path)
+  end
+
+  def <=>(other)
+    `self.path > other.path ? 1 : (self.path < other.path ? -1 : 0)`
```
IlyaBylich: I'm pretty sure it can be replaced with `self.path <=> other.path`

Mogztter: @iliabylich Thanks, I will try :+1:

stdlib/pathname.rb
```diff
@@ -1,7 +1,14 @@
+require 'corelib/comparable'
+
 class Pathname
+  include Comparable
+
   def initialize(path)
-    raise ArgumentError if path == "\0"
```
elia: you should probably retain this line

stdlib/pathname.rb
```diff
@@ -37,7 +44,7 @@ def cleanpath
   end
 
   def to_path
-    @path
+    @path.to_s
```
elia: what about doing this in the initialize, I think many methods below assume it's a string anyway

stdlib/pathname.rb
Mogztter: could you please explain the purpose of this line ? I don't remember exactly but this cause me troubles :neutral_face:

stdlib/pathname.rb
Mogztter: good idea, I'm just a little bit concerned with `<=>` method but I think it's better to do it in the initialize for consistency.

elia:
:+1:

PR #1410, opened 2016-03-23
Making generated code "use strict" compliant

elia:
Actually I thought I had "use strict" enabled already, but, admittedly, was too easy…

_Thanks to @wied03 for pointing out that the commit was bogus (a63161feeb63b67483f3ffb471964b98ffe02d02)!_

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -12,6 +12,7 @@
   //
   //   The way the code is digested before going through Yardoc is a secret kept
   //   in the docs repo (https://github.com/opal/docs/tree/master).
+  "use strict"
 
   if (typeof(this.Opal) !== 'undefined') {
     console.warn('Opal already loaded. Loading twice can cause troubles, please fix your setup.');
--- a/lib/opal/nodes/helpers.rb
+++ b/lib/opal/nodes/helpers.rb
@@ -44,7 +44,7 @@ module Opal
       end
 
       def valid_ivar_name?(name)
-        not (PROTO_SPECIAL_PROPS =~ name or PROTO_SPECIAL_METHODS =~ name)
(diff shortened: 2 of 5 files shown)


jirutka:
What's the state of this? Have you hit some hard to overcome problem with making Opal `"use strict"` compliant?

hmdne:
This patchset became much lesser after a rebase :D

opal/corelib/runtime.js
```diff
@@ -788,6 +788,14 @@
   // @param object [Object]
   // @return [Class]
   Opal.build_object_singleton_class = function(object) {
+    if (object.$$is_number) {
+      throw Opal.TypeError.$new("can't define singleton");
+    }
+
+    if (object.$$is_string) {
```
hmdne, line 795: This check needs to be a little bit more extensive.

lib/opal/nodes/literal.rb
```diff
@@ -156,7 +156,15 @@ def compile_static_regexp
           message = "named captures are not supported in javascript: #{value.inspect}"
           push "self.$raise(new SyntaxError('#{message}'))"
         else
-          push "#{Regexp.new(value).inspect}#{flags.join}"
+          regexp = Regexp.new(value)
+          flags = self.flags
+
+          if flags && flags !~ /^[gim]*$/
+            compiler.warning "Ignoring unsupported flags #{flags.inspect} found for regexp #{regexp.inspect}"
```
hmdne, line 163: Nowadays JS supports some more flags. For example: u. (All supported: dgimsuy).

In Ruby `u` is set by default, denotes the UTF-8 regexp encoding, in JS `u` enables the \p{..} and treats the string as UTF-16 and not UCS-2 (ie. support for characters 0x010000 - 0x10ffff, eg. emoji).

From what I know, this is used in Asciidoctor for instance.

Issue #105, opened 2012-06-21
File name and line numbers in javascript code


elia:
Adding file name and line numbers of the original file to the generated js would help debugging very much.

Also can open the possibility to [sourcemaps](http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/) support (I'm already looking into [ruby-source_map](https://github.com/ConradIrwin/ruby-source_map))

PR #1411, opened 2016-03-23
Experiment with adding programmatic line/file context info

wied03:
This is kind of a rehash of https://github.com/opal/opal/issues/105 but I think it still might be relevant for other libraries like opal-rspec (not for humans).

Source maps are great for exceptions and debugging. Exceptions slow down the code a lot though when you want code location info **outside of an exception**.

This would be an opt in compiler flag that adds this context to each block (included in this PR) and each method body (either via compiler trickery or in this rough sketch via the `defn` function in runtime, but I didn't go very far).

This could be used to implement `caller` in a more performant way or it could just be simple thing where code can access `functionObject.$$sourcemap` of a block/method to get its location.

Thoughts?

FYI, This would make https://github.com/opal/opal-rspec/issues/57 possible

### DIFF
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -118,6 +118,12 @@ module Opal
           push ')'
         elsif scope.top?
           unshift "Opal.defn(Opal.Object, '$#{mid}', "
+          # TODO: Compiler config, default of false
+          include_source_in_code = true
+          if include_source_in_code
+            file_path = File.expand_path("#{compiler.file}.rb")
+            push ", {line: #{@sexp.line}, column: #{@sexp.column}, file: '#{file_path}'}"
+          end
           push ')'
         else
           raise "Unsupported use of `def`; please file a bug at https://github.com/opal/opal reporting this message."
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -122,7 +122,7 @@ module Opal
           include_source_in_code = true
           if include_source_in_code
             file_path = File.expand_path("#{compiler.file}.rb")
-            push ", {line: #{@sexp.line}, column: #{@sexp.column}, file: '#{file_path}'}"
+            push ", {line: #{@sexp.line}, column: #{@sexp.column}, file: '#{file_path}', method: '#{function_name.strip}'}"
           end
           push ')'
         else
--- a/lib/opal/nodes/iter.rb
+++ b/lib/opal/nodes/iter.rb
@@ -44,6 +44,12 @@ module Opal
           unshift "(#{identity} = function(){"
         end
         push "}, #{identity}.$$s = self,"
+        # TODO: Compiler config, default of false
+        include_source_in_code = true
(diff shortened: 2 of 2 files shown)


wied03:
As an example, Ruby:

```ruby
stuff = lambda { puts 'howdy' }
stuff.call

def howdy
    puts 123
end
```
compiled:
```js
(function(Opal) {
  var $a, $b, TMP_1, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, stuff = nil;

  Opal.add_stubs(['$lambda', '$puts', '$call']);
  stuff = ($a = ($b = self).$lambda, $a.$$p = (TMP_1 = function(){var self = TMP_1.$$s || this;

  return self.$puts("howdy")}, TMP_1.$$s = self, TMP_1.$$sourcemap = {line: 1, column: 17, file: '/host/stuff.rb'}, TMP_1), $a).call($b);
  stuff.$call();
  return (Opal.defn(Opal.Object, '$howdy', function ːhowdy() {
    var self = this;

    return self.$puts(123);
  }, {line: 4, column: 0, file: '/host/stuff.rb'}), nil) && 'howdy';
})(Opal);

(function(Opal) {

  Opal.add_stubs(['$exit']);
})(Opal);
```

elia:
(@wied03 fun fact, the first implementation of sourcemaps (by yours, truly) was based on `/* some/file.rb:123 */` kind of comments then parsed and matched into a JS map file. :smile:)

wied03:
@elia - was there any reason for going away from that other than changing to do things the browser/standard way?

elia:
@wied03 well, the current fragment system is much better and can relate single ruby tokens to their js counterpart, that's how I can hover on a variable in the ruby source-mapped code in safari and see its javascript contents :)

Ideally I'd say that the best would be to take the js stacktrace and apply the sourcemap to it in order to get the original locations, but I don't think there's any api for that yet…

wied03:
@elia - Understood, except the idea of this PR is pointless if stack traces are used. The only reason to do this is to get fast `caller` info in non error situations. creating a stack trace to get that is slow.

wied03:
Only case I can think of this is for rspec which tells you where tests are (success or fail). I'm planning on removing the caller hack that's in there now because it slows down things a lot with 1 exception throw per example and failures are covered by the JS source map stuff in Karma, etc. Without this PR though, you only know about where a test is unless it fails (which isn't terrible, just not ideal).

wied03:
@elia  - closing this, could do this on a more targeted basis using RSpec metadata, see the 57 issue.

elia:
@wied03 which 57 issue? doesn't look like #57

wied03:
Sorry opal-rspec's 57 issue.

elia:
→ opal/opal-rspec#57

wied03:
Yep, that one. It would be more consistent with how you advised the `require_dependency` stuff be done. It's also less invasive and more targeted to what it's needed for.

PR #1415, opened 2016-03-31
Allow method definition in method body.

iliabylich:
Added ruby/spec/language/def_spec to the test suite.


iliabylich:
Not sure where should I put `public/private` methods (atm both are in the `opal/main.rb`). Suggestions? I feel like they should be in the `unsupported.rb` :smile:

elia:
@iliabylich makes sense

elia:
:tada: :+1:

PR #1416, opened 2016-03-31
Fixed setting mlhs arg to nil when JS-falsey value is passed.

iliabylich:
My regression, caused by https://github.com/opal/opal/pull/1395.
Can be easily reproduced with:
``` ruby
[[false], [""], [0]].each { |(a)| p a }
```


elia:
:beers:

iliabylich:
@elia Time for 0.10.0.beta2? :smile:

elia:
Let's do it! :smile:

PR #1417, opened 2016-04-01
Optimized rescue body by extracting it to a separate inline function on the compilation stage.

iliabylich:
Resolves https://github.com/opal/opal/pull/1393
``` ruby
# opal -c test.rb | node --trace_opt --trace_deopt --allow-natives-syntax

class A
  def m
    1 + 1
  rescue
  end
end

%x{
a = Opal.A.$new();

function printStatus(fn) {
    switch(%GetOptimizationStatus(fn)) {
        case 1: console.log("Function is optimized"); break;
        case 2: console.log("Function is not optimized"); break;
        case 3: console.log("Function is always optimized"); break;
        case 4: console.log("Function is never optimized"); break;
        case 6: console.log("Function is maybe deoptimized"); break;
        case 7: console.log("Function is optimized by TurboFan"); break;
        default: console.log("Unknown optimization status"); break;
    }
}

a.$m();
a.$m();

%OptimizeFunctionOnNextCall(a.$m.$$rescue_body);

a.$m();
a.$m();

printStatus(a.$m.$$rescue_body);
}

# output:
# Function is optimized
```
Ruby method arguments MUST be passed as arguments one by one to `$$rescue_body` function, otherwise JS engine optimizes it, so closure becomes static (or cached), and by calling method with different arguments JS invokes `$$rescue_body` always with the same closure

cc @elia @meh @jgaskins This implementation is slower than inline code invocation, I think I'm missing something,  the code below shows that I was wrong in https://github.com/opal/opal/pull/1393#issuecomment-204459976, but I don't understand where's the problem in this PR. Any ideas?
``` ruby
%x{global.$$rescue_body = function() { return 1 + 1 }}

def m1
  %x{global.$$rescue_body()}
rescue
end

def m2
  %x{return 1 + 1}
rescue
end

def bm(title, &block)
  start = Time.now
  10_000_000.times(&block)
  puts "#{title}: #{Time.now - start}"
end

bm 'inline' do
  m1
end

bm 'inline function' do
  m2
end
```
Output:
```
inline: 0.837
inline function: 0.796
```

### DIFF
--- a/lib/opal/nodes/def.rb
+++ b/lib/opal/nodes/def.rb
@@ -9,11 +9,32 @@ module Opal
       children :recvr, :mid, :args, :stmts
 
       attr_accessor :block_arg
+      attr_accessor :arg_names
+      attr_accessor :inline_rescue_body
+
+      def initialize(*)
+        super
+        extract_block_arg
+        @arg_names = extract_arg_names
+      end
 
       def extract_block_arg
         if args.last.is_a?(Sexp) && args.last.type == :blockarg
           @block_arg = args.pop
         end
+      end
+
+      def extract_arg_names(args_group = args.children)
+        result = []
+        args_group.map do |arg|
+          case arg.type
+          when :mlhs
+            result.concat extract_arg_names(arg.children)
+          else
+            result << arg[1].to_s if arg[1]
+          end
+        end
+        result
       end
 
       def argc
@@ -28,7 +49,6 @@ module Opal
       end
 
       def compile
-        extract_block_arg
         split_args
 
         inline_params = nil
@@ -44,6 +64,8 @@ module Opal
         end
 
         in_scope do
+          scope.identify!
+
           scope.mid = mid
           scope.defs = true if recvr
 
@@ -99,6 +121,13 @@ module Opal
         unshift "function#{function_name}("
         unshift "#{scope_name} = " if scope_name
         line "}"
+
+        if inline_rescue_body
+          push ","
+          inline_rescue_body.each do |code|
+            push code
+          end
+        end
 
         if recvr
           unshift 'Opal.defs(', recv(recvr), ", '$#{mid}', "
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -102,6 +102,19 @@ module Opal
       children :body
 
       def compile
+        processed_body = process(body_code, @level)
+        if inline_body_function?
+          # Current rescue node is the only node in the def,
+          # and it's a simple begin/rescue(s)/end statement,
+          # here we can optimize it by extracting the code
+          # between begin and rescue to the separate inline function
+          scope.inline_rescue_body = [
+            "#{scope_name}.$$rescue_body = #{scope_name}.$$rescue_body || function(#{inline_body_arguments}) {",
+            processed_body,
+            "}"
+          ]
+        end
+
         scope.rescue_else_sexp = children[1..-1].detect { |sexp| sexp.type != :resbody }
         has_rescue_handlers = false
 
@@ -109,9 +122,13 @@ module Opal
           line "var $no_errors = true;"
         end
 
-        push "try {"
+        line "try {"
         indent do
-          line process(body_code, @level)
+          if inline_body_function?
+            line "return #{scope_name}.$$rescue_body(#{inline_body_arguments});"
+          else
+            line processed_body
+          end
         end
         line "} catch ($err) {"
 
@@ -170,6 +187,21 @@ module Opal
       #
       def handle_rescue_else_manually?
         !scope.in_ensure? && scope.has_rescue_else?
+      end
+
+      def scope_name
+        scope.identity
+      end
+
+      def inline_body_function?
+        scope.def? && scope.stmts.children == [@sexp]
+      end
+
+      def inline_body_arguments
+        result = ['self']
+        result.concat(scope.arg_names)
+        result << scope.block_name if scope.uses_block?
+        result.uniq.join(',')
       end
     end
 
--- a/spec/lib/compiler/call_spec.rb
+++ b/spec/lib/compiler/call_spec.rb
@@ -114,9 +114,9 @@ if (a == null) a = nil;
 
                 it do
                   is_expected.to include <<-CODE
-    return ($b = ($c = self).$call_method, $b.$$p = (TMP_1 = function(a){var self = TMP_1.$$s || this;
+    return ($b = ($c = self).$call_method, $b.$$p = (TMP_2 = function(a){var self = TMP_2.$$s || this;
 if (a == null) a = nil;
-    return self.$foobar()}, TMP_1.$$s = self, TMP_1), $b).apply($c, Opal.to_a(stuff));
+    return self.$foobar()}, TMP_2.$$s = self, TMP_2), $b).apply($c, Opal.to_a(stuff));
                   CODE
                 end
               end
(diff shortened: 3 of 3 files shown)


iliabylich:
Alright, by calling `TMP_.$$rescue_body = function ...` **outside** of the method definition makes it faster, and I'm quite sure it can be implemented.

jgaskins:
I wonder if it would be different if we used a much more complex function.

iliabylich:
@jgaskins I've got same bm results for `Number#<=>`

iliabylich:
Extracted `$$rescue_body = ...` outside of the method definition, now it's faster than master :smile:
``` ruby
(0..1_000_000).to_a.reverse.sort
``` 
master: 5,87s
patch: 5,65s
Not sure whether this PR should be merged, it improves a little bit performance in some small cases, but generated code looks more complex for me.
`opal -c --no-opal -e "def m; 1 + 1; rescue; end"`:
``` js
(function(Opal) {
  function $rb_plus(lhs, rhs) {
    return (typeof(lhs) === 'number' && typeof(rhs) === 'number') ? lhs + rhs : lhs['$+'](rhs);
  }
  var TMP_1, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$+']);
  return (Opal.defn(Opal.Object, '$m', TMP_1 = function ːm() {
    var self = this;

    try {
      return TMP_1.$$rescue_body(self);
    } catch ($err) {
      if (Opal.rescue($err, [$scope.get('StandardError')])) {
        try {
          return nil
        } finally { Opal.pop_exception() }
      } else { throw $err; }
    };
  },TMP_1.$$rescue_body = TMP_1.$$rescue_body || function(self) {return $rb_plus(1, 1)}), nil) && 'm'
})(Opal);
```

iliabylich:
oops, there's an issue with this implementation, Opal compiles `locals` and `temps` in the ruby-method function, some of them can be used in the rescue body. @elia There's no way to fix it without modifying 1000 locs, right? :disappointed: 

``` js
Opal.defn(self, '$==', TMP_2 = function(other) {
      var $a, self = this, cmp = nil;

      
      try {
        return TMP_2.$$rescue_body(self,other);
      } catch ($err) {
        if (Opal.rescue($err, [$scope.get('StandardError')])) {
          try {
            return false
          } finally { Opal.pop_exception() }
        } else { throw $err; }
      };
    },TMP_2.$$rescue_body = TMP_2.$$rescue_body || function(self,other) {if ((($a = self['$equal?'](other)) !== nil && (!$a.$$is_boolean || $a == true))) {
        return true};
      
      if (self["$<=>"] == Opal.Kernel["$<=>"]) {
        return false;
      }

      // check for infinite recursion
      if (self.$$comparable) {
        delete self.$$comparable;
        return false;
      }
    
      if ((($a = cmp = (self['$<=>'](other))) !== nil && (!$a.$$is_boolean || $a == true))) {
        } else {
        return false
      };
      return $scope.get('Comparable').$normalize(cmp) == 0;});
```

jgaskins:
If you define the rescue wrapper internally and memoize it, you might be able to use it as a closure.

```ruby
def foo(bar)
  baz = bar + ' lol'
rescue
  'lol'
end
```

```javascript
Opal.defn(self, '$foo', function(bar) {
  var self = this;
  var baz;
  self.$foo.$$regular_method_body = self.$foo.$$rescue_body || function() {
    return baz = bar['$+'](' lol');
  };
  self.$foo.$$rescue_body = self.$foo.$$rescue_body || function() {
    return 'lol';
  };

  try {
    self.$foo.$$regular_method_body();
  } catch(e) {
    self.$foo.$$rescue_body();
  }
});
```

I have no idea how hard it would be to turn that Ruby code into that JS code (or something resembling it), but it may be useful.

Caveats:
- It will likely cause increased memory usage because all method locals will be kept around for the life of the app. This may not be problematic, but is something to consider.
- It also may not work at all on the second invocation because the closure variables may not reference the new variables after the first time the method is called. Lexical scoping is confusing sometimes.

jgaskins:
The closure is the only way I could think of to get locals into the wrapped functions without using `with` — which would cause the contained functions to be unoptimizable and negate everything we're trying to do here. :-)

elia:
_/me catching up_

elia:
proposal:

```rb
def foo(bar, baz = 123)
  baz = baz.to_s + bar + ' lol'
rescue
  'lol' + bar
end
```

```js
return (Opal.defn(Opal.Object, '$foo', function ːfoo() {
  var $$rescue;
  TMP_1.$$body = TMP_1.$$body || function(bar, baz) {
    var self = this, $$return;

    $$rescue = function() {
      return $rb_plus("lol", bar)
    };

    if (baz == null) {
      baz = 123;
    }
    return baz = $rb_plus($rb_plus(baz.$to_s(), bar), " lol")
  };
  try {
    return TMP_1.$$body.call(this, arguments);
  } catch ($err) {
    if (Opal.rescue($err, [$scope.get('StandardError')])) {
      try {
        return $$rescue.call(this, arguments);
      } finally { Opal.pop_exception() }
    } else { throw $err; }
  }
});
```

but I'm not sure it works with that `$$rescue` in the closure…

iliabylich:
@elia You may want to use method arguments in the `rescue` statement:
``` ruby
def m(a)
  a.method1
rescue
  a.method2
end
```
(Don't even know how to detect this case on the compilation stage)

iliabylich:
@jgaskins benchmarks for something similar to

``` ruby
def m(a,b=1)
  a + b
rescue => e
  puts e
end
```
Inline code: 0,17s
``` js
var TMP_1 = function(a, b) {
  if (b == null) {
    b = 1;
  }

  try {
    return a + b;
  } catch(e) {
    console.log(e);
  }
}

for (var i = 0; i < 10 * 1000 * 1000; i++) {
  TMP_1(i);
}
```
Inline function: 1,13s
``` js
var TMP_1 = function(a, b) {
  if (b == null) {
    b = 1;
  }

  TMP_1.$$rescue_body = TMP_1.$$rescue_body || function(a, b) {
    return a + b;
  }

  try {
    return TMP_1.$$rescue_body(a, b);
  } catch(e) {
    console.log(e);
  }
}

for (var i = 0; i < 10 * 1000 * 1000; i++) {
  TMP_1(i);
}
```
Function defined once and outside of the method body: 0,21s
``` js
var TMP_1 = function(a, b) {
  if (b == null) {
    b = 1;
  }

  try {
    return TMP_1.$$rescue_body(a, b);
  } catch(e) {
    console.log(e);
  }
}

TMP_1.$$rescue_body = function(a, b) {
  return a + b;
}

for (var i = 0; i < 10 * 1000 * 1000; i++) {
  TMP_1(i);
}
```
Does this optimization work only for a big `try/catch` statement or am I missing something? :smile:

jgaskins:
You're right, it looks like the closure's performance is just as bad, checked with Safari and Chrome. I wonder if the VMs are compiling the inner function on every execution of the outer function.

This is starting to look like it can't be optimized in a generic way in the compiler. :-)

elia:
@iliabylich I think we can close this unless you think we can save anything or you wanna do some more experiments on the matter.

@jgaskins let's merge the original PR and maybe do a general check for similar situations in the core/std lib…

jgaskins:
Cool. I bet all the `<=>` methods will have it and I'm sure Ruby has plenty other methods that return `nil` instead of raising exceptions. I don't know how many of them will be necessary to optimize, but I imagine anything you could expect to be called on elements in an array. This whole thing came up because I was sorting models to render them in a `ul` and realized I was spending like 15-18% of my profiling time inside `<=>`.

I'm glad we had this discussion, though. If someone can figure out how to handle this automatically in the future, I'm happy to have it again.

elia:
@jgaskins Just keep a record of what we've learned a section regarding optimizations can be added to HACKING.md or as a separate guide (hint hint)

PR #1420, opened 2016-04-02
Add require_dependency alias for require

wied03:
When sharing code between Opal and Rails, server side, Rails autoload does not like `require`, but will work OK if you 'require_dependency'. Added alias that treats require_dependency just like require


elia:
@wied03 makes total sense, please retarget this to opal-rails, I'm ok with it outside of Opal even if it uses internal APIs, as long as there are specs.

wied03:
Here is the separate project for this - https://github.com/wied03/opal-require-dep

PR #1422, opened 2016-04-06
Fix Regexp.new edge case with trailing escaped backslashes

wied03:
Ruby specs for this

https://github.com/ruby/spec/pull/227

With that new rubyspec, the (as of yet) unimplemented compile method fails with these reused specs, go ahead and exclude that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ Whitespace conventions:
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


wied03:
If https://github.com/opal/opal/pull/1423 is merged first, remember to remove the 1 commented regexp filter after rebasing

elia:
thanks! :+1:

PR #1423, opened 2016-04-07
Fix more super issues related to basic objects, singletons, anon modules

wied03:
Also fixed issue with filtered alias spec (special treatment isn't needed for singleton cases)

https://github.com/ruby/spec/pull/229 - (part of this PR)
Also covered in
https://github.com/ruby/spec/pull/231 - (part of this PR)
https://github.com/ruby/spec/pull/232 - (part of this PR)


opal/corelib/runtime.js
```diff
@@ -1068,24 +1063,50 @@
...
+        // case is also hit with multiple module includes
+        break;
+      }
+      else if (klass.$$iclass && klass.$$module === owner) {
+        // module has donated to other classes but klass isn't one of those
+        break;
+      }
+      else if (owner.$$is_singleton && klass === owner.$$singleton_of.$class()) {
+        // cases like stdlib `Singleton::included` that use a singleton of a singleton
+        break;
```
wied03: @elia or @meh - Open to suggestions if there is a better way for this singleton case. Only a few weird cases hit this. Another one besides `Singleton` in stdlib was the alias case here: https://github.com/ruby/spec/commit/bbe5d48e6eff38c980010b60b7c42d1acd6316fd

I did at least try and make the boolean flag first to avoid the `$class()` function call unless we need to.

elia: I looked into opal's code and actually I can't find a place where `#class` is defined to anything different than `self.$$class`, so I'd say it can be changed to `owner.$$singleton_class.$$class`, what am I missing?

wied03:
@elia - I might be a little out of my knowledge realm, but in that case, I needed the class the singleton class was created from, not the object's superclass. I was able to cut off the function call though and just use `$$class` rather than than `$class()` in this latest squash.

PR #1425, opened 2016-04-08
Fixed Method#arity for negative returning values. Added Method/Proc#parameters

iliabylich:
Returned back `OPAL_CONFIG`, we need an information about `arity_check` in runtime to call `Kernel#eval` with correct options. Before this patch all specs with `evaluate` were compiling the code without arity checking :smile: 

cc @elia @wied03 

Upd: Added Method#parameters and Proc#parameters. These methods work only when arity check is enabled. Otherwise returns a blank array


iliabylich:
Don't merge it yet please. This code definitely should be in `NodeWithArgs`, so `Proc#arity` can be fixed too

wied03:
Fast! How is this fixing `Method#arity`? I see the test filters being removed there but I don't see any changes to method.rb. I just took a quick glance though.

iliabylich:
Existing implementation of `Method#arity` uses `self.$$arity`. However, there were no code for populating this property :smile:

wied03:
```ruby
class Method
  attr_reader :owner, :receiver, :name

  def initialize(receiver, method, name)
    @receiver = receiver
    @owner    = receiver.class
    @name     = name
    @method   = method
  end
end
```

Was/is `@method` an instance of Proc?

iliabylich:
`@method` is a method body from `Opal.def(owner, method_name, method_body)`. It's a function (and a proc of course)

wied03:
Ahh, I see. I noticed Proc native bridges to function but didn't make the connection.

iliabylich:
Finally green. Ready for review.

elia:
@iliabylich ok, discard the OPAL_CONFIG comment in the other thread, I have another better (?) proposal. Let's make OPAL_CONFIG a special var inside ruby code and dump its value inline whenever it's used in the code.

The behavior is similar to `__FILE__` and `__LINE__` but I'd keep the implementation in the compiler. Optionally it could be `__COMPILER_ARITY_CHECK_ENABLED__`, one for each option which would keep the compiled code even shorter and will save us from inlining a hash each time.

That's also the reason I had for removing OPAL_CONFIG in the first place, it was compiled for every top node and virtually never used taking up both visual and byte space.

how does it sounds?

iliabylich:
@elia Sounds good

elia:
@iliabylich :+1: will push a PR that adds `__OPAL_COMPILER_CONFIG__` or something like that today (10:40am here now). The magic constant will contain an hash with the options passed to `Opal::Compiler` to compile the file in which is used.

iliabylich:
@elia Done!

iliabylich:
@elia Thanks!

elia:
@iliabylich kudos to you, really great stuff!

Issue #1075, opened 2015-08-25
stdlib/source_map/vlq is broken


zetachang:
```ruby
require "source_map/vlq"
puts SourceMap::VLQ.encode([0])
```
It should output "A", but instead we got empty string. 

After investigation, the problem is due to `begin; end while` block around https://github.com/opal/opal/blob/master/stdlib/source_map/vlq.rb#L31-L36.

According to https://github.com/opal/opal/issues/575, Opal implemented differently compared to MRI.

A solution may be using a `loop do; break; end` to replace it.

Some context: I am working on https://github.com/zetachang/opalrb-loader, thus source map support will be utilized.

elia:
Apparently I fixed this upstream but forgot to apply in our copy 

https://github.com/maccman/sourcemap/issues/10

elia:
fixed by https://github.com/opal/opal/pull/1426

PR #1426, opened 2016-04-08
Fix compiler error handling for bootstrap case / run compiler specs on Opal

wied03:
The exception started slipping through the rescue in compiler.rb

Ultimately that needs to be fixed (along with validating whether SyntaxError or RuntimeError should be what's raised). For now, make the compiler specs runnable in Opal and filter what does not currently
pass (regex, `__FILE__` issues)

Add matchers to fill in for RSpec, stub out RSpec helper, fix not_to bridge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ Whitespace conventions:
 
 ### Fixed
 
+- `SourceMap::VLQ` patch (#1075)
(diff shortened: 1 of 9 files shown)


wied03:
I should add that before this, when compiling in a bootstrapped case, you didn't even see the "in file" error message from compiler.rb. At least now that's back

wied03:
Also fixes https://github.com/opal/opal/issues/1075

PR #1429, opened 2016-04-12
File::dirname and File::basename - RubySpec compliant

wied03:
Converted the MRI C implementation to JS and pass the Rubyspecs.

### DIFF
--- a/opal/corelib/file.rb
+++ b/opal/corelib/file.rb
@@ -22,12 +22,204 @@ class File < IO
     end
     alias realpath expand_path
 
+    %x{
+      function chompdirsep(path) {
+        var last;
+
+        while (path.length > 0) {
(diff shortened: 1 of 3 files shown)


opal/corelib/file.rb
```diff
@@ -22,20 +22,212 @@ def expand_path(path, basedir = nil)
...
+          }
+        }
+        return path;
+      }
+
+      function inc(a) {
+        return a.substring(1, a.length);
+      }
+
+      function skipprefix(path) {
```
elia: Maybe this can be removed altogether

wied03:
@elia - probably need it. see #1430 . Could just use that PR but I wanted to break this up into chunks.

elia:
Tried out the new squash and merge, looks great :)

Issue #1424, opened 2016-04-07
Arity Regression from 0.9 to 0.10


wied03:
It's great that a lot of args stuff is supported now. That said, since then, arity broke in several cases.

On 0.10 master
Execute: `bundle exec opal foo.rb`
Response:
```
arity
1
```

On 0.9 with same command line:
```
arity
0
```
foo.rb code:
```ruby
dummy = Class.new
class_method = dummy.method(:new)
puts 'arity'
puts class_method.arity
```

No surprise here. Since restargs is being used to implement splat stuff for the `Class` `new(*args, &block)` method, it broke.

Old: `Opal.defn(self, '$new', TMP_2 = function() {`
New: `Opal.defn(self, '$new', TMP_2 = function($a_rest) {`

wied03:
I think the new approach is sensible (passing params explicitly rather than using the function object). That said. I think we need to get arity checking in gear though so it doesn't break like this.

Approach:
1. `Method#arity` becomes a noop (or throws exception) if not compiled with arity checking enabled
2. When arity checking is enabled, pass in the MRI style `parameters` info into `Opal.defn`
3. Modify `Opal.defn` etc. to take in an optional params parameter that the compiler drops in and set that (and arity) on the function object such that `Method` can pick it up.

Thoughts?

elia:
The proposed approach makes sense to me.

```js
  (Opal.defn(Opal.Object, '$foo', TMP_1 = [-2, [["req","a"],["opt","b"],["rest","c"]]], function ːfoo(a, b, $a_rest) {
    …
    var $arity = TMP_1[0]; …
  }), nil) && 'foo'
```
cc @iliabylich

wied03:
@elia @iliabylich - so that it doesn't silently produce bad results, any problem with using throws/raise in those methods when arity is off? I can't think of why someone would call those arity methods unless they are testing (e.g. RSpec).

iliabylich:
@wied03 Well, one more regression caused by https://github.com/opal/opal/pull/1412 :smile: . I was trying to join the functionality of method params and block params in `NodeWithArg` class. Why do we need this parameter `$a_rest`? Because blocks expand their arguments when a single array item is provided (like `proc { |a,b| [a,b] }.call([1,2]) == [1,2]`). And of course it changes `function.length` for the method body. I'd say that we still need these `$*_rest arguments, instead we should modify `Method#arity`.

[...]

PR #1431, opened 2016-04-12
Fix Method#arity and Method#parameters for methods generated using attr_reader/attr_writer.

iliabylich:
Closes https://github.com/opal/opal/issues/1424.

cc @wied03 @elia

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -135,6 +135,9 @@ class Module
         // initialize the instance variable as nil
         proto[ivar] = nil;
 
+        body.$$parameters = [];
+        body.$$arity = 0;
+
         if (self.$$is_singleton) {
           proto.constructor.prototype[id] = body;
         }
@@ -163,6 +166,9 @@ class Module
             return this[ivar] = value;
           }
         })(ivar);
+
+        body.$$parameters = [['req']];
+        body.$$arity = 1;
 
         // initialize the instance variable as nil
         proto[ivar] = nil;
--- a/spec/opal/core/method/arity_spec.rb
+++ b/spec/opal/core/method/arity_spec.rb
@@ -1,0 +1,10 @@
+describe 'Method#arity' do
+  # regression test https://github.com/opal/opal/issues/1424
+  context 'when method was defined through attr_reader/attr_writer' do
+    klass = Class.new do
+      attr_accessor :attribute_name
+    end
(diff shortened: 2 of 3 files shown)


wied03:
@iliabylich - Why not submit these as Rubyspecs instead?

iliabylich:
@wied03 This issue is Opal-specific. In an ideal world we would have something like
``` ruby
def attr_reader(*attrs)
  attrs.each do |attr|
    define_method attr do
      instance_variable_get("@#{attr}")
    end
  end
end
```
(`ruby/spec` has enough tests for methods defined through `define_method`)

All the tests in this PR a regressions tests for Opal, all the issues were caused by internals of Opal. I don't think that other ruby implementations may have the same bug and most probably `ruby/spec` doesn't need these tests.

cc @eregon

iliabylich:
@wied03 These tests can be compared to something like `testing that method created through attr_reader can be binded to another object` or `testing that method created through attr_reader has correct #name`. Yes, we may have these bugs, and I'd say all of them should be placed to `spec/opal`

wied03:
@iliabylich - How do we know for sure that other Rubies might not have the same issue at some point? I think if they demonstrate behavior that all Rubies should have (not JS bridging for example), then it might be a good reason to put them in there.

iliabylich:
@wied03 This particular issue comes from Opal internals. `Module#attr_reader` directly invokes `Opal.def` that doesn't know anything about `$$arity/$$parameters` stuff. You are right, other rubies might have the same issue, but we don't know this for sure :smile: I'm not against putting these specs to `ruby/spec`, let's wait for @eregon 's response

elia:
Merging in the meanwhile 👍🏼

eregon:
@wied03 @iliabylich This would be perfectly fine for ruby/spec, attr_* are usually implemented differently than other methods for many reasons, so ensuring that Method introspection works would be a good test :smiley: 
I don't see anything Opal-specific in the specs per se. Of course most bugs are caused by bugs in implementation internals, but that makes the specs even more interesting to catch bugs or verify behavior on others implementations.
Some complicated regression tests involving 3+ methods might be difficult to translate to a spec, or they would need to be isolated to a smaller failing scenario. But in general, I'm all for having regression tests written as specs in ruby/spec. For instance, jruby+truffle does not have a separate test suite for regressions tests, it only has some additional specs for extra behavior (like integration with other languages).

Issue #1311, opened 2016-01-30
SyntaxError: Unexpected token break


vais:
Blocks Opal from being able to run `rubyspec/core/integer/times_spec.rb`
```
$ cat x.rb
while true
  nil or break
end
```
```
$ bundle exec opal x.rb 
/private/var/folders/pt/kr0lwm4d20jgc_jvpc5fyk1r0000gn/T/opal-nodejs-runner-20160130-5485-1dq57le:18691
  ((($b = nil) !== false && $b !== nil) ? $b : break;)}
                                               ^^^^^

SyntaxError: Unexpected token break
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3
```

PR #1434, opened 2016-04-13
Add ruby/core/integer/times_spec to the test suite.

iliabylich:
Closes https://github.com/opal/opal/issues/1311

### DIFF
--- a/lib/opal/nodes/logic.rb
+++ b/lib/opal/nodes/logic.rb
@@ -111,12 +111,30 @@ module Opal
       end
     end
 
-    class OrNode < Base
+    class BinaryOp < Base
+      def compile
+        if rhs.type == :break
+          compile_if
+        else
+          compile_ternary
+        end
+      end
+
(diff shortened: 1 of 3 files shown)


elia:
This is a big one for code coming from MRI's stdlib as #1311 it's very frequent :+1: :smile:

iliabylich:
@elia I've made a fix only for simple cases like `true and break` or `false or break`. By complex case here I mean something like `1 or (2 + 2; break)`. Probably it can be fixed by wrapping an `rhs` here in the `(function() { ... })()` and using always an `Opal.brk`. Also don't forget about the stuff like `false or redo` and `true and next` :smile:. There's a huge difference between JS' and MRI's syntaxes.

elia:
uhm, I guess `true and next` would be an important one too, I just remember I changed that type of style multiple times in stuff coming from MRI, but don't remember what was exactly…

Issue #1343, opened 2016-02-14
String#to_proc triggers function deoptimizations


jgaskins:
I'm writing a real-time dashboard app in Clearwater that receives thousands of updates over a websocket per minute (porting it from one of our React apps). The good news is that it's handling it really well, considering literally every single order in progress can potentially be shown in this app, but I noticed while profiling it that about 15% of CPU time is spent in `String#to_proc` and that that function is deoptimized due to "Assignment to parameter in arguments object", which is caused by reassigning one of the function arguments. So it could be going even faster.

It looks like it's happening in [`Opal::Nodes::IterNode`](https://github.com/opal/opal/blob/723f47c5c07f918c0ee410c58df2b95cd53b8f71/lib/opal/nodes/iter.rb#L53), where all of the params to the function (including splat params) are injected.

While investigating, I also noticed that it's using the [`Array.prototype.slice.call(arguments)` thing](https://github.com/opal/opal/blob/723f47c5c07f918c0ee410c58df2b95cd53b8f71/lib/opal/nodes/iter.rb#L34), which triggers another deoptimization for the same call.

Methods used to suffer these same deoptimizations, but that was fixed in 121eff369b4de925e8a86f50927dd1635ca33b4b. Block splat args seem to be handled a bit differently. I spent most of last night trying to figure out how to remove those deoptimizations, but I couldn't figure it out. I tried to extract the common functionality to `Opal::Nodes::ScopeNode` so `Opal::Nodes::DefNode` and `Opal::Nodes::IterNode` could both use the same functionality, but I couldn't get it working for both regular and splat args. :-\

iliabylich:
This is the proc that `String#to_proc` returns:
``` ruby
proc do |*args, &block|
  raise ArgumentError, "no receiver given" if args.empty?
  obj = args.shift
  obj.__send__(sym, *args, &block)
end
```
It takes a splat, that's why we have to call `$slice.call(arguments)` here. MRI says that proc returned by `Symbol#to_proc` has `arity = -1` (which means that most probably proc's parameters are correct)

Maybe we can change it to something like
``` ruby
def to_proc
  # Give name to self in case this proc is passed to instance_eval
  sym = self

  result = proc do |obj, &block|
    raise ArgumentError, "no receiver given" unless obj
    obj.__send__(sym, &block)
  end

  `result.$$arity = -1;`

  result
end
```
Just tested it on the latest master and it works, all specs from `ruby/spec` are green.

cc @elia What do you think?

[...]

PR #1436, opened 2016-04-13
Optimize restarg according to #1096.

iliabylich:
From what I see locally `ruby/spec` test suite is now 10% faster :smile: 
Closes https://github.com/opal/opal/issues/1343.


elia:
:beers: :smile:

Issue #1108, opened 2015-09-18
Class/method names in source maps


jegt:
As far as I can see there is currently no "names" data added to source maps in Opal. Is this something that would be possible to do?

jegt:
I figured out how to get this information from the with the help of the Error.prepareStackTrace callback. At least on V8.

elia:
@jegt Great! Would you mind sharing your findings? I'd like to add them to the stdlib Nodejs support files

jegt:
Sure. This is just on the prototype stage but here is my extension to Error: https://gist.github.com/jegt/1f14990f82d8672380e3

It will give us stack traces like this:
NoMethodError: undefined method `foo' for #<Router:0x2fa>
      at Router.define_mappings (lib/router.rb:9)
      at HTMLDocument.<anonymous> (application.rb:105)

We are writing a desktop application on top of Electron and most of the code is Opal instead of JS.

elia:
Cool! just started doing the same the other day for a [side project](https://github.com/timesampler/timesampler-desktop)…

any plan to open source stuff from the project? 
(e.g. I'm very curious about how you pass source maps to nodejs)

jegt:
There will probably be opportunities to extract code to open source later but I can share how I handle source maps.  This is out build process: https://gist.github.com/jegt/a5a66c7f9518da4c499b

We compile each ruby file inte it's own js file with a source map for each. Then we load all those source maps and create a lookup table that we save to a json file. This json file is then loaded by the Electron app and used to replace the file names and line numbers in the stack traces. 

Our app is currently in public beta and we have 4 devs working on it. I could give you a link but it's only available in Swedish at the moment.  =)

tobymao:
So I tried generating the source maps with the builder.

builder = Opal::Builder.new
app = builder.build("app").to_s
source_map = builder.source_map
app << '//# sourceMappingURL=app.js.map'
File.write(SOURCE_MAP_PATH, source_map)
File.write(APP_PATH, app)

This sort of works, but as @jegt says, no name is generated in the source maps... this still seems broken to me. Googling online, it seems like everyone has to use sprockets in order to have working source maps. Is there a way to get this to work without sprockets?

PR #1440, opened 2016-04-15
Add method names to source maps

wied03:
I realize this doesn't have class names, etc. in yet but it's a start (I hope better than nothing).


wied03:
Could provide more of a resolution to https://github.com/opal/opal/issues/1108

elia:
Great stuff, especially seeing some new specs for source maps and line numbers!

PR #1441, opened 2016-04-18
Source map method name test improvements

wied03:
Try and keep this test from becoming brittle

### DIFF
--- a/spec/opal/core/source_map_spec.rb
+++ b/spec/opal/core/source_map_spec.rb
@@ -9,6 +9,12 @@ describe Opal::SourceMap do
     @map     = Opal::SourceMap.new(compiler.fragments, pathname)
   end
 
+  def js_line_for(code)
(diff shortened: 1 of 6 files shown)


spec/opal/core/source_map_spec.rb
```diff
@@ -9,6 +9,12 @@
     @map     = Opal::SourceMap.new(compiler.fragments, pathname)
   end
 
+  def js_line_for_123
+    index = @source.split("\n").index {|line| line.include?('return 123')}
+    # 1 based line numbers
+    index + 1
+  end
+
```
elia: What about changing to:

```
def js_line_for(code)
  … .index {|line| line.include?(code)}
end 

…

match.generated.line.should == js_line_for('return 123')
```

wied03:
@elia - might be good for making this more generic later even though the before hook is coupled to that right now. I added that in there.

PR #1444, opened 2016-04-19
Remove Node non I/O File methods now that File corelib has a rubyspec

wied03:
passing implementation

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -35,14 +35,6 @@ class File < IO
     end
   end
 
-  def self.basename(path, ext = undefined)
-    `__path__.basename(#{path}, #{ext})`
-  end
-
-  def self.dirname(path)
-    `__path__.dirname(#{path})`
-  end
-
   def self.join(*paths)
     `__path__.join.apply(__path__, #{paths})`
   end


elia:
Are those methods already implemented in the corelib version of the file?

wied03:
@elia - They were just implemented as part of https://github.com/opal/opal/pull/1429

Mogztter:
Even if this two methods are implemented in Ruby, it's not better to keep the native implementation for performance and readability ? (just asking)

wied03:
I suppose it's a question of performance and meeting the Rubyspecs. This isn't I/O based but if Node's implementation is written in C, it could be faster than a JS one.

As to Rubyspecs, from what I can tell, we do not currently `require` the `nodejs` stuff from stdlib, so I can't say if the Node implementations of those pass the Rubyspecs.

@elia - What is your guidance on how these nodejs stdlib wrappers ought to be tested?

elia:
@wied03 the idea is to duplicate the cruby_tests task for nodejs and with time for the other platforms as well. There we can add targeted tests for each platform, I think we started to do that in some PR with @Mogztter tho I don't have the link at hand.

elia:
> Even if this two methods are implemented in Ruby, it's not better to keep the native implementation for performance and readability ? (just asking)

I agree that could be good for performance but compliance with MRI is more important, by that I mean that for performance critical stuff one can always drop to JS and use the node stuff directly but when used through the Ruby api I expect those checks about types etc. that Ruby provides.

So I'm merging this, yet I'm ok with perf improvements leveraging the Nodejs C implementation as long as they provide the same API (roughly: they pass the ruby tests/specs for those methods).

wied03:
@elia - In order to know if the Node version passes, we'd at least need to run some subset of the rubyspecs with `-rnodejs`. Since rubyspecs are easier to read (result wise) than cruby tests, I would think it would make sense to have a smaller rubyspec task that runs with `-rnodejs`, no?

For classes truly in stdlib (not `File` like this PR), there is not much rubyspec coverage and thus the cruby route makes sense.

wied03:
@elia @Mogztter - As the cruby tests advance (or rubyspec coverage grows), I think it would be great to add these back in if we can prove they pass the specs and that there is a performance gain by using Node's functions as the implementation.

PR #1446, opened 2016-04-20
Added arity check to lambdas.

iliabylich:
Also changed `lamda` to `proc` in a few places in core/std -lib.

This change can easily wait for `0.10.0` realease


elia:
I'll merge this now as it looks as a nice addition to the new params stuff 😄

PR #1450, opened 2016-04-21
Remove breaks from Enumerable methods

jgaskins:
The Ruby `break` keyword calls `Opal.new_brk`, which is very, very slow. In many cases, there is a much more performant way to do it. In the case of short-circuiting `Enumerable` methods, it's just a matter of returning.

After talking to @wied03 about the performance of `opal-rspec`, I ran a profile of 1k empty specs. It took over 8 seconds:

screen shot 2016-04-20 at 9 10 32 pm

Well over half of that runtime was spent in `Opal.new_brk`. Considering that profiler run also includes the load time of RSpec into the VM, I figured it might actually be more. Updating the `Enumerable` methods not to use `break` gave that a 4x performance boost:

screen shot 2016-04-20 at 10 00 39 pm

Improving `Enumerable` performance will have massive effects throughout the entire runtime. For example, compare the [individual build line run times of this branch](https://travis-ci.org/jgaskins/opal/builds/124638214) vs [current master builds](https://travis-ci.org/opal/opal/builds/124520964).

### DIFF
--- a/opal/corelib/enumerable.rb
+++ b/opal/corelib/enumerable.rb
@@ -1,13 +1,10 @@
 module Enumerable
   def all?(&block)
-    result = true
-
     if block_given?
 
       each do |*value|
         unless yield(*value)
-          result = false
-          break
+          return false
         end
       end
 
@@ -15,25 +12,21 @@ module Enumerable
 
       each do |*value|
         unless Opal.destructure(value)
-          result = false
-          break
+          return false
         end
       end
 
     end
 
-    result
+    true
   end
 
   def any?(&block)
-    result = false
-
     if block_given?
 
       each do |*value|
         if yield(*value)
-          result = true
-          break
+          return true
         end
       end
 
@@ -41,14 +34,13 @@ module Enumerable
 
       each do |*value|
         if Opal.destructure(value)
-          result = true
-          break
+          return true
         end
       end
 
     end
 
-    result
+    false
   end
 
   def chunk(state = undefined, &original_block)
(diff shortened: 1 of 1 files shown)


wied03:
👍  👍

jgaskins:
This does leave a `break` in `Enumerable#detect`. The only ways I could think off off the top of my head that would improve performance of that method would also make it considerably uglier. Maybe I'll look at it when I've had more sleep. :-)

wied03:
@elia - Could this one make it into 0.10? :)

elia:
Good stuff here!

@wied03 sure, I have no problem with that!

@jgaskins will be AFK for some days, but will merge this & other PRs once I'm back to full speed.

jgaskins:
@elia All good. That'll give me time to figure out how to get rid of the `break` call in `Enumerable#detect`, too.

Just wanted to note what I discussed with @wied03 about this patch so it's out in the open, as well. `break` isn't slow due to deoptimization, which is what causes nearly every other bottleneck in Opal. Our theory is that most of the time spent in `Opal.new_brk` is in generating the stack trace for the `Error` object. I don't know if there's another decent way to implement `break` (unlikely given its semantics), but it might be a good idea to avoid it wherever it's feasible.

jgaskins:
Good news, I was able to remove the `break` from `Enumerable#detect` pretty trivially, too. This was the last one in the file and when I run `ack break opal/corelib`, the only ones in the entire core lib appear to be JS breaks, so this should be all that needs to be done here.

elia:
🎉 finally 😄

iliabylich:
:beers: Great Job!

Issue #1449, opened 2016-04-21
Singleton Object/Class Ancestor Regression


wied03:
When a module defines the case compare operator, it also breaks usages of the case compare operator when that module is on the RHS of the operator.

```ruby
module Foo
  def self.===(*)
    raise 'method contents are irrelevant to bug'
  end
end

puts Class === Foo
```
MRI:
```
false
```
Opal master via `bundle exec opal -A ancestors.rb`:
```
true
```
Opal 0.9.2 executed same way
```
false
```

wied03:
Might be a side effect of https://github.com/opal/opal/commit/bf22586c0fc2a4a98f5f9b6c31aa39f65ea64493

wied03:
I created a Rubyspec for this - https://github.com/ruby/spec/pull/244

wied03:
I keep trying to find alternatives to this code in `Opal.is_a` to fix this:
```js
    ancestors = Opal.ancestors(object.$$meta);
    for (i = 0, length = ancestors.length; i < length; i++) {     
      if (ancestors[i] === klass) {
        return true;
      }
    }
```
I thought about a guard to bail out of the function to if `===` is defined on object before iterating over the meta ancestors might work. Except I'm not sure that would really fix this since a module like the one in this example might be used a singleton case like in the specs.

Should `Class` be an ancestor of the singleton class for `Foo` in this case? Maybe that's the issue.

wied03:
Along those lines, ran this test:
```ruby
module Foo
  def self.===(*)
    raise 'method is irrelevant to bug'
  end
end

class << Foo
  puts self.ancestors
end
```
MRI:
```
#<Class:Foo>
Module
Object
Kernel
BasicObject
```
Opal:
```
[#<Class:0x10c>, #<Class:0x16>, #<Class:0x14>, #<Class:0x12>, Class, Module, Object, Kernel, BasicObject]
```

Class isn't there in the MRI version.

jgaskins:
This may just be a really weird edge case. An instance of `Class` isn't necessarily a subclass of `Class`. In fact, you should probably never _actually_ inherit from `Class`. :-)

I wonder why it's showing up in the ancestor chain.

wied03:
I looks more like the `===` stuff is just a symptom of the problem and that `Class` being in the ancestor chain is the real problem

wied03:
https://github.com/ruby/spec/pull/245 - more directly target the ancestor issue

PR #1445, opened 2016-04-20
Implement Pathname#relative_path_from

wied03:
Rubyspecs - https://github.com/ruby/spec/pull/241

PR #1451, opened 2016-04-21
Fix underlying ancestor issue that caused case compare regression

wied03:
When a singleton class for a module is used, don't traverse the class hierarchy

https://github.com/ruby/spec/pull/245 should cover this one fairly well (and Opal will pass all of them when it's merged)

Fixes https://github.com/opal/opal/issues/1449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ Whitespace conventions:
 
 ### Fixed
 
+- `Module#ancestors` and shared code like `====` and `is_a?` deal with singleton class modules better (#1449)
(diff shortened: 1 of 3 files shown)


wied03:
@iliabylich - Does this cover it?

wied03:
@elia - We have some congestion with rubyspecs (3 dependent PRs) - Rather than add two sets of exclusions for pending PRs (already 1 in here), can you merge (or reject) https://github.com/opal/opal/pull/1445 first, then I can rebase this and it will pass?

wied03:
@elia - rebased, once travis completes, this one should be green

PR #1452, opened 2016-04-21
Class#to_s - Handle singleton class names

wied03:
This made it a lot easier to debug https://github.com/opal/opal/pull/1451

Rubyspecs are https://github.com/ruby/spec/pull/246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ Whitespace conventions:
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


wied03:
Will wait to update rubyspecs based on what happens with https://github.com/opal/opal/pull/1451

wied03:
Before this is merged, ensure https://github.com/opal/opal/pull/1451 is merged, then rebase and remove the rubyspec to_s filter

PR #1453, opened 2016-04-21
Enable CRuby tests again

wied03:
(can catch some things ruby specs do not test yet).

Also remove the minitest task since all it did was call the CRuby tests.
This way it's more clear in Travis what's going on.


elia:
Great thanks, just one request, can you put back the Minitest task? That array with only CRuby_test in it is meant to be filled by other stuff with time and has parity with the other test runners we have. Also CRuby_test should be split, but that's beyond this PR 😊

wied03:
@elia - done

Issue #1286, opened 2016-01-13
code execution bug


Thermatix:
I'm not sure if this is a bug or not but when I do:
```ruby
require './example'`
```
 the code in file example.rb doesn't seem to execute 
but if i do 
```ruby
`require_relative './example'`
```
 it does, or have I missed somthing?

wied03:
I can confirm that MRI lets you do `require './example'`. In that case, it seems to function like `require_relative` and does not look for 'example' in the load path, it looks relative to the requiring file.

elia:
Actually `require_relative` has a slightly different behavior, it requires relative from the file in which it's used (vs. the current dir). E.g.:

### `require "./foo"`

```
⤑ cd /tmp/                                                                     ~
⤑ echo "p 1" > foo.rb                                                       /tmp
⤑ mkdir bar                                                                 /tmp
⤑ echo "p 2" > bar/foo.rb                                                   /tmp
⤑ echo "require './foo'" > bar/baz.rb                                       /tmp
⤑ ruby bar/baz.rb                                                           /tmp
1
⤑ echo "require_relative './foo'" > bar/qux.rb                              /tmp
⤑ ruby bar/qux.rb                                                           /tmp
2
⤑       
```

While running opal code current dir info is seldom available, and often doesn't make sense at all (e.g. in a browser), but maybe we could assume "." and support that too, so that `require "./foo"` is the same as `require "foo"`.

wied03:
@elia - I saying that require's behavior is unique in that case. Here's an example:

Let's say I have this code. example.rb looks like this
```ruby
puts 'hello there'
```
When I run `ruby foo.rb` from the junk directory, I get a predictable cannot load such file -- ./example error

When I run `ruby ../foo.rb` from the stuff directory, I get `hello there`. Different behavior than `require_relative`.
![image](https://cloud.githubusercontent.com/assets/297123/14619481/fc66ecb2-0574-11e6-9606-362ba9971a45.png)

elia:
Yeah, sorry, I guess we all agree, I initially misunderstood your reply 😊

PR #1454, opened 2016-04-21
Allow require to work with current directory

wied03:
Replicate MRI behavior, probably needs some discussion before squashing.

Should fix https://github.com/opal/opal/issues/1286 - see discussion there

This approach avoids adding the current directory to the load path (which would be incompatible with MRI) and it also scrubs the ".." and "." references out of `Opal.modules` and `self.$require` calls.

The main remaining issue is there are a few bugs in Pathname that prevent the compiler_spec changes here from running under Opal (as in bootstrap compiling) and thus the filter. One of those bugs is fixed by https://github.com/opal/opal/pull/1445 but there are others.

**Alternative approach:**
Add the current directory to the load path, but ignore it unless '..' or '.' is used in the require statement. You'd still need to clean up `Opal.modules` and `self.$require` though and you'd probably do just as much if not more Hike hacking than what's done here.

### DIFF
--- a/lib/opal/hike_path_finder.rb
+++ b/lib/opal/hike_path_finder.rb
@@ -16,6 +16,7 @@ module Opal
       super
     end
 
+    # Cover require './something' and require '../something' cases, which should use the current directory
     def find_relative_current_dir path
       # Hike is great, but doesn't seem to do very well when trying to deal with relative paths (.. in particular)
       basename = File.basename(path)
@@ -39,15 +40,11 @@ module Opal
     private
 
     def sort_matches(matches, basename)
-      aliases = []
-
       matches.sort_by do |match|
         extnames = match.sub(basename.to_s, '').to_s.scan(/\.[^.]+/)
         extnames.inject(0) do |sum, ext|
           if i = extensions.index(ext)
             sum + i + 1
-          elsif i = aliases.index(ext)
-            sum + i + 11
           else
             sum
           end
--- a/lib/opal/hike_path_finder.rb
+++ b/lib/opal/hike_path_finder.rb
@@ -35,6 +35,8 @@ module Opal
           return pathname.expand_path.to_s
         end
       end
+      # not found
+      nil
     end
 
     private
--- a/lib/opal/hike_path_finder.rb
+++ b/lib/opal/hike_path_finder.rb
@@ -14,5 +14,17 @@ module Opal
       return path if pathname.absolute? and pathname.exist?
       super
     end
+
+    def find_relative_current_dir path
+      current_dir_index.find path
+    end
+
+    private
+
+    def current_dir_index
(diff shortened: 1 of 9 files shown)


wied03:
https://travis-ci.org/opal/opal/builds/124867250 in case build is not visible due to CHANGELOG commit.

lib/opal/nodes/top.rb
```diff
@@ -38,7 +38,12 @@ def compile
 
       def opening
         if compiler.requirable?
-          path = Pathname(compiler.file).cleanpath.to_s
+          path = Pathname(compiler.file)
+          # Don't want relative paths in modules
+          if path.to_s =~ /^\.\./
```
elia: `path.to_s.start_with? '..'` looks more clean here, wdyt?

wied03: No good reason for not doing that. Got in a bad regex routine I guess.

wied03:
@elia - Any other issues? Is the general approach here a good one?

lib/opal/hike_path_finder.rb
```diff
@@ -7,12 +7,61 @@ def initialize(paths = Opal.paths)
...
+
+    def pattern_for(basename)
+      @patterns[basename] ||= build_pattern_for(basename)
+    end
+
+    def build_pattern_for(basename)
+      basename_re = Regexp.escape(basename.to_s)
+      extension_pattern = extensions.map { |e| Regexp.escape(e) }.join("|")
+      /^#{basename_re}(?:#{extension_pattern})*$/
+    end
```
elia: Here I'd do

```ruby
def pattern_for(basename)
  @patterns[basename] ||= begin
    basename_re = Regexp.escape(basename.to_s)
    extension_pattern = extensions.map { |e| Regexp.escape(e) }.join("|")
    /^#{basename_re}(?:#{extension_pattern})*$/
  end
end
```

But it's ok as it is too

lib/opal/hike_path_finder.rb
```diff
@@ -7,12 +7,61 @@ def initialize(paths = Opal.paths)
       super()
       append_paths(*paths)
       append_extensions '.js', '.js.rb', '.rb', '.opalerb'
+      @patterns = {}
     end
 
-    def find path
+    def find path, options={}
```
elia: Is this for API parity with something else?

wied03: Originally, path_reader's `expand` method was using this with an additional option but ended up having to write a new method anyways. This could be removed.

wied03:
@elia - Besides your thoughts on the general approach here (still curious), I do think it's important to get the `Pathname` issues addressed so that this doesn't break bootstrap compilation before merging it.

PR #1462, opened 2016-04-26
Small Date improvements.

iliabylich:


stdlib/date.rb
```diff
@@ -338,6 +338,10 @@ def parse(string, comp = true)
     def today
       wrap `new Date()`
     end
+
+    def gregorian_leap?(year)
+      `!(new Date(#{year}, 1, 29).getMonth()-1)`
```
elia: the failure seems due to this line, the JSHint approved version is probably:

`(new Date(#{year}, 1, 29).getMonth()-1) === 0`

PR #1348, opened 2016-02-14
Implement IO.each_line & IO.read method for Node.js

Mogztter:
Resolves #1346 & resolves #1221

PR #1464, opened 2016-04-27
Implement `File.new("path").mtime`, `File.mtime("path")`, `File.stat("path").mtime`

Mogztter:


CHANGELOG.md
```diff
@@ -46,12 +46,12 @@ Whitespace conventions:
 - Implement `Pathname#join` and `Pathname#+` methods (#1301)
 - Added support for `begin;rescue;else;end`.
 - Implement `File.extname` method (#1219)
-- Implement File.extname method (#1219)
```
Mogztter: Not related but boy scout rule :tent:

wied03:
@Mogztter - I'm considered about test coverage here. Shouldn't we get either the cruby tests or some rubyspecs running for this before we add it? Otherwise we're setting ourselves up for bugs down the road where a method appears to be implemented but we get unexpected behavior. I'd rather have no method than rely on one assuming it behaves like the Ruby method should but then I get surprised in a subtle way.

Mogztter:
@wied03 Sure, I've created a cruby task for Node.js in https://github.com/opal/opal/pull/1348/files#diff-31d0c7df77415700019f4ac1f555d007R243
Once this pull request is merged I will be able to add/activate tests for `mtime` method

Issue #1235, opened 2015-12-09
Switch to Parser


elia:
ref #604 

- [ ] use the `ast` gem, which is the same one used by `parser`
- [ ] ensure source maps still work
- [ ] move to an AST structure as similar as possible to the one emitted by `parser`
- [ ] switch to `parser` 
- [ ] implement all the missing stuff in the compiler

_this list is a draft, it's expected to change down the road and adjust to how things turn out be *in the "real world"™*_

--- 

**Notes**

- a fork of the AST gem with passing specs, the only actual changes are for `#inspect` like methods that were using `String#<<` https://github.com/elia/ast

elia:
@alexdowad I drafted the list of stuff that needs to be done to switch to parsers, please edit if you see anything that can be improved or done differently

wied03:
@elia - Can we add that it should be bootstrap friendly?

elia:
@wied03 what do you mean?

wied03:
@elia - that this change does not break the ability to run the Opal compiler in Node, which currently works

whitequark:
@elia I'm confused about your changes to `ast`. What's that for?

elia:
@whitequark were mostly related to make the tests pass under opal, which translates to avoiding string mutation and string/symbol difference. I basically replaced `String#<<` with `Array#<<`+`Array#join` and tweaked a couple of specs to use `:symbol.inspect` instead of `":symbol"` so that it passes when `String` == `Symbol`. Finally I think I have some local bacon changes that I still need to push.

Feel free to point out specific lines in https://github.com/elia/ast/commit/6f76f4e4ba6871f5f638285f478991e2d93d9e75 if there's anything else that's unclear.

That said I think all of that will stay on hold until #1465 is in a good shape.

elia:
solved by #1465

PR #1465, opened 2016-05-03
Replace Opal::Parser/Opal::Lexer/Opal::Sexp with parser/ast gems.

iliabylich:
For now:
+ I can compile and run corelib locally :smile: `opal -A -e 'p 123'`
+ A lot of edge cases are not covered yet, didn't have a chance to compile and run stdlib yet.
+ Compiler contains some mass assignment stuff that must be resolved before compiling `opal-parser`
+ After this _probably_ we can even run some ruby specs (at least from `/language` dir)
+ There is a serious difference in generated AST between opal/parser parsers. We definitely can remove some nodes after migrating to `parser`
+ Stuff like `.JS.` and `.JS[]` can be easily implemented through spacial calls, it's just a chain of two `:send` nodes.
+ /cc @elia One of the items in #1235 was literally "to not break source maps". I've never used sourcemap stuff, so I have no idea how to fix/check it. Could you try it locally/point to the missing place in the source code?
+ @opal/core Patches/code reviews are always welcome

I'll create a more detailed todo list a little bit later.


wied03:
@iliabylich - there are a couple source map specs that run (see https://github.com/opal/opal/pull/1440)

vais:
> +653 −11,016

LOL 👍

alexdowad:
@whitequark Looks like you will have another "customer" soon.

iliabylich:
Alright, compiler now can be compiled to a valid JS.

```
bin/opal -gast -gparser -e "require 'opal-parser'; p 42"
# => 42
```

Got a problem:
1. `parser/ruby23` uses `module_eval(string)` inside as inline method call to define all methods
2. So if we want to do `require 'parser/ruby23'` we have to call `Module#module_eval`
2. And `Module#module_eval` requires `opal-parser` to be `require`-d
3. And `opal-parser` `require`-s `opal/compiler` (which obviously requires `parser/ruby23`)

Everything would be ok without that inline invocation of `module_eval`. Does anyone have any ideas how to solve that? `parser` gem provides compiled grammar for ruby23, I can't modify its source.

@alexdowad Any thoughts?

whitequark:
The `module_eval` invocation is only used for nice line numbers in backtraces. I actually don't think it's even necessary in release builds.

alexdowad:
Hmm. `parser`'s grammars are compiled using Racc. Racc is what is actually putting those `module_eval` calls in the parsers.

I was working on enhancing Racc at one point, but never reached the 2.0 release, and it's too much work to finish right now. It would be easy to make a tweaked Racc which would not insert those `module_eval` calls. Then you could compile the `parser` grammars yourself to make your own parser files.

What do you think? Is there a better way? @whitequark?

whitequark:
`racc --no-line-convert` ?

alexdowad:
Uhhh... yeah, that would do the trick too.

elia:
`--no-line-convert` (as `-l`) is what is used to generate the current grammar:

https://github.com/opal/opal/blob/c3c4b87db347e7283e3b38cc328f2995e28625e0/tasks/building.rake#L52

iliabylich:
@whitequark @alexdowad Is it possible to tell `parser` to yield raw x-string content? Current opal parser+lexer converts the file containing `%x{/\d/}` to `/\d/`. With the `parser` gem I'm getting `/d/`.

Parser's lexer skips the `\` character. I know, this is a default MRI behavior, but in Opal x-string means a raw string containing JS code that should be injected directly to the output file. And I guess it can't be fixed by AST rewritting because lexer returns 
```
read    :tXSTRING_BEG(tXSTRING_BEG) ["%x{", #<Parser::Source::Range test 0...3>]
...
read    :tSTRING_CONTENT(tSTRING_CONTENT) ["/d/", #<Parser::Source::Range test 3...7>]
```

Of course, there's a simple solution: we can escape those `\` characters everywhere. Do you have any ideas how to fix it without duplicating every single `\` character?

iliabylich:
Another alternative solution for Opal could be explicit regexp interpolating to the x-string
``` ruby
`#{/\d\\/}`
```
@elia What do you think?

alexdowad:
@iliabylich, probably @whitequark may have a better idea, but what comes to mind is either monkey-patching `Parser::Lexer::Literal#interpolate?` or modifying `Parser::Lexer::Literal::TYPES` like:

    Parser::Lexer::Literal::TYPES['%x'] = [:tXSTRING_BEG, false]

iliabylich:
@alexdowad already tried patching `Literal` without any success, x-strings should support interpolation, [here is a quick example](http://opalrb.org/try/?code:%25x%7B%2F%5Cd%2F%3B%20%23%7B123%7D%7D)

iliabylich:
I guess skipping `\` is a part of interpolation, right?
``` ruby
2.3.0 :001 > "\d"
 => "d" 
2.3.0 :002 > '\d'
 => "\\d"
```

alexdowad:
Yes, it is.

alexdowad:
Hmm. I wouldn't want to suggest that you should change the Opal dialect of Ruby just to suit the `parser` gem, but there would be advantages in being more consistent with MRI Ruby's treatment of strings.

whitequark:
@iliabylich you can call `node.loc.source` to get the code of the node exactly as it was in the source code.

whitequark:
@iliabylich you can call `node.loc.source` to get the code of the node exactly as it was in the source code.

iliabylich:
@whitequark Thanks, it works!

iliabylich:
@whitequark Could you release the latest version of `ast` please? Opal needs it as a runtime dependency in gemspec.

whitequark:
ast `2.3.0`

iliabylich:
Removed WIP label, ready for review
cc @opal/core

iliabylich:
Probably I should also mention the following changes:
1. `.JS.`/`.JS[]` syntax is not native node anymore. Opal handles it through `handle_special` stuff.
2. Extracted `defs` (`def self.m`) to a separate node, `parser` handles it very well
3. Opal now can compile `retry` to `$retry()` :smile: - anyway there were no support for this node, it's just a stub for now
4. Parser can't parse some unicode characters - opal skips them with a warning
5. `while_post` and `until_post` are separate nodes (thanks to `parser`)
6. Opal calls `use_gem 'parser'` and `use_gem 'ast'` by default

lib/opal/nodes/case.rb
```diff
@@ -8,34 +8,39 @@ class CaseNode < Base
...
+          push "$case = ", expr(condition), ";"
+          empty_line
+        end
+
+        case_parts.each_with_index do |wen, idx|
+          if wen and wen.type == :when
+            wen = compiler.returns(wen) if needs_closure?
+            push "else " unless idx == 0
+            push stmt(wen)
+          elsif wen # s(:else)
```
meh: I'm not sure I like this.

I think it would be better to check `if wen` and then do a `case`.

lib/opal/nodes/constants.rb
```diff
@@ -5,25 +5,28 @@ module Nodes
     class ConstNode < Base
       handle :const
 
-      children :name
+      children :const_scope, :name
 
       def compile
-        if name == :DATA and compiler.eof_content
+        if const_scope.nil? && name == :DATA and compiler.eof_content
```
meh: This one is pretty hard to follow as well.

whitequark:
> 4. Parser can't parse some unicode characters - opal skips them with a warning

Some *non*-Unicode characters? If Parser doesn't handle any Unicode characters it's certainly a bug.

iliabylich:
@meh See https://github.com/whitequark/parser/issues/290 and https://github.com/opal/opal/pull/1465/files#diff-bbd9f11c4c273494a8e1eaf9b27956d4R66

whitequark:
@iliabylich The first case refers to non-UTF-8 byte sequences (more specifically, byte sequences incompatible with file encoding), which is exactly the problem. Though you can override the corresponding method in `AST::Builder` to do something sane with them, assuming you know what that is.

The second case is actually just invalid. You can represent astral plane characters in UTF-16, which is what JavaScript uses, using a surrogate pair. Surrogate pairs are invalid in `\u{hex}` notation as parsed by MRI though, so you probably should only put them in the JavaScript strings as literal characters, or something like that. Try evaluating `"\ud83d\ude0d"` in a JavaScript console.

elia:
🎉 😄

PR #1469, opened 2016-05-06
Opal Compiled Ruby docs fix example in Native section

roupen:
rename opal variable to win - window was causing error


elia:
Thanks!

PR #1472, opened 2016-05-23
Allow methods to be marked as pristine from corelib

### LINKED ISSUE #1459: Opal 0.10.0.beta3 is considerably slower than Opal 0.9.0.beta2
Hello,

We've setup a benchmark in Asciidoctor.js and Opal 0.10.0.beta3 is considerably slower than Opal 0.9.2.beta2. See: https://github.com/asciidoctor/asciidoctor.js/issues/183

I've created a repository to reproduce this issue: https://github.com/Mogztter/asciidoctor.js-benchmark
Any help to find the root cause will be greatly appreciated :smile: 

NOTE: We've also found that Opal 0.9.2 is slightly faster than Opal 0.9.0.beta2 :+1:

elia:
This allows to make certain assumptions and possibly provide a native
implementation instead of going through the complete call process.

An example is showed for improving `Array#dup` performance.

fixes #1459

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -795,6 +795,19 @@ class Array < `Array`
     }
   end
 
+  def dup
+    %x{
+      if (
+        self.$$class === Opal.Array &&
+        self.$allocate.$$pristine &&
+        self.$copy_instance_variables.$$pristine &&
+        self.$initialize_dup.$$pristine
+      ) return self.slice(0);
+    }
+
+    super
+  end
+
   def each(&block)
     return enum_for(:each){self.size} unless block_given?
 
@@ -2231,4 +2244,7 @@ class Array < `Array`
   def instance_variables
     super.reject { |ivar| `/^@\d+$/.test(#{ivar})` || ivar == '@length' }
   end
+
+  Opal.pristine self, :allocate, :copy_instance_variables, :initialize_dup
 end
+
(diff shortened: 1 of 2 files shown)


elia:
@opal/core @Mogztter @jgaskins 

Please review and let me know about this pattern and naming (`$$pristine` vs. ?), if all is good I'll add proper docs and merge.

iliabylich:
@elia So this optimization works only for `Array#dup` and for `Array` subclasses having original `allocate`, `copy_instance_variables`, `initialize_dup` methods, right?

I remember in some issue you mentioned that it would be nice to have a separate `opal/micro` (ignore the naming :smile: ) that has some widely used subset of functionality defined in `opal/corelib`, like it should ignore stuff with subclasses  or `String#grep(&blk)` (because nobody actually uses it). As I remember, the idea was taken from MRuby.

I'm completely ok with your idea, but is there any common solution for this optimization problem? I just think that we may have to put similar code to most of corelib methods (which is obviously bad)

elia:
> So this optimization works only for Array#dup and for Array subclasses having original allocate, copy_instance_variables, initialize_dup methods, right?

yes

>  the idea was taken from MRuby.

yes, that's correct, the only barrier I see before switching to a layered approach is testability as ruby/spec will be useful just for the complete version while I expect most browser usage would go with the "micro" version.

That said I'm not sure we can actually find a combined approach because I still desire `Array#dup` to be performant even when using the "full featured" opal.

Mogztter:
Thanks @elia I will give it a try as soon as possible !

PR #1473, opened 2016-05-30
Optimize Enumerable#count

jgaskins:
One of the most common uses of `Enumerable#count` is as an alias for `length` or `size`. Without a block or argument, we can short-circuit this to `size`. I also switched out the passing of `arguments` to `Opal.yieldX` so the function could be optimized.

I broke some tests for what looks like the more obscure cases, so I'll need to check into that, but it seems to work for the more common cases.

All benchmarks are iterations/sec after a 100ms warmup. The array has 1000 elements. Higher results are better. I included `Array#size` to show the performance ceiling.

#### Before

```
array.size
  1779658
array.count
  4931
array.count(1)
  1958
array.count(&:odd?)
  480
array.count { |i| i.odd? }
  2042
```

#### After

```
array.size
  1546681
array.count
  1471198
array.count(1)
  2873
array.count(&:odd?)
  464
array.count { |i| i.odd? }
  2572
```

#### Before

```
array.size
  3197569
array.count
  3905
array.count(1)
  2087
array.count(&:odd?)
  397
array.count { |i| i.odd? }
  1868
```

#### After

```
array.size
  2191544
array.count
  2100559
array.count(1)
  3324
array.count(&:odd?)
  567
array.count { |i| i.odd? }
  3134
```

#### Before

```
    array.size
        1642654
    array.count
        1333
    array.count(1)
        693
    array.count(&:odd?)
        359
    array.count { |i| i.odd? }
        840
```

#### After

```
    array.size
        1840939
    array.count
        1434968
    array.count(1)
        1929
    array.count(&:odd?)
        458
    array.count { |i| i.odd? }
        1933
```

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -554,6 +554,14 @@ class Array < `Array`
     self
   end
 
+  def count(object = nil, &block)
+    if !(object || block) && respond_to?(:size)
+      size
+    else
+      super
+    end
+  end
+
   def initialize_copy(other)
     replace other
   end
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -555,10 +555,10 @@ class Array < `Array`
   end
 
   def count(object = nil, &block)
-    if !(object || block) && respond_to?(:size)
-      size
-    else
+    if object || block
       super
+    else
+      size
     end
   end
 
--- a/opal/corelib/enumerable.rb
+++ b/opal/corelib/enumerable.rb
@@ -111,11 +111,11 @@ module Enumerable
     map { |item| yield item }.flatten(1)
   end
 
-  def count(object = nil, &block)
+  def count(object = undefined, &block)
     %x{
       var result = 0;
 
-      if (#{Opal.truthy?(`object`)}) {
+      if (object != null) {
         block = function(arg) {
           return #{`arg` == `object`};
         };
--- a/opal/corelib/enumerable.rb
+++ b/opal/corelib/enumerable.rb
@@ -112,35 +112,21 @@ module Enumerable
   end
 
   def count(object = undefined, &block)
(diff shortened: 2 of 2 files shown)


jgaskins:
This whole thing started as a look into why `String#to_proc` was slower than an explicit block. :-)

jgaskins:
I noticed as I iterated on this that it started to resemble an idiomatic Ruby method post-compile, so I went ahead and ported it. No performance was lost.

JS:
screen shot 2016-05-31 at 10 49 44 pm

Ruby:
screen shot 2016-05-31 at 10 49 55 pm

Higher iteration counts in a couple of those is due to longer dry runs to fully invoke the optimizer.

elia:
Ruby FTW!

I'll squash/merge given the change is small (vs. the commit history)

--

_btw, was "researching" good ways to write the arguments to args loop and wanted to share this trick I found which is slightly fastwe but still readable:_

```js
var args = [];
for(var i = 0, l = arguments.length; i < l; i++) { args[i] = arguments[i] }
```

https://albertxing.com/stool/#215066046ca3427b8faa15c012425314

jgaskins:
> btw, was "researching" good ways to write the arguments to args loop and wanted to share this trick I found which is slightly fastwe but still readable

Totally, caching object properties that don't change is not a bad idea. There's doesn't seem to be a detectable difference between the two in Chrome/Firefox, but it looks to be about a 3-4% performance boost in Safari 9.1.

Also, I'm really glad there's something else for benchmarking JS out there. JSPerf has been down for aaaaages.

elia:
> Also, I'm really glad there's something else for benchmarking JS out there. JSPerf has been down for aaaaages.

Yeah, was on the brink of reinventing the wheel myself before finding it. 😄

PR #1474, opened 2016-06-01
Fixes Native Hash initialize to work with React.rb

adamcreekroad:
Currently in 0.9 when using alongside React.rb, component rendering will break and return this error
```js
TypeError: Cannot read property 'constructor' of null
```

This traces back to the Native Hash initialize method
```ruby
def initialize(defaults = undefined, &block)
  %x{
    if (defaults !== undefined && defaults.constructor === Object) {
```

In the current 0.8, this method checks
```ruby
def initialize(defaults = undefined, &block)
  %x{
    if (defaults != null) {
      if (defaults.constructor === Object) {
```

I'm not sure exactly why it was changed, but changing it to

```ruby
def initialize(defaults = undefined, &block)
  %x{
    if (defaults !== undefined && defaults != null && defaults.constructor === Object) {
```
appears to fix it and allows me to go on with life.

### DIFF
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -526,7 +526,7 @@ class Hash
 
   def initialize(defaults = undefined, &block)
     %x{
-      if (defaults !== undefined && defaults.constructor === Object) {
+      if (defaults !== undefined && defaults != null && defaults.constructor === Object) {
         var smap = self.$$smap,
             keys = self.$$keys,
             key, value;
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -526,7 +526,7 @@ class Hash
 
   def initialize(defaults = undefined, &block)
     %x{
-      if (defaults !== undefined && defaults != null && defaults.constructor === Object) {
+      if (defaults != null && defaults.constructor === Object) {
         var smap = self.$$smap,
             keys = self.$$keys,
             key, value;


vais:
`defaults !== undefined && defaults != null` should probably be `defaults !== undefined && defaults !== null` or, as much as I hate to admit it, simply `defaults != null`. In JS, `undefined === null` #=> `false`, but `undefined == null` #=> `true`, so simply checking with `defaults != null` is sufficient to catch both cases.

jgaskins:
@vais Should `null` be an acceptable value to pass to the `Hash` constructor, though? The `undefined` check is there specifically to see if a value was passed in so that `Hash.new` just generates an empty hash. Passing `null` is explicit, which makes me think this isn't the right place to be handling that value.

vais:
@jgaskins well, this code is inside stdlib's `Native` module monkey-patching of `Hash`, so it's trying to be javascript-friendly, i.e. there's no telling what will be coming over from JS-land, and `defaults != null` seems to be the least verbose way of handling both `null` and `undefined` in one go.

jgaskins:
If it were a matter of handling `undefined` and `null` as explicit args, that would make perfect sense. However, `undefined` is a sentinel value to indicate they omitted the args. With `null`, we know they passed one in.

elia:
just for the record the original change happened here: https://github.com/opal/opal/commit/bd61c41e#diff-9da39d7da8fe5f828014243ec8c40f48L449

meh:
Personally I think `null` should be an acceptable value for `Hash.new`, and be treated the same as `undefined`.

The rationale is `Hash.new(nil) => {}`.

jgaskins:
@meh In that case, wouldn't we be interpreting `nil` as the default value for `Hash#[]`? Should this be true?

```ruby
Hash.new(`null`)[:foo] == `null`
```

meh:
@jgaskins the `undefined/null` should be treated as `nil` as default, which is already what would be happening.

elia:
The specs were broken because they were using a newer version of mspec that wasn't compatible

This was the fix: https://github.com/opal/opal/commit/6fd025649c6f371218ddf58bac65791ba1ba0ea6

@adamcreekroad 
thanks for your contribution! 
can you send a PR with to same change targeting master?

I'll issue a new 0.9 release soon.

jgaskins:
> the `undefined/null` should be treated as `nil` as default

@meh Ah, I see. I didn't expect it be intended to cast it. 👍

PR #1478, opened 2016-06-03
super fixes for replace-opal-parser branch

wied03:
Fixes all the call_specs and removed filters for range min/max since those pass now too. By all means do a code review since there might be better ways to do some of this.

Besides the scope.rb change:
* Removed semicolons from call_spec for less brittle specs and to focus more on real failures
* Take advantage of `extract_iter` inheritance from call.rb and utilize that properly for the zsuper case
* zsuper's `@arguments_without_block = []` else clause was not being used anymore since that was all removed, so clean that up

### DIFF
--- a/lib/opal/nodes/scope.rb
+++ b/lib/opal/nodes/scope.rb
@@ -265,11 +265,10 @@ module Opal
             chain << scope.identify!
             scope = scope.parent if scope.parent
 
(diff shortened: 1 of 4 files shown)


iliabylich:
A huge :+1:

meh:
Looks good to me.

PR #1479, opened 2016-06-07
Fix from 0.9-stable that changes Native Hash initialize check for null

adamcreekroad:
Sending a PR to master with the same fix.


elia:
Thanks!

PR #1481, opened 2016-06-11
Make runtime.js compatible with opal/mini

jgaskins:
It was calling `Kernel#respond_to?`, which doesn't come with opal/mini. Instead, that call needs to use the lower-level implementation of that method.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1321,7 +1321,7 @@
       // A splatted array must be copied
       return value.slice();
     }
-    else if (value['$respond_to?']('to_a', true)) {
+    else if (typeof(value.$to_a) === 'function' && !value.$to_a.$$stub) {
       var ary = value.$to_a();
       if (ary === nil) {
         return [value];


wied03:
Ideally we add a simple smoke test of opal/mini to the travis build to catch this stuff in the future. @elia - since there are probably people that depend on opal/mini, should this get fixed before the 0.10 release?

opal/corelib/runtime.js
```diff
@@ -1321,7 +1321,7 @@
       // A splatted array must be copied
       return value.slice();
     }
-    else if (value['$respond_to?']('to_a', true)) {
+    else if (typeof(value.$to_a) === 'function' && !value.$to_a.$$stub) {
```
iliabylich: This code is not 100% correct for MRI. It **must** call `respond_to?`. I think it could be replaced to
``` js
else if (
  (
    // checking through 'respond_to?' if possible
    typeof(value['$respond_to?']) === 'function' &&
    !value['$respond_to?'].$$stub &&
    value['$respond_to?']('to_a', true)
  ) || (
    // we are in 'opal/mini', checking directly through 'to_a'
    typeof(value.$to_a) === 'function' &&
    !value.$to_a.$$stub
  )
) {
  // calling to_a
}
```

There are 4 occurrences of `respond_to?` in `runtime.js`, so it could be extracted to something like
``` js
Opal.respond_to = function(method_name, include_private) {
  // the code here
}
```
And also we could make some flag on `Opal.config` (js object) that indicates whether we running on `opal/full` or `opal/mini`, it can simplify the logic.

jgaskins: > It **must** call `respond_to?`

Why do you say that?

iliabylich: Because MRI calls it. There's a [test for this scenario](https://github.com/ruby/spec/blob/master/language/variables_spec.rb#L385-L392) in rubyspec and it fails [in your CI build](https://travis-ci.org/opal/opal/jobs/136962065#L316).

Actually `opal/mini` doesn't need it, but to make it compatible with `opal/full` there should be some wrapping `if` condition.

jgaskins: Maybe the solution is to add `respond_to?` to the smaller Opal builds, then. If it's a dependency of basic syntax, it should be available regardless of how the runtime is loaded.

iliabylich:
@wied03 Do you have any ideas what should be verified in the smoke test?

iliabylich:
Wait, `opal/mini` contains `corelib/kernel`, so it has both `Kernel#puts` and `Kernel#respond_to?`. However `Kernel#puts` depends on `corelib/io` (which is not a part of `opal/mini`) so it doesn't work - this is the only thing that should be fixed.

iliabylich:
Can we fix `Kernel#puts` by adding `return unless $stdout`? (Without `corelib/io` we can't even do some `warn` :smile: ) Or can we add a `require 'corelib/io'` to `opal/mini`? The first solution is ugly, the second one increases the resulting size of `opal/mini`

meh:
I think it makes sense to add `opal/io` to `opal/mini`, it's really small and kind of basic.

PR #1484, opened 2016-06-14
Fix implementation of class variables.

iliabylich:
The following methods are 100% compliant with rubyspec:
  - Module#class_variables
  - Module#class_variable_get
  - Module#class_variable_set
  - Module#remove_class_variable


iliabylich:
@opal/core Any comments here?

elia:
@iliabylich just add a changelog entry (under 0.11) and then merge it. 

After this we're not officially discouraging evil `@@cvars` anymore! 😄

PR #1486, opened 2016-06-21
Some improvements to Range class

taw:
I needed working `Range#step`, but since I was already hacking it I did a bunch of other methods as well.
It's still quite short of fully compliant `Range` (which has a ton of special cases for `Float`s and arbitrary objects, not just `Integer`s this implementation quietly assumes half of the time), but a lot of issues got fixed.

* a few things which will never work (tainted/untrusted/private) moved from bugs to unsupported
* `Range#hash` returning `Fixnum` bug added, as it was masked by `Range#hash` not existing


opal/corelib/range.rb
```diff
@@ -85,16 +93,26 @@ def exclude_end?
     @exclude
   end
 
-  alias :first :begin
+  def first(n=(n_specified=true; nil))
```
iliabylich: Probably it can be improved by passing default value = `undefined`:
``` ruby
def m(n = undefined)
  %x{
    if (n == null) {
      console.log('not passed');
    } else if (n === nil) {
      console.log('nil passed');
    } else {
      console.log('value passed');
    }
  }
end

m
m(nil)
m(123)
```

Out:
```
not passed
nil passed
value passed
```

opal/corelib/range.rb
```diff
@@ -126,18 +146,55 @@ def size
   end
 
   def step(n = 1)
-    raise NotImplementedError
+    return enum_for(:step, n) unless block_given?
+    raise ArgumentError.new("step can't be negative") if n < 0
```
iliabylich: Could you use the same syntax `raise ErrorClass, 'error message'` everywhere?

opal/corelib/range.rb
```diff
@@ -126,18 +146,55 @@ def size
...
+    raise ArgumentError.new("step can't be 0") unless n > 0
+    if @begin.is_a?(Numeric) and @end.is_a?(Numeric)
```
iliabylich: Could you use `Numeric#===` here? `Object#is_a?` can be overwritten, the logic flow must be always defined by `Class#===`

opal/corelib/range.rb
```diff
@@ -126,18 +146,55 @@ def size
...
+        if @exclude
+          break if current >= @end
+        else
+          break if current > @end
+        end
+        yield(current)
+        i += 1
+      end
+    else
+      raise TypeError, "step must be an integer" unless n.is_a?(Numeric) && n.to_int == n
```
iliabylich: You can use `n = Opal.coerce_to(n, Integer, :to_int)` here (I guess you will get some passing specs in `spec/ruby` after this change)

opal/corelib/range.rb
taw: Done

iliabylich:
In general looks good, just a few styling issues. Hopefully travis will run next build :smile:

opal/corelib/range.rb
taw: For what it's worth MRI uses `rb_obj_is_kind_of ` and rubinius uses `kind_of?` checks here (`is_a?` and `kind_of?` are synonyms). Both of them can be overriden, this seems closer to semantics of other rubies, even if specs don't check it.

opal/corelib/range.rb
taw: Done

opal/corelib/range.rb
taw: Like this?

```
  def first(n=undefined)
    return @begin if `n === undefined`
    super
  end
```

The code I had here was standard Ruby hack (even if it looks ugly), I'm not totally sure when `undefined` can happen in Opal code.

iliabylich: `== null` would be enough here.

`undefined` is a special value of any optional argument (https://github.com/opal/opal/blob/master/lib/opal/nodes/args/optarg.rb#L17). It compiles to nothing, so your argument is `undefined` when you pass no value to your method.

``` js
function compiledOpalMethod(optional_argument) {
  console.log(optional_argument);
}

compiledOpalMethod();
// => undefined
```

taw: I'll just leave it up to you to change it to whichever form is most syntactically appropriate.

I know how `null/undefined` works in JS, and how `nil` and optional arguments work in Ruby, but I have no idea how they both interact in Opal.

iliabylich: `=== undefined` doesn't handle `null`. `== null` handle both

``` ruby
  def first(n=undefined)
    return @begin if `n == null`
    super
  end
```

taw: Is that going to be `null` or `undefined` if we do `(n=undefined)`. Is `undefined` the same as ``` `undefined` ```?

iliabylich: `(n = undefined)` produces uninitialized function argument, it is `undefined` (from JS world) if you don't pass any value.

You can use `undefined` (from Ruby) only in optional arguments. By using it inline you will get a method call :smile: 
``` ruby
$ opal --compile --no-opal -e "a = undefined"

return (a = self.$undefined())

```

taw: No that `n` can't be `null` ever when passed a legit ruby value (unless explicitly called with backticks like ```first(`null`)```)?

iliabylich: Any ruby variable may have `null` inside.

iliabylich:
:+1: Great job!

@opal/core Any comments?

@taw You also need to add a changelog entry (@elia always asks about it anyway :smile:) and make a rebase.

taw:
What do you mean by rebase here? The whole branch is forked from most recent `master`.

I should probably leave things like changelog to you guys, as every project does it differently, especially if you want anything more specific than:

* Fixed a bunch of things about `Range`

iliabylich:
@taw I was going to say "squash your commits please" :smile:

taw:
Squashed.

elia:
🎩 Great job indeed! 

I'll do the changelog routine, no problem.

PR #1487, opened 2016-06-23
Add support for complex and rational number literals.

iliabylich:


lib/opal/nodes/literal.rb
```diff
@@ -288,5 +288,29 @@ def compile_range_initialize
...
+
+    # 0b1111r -> s(:rational, (15/1))
+    # -0b1111r -> s(:rational, (-15/1))
+    class RationalNode < Base
+      handle :rational
+
+      children :value
+
+      def compile
+        push "Opal.Rational.$new(#{value.numerator}, #{value.denominator})"
```
iliabylich: Should it be `Opal.get('Rational')` to make it compatible with `opal/mini`? Well, not compatible. It should raise a proper error when `Rational` is not defined, right?

elia: Would be nicer but I'd rather keep it optimized for the happy path, which is using those literals when Rational is there.

Issue #1493, opened 2016-06-26
--map command line not working


h4ck3rm1k3:
--map and -m do not work
```
mdupont@mdupont-Aspire-7750G:~/experiments/opal$ ruby -I lib/ bin/opal --map
/home/mdupont/experiments/opal/lib/opal/cli.rb:64:in `initialize': unknown options: {:map=>true} (ArgumentError)
	from bin/opal:15:in `new'
	from bin/opal:15:in `<main>'
```

```
mdupont@mdupont-Aspire-7750G:~/experiments/opal$ ruby -I lib/ bin/opal -m   
/home/mdupont/experiments/opal/lib/opal/cli.rb:64:in `initialize': unknown options: {:map=>true} (ArgumentError)
	from bin/opal:15:in `new'
	from bin/opal:15:in `<main>'
```

elia:
is that on master?

h4ck3rm1k3:
yes, all on master. 
```
commit df31c8a1d2f1d0b441ca37ae7dbc6183d437bd04 (HEAD -> refs/heads/master, refs/remotes/origin/master, refs/remotes/origin/HEAD)
Author: Ilya Bylich <ibylich@gmail.com>
Date:   Thu Jun 23 20:15:42 2016 +0300

    Added support for complex and rational number literals. (#1487)
```

elia:
This is a genuine bug, some glue code between cli_options.rb and cli.rb is missing.

h4ck3rm1k3:
I dont know if this works... but it does not crash. how to verify?

h4ck3rm1k3:
I have gotten the map to be emitted, see pr, but it is not what I am looking for. I want annotated source with source lines or some line by line mapping.

elia:
fixed by https://github.com/opal/opal/commit/80d5938e9a8d657b18d9139f9328b41436e1649e and https://github.com/opal/opal/commit/80d5938e9a8d657b18d9139f9328b41436e1649e.

There's still room for improvement but at least it should be working now, thanks!

PR #1498, opened 2016-06-28
adding cli

h4ck3rm1k3:
https://github.com/opal/opal/issues/1493


h4ck3rm1k3:
now the map is being called, but the results make no sense.

elia:
fixed by https://github.com/opal/opal/commit/80d5938e9a8d657b18d9139f9328b41436e1649e and https://github.com/opal/opal/commit/80d5938e9a8d657b18d9139f9328b41436e1649e.

There's still room for improvement but at least it should be working now, thanks!

PR #1502, opened 2016-07-02
Source lines

h4ck3rm1k3:
This patch, and I don't expect it to be merged, contains the start of adding source lines to the output js that can be interpreted by emacs in ruby build mode as ruby eror logs

   Level: source:line:column at blah

this will help you understand where the code comes from. Next step is to create a map of the short names to the full path names and emit the absolute file names.


h4ck3rm1k3:
actually the debug lines are causing the javascript code to fail!

h4ck3rm1k3:
I am starting on a pure ruby stub for standard lib functions.

PR #1504, opened 2016-07-02
core compiles to coffescript without warnings

h4ck3rm1k3:
changes to core lib to remove js2coffee warnings.


meh:
The `null` checks are actually proper as they are, it's checking for `null` and `undefined` not just for `null`.

The other changes make the code less readable too, and I don't see why we should decrease readability to fix warnings for something like js2coffee.

h4ck3rm1k3:
ok, thanks for the feedback.

PR #1506, opened 2016-07-04
basic file open

h4ck3rm1k3:
start of a file open implementation for node.

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -1180,11 +1180,16 @@ module Kernel
   end
 
   # basic implementation of open, delegate to File.open
-  def open(*args, &block)
-    File.open(*args, &block)
-  end
-end
 
+  def open(*args, &block)
+    print "Going to do_open: #{args}\n"
+    %x{
+        Opal.open_file(#{args[0]},#{args[1]});
+      }
+  end
+  
+end
+                               
 class Object
   include Kernel
 end
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -92,6 +92,16 @@
     return unique_id;
   };
 
+  Opal.file_error = function (err, fd) {
+    console.log( "open failed" + err);
+  }
+
+  Opal.open_file = function(filename, flags) {
+    fs = require('fs');
+    console.log( "going to open " + filename);
+    return fs.open(filename, flags, Opal.file_error);
+  };
+
   // Table holds all class variables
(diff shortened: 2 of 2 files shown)


iliabylich:
node-specific code should be placed to https://github.com/opal/opal/tree/master/stdlib/nodejs

h4ck3rm1k3:
@iliabylich sounds good, so here I am adding in javascript to the core, how to do that for only node? If I don't put the open method into core_lib/kernel it crashes. Can you save me some time and please explain how this all fits together?

iliabylich:
By default opal contains only browser-specific and platform-independent code. `File` depends on the platform, so `corelib/file` has only stuff like `File#join` that can work on any platform and doesn't depend on IO/File interfaces/implementation.

`nodejs/file` has `File.open` - https://github.com/opal/opal/blob/master/stdlib/nodejs/file.rb#L80-L91, if you can't call it make sure you have `require 'opal/platform'`. This file contains a code that detects a platform and requires platform-specific implementations of `File`, `IO` and other classes.

h4ck3rm1k3:
Thanks, so by default calling file open creates an endless recursion. For the browser it would be better to define something that just fails it gracefully.

h4ck3rm1k3:
this also produces a endless loop : 
```
require 'opal/platform';
File.open("test");
```

iliabylich:
@h4ck3rm1k3 
``` sh
echo 'text' > /tmp/test
opal -rnodejs -e "File.open('/tmp/test', 'r') { |f| p f.read }"
# => 'text'
```

I was wrong, for a `nodejs` integration you should better require `nodejs`. :smile: 

@Mogztter `opal/platform` doesn't require `nodejs/file` for a node platform. You were working a lot on integrating opal with node, do you have any ideas why it's made this way? Does `opal/platform` require a minimal integration with `nodejs`?

h4ck3rm1k3:
```
require 'nodejs';
File.open("test",'a');
```
this works, thanks.

Mogztter:
@iliabylich To be honest I don't remember why we didn't require `nodejs/file` in `opal/platform`, most like an oversight.

> Does opal/platform require a minimal integration with nodejs?

I think @elia "favors" a minimal integration but IMO `file` is really useful in a Node environment.

elia:
> I think that @elia "favors" a minimal integration

@Mogztter That's true, but open for debate, not strict about that.
Feel free to open a new issue to track that.

PR #1511, opened 2016-07-06
Fix CLI -L option that was unusable

### LINKED ISSUE #1510: CLI error when requesting library only
When I try to get only the library code using the `opal` CLI, I get following error, even if I don't pass any other flag:

```
C:\Users\Bryan>opal -L -c
C:/Ruby22/lib/ruby/gems/2.2.0/gems/opal-0.10.0/lib/opal/cli.rb:64:in `initialize': can't accept evals or file in `library only` mode (ArgumentError)
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/opal-0.10.0/bin/opal:15:in `new'
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/opal-0.10.0/bin/opal:15:in `<top (required)>'
        from C:/Ruby22/bin/opal:22:in `load'
        from C:/Ruby22/bin/opal:22:in `<main>'

C:\Users\Bryan>echo %ERRORLEVEL%
1
```

I'm using Ruby 2.2.4p230 from RubyInstaller:

```
C:\Users\Bryan>ruby -v
ruby 2.2.4p230 (2015-12-16 revision 53155) [i386-mingw32]

C:\Users\Bryan>gem -v
2.6.6

C:\Users\Bryan>opal -v
Opal v0.10.0
```

[EDIT]
Trying to debug a bit, the issue is in [cli.rb](https://github.com/opal/opal/blob/master/bin/opal#L16) file, when requesting for `ARGV.file`, which is the [file](http://ruby-doc.org/core-2.0.0/ARGF.html#file-method) being run itself for that case!

elia:
fixes #1510

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,16 @@ Whitespace conventions:
 
 
 
+## [0.10.1] - 2016-07-06
+
+
+### Fixed
+
+- Fixed `-L` option for compiling requires as modules (#1510)
+
+
+
+
 ## [0.10.0] - 2016-07-04
 
 
--- a/bin/opal
+++ b/bin/opal
@@ -12,11 +12,13 @@ rescue OptionParser::InvalidOption => e
 end
 
 require 'opal/cli'
-cli = Opal::CLI.new options.options.merge(
+options_hash = options.options
+options_hash.merge!(
   file:     ARGF.file,
   filename: ARGF.filename,
   argv:     ARGV.dup
-)
+) unless options_hash[:lib_only]
+cli = Opal::CLI.new options_hash
 
 begin
   cli.run
--- a/lib/opal/cli_options.rb
+++ b/lib/opal/cli_options.rb
@@ -97,10 +97,6 @@ module Opal
         options[:no_exit] = true
       end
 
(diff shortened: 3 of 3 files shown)


iliabylich:
:+1:

bryanjhv:
It's still not fixed.

I suppose the `-L` switch needs to generate the Opal library code and print it (or save to a file), but for now what I'm getting is empty output:

```
C:\Users\Bryan>opal -L


```

Maybe the problem is the `:skip_opal_require` (the implicit `-O` switch)? Or it's meant to work like that?

elia:
@bryanjhv I agree that's debatable but I guess most of the time you'll want to use this to compile a lib for distribution and so made sense to me to ask for an explicit Opal require when they wanted that. 

Although it's also true that an error should be displayed if there's nothing to compile… (great first pull-request! HINT HINT 😄)

`C:\Users\Bryan>opal -L -r opal` should do the job

bryanjhv:
Oh, didn't know what `Kernel#exit` thing.

And yes, you're right with the approach you took, that way we can build `opal`, `opal-parser` and others just with one command.

PR #1512, opened 2016-07-06
Improve Numeric#step.

iliabylich:


elia:
You gotta love when the deleted lines form `ruby_specs` are more than the ones added to corelib 😄

PR #1513, opened 2016-07-07
Optimize Hash#merge! type check

jgaskins:
Rather than looking up the ancestor chain with Ruby, might as well utilize the browser's native prototype chain, which is a lot faster.


elia:
legit! 👍🏼

PR #1515, opened 2016-07-07
Allow passing dynamic symbols to undef/alias.

### LINKED ISSUE #1127: undef doesn't take a symbolized string

iliabylich:
Reincarnation of https://github.com/opal/opal/pull/1128 :smile: + fix for `alias :dynamic_symbol` + minor refactoring of `Compiler#returns`
Closes #1127
Closes #1128

### DIFF
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -340,23 +340,20 @@ module Opal
       return returns s(:nil) unless sexp
 
       case sexp.type
-      # Undefs go from 1 ruby undef a,b,c to multiple JS Opal.udef() calls, so need to treat them as individual statements
-      # and put the return on the last one
       when :undef
-        sexp.updated(nil,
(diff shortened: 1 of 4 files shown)


elia:
_upated description to close #1128 too when merged_

jgaskins:
Nice. Much cleaner than my implementation. :-)

Issue #1518, opened 2016-07-07
Opal Master Return with a begin...else is broken


wied03:
I'll try and work on a PR:

Note that in the `finally/$no_errors if`, the true used to have a return statement and now it doesn't.

Ruby:
```ruby
def match_unless_raises(expected_exception=Exception, &match_block)
  define_user_override(:matches?, match_block) do |actual|
    @actual = actual
    begin
      super(*actual_arg_for(match_block))
    rescue expected_exception => @rescued_exception
      false
    else
      true
    end
  end
end
```

Compiled with `opal -c --no-opal -A`

Opal 0.10:
```js
/* Generated by Opal 0.10.1 */
(function(Opal) {
  var TMP_1, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$define_user_override', '$actual_arg_for']);
  return (Opal.defn(Opal.Object, '$match_unless_raises', TMP_1 = function ːmatch_unless_raises(expected_exception) {
    var $a, $b, TMP_2, self = this, $iter = TMP_1.$$p, match_block = $iter || nil;

    if (expected_exception == null) {
      expected_exception = $scope.get('Exception');
    }
    TMP_1.$$p = null;
    var $arity = arguments.length; if ($arity > 1) { Opal.ac($arity, -1, this, "match_unless_raises"); }
    return ($a = ($b = self).$define_user_override, $a.$$p = (TMP_2 = function(actual){var self = TMP_2.$$s || this, $c, $d;
if (actual == null) actual = nil;
      if (TMP_2.$$is_lambda || TMP_2.$$define_meth) {
        var $arity = arguments.length;
        if ($arity !== 1) { Opal.block_ac($arity, 1, 'match_unless_raises'); }
      }
    self.actual = actual;
      
      var $no_errors = true;try {
        return ($c = ($d = self, Opal.find_iter_super_dispatcher(self, 'match_unless_raises', (TMP_2.$$def || TMP_1), false, false)), $c.$$p = null, $c).apply($d, Opal.to_a(self.$actual_arg_for(match_block)))
      } catch ($err) {
        $no_errors = false;
        if (Opal.rescue($err, [expected_exception])) {self.rescued_exception = $err;
          try {
            return false
          } finally { Opal.pop_exception() }
        } else { throw $err; }
      }finally {
        if ($no_errors) { 
          return (function() {
            return true
          })();
[...]

PR #1519, opened 2016-07-07
Fix begin..else issue

wied03:
Update to latest ruby specs which test for this
Wasn't broken for blocks, only methods
Exclude new ruby specs that fail

https://github.com/opal/opal/issues/1518

### DIFF
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -68,7 +68,7 @@ module Opal
 
       def rescue_else_code
         rescue_else_code = scope.rescue_else_sexp
-        rescue_else_code = compiler.returns(rescue_else_code) unless stmt?
+        rescue_else_code = compiler.returns(rescue_else_code)
         rescue_else_code
       end
     end
@@ -140,7 +140,7 @@ module Opal
 
       def rescue_else_code
         rescue_else_code = scope.rescue_else_sexp
-        rescue_else_code = compiler.returns(rescue_else_code) unless stmt?
+        rescue_else_code = compiler.returns(rescue_else_code)
         rescue_else_code
       end
 
--- a/lib/opal/nodes/rescue.rb
+++ b/lib/opal/nodes/rescue.rb
@@ -2,7 +2,13 @@ require 'opal/nodes/base'
 
 module Opal
   module Nodes
-    class EnsureNode < Base
+    class BaseRescueEnsure < Base
+      def rescue_else_code
+        compiler.returns(scope.rescue_else_sexp)
+      end
+    end
+
+    class EnsureNode < BaseRescueEnsure
       handle :ensure
 
       children :begn, :ensr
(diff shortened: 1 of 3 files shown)


wied03:
@iliabylich - I looked and the old/pre-AST code didn't seem to check `stmt?` when it called `compiler.returns` for the rescue else code. Everything, including new specs, passed when I took it out. Do you see any reason why we can't remove it?

iliabylich:
@wied03 The bug is in the `Compiler#returns`. For `:rescue` node it should wrap `else_sexp` with `returns()`. `begin; rescue; else` can be a statemnt, in this case we don't need to return `else`-value.

Try to compile the following code, looks like it doesn't work as expected with your patch:
``` ruby
begin
  1
rescue
  2
else
  3
end

:returning_value
```

Only `:expr` and `:recv` (i.e. not `:stmt`) begin-rescue-else nodes must return their `else`-values.

I guess the bug was introduced in #1465, AST was changed from `s(:rescue, s(:resbody, body_code))` to `s(:rescue, body_code)` and we missed it.

wied03:
Perhaps, but how is the bug in compiler returns? The reason this didn't work in the first place is because compiler returns isn't called due to the guard condition. I don't know enough about the stmt piece here.

wied03:
And if this PR leaves things in a broken state, we need more rubyspecs :)

wied03:
But I do see that we're returning `3` with the PR instead of `returning_value`. I'll add some more rubyspecs and then can resume working on this.

iliabylich:
@wied03 
``` ruby
obj.method(value)
```
`obj` here is a `:recv` (receiver)
`value` is an `:expr` (expression)
`object.value` is a `:stmt` (statement)

It could be possible to use only `:expr` and `:stmt`, but JS is JS.

There are only 3 places where Opal needs `:recv`:
+ numbers (to prevent generating a code like `1.toString()` - should be `(1).toString()`)
+ x-string (because it can contain absolutely everything)
+ dynamic string (same as x-string)

`:stmt` is the simplest level, this is the default level of compilation for each node. `:expr` and `:recv` are mostly used to wrap the code with parentheses and anonymous functions + JS returns.

In your case when `begin; rescue; else; end` is a statement there's no need to put `js_return` to the AST because nothing should be returned.
But when it's not a statement (i.e. receiver or expression) it must return rescue-value or else-value + sometimes it must be wrapped with anonymous function.

wied03:
@iliabylich - In your last paragraph, are you disputing the bug that prompted this PR or just explaining?

iliabylich:
Just saying what I accidentally did wrong in https://github.com/opal/opal/pull/1465 :smile:
``` ruby
# Compiler#returns
when :rescue
  body_sexp, *resbodies, else_sexp = sexp.children

  resbodies = resbodies.map do |resbody|
    returns(resbody)
  end

  sexp.updated(nil, [
    returns(body_sexp),
    *resbodies,
    returns(else_sexp)
  ])
```

wied03:
The reason we have a bug though is this IS a statement AND it needs the return statement.

wied03:
Since @iliabylich decided to fix this on their own, I'll close this.

PR #1523, opened 2016-07-09
Throw error when lib_only and no libraries

### LINKED ISSUE #1521: Display an error when -L is used from the CLI but no require is provided
_see https://github.com/opal/opal/pull/1511#issuecomment-231018048_

Currently the CLI just outputs nothing.

bryanjhv:
Fixes #1521 and trying to end thread on #1510

### DIFF
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -60,6 +60,7 @@ module Opal
         end.compact.flatten
       ]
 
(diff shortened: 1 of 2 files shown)


elia:
Thank you!

bryanjhv:
You're welcome.

Now, there's a problem. In the tests I added (and modified), there's a need to use `requires: ['opal']` explicitly, which may look a bit ugly than using `skip_opal_require`.

PR #1525, opened 2016-07-10
update browser opal-rspec example

tongueroo:
Following the docs and couldn't the In a Browser example working without creating a fake env `sprockets_env = Opal::RSpec::SprocketsEnvironment.new` first.

### DIFF
--- a/docs/rspec.md
+++ b/docs/rspec.md
@@ -51,7 +51,8 @@ $ bundle exec rake
 require 'bundler'
 Bundler.require
 
-run Opal::Server.new { |s|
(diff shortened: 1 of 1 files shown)


elia:
Thanks!

Although in the end it's probably worth replacing the whole guide with a ref to the opal-rspec readme.

cc @wied03

tongueroo:
Ah that makes a lot of sense. Thanks!

wied03:
@elia - 6 month backlog - https://github.com/opal/opal.github.io/pull/36 :)

Issue #1520, opened 2016-07-08
Setter Return Value


wied03:
Should ignore the return value from the method, worked OK in Opal 0.10.1. Seems to not happen if it's a standalone method, had to put it in a class.

```ruby
class Foo
  def foobar=(*)
    1
  end
end

puts (Foo.new.foobar = 'bar')
```

Just ran with `opal foo.rb`

MRI output: `bar`
0.10.1 output: `bar`
Opal master output: `1`

wied03:
compiled code, master:

```js
return self.$puts($scope.get('Foo').$new()['$foobar=']("bar"));
```

0.10:
```js
return self.$puts((($a = ["bar"]), $b = $scope.get('Foo').$new(), $b['$foobar='].apply($b, $a), $a[$a.length-1]));
```

elia:
Reopening as it's just _partially_ closed by #1528

iliabylich:
`[]=` method doesn't work properly (i.e. `a[0] = 1` must return `1`). All other things (including mass-assignment,  I was wrong in [this comment](https://github.com/opal/opal/pull/1528/files#diff-94fb9a69ec1d65114a3fd37d37b81b79R44)) are working on the latest master.

To fix it:
+ [this check](https://github.com/opal/opal/blob/master/lib/opal/rewriters/explicit_writer_return.rb#L15) must be changed to also modify a `:[]=` method call
+ the returning value must be `arguments[-1]` because `[]=` takes two arguments (it also covers a simple case of `self.method = value`)

wied03:
@iliabylich - Added https://github.com/ruby/spec/pull/272

wied03:
This is now in rubyspecs, commit 0a65ffe

iliabylich:
What should be a returning value of `a.JS.b = 1`? And `a.JS[:b] = 1`?

elia:
`a.JS.b = 1` compiles to invalid code in 0.10, and anyway I'm not sure what behavior it can have:

- `a.b=(1)` which is what currently compiles to is invalid
- `a['b='](1)` doen't sound very useful when interoperating with JS
- `a.b = 1` seems useful but confuses things as `a.JS.b` is a method call and the proper property assigment is `a[:b] = 1`
- raising an error during compilation seems a better choice to me

-- 

As for `a.JS[:b] = 1` I think it should just leave it to javascript, I don't think it makes sense to have a translation for, say, `a.JS[:b] = *[1,2,3]` and friends

[...]

PR #1527, opened 2016-07-11
Add opal-rspec smoke test to aid in finding rubyspec coverage gaps

wied03:


wied03:
@elia - Not sure what I think about the allow_failure part. If we start ignoring it, then this is pointless. You mentioned keeping an eye on it though.

wied03:
Latest build fails "properly". Depends on #1518 and #1520 regressions being fixed

elia:
👍🏼 you did it finally! 🎉

PR #1529, opened 2016-07-12
Add ruby/language/constants_spec to the test suite.

iliabylich:


iliabylich:
Two things I'm worrying about:
+ performance. Time takes for a whole rubyspec test suite on the master branch: 57s. Patch: 60s. @vais @elia Are there any benchmarks for a constant lookup on the chain of nested modules?
+ namings :smile: Any ideas how to rename these `Opal.const_get_multi`?

vais:
@iliabylich there does not seem to be benchmarks specific to nested modules constant lookup [here](https://github.com/ruby/ruby/tree/trunk/benchmark) or [here](https://github.com/jruby/jruby/tree/master/bench)... [this](https://github.com/rubinius/rubinius-benchmark) may be worth exploring (there's a bunch of stuff there), but no guarantee you'll find what you're looking for. Good luck 👍

elia:
> Any ideas how to rename these Opal.const_get_multi?

@iliabylich Was about to suggest `Opal.get_many` but then realized we now have both `Opal.get` & `Opal.const_get`, do they do the same thing?

iliabylich:
`Opal.get` tries to find a constant directly on `self` / `Opal.const_get` doesn't raise `NameError`

iliabylich:
How about merging everything to a single `Opal.get` function that:
+ takes also a `raise` parameter that indicates whether we need to throw an error if nothing was found
+ takes an array of scopes instead of a single scope

elia:
@iliabylich makes sense to me

Mogztter:
> How about merging everything to a single Opal.get function

In Scala we are using the word `try` to differentiate methods:

 * `Opal.get()` will throw an exception if nothing was found
 * `Opal.tryGet()` will *not* throw an exception if nothing was found

But I'm fine with a single method and a `raise` parameter :+1:

iliabylich:
+ Moved all logic to `Opal.const_get`
+ `Opal.get` is a one line function that invokes `Opal.const_get`  - mostly for a backward compatibility, like `Opal.A.$$scope.get("B")`

Added benchmarks:
+ `bm_module_definition_small` - 15 nested modules
+ `bm_module_definition_big` - 200 nested modules
+ `bm_constant_lookup_small` - 15 nested modules, each defines a constant, each takes a constant from toplevel
+ `bm_constant_lookup_big` - same but with 200 modules / 200 constants


```
$ BMFILES="benchmark/bm_module_def*,benchmark/bm_constant_lookup*" && rake bench:clear || true && gco master && rake bench:opal[$BMFILES] && gco fix-constant-lookup && rake bench:opal[$BMFILES] && rake bench:report

Benchmark                                Opal2  Opal1
benchmark/bm_module_definition_big.rb    0.015  0.013
benchmark/bm_module_definition_small.rb  0.001  0.001
benchmark/bm_constant_lookup_big.rb      0.110  0.008
benchmark/bm_constant_lookup_small.rb    0.001  0.001
```

Opal1 = master, Opal2 = patch. In a real world (< 10 nested modules) there's almost no difference (`bm_module_definition_small` / `bm_constant_lookup_small` contain 15 nested modules). For 200 modules, of course,  there's a serious degradation (since iterate over 200 scopes).

Is there anything else that I should measure?

opal/corelib/runtime.js
```diff
@@ -124,21 +124,139 @@
...
+  // class/module's `#const_missing` is called, which by default will raise an
   // error.
   //
   // @param name [String] the name of the constant to lookup
   // @return [Object]
   //
-  Opal.get = function(name) {
-    var constant = this[name];
+  Opal.get = function(const_name) {
+    return Opal.const_get([this], const_name, true, true);
```
elia: I think using `Opal` instead of `this` would be better. 

I don't have any proof but I sense it's easier to inline something for JS runtimes if it has no `this` in the function.

opal/corelib/runtime.js
```diff
@@ -124,21 +124,139 @@
...
+  //
+  //     Opal.const_get(scopes, const_name, inherit, true)
+  //
+  // @param scopes [Array<$$scope>] a list of scopes
+  // @param const_name [String] the name of the constant
+  // @param inherit [Boolean] flag to perform a search in the parents
+  // @param raise [Boolean] flag to trigger "const_missing" if nothing was found
+  // @return [Object]
+  Opal.const_get = function(scopes, const_name, inherit, raise) {
+    function getSingle(scope, const_name, inherit) {
```
elia: @iliabylich Can any of these functions be moved outside to avoid multiple evaluation?

(Please correct me if you already know that it doesn't affect perfs!)

opal/corelib/runtime.js
iliabylich: This method can be used on any `$$scope`, i.e. :
``` js
Opal.get('String').$$scope.get('Wrapper').$name()
=> "String::Wrapper"
```
`this` is a `$$scope`, so we can't avoid it (and I guess we shouldn't)

opal/corelib/runtime.js
iliabylich: Done. These functions are optimized by V8, but my simple benchmark shows that defining inline functions dynamically is ~2x slower.

opal/corelib/runtime.js
elia: My bad, forgot that `Opal` is also the prototype for scopes 😅

opal/corelib/runtime.js
elia: TMYK, that means probably that some of the bm above changed for good! 🏎

PR #1531, opened 2016-07-13
Extact detecting reserved JS var names and props to a separate AST rewriter.

iliabylich:


wied03:
@iliabylich  - Is this for #1520 ? Are you going to update the rubyspec submodule per my comment there?

elia:
@iliabylich I have just one doubt, and it's about throwing away knowledge of the original name, any thoughts on that?

iliabylich:
@elia From what I see source maps still use the original name of the variable (and compiled JS contains a var name is `$` suffix). It happens because we don't create a new AST node, we [update an existing one](https://github.com/iliabylich/opal/blob/28159524c4fdf58d98752b0a9c751670c7bd4419/lib/opal/rewriters/js_reserved_words.rb#L51).

@wied03 Not really, this PR is just a refactoring. I'd like to extract as many common things from `opal/compiler` and `opal/nodes` as possible (hopefully @elia has nothing against it :smile: ). I have a fix for #1520 locally but it doesn't handle `.JS.` syntax. To fix it we need to extract `.JS.` stuff to a rewriter too (have it also locally, need to combine two commits).

PR #1532, opened 2016-07-13
Add safe navigator support.

iliabylich:
@eregon Rubyspec doesn't have any specs for a safe navigator, right? [This issue](https://github.com/ruby/spec/issues/175) says that its WIP. I've added some specs in this PR, should I send also a PR to ruby/spec?

@opal/core Two questions:
+ Currently safe navigator handles only `nil`. Is it a good idea to handle also `undefined` and `null`?
+ Do we need a support for `.JS.` syntax? And how it may look like? `.JS&.js_method` ?


eregon:
@iliabylich Indeed, it seems they are no specs for it yet. Yes, it would be very useful to make a PR with the specs to ruby/spec, and then it can easily be updated in opal.

elia:
@iliabylich supporting `undefined` and `null` sounds like a good idea, as for the syntax the one you proposed makes sense, but to me it's ok to handle `.JS` calls in a separate PR.

cc @jeremyevans

jeremyevans:
I agree that handling `undefined` and `null` would probably be best.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Whitespace conventions:
     * `Hash#fetch_values`
     * `Hash#to_proc`
     * `Struct#dig`
+- Added save navigator (`&.`) support. (#1532)
```
dougo: Typo, should be "safe".

iliabylich:
MRI has some strange behavior with `obj&.m += value`. Looks like it doesn't use safe navigator in this case:
``` ruby
class TestClass
  attr_accessor :m

  def initialize
    @m = 0
  end
end

obj = TestClass.new

puts RUBY_DESCRIPTION

obj&.m += 1
p obj.m

obj&.n += 1
```

Output:
```
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
1
test.rb:16:in `<main>': undefined method `n' for #<TestClass:0x00000001280fe0 @m=1> (NoMethodError)
```

However ripper parses it:
``` ruby
require 'ripper'
require 'pp'

pp Ripper.sexp("obj&.m += 1")
```
Output:
``` ruby
[:program,
 [[:opassign,
   [:field, [:vcall, [:@ident, "obj", [1, 0]]], :"&.", [:@ident, "m", [1, 5]]],
   [:@op, "+=", [1, 7]],
   [:@int, "1", [1, 10]]]]]
```

`parser` can't parse it because of [this condition](https://github.com/whitequark/parser/blob/master/lib/parser/builders/default.rb#L490), `csend` is missing . @whitequark is it something that should be fixed? And keeping in mind that MRI doesn't call safe navigator in this case which AST should be created? `csend` or `send`?

@eregon Do you know anything about it? Should I add some tests for these cases to rubyspec?

iliabylich:
@elia Added support for `null` and `undefined`. I'd like to add a support for `.JS.` syntax in a separate PR :smile:

eregon:
@iliabylich The safe navigation operator only do sth different if the receiver is nil, isn't it?
It's not about whether the method is defined or not AFAIK.
```ruby
pry(main)> nil.to_i
=> 0
pry(main)> nil&.to_i
=> nil # So even if the method is defined it doesn't matter, receiver is nil.
```

eregon:
BTW, we [are implementing](https://github.com/jruby/jruby/commit/e9625516d842e4821501b2d6b91d6c479000fc34) this in JRuby+Truffle.
Indeed the special op assign case needs some care, because desugaring
`a.b += c` in `a.b = a.b + c` should not end up in "No method + defined on nil".

iliabylich:
I am dumb :smile: need to think twice before submitting comments.
Well, it works as expected on MRI.
@whitequark Then it's definitely a bug, I'll send a PR
This part is unsupported by Opal for now.

eregon:
> I am dumb :smile:

No worry, the semantics of &. were also discussed with various designs inspired by Rails' `try`.

iliabylich:
Anyway, PR is ready for merging if everyone is ok with this implementation

elia:
w00t! 👏

Issue #1536, opened 2016-07-20
Shared state between tests


Mogztter:
> @iliabylich: Opal runs all specs in the context of single object, this static object.
Reference: https://github.com/ruby/spec/issues/275

In our case:
 * When `rescue_spec` runs before `return_spec` everything is fine.
 * When `return_spec` runs before `rescue_spec` this test fails.

I think we need to create an `@env` object per spec file.
Currently we have a single static instance defined in [mspec-opal/runner.rb](https://github.com/opal/opal/blob/master/spec/mspec-opal/runner.rb#L87-L88).

PR #1537, opened 2016-07-20
Sort specs files to be consistent

Mogztter:
Currently tests are sharing a global state `@env`.
Sorting specs files is important to make errors reproducible by everyone.

Ref #1536

### DIFF
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -58,14 +58,14 @@ module MSpecSuite
 
     # Filters must be added first
     suite_filters = suite == 'opal' ? opalspec_filters : rubyspec_filters
-    add_specs["#{suite} filters", suite_filters]
+    add_specs["#{suite} filters", suite_filters.sort]
 
     if pattern
-      add_specs["PATTERN=#{pattern}", userspecs]
+      add_specs["PATTERN=#{pattern}", userspecs.sort]
     elsif suite == 'opal'
-      add_specs['spec/opal', opalspecs]
+      add_specs['spec/opal', opalspecs.sort]
     elsif suite == 'ruby'
-      add_specs['spec/ruby', rubyspecs]
+      add_specs['spec/ruby', rubyspecs.sort]
     else
       warn 'Please provide at lease one of the following environment variables:'
       warn 'PATTERN # e.g. PATTERN=spec/ruby/core/numeric/**_spec.rb'


iliabylich:
> Sorting specs files is important to make errors reproducible by everyone

But it also hides possible errors. Currently when you run specs in the alphabet order you get no errors. But when you run it in the reverse order you get an error in `rescue_spec`.

I thought it's a good practice to run specs in the random order.

MSpec has on option called [randomize](https://github.com/ruby/mspec/blob/master/lib/mspec/runner/mspec.rb#L226-L232) that [shuffles](https://github.com/ruby/mspec/blob/master/lib/mspec/runner/mspec.rb#L51) files, I think it makes sense to enable it to catch possible conflicts between specs.

Mogztter:
You are right but I don't like random failures because it's pretty hard to reproduce and make the build unstable.
I would rather prefer a consistent behavior for every developers.

Catching bugs between specs is nice but specs are mainly designed to catch bugs in the "production" code not in the "testing" code :bug:

iliabylich:
:+1:

+ If we want to run a test suite in a normal (i.e. sequential) order, we simply run it as is. The order can't depend on the platform or locale(?), this is wrong.
+ If we want to a random order, we add `MSpec.randomize(true)` to the `spec_helper` by setting some environment variable.

@elia Any thoughts?

tasks/testing.rake
```diff
@@ -58,14 +58,14 @@ module MSpecSuite
 
     # Filters must be added first
     suite_filters = suite == 'opal' ? opalspec_filters : rubyspec_filters
-    add_specs["#{suite} filters", suite_filters]
+    add_specs["#{suite} filters", suite_filters.sort]
```
iliabylich: By the way, there's no need to sort spec filters, they can't conflict

elia:
Ideally we should add a shuffle after seeding srand with a number we also print out, that should make the random order reproducible, but still random. 

```rb
> srand 123; (1..10).to_a.shuffle
=> [5, 1, 8, 6, 9, 4, 2, 7, 10, 3]
> srand 123; (1..10).to_a.shuffle
=> [5, 1, 8, 6, 9, 4, 2, 7, 10, 3]
> srand 123; (1..10).to_a.shuffle
=> [5, 1, 8, 6, 9, 4, 2, 7, 10, 3]
```

Mogztter:
> Ideally we should add a shuffle after seeding srand with a number we also print out, that should make the random order reproducible, but still random.

Yes that's a good idea! Reproducible is the key :smile:

elia:
> Yes that's a good idea! Reproducible is the key 😄

To be fair, it's RSpec team's idea💡 😄

elia:
sure thing

PR #1538, opened 2016-07-20
MSpec intergation: run every spec file in a separate environment to avoid potential conflicts.

### LINKED ISSUE #1536: Shared state between tests

iliabylich:
Closes #1536

The idea of this PR is to run every file in a separate environment, so dynamically defined methods will not conflict. This can be achieved using a default MSpec behavior. In fact, this PR doesn't introduce any new mechanism.

+ it uses `MSpec.register_files` to populate MSpec's configuration.
+ it uses `MSpec.files` to run every single file in it's own environment
+ `MSpec.files` dynamically loads every file, which is not allowed by Opal. To make it working added `requirable_spec_file` special call (available only in the MSpec test suite) that explicitly adds passed filepath to compiler's list of required files. By doing this we can automatically inject spec's code to the generated file and require it later (when MSpec needs it).

New `mspec_nodejs.rb` looks like:
``` ruby
require 'spec_helper'
require 'opal/full'

# ... conditionally injected code for benchmarks ...

require 'filters/unsupported/...'
# ... other files from 'unsupported' and 'bugs' directories ...

requirable_spec_file 'ruby/language/lambda_spec.rb'
# ... other files from rubyspec/opal suite ...

MSpec.register_files [
  'ruby/language/lambda_spec.rb',
  # ... other files from rubyspec/opal suite ...
]

OSpecRunner.main.start_suite
OSpecFilter.main.unused_filters_message(list: false)
OSpecRunner.main.did_finish
exit MSpec.exit_code
```

@eregon Is it closer to how it should be? :smile: 
@Mogztter This branch is in the Opal's repo. Could you test it please?

### DIFF
--- a/spec/mspec-opal/runner.rb
+++ b/spec/mspec-opal/runner.rb
@@ -82,13 +82,6 @@ class Object
   end
 end
 
-module MSpec
-  def self.opal_runner
-    @env = Object.new
-    @env.extend MSpec
-  end
-end
-
 class OSpecFormatter
   def self.main
     @main ||= self.new
@@ -112,35 +105,18 @@ class OSpecFormatter
   end
 end
 
-class OSpecRunner
+class OpalBM
   def self.main
     @main ||= self.new
   end
 
-  def initialize
-    register
-    run
-  end
-
-  def register
-    OSpecFormatter.main.register
-    OSpecFilter.main.register
-  end
-
-  def run
-    MSpec.opal_runner
-  end
-
-  def will_start
-    MSpec.actions :start
-  end
-
-  def bm!(repeat, bm_filepath)
+  def register(repeat, bm_filepath)
     `self.bm = {}`
     `self.bm_filepath = bm_filepath`
     MSpec.repeat = repeat
     MSpec.register :before, self
     MSpec.register :after,  self
+    MSpec.register :finish, self
   end
 
   def before(state = nil)
@@ -159,7 +135,7 @@ class OSpecRunner
     }
   end
 
-  def did_finish
+  def finish
     %x{
       var obj = self.bm, key, val, report = '';
       if (obj) {
@@ -172,7 +148,6 @@ class OSpecRunner
         require('fs').writeFileSync(self.bm_filepath, report);
       }
     }
-    MSpec.actions :finish
   end
 end
 
@@ -191,3 +166,6 @@ module OutputSilencer
     end
   end
 end
+
+OSpecFormatter.main.register
+OSpecFilter.main.register
--- a/spec/opal/core/kernel/respond_to_spec.rb
+++ b/spec/opal/core/kernel/respond_to_spec.rb
@@ -12,6 +12,9 @@ class RespondToSpecs
   def some_method
     :foo
   end
+
+  def foo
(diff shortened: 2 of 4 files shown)


Mogztter:
@iliabylich Sure thanks for your work! Once this is merged we should definitely sort specs files. Otherwise I really like your idea about activating `MSpec.randomize(true)` on an environment variable to catch possible conflicts between specs :+1:

eregon:
Looks good, and it's now more similar to the [original code](https://github.com/ruby/mspec/blob/master/lib/mspec/commands/mspec-run.rb#L88-L93) of MSpec's run command!

`MSpec.process` is https://github.com/ruby/mspec/blob/master/lib/mspec/runner/mspec.rb#L42-L46.
Using it and making OSpecRunner use a `finish` hook looks like it could avoid duplication even further :)

Mogztter:
> This branch is in the Opal's repo. Could you test it please?

@iliabylich I can confirm that this is working great!

Mogztter:
@iliabylich Since I've created conflicts with you branch I can rebase on master if you want :wink:

iliabylich:
Updated:
``` ruby
require 'spec_helper'
require 'opal/full'

# bm code
# require filters code

requirable_spec_file 'ruby/core/module/to_s_spec.rb'

MSpec.register_files [
  'ruby/core/module/to_s_spec.rb'
]

srand(32152)
MSpec.randomize(true)

MSpec.process
OSpecFilter.main.unused_filters_message(list: false)
exit MSpec.exit_code
```

Randomization is enabled by default. `testing.rake` takes ENV variable `RANDOM_SEED` or manually generates it and injects into `mspec_nodejs.rb`. Also `rake mspec_#{ruby,opal}_#{platform}` prints this value before compiling.

atm Travis is lagging, I hope we will see it in action soon.

elia:
@iliabylich I see this error in Travis:

```
Kernel#respond_to? returns true if a method exists FAILED
Expected false to be true
```

_form `rake mspec_opal_nodejs`_

iliabylich:
@elia There were [two](https://github.com/opal/opal/blob/master/spec/opal/core/kernel/respond_to_spec.rb#L29-L31) [conflicting](https://github.com/opal/opal/blob/master/spec/opal/core/kernel/respond_to_spec.rb#L37-L40) specs in `spec/opal/core/kernel/respond_to_spec.rb`. It was working like:
``` ruby
class A
  def do
  end
end

a = A.new
`a.$do.$$stub = true`

b = A.new
puts `b.$do.$$stub`
```

When we run them in the direct order it's fine, but the second test makes the first one broken.
I've changed these tests to use different methods.

PR #1551, opened 2016-08-13
Resolves #1550, define the default open mode on the File constructor

### LINKED ISSUE #1550: flags can be undefined when creating a new File
It's possible to instantiate a new File without a `flags` argument.
In this case the following error will be thrown at runtime (Node.js):

```
    TypeError: Cannot read property '$match' of undefined
  Stack:
    TypeError: Cannot read property '$match' of undefined
        at $File.ːinitialize (/home/guillaume/workspace/opensource/asciidoctor.js/node_modules/opal-runtime/src/nodejs.js:218:23)
        at Opal.defn.TMP_4 [as $new] (/home/guillaume/workspace/opensource/asciidoctor.js/node_modules/opal-runtime/src/opal.js:3140:23)
        at module_constructor.ːload_file (/home/guillaume/workspace/opensource/asciidoctor.js/build/npm/asciidoctor-core.js:12355:44)
        at Object.<anonymous> (/home/guillaume/workspace/opensource/asciidoctor.js/spec/npm/asciidoctor-all.spec.js:20:29)
```

The reason is that we are using `match` method on a undefined argument: https://github.com/opal/opal/blob/master/stdlib/nodejs/file.rb#L108-L112

Mogztter:
Also adds tests for Node.js on File.new and File.mtime

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -101,7 +101,7 @@ class File < IO
 
   # Instance Methods
 
-  def initialize(path, flags)
+  def initialize(path, flags = 'r')
     binary_flag_regexp = /b/
     encoding_flag_regexp = /:(.*)/
     # binary flag is unsupported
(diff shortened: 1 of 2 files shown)


elia:
👍🏼 :)

PR #1556, opened 2016-08-30
Evaluate dynamic method args before assigning block to the method body.

### LINKED ISSUE #1173: Calling a method that takes a block after just calling another method that takes a block blows up
When a method executes a block, unless that method's return value is stored in a local variable first, it cannot be passed to another method as a parameter.

```ruby
def do_ival(something, &block)
  puts "about to call block #{something}"
  block.call
end

def method1
  do_ival(1) do
    puts 'block 1'
    :foo
  end
end

do_ival(method1) do
  puts 'block 2'
end

# this approach works on Opal
# m = method1
# do_ival(m) do
#   puts 'block 2'
# end
```
Output:
```
root@default:/host/opal# bundle exec opal ../bah.rb 
about to call block 1
block 1
about to call block foo

/tmp/opal-nodejs-runner-20151031-75-refeve:4219
      throw exception;
            ^
call: undefined method `call' for nil
    at OpalClass.Opal.defs.self [as $new] (/tmp/opal-nodejs-runner-20151031-75-refeve:4436:15)
    at NilClass.Opal.defn.TMP_1 (/tmp/opal-nodejs-runner-20151031-75-refeve:3032:54)
    at NilClass.method_missing_stub [as $call] (/tmp/opal-nodejs-runner-20151031-75-refeve:886:35)
    at Object.Opal.defn.TMP_1 [as $do_ival] (/tmp/opal-nodejs-runner-20151031-75-refeve:18884:18)
    at Opal.dynamic_require_severity (/tmp/opal-nodejs-runner-20151031-75-refeve:18896:64)
    at Object.<anonymous> (/tmp/opal-nodejs-runner-20151031-75-refeve:18897:3)
    at Function.Module._load (module.js:312:12)

root@default:/host/opal# ruby ../bah.rb 
about to call block 1
block 1
about to call block foo
block 2
```

### LINKED ISSUE #1176: Can't yield a block in initialize
I think this a bug, pass a block to initialize but can't yield. `block_given?` is also `false`

```ruby
class A
  def initialize(opts)
    yield "OK"
  end
end

x = Native(`{a: {b: "asd"}}`)

A.new(x.a.b) do |s|
  puts s
end
```

Throws exception `Uncaught LocalJumpError: no block given`

And little work around:

```ruby
class A
  def initialize(opts)
    yield "OK"
  end
end

x = Native(`{a: {b: "asd"}}`)

c = x.a.b

A.new(c) do |s|
  puts s
end
```

### LINKED ISSUE #1181: Redesign block and metadata passing.
Right now blocks are handled with a property on the `Function` object that expects a block, `$$p` is the block, and `$$s` is the `self` value to use.

This can cause very weird bugs in case the special properties are cleared or set incorrectly.

I'm pretty sure the following bugs are caused by the current design: #1176, #1173, #811.

Just fix bugs
-------------
We could litter the whole codebase with try/catch blocks to properly reset or set to the previous values all the block calls, this is going to have a huge performance hit.

Make `$$p` and `$$s` a stack
----------------------------
If `$$p` and `$$s` were a stack, the prelude for the block could pop the values that have been set from the caller, thus removing the problem of resetting the values and no need for try/catch because the handling would be before any exceptions can happen.

This has the downside of keeping around a lot of values that might be useless, and if the call stack gets deep enough the block stack could get fairly big.

I'm also not sure this would actually work properly, because if every block in the chain automatically pops and there was no argument set, then you'd find yourself in even a worse situation.

Pass metadata as last parameter
-------------------------------
Alternatively the `&block`, `self` and other metadata like `$$eval` in the case of `class_eval` and friends could be passed as last parameter to blocks, methods and whatnot.

The downside here is that the prelude will get a little more complex, because in the case of `*args` it will have to ignore the last argument, and in case of optional parameters it will have to check all of them to reset the default value and move the block where it has to be.

In practice the downside shouldn't be too big, because we're already generating a prelude to set the default value, so it's just another check.

### LINKED ISSUE #1499: obscure case of block params getting whacked
Try this:
```ruby
class Foo 
  def self.register(object, name, &block)
    puts "defining method #{name} block = #{block}"
    define_method name do 
      puts "calling #{name}: object = #{object} block = #{block}"
      block.call if block
    end
  end
end

class BreakFoo
  def initialize(name)
    @klass = Class.new(Foo)
    @klass.class_eval { register("breaker", :breaker) { puts "say hi" } }
  end
end

class Baz < Foo
  register(BreakFoo.new, :doit) { puts "another block" }
end

class Works < Foo 
  @x = BreakFoo.new
  register(@x, :doit2) { puts "working block" }
end

Baz.new.doit
Works.new.doit2
```

(try it [here](http://opalrb.org/try/?code:class%20Foo%20%0A%20%20%0A%20%20def%20self.register(object%2C%20name%2C%20%26block)%0A%20%20%20%20puts%20%22defining%20method%20%23%7Bname%7D%20block%20%3D%20%23%7Bblock%7D%22%0A%20%20%20%20define_method%20name%20do%20%0A%20%20%20%20%20%20puts%20%22calling%20%23%7Bname%7D%3A%20object%20%3D%20%23%7Bobject%7D%20block%20%3D%20%23%7Bblock%7D%22%0A%20%20%20%20%20%20block.call%20if%20block%0A%20%20%20%20end%0A%20%20end%0A%20%20%0Aend%0A%0Aclass%20BreakFoo%0A%20%20%0A%20%20def%20initialize(name)%0A%20%20%20%20%40klass%20%3D%20Class.new(Foo)%0A%20%20%20%20%40klass.class_eval%20%7B%20register(%22breaker%22%2C%20%3Abreaker)%20%7B%20puts%20%22say%20hi%22%20%7D%20%7D%0A%20%20end%0A%20%20%0Aend%0A%0Aclass%20Baz%20%3C%20Foo%0A%20%20%0A%20%20register(BreakFoo.new%2C%20%3Adoit)%20%7B%20puts%20%22another%20block%22%20%7D%0A%20%20%0Aend%0A%0Aclass%20Works%20%3C%20Foo%20%0A%20%20%0A%20%20%40x%20%3D%20BreakFoo.new%0A%20%20%0A%20%20register(%40x%2C%20%3Adoit2)%20%7B%20puts%20%22working%20block%22%20%7D%0A%20%20%0Aend%0A%0ABaz.new.doit%0AWorks.new.doit2)

one would expect both Baz and Works to behave the same, but sadly in Baz the block passed to `register` is nil!!!

iliabylich:
Closes #1499 
Closes #1181
Closes #1176
Closes #1173
Closes #1499

Example from #1499 (by @catmando):
MRI:
```
defining method breaker block = #<Proc:0x000000016ae040@test.rb:17>
defining method doit block = #<Proc:0x000000016ad500@test.rb:24>
defining method breaker block = #<Proc:0x000000016abdb8@test.rb:17>
defining method doit2 block = #<Proc:0x000000016ab8e0@test.rb:32>
calling doit: object = #<BreakFoo:0x000000016ae748> block = #<Proc:0x000000016ad500@test.rb:24>
another block
calling doit2: object = #<BreakFoo:0x000000016ac510> block = #<Proc:0x000000016ab8e0@test.rb:32>
working block
```
Opal:
```
defining method breaker block = function (){var self = TMP_4.$$s || this;

        return self.$puts("say hi")}
defining method doit block = function (){var self = TMP_Baz_6.$$s || this;

    return self.$puts("another block")}
defining method breaker block = function (){var self = TMP_4.$$s || this;

        return self.$puts("say hi")}
defining method doit2 block = function (){var self = TMP_Works_7.$$s || this;

    return self.$puts("working block")}
calling doit: object = #<BreakFoo:0x120> block = function (){var self = TMP_Baz_6.$$s || this;

    return self.$puts("another block")}
another block
calling doit2: object = #<BreakFoo:0x126> block = function (){var self = TMP_Works_7.$$s || this;

    return self.$puts("working block")}
working block
```

Example from https://github.com/opal/opal/issues/1499#issuecomment-232087091 :
MRI
```
false
1
false
2
```
Opal
```
false
1
false
2
```

The idea is to evaluate method arguments *before* assigning `$$p` property on the method body (this PR fixes a spec `A method call evaluates block pass after arguments`).

Method calls without splats and blocks are getting compiled as usually (i.e. `obj.$method(a,b,c)`)

There are 53 failing tests in `spec/lib/compiler/call_spec.rb`, will fix them after :+1:


elia:
given the ever increasing noise in compiled code I'm wondering if using an inline-able function like `$rb_plus` could help without sacrificing perfs, eg:

```js
(function(Opal) {
  function $block_send(obj, method, args, block) {
    method.$$p = block;
    return method.apply(obj, args);
  }
  var self = Opal.top, $scope = Opal, $scopes = [Opal], nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$foo']);
  return $block_send(self, self.$foo, function(){ … });
})(Opal);
```

iliabylich:
@elia I like your idea, will send an update soon

iliabylich:
PR udpated. Added two methods two `Opal` object:
+ `send_method_name` - used by a plain method call, takes a method name, fires `method_missing`
+ `send_method` - used by `super`, takes a function that represents a method body, doesn't fire `method_missing`

There are no performance changes for method calls that don't take a block or a splat.

Benchmarks for splat + block:
``` ruby
def m(v, &blk)
  v + blk.call
end

10_000_000.times { |i| m(*[i]) { i } }
```
Master:
``` sh
$ tail -n14 bm_master.js | head -n5
  return ($a = ($b = (10000000)).$times, $a.$$p = (TMP_2 = function(i){var self = TMP_2.$$s || this, $c, $d, TMP_3;
if (i == null) i = nil;
  return ($c = ($d = self).$m, $c.$$p = (TMP_3 = function(){var self = TMP_3.$$s || this;

    return i}, TMP_3.$$s = self, TMP_3.$$arity = 0, TMP_3), $c).apply($d, Opal.to_a([i]))}, TMP_2.$$s = self, TMP_2.$$arity = 1, TMP_2), $a).call($b);
```

``` sh
$ repeat 5 time node bm_master.js
node bm_master.js  16,09s user 0,02s system 100% cpu 16,009 total
node bm_master.js  16,55s user 0,02s system 100% cpu 16,479 total
node bm_master.js  15,57s user 0,04s system 100% cpu 15,511 total
node bm_master.js  14,87s user 0,03s system 100% cpu 14,811 total
node bm_master.js  14,23s user 0,04s system 100% cpu 14,176 total
```
Patch (using `Opal.send_method_name`):
``` sh
$ tail -n14 bm_patch.js | head -n5
  return $send_method_name((10000000), 'times', [], (TMP_2 = function(i){var self = TMP_2.$$s || this, TMP_3;
if (i == null) i = nil;
  return $send_method_name(self, 'm', Opal.to_a([i]), (TMP_3 = function(){var self = TMP_3.$$s || this;

    return i}, TMP_3.$$s = self, TMP_3.$$arity = 0, TMP_3))}, TMP_2.$$s = self, TMP_2.$$arity = 1, TMP_2));
```

``` sh
$ repeat 5 time node bm_patch.js
node bm_patch.js  15,25s user 0,06s system 100% cpu 15,215 total
node bm_patch.js  16,04s user 0,04s system 100% cpu 15,983 total
node bm_patch.js  16,76s user 0,04s system 100% cpu 16,692 total
node bm_patch.js  15,17s user 0,04s system 100% cpu 15,118 total
node bm_patch.js  15,27s user 0,03s system 100% cpu 15,205 total
```

iliabylich:
Updated PR.
Merged `Opal.send_method_name` and `Opal.send_method` to a single `Opal.send` that may take method or method name.

Closes #1181 #1176 #1173 #1499.

opal/corelib/runtime.js
```diff
@@ -1558,51 +1558,66 @@
...
 
-  // Call a ruby method on a ruby object with some arguments:
+  // Call a ruby method on a ruby object with some arguments and block.
+  // Used internally by Opal to invoke methods that take a block or a splat.
   //
   // @example
   //   var my_array = [1, 2, 3, 4]
-  //   Opal.send(my_array, 'length')     # => 4
-  //   Opal.send(my_array, 'reverse!')   # => [4, 3, 2, 1]
+  //   Opal.send_method_name(my_array, 'length')     # => 4
```
meh: Stuff still talks about `send_method_name`.

iliabylich: Fixed, thanks!

meh:
Do we have any performance numbers after the `send` merge? Other than that looks good to me.

iliabylich:
``` sh
$ repeat 5 time node bm_master.js
node bm_master.js  14,81s user 0,04s system 100% cpu 14,748 total
node bm_master.js  15,35s user 0,04s system 100% cpu 15,298 total
node bm_master.js  14,40s user 0,04s system 100% cpu 14,343 total
node bm_master.js  15,28s user 0,04s system 100% cpu 15,234 total
node bm_master.js  14,10s user 0,05s system 100% cpu 14,057 total

$ repeat 5 time node bm_patch.js 
node bm_patch.js  14,44s user 0,05s system 100% cpu 14,394 total
node bm_patch.js  15,72s user 0,05s system 100% cpu 15,667 total
node bm_patch.js  14,41s user 0,04s system 100% cpu 14,357 total
node bm_patch.js  14,44s user 0,04s system 100% cpu 14,387 total
node bm_patch.js  14,82s user 0,05s system 100% cpu 14,775 total

$ grep '$send' bm_master.js | wc -l
3
$ grep '$send' bm_patch.js | wc -l
253
```

meh:
:+1:

PR #1557, opened 2016-09-03
[WIP] Unicode pack and unpack

camertron:
The current implementations of String#unpack and Array#pack don't take surrogate pairs into consideration. This PR is an attempt to add support for them. (Just before submitting I realized my method and variable names are supposed to be snake_cased. I'll try to fix them this weekend).

I believe this work should be part of a wider effort to support UTF-safe string operations in Opal-generated code, but I don't really know where to start or whether such an effort would be well-received by the Opal maintainers and community. I did a little Google searching but couldn't find a comprehensive Javascript library for doing this, so I wrote [utfstring](https://github.com/camertron/utfstring). I think much of its functionality could be incorporated into Opal. Please let me know what you think!

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -1696,23 +1696,108 @@ class String < `String`
   def unpack(pattern)
     %x{
       function stringToBytes(string) {
-        var i,
-            singleByte,
-            l = string.length,
-            result = [];
+        var result = [];
 
-        for (i = 0; i < l; i++) {
-          singleByte = string.charCodeAt(i);
-          result.push(singleByte);
+        for (var i = 0; i < string.length; i ++) {
+          var chr = string.charCodeAt(i);
+          var byteArray = [];
+
+          while (chr > 0) {
+            byteArray.push(chr & 0xFF);
+            chr >>= 8;
+          }
+
+          // all utf-16 characters are two bytes
+          if (byteArray.length == 1) {
+            byteArray.push(0);
+          }
+
+          // assume big-endian
+          result = result.concat(byteArray.reverse());
         }
+
         return result;
       }
+
+      function stringToCodePoints(string) {
+        var result = [];
+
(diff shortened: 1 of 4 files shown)


opal/corelib/string.rb
```diff
@@ -1696,23 +1696,113 @@ def self._load(*args)
...
         }
+
         return result;
       }
+
+      function stringToCodePoints(string) {
+        var result = [];
+
+        for (var i = 0; i < string.length; i ++) {
+          codePoint = charCodeAt(string, i);
```
vais: `codePoint` is a global variable here because it is not defined with `var` in this function (see https://travis-ci.org/opal/opal/jobs/157245342#L298)

camertron: Good catch, thanks @vais :)

vais:
@camertron 🍻  ...but the other build failures are all over the map. Are you looking into those?

Have you looked into whether there are any existing Ruby Specs that are currently filtered out that Opal can start passing after this change? (As opposed to the new Opal-specific specs that you added. If not, it would be awesome to submit those to the Ruby Spec project instead, so everyone can benefit from them, including Opal once they are accepted and submodules are updated.)

camertron:
@vais no, I wasn't looking at them. They don't seem to have anything to do with my changes? I guess master is passing tho... I can take a look.

karimmtarek:
Hey @camertron, are you still working on this? If not, I'd like to take over. Please let me know, thanks.

camertron:
@karimmtarek Unfortunately I was sidetracked by other projects. The main blocker was trying to get the tests to pass since the failures never made any sense to me. Glad you're willing to take up the mantle here - go for it!

hmdne:
I'm closing this PR as Opal already got a (bigger) pack/unpack implementation. Also, #2235 improves the U* support a lot.

PR #1558, opened 2016-09-04
Optimize Method#to_proc

jgaskins:
This replaces a function wrapper with a bound version of the same function, which AFAIK inherits optimizations from the original. 

I don't think the fact that the previous version leaked an `arguments` object had much effect on performance in this scenario. Calling `foo(&method(:bar))` only ever executed that function once, so optimization couldn't happen anyway. I think most of the performance difference here is from inheriting previous optimizations.

Benchmark:

``` ruby
Measurement.new('Method#to_proc', &method(:noop))
```

Results in iterations per second:

Safari 10: 3.3M up from 2.26M (46% faster)
Chrome 53: 1.36M up from 925K (47% faster)
Firefox 48: 1.6M up from 483K (231% faster)

### DIFF
--- a/opal/corelib/method.rb
+++ b/opal/corelib/method.rb
@@ -32,7 +32,7 @@ class Method
 
   def to_proc
     %x{
-      var proc = function () { return self.$call.apply(self, $slice.call(arguments)); };
+      var proc = self.$call.bind(self);
       proc.$$unbound = #@method;
       proc.$$is_lambda = true;
       return proc;


elia:
Looks really cool!

The only objection that I have is that [Function.prototype.bind is IE9+](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Specifications) and we still [advertise IE6 support in the readme](https://github.com/opal/opal#browser-support). 

I think it's time to update the readme but I'm not completely sure on where to ✂️.

An alternative could be to provide an extra file stuffed with polyfills for APIs we rely on and offer that to support older browsers.

@opal/core any considerations on this?

vais:
@elia maybe it's time to say goodbye to IE6, 👍 

I also don't know where to make the cut on the supported browsers - I'm not even sure Opal still runs on IE6 since I gave up trying to get the specs to run in IE8 a long while back.

In other words, *viva la progress*!

svoboda-jan:
👍  for __keeping__ IE6 support.

I am currently developing and have opal applications in production that depend on IE 6.

Using polyfills is a nice solution, but in this case it would be better to use them as __progressive enhancement__.

svoboda-jan:
__Update / Idea__ : Since opal is a source-to-source compiler, we could have `targets` like `ECMAScript 5.1 (ECMA-262)` where `Function.prototype.bind` has a status of `Standard`.

This makes more sense because it abstracts the notion of browsers / runtime.

meh:
But all the work I did to make it work on there! :rage4:

jdalton:
You may be able to drop support for older environment by simply having folks that want to support them use language shims like [es-shim](https://github.com/es-shims/es5-shim).

jdalton:
It's also worth noting the MS no longer supports IE < 11 and that many popular libraries in the JS ecosystem, like jQuery, Prototype, Ember, Angular, React, & Lodash, have dropped support for IE 8 and below.

elia:
@jdalton good point

_…and good timing_

After having [restored browser testing](https://github.com/opal/opal/blob/dbf166568deef60d696555e67e54c53a53581204/tasks/testing.rake#L356-L376) and tried to run the [`opal` suite](https://github.com/opal/opal/blob/dbf166568deef60d696555e67e54c53a53581204/tasks/testing.rake#L133) on IE6 with the only result of getting an "Out of memory" error 😖, I finally agree on dropping support for IE8 and below 😄. 

For people still requiring that kind of antiquities (and I'm sure they're out there) we still have older versions of Opal. 

@meh, let me add that all the work you've done 🎖 is still valuable for anyone trying to run Opal with the help of the shims (e.g. avoiding trailing commas in compiled code etc.).

meh:
@elia yeah, go ahead, I was mostly joking 🐼

Issue #336, opened 2013-07-30
Method#source_location


peter-leonov:
Ha? ;)

iliabylich:
Fixed in #1560

PR #1560, opened 2016-09-07
Add Method#source_location and Method#comments

iliabylich:
+ `Method#source_location` is available only when --enable-source-location/enable_source_location = true option specified.
+ `Method#comments` is available only when --parse-comments/parse_comments = true specified.

Both of them are disabled by default.

```
$ cat -n test.rb
     1	# Does "something"
     2	# and something 'else'
     3	def m
     4	end
     5	
     6	p method(:m).source_location
     7	p method(:m).comments
     8	
     9	class Foo
    10	  # Instance method
    11	  # that does the following:
    12	  #
    13	  def m
    14	  end
    15	end
    16	
    17	p Foo.instance_method(:m).source_location
    18	p Foo.instance_method(:m).comments

$ opal --parse-comments --enable-source-location test.rb
["test.rb", 3]
["# Does \"something\"", "# and something 'else'"]
["test.rb", 13]
["# Instance method", "# that does the following:", "#"]
```


iliabylich:
Finally green. Any comments?
I'm not sure about `Method#comments` - there's no such method in MRI :smile: 

Closes #336.

fkchang:
@iliabylich I like it, I was hoping to extract the actual code from the method for some cool tooling. I suppose I could work something out though I'll need to think about how I can do it both embedded in the app and from opal-console (not exactly sure how to get chrome devtools to give me 'arbitrary access' to files on the user's computer -- might be non secure)

elia:
@iliabylich 👍🏼 "Squash and merge" at will

(too bad GitHub doesn't listen to .gitattributes and marked the PR as with conflicts for the Changelog change)

iliabylich:
@elia I guess it was :+1: :smile:

PR #1561, opened 2016-09-08
Add JS::Error class that can handle any JS errors in rescue.

### LINKED ISSUE #681: Opal does not catch Object exceptions from Javascript
Running the following code in http://opalrb.org/try/# indicates that Opal does not catch exceptions from Javascript which are objects. 

It ends in

```
called foo
unhandled exception
the end
```

It works with String exceptions. (remove the comment on 

```// throw "exception from foo"```

This ends in 

```
called foo
caught exception: exception from foo
the end
```

This is the code to play with

```ruby
def foo(m, at)
  puts "called foo"
  %x[
      //throw "exception from foo";
      throw {message: m, pos: at}
   ]
  nil
end

begin
foo("calling foo at", 10)
rescue Object => e
  puts "caught exception: #{Native(e)}"
rescue 
  puts "unhandled exception"
end

puts "the end"
```

### LINKED ISSUE #1280: Empty rescue with native JS Exceptions
Before https://github.com/opal/opal/commit/73707394c288ed43a134b62559b92ba924c73fe6, the rescue worked as noted below

```ruby
begin
  `throw 42`
  return
rescue
  puts 'rescue ran!'
end
```

After that commit, `rescue Exception =>...` is needed because a native JS exception is not considered an instance of `StandardError`.

What's the best way to handle this? Modify the compiler to add an 'empty_rescue' flag and then have the `Opal.rescue` function look at the flag and match on StandardError or native exception?

iliabylich:
Fixes #1280 
Fixes #681

### DIFF
--- a/spec/opal/core/runtime/rescue_spec.rb
+++ b/spec/opal/core/runtime/rescue_spec.rb
@@ -68,4 +68,39 @@ describe "The rescue keyword" do
     a.should == "I got foo"
   end
 
+  context 'JS::Error' do
+    it 'handles raw js throws' do
+      begin
+        `throw { message: 'foo' }`
+        nil
+      rescue JS::Error => e
+        e.JS[:message]
+      end.should == 'foo'
+    end
+
+    it 'handles other Opal error' do
+      begin
+        raise 'bar'
+      rescue JS::Error => e
+        e.message
+      end.should == 'bar'
+    end
+
+    it 'can be combined with other classes to catch js errors' do
+      begin
+        `throw { message: 'baz' }`
+        nil
+      rescue JS::Error, RuntimeError => e
+        e.JS[:message]
+      end.should == 'baz'
+    end
+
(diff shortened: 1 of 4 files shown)


spec/opal/core/runtime/rescue_spec.rb
```diff
@@ -68,4 +68,39 @@ def empty_rescue
...
+    it 'can be combined with other classes to catch js errors' do
+      begin
+        `throw { message: 'baz' }`
+        nil
+      rescue JS::Error, RuntimeError => e
+        e.JS[:message]
+      end.should == 'baz'
+    end
+
+    it 'can be combined with other classes to cath Opal errors' do
```
elia: typo: `catch Opal errors`

iliabylich:
Rebased. @elia Any other comments?

elia:
> @elia Any other comments?

Just one: great job 😉

PR #1567, opened 2016-09-24
docs: Update compiled code with latest gem

youchan:
This pull request is for updating compiled code in the document, because the compiled codes in document differ from  real outputs.
I guess it is not updated from old version.
I tried to compile the examples in `docs/compiled_ruby.md` with the latest gem version `0.10.2`.


elia:
Thank you very much!

PR #1570, opened 2016-09-30
Properly convert `{"a":[null]}` to {a: [nil]}

jgaskins:
`native`'s `Hash#new` recursion for arrays didn't Rubify values directly within arrays, only for values of hashes within arrays. This means that the `null` in `{ "a": [null] }` would not be converted to `nil` — it would remain `null`.

Explaining data structures in prose is terrible. The spec changes in the diff are so much more expressive than my description here.

This was actually my bug from when I added this array recursion in the first place.

/cc @catmando

### DIFF
--- a/spec/opal/stdlib/native/hash_spec.rb
+++ b/spec/opal/stdlib/native/hash_spec.rb
@@ -12,6 +12,7 @@ describe Hash do
         e: [
           {
             f: 'g',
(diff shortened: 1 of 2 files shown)


elia:
@catmando let me know if you need this to be backported to 0.10 or 0.9

PR #1572, opened 2016-10-06
JSON#parse raises JSON::ParserError on invalid input

### LINKED ISSUE #1545: JSON.parse not throwing StandardError
This works in IRB
`JSON.parse('foo') rescue 'foo'`
but not in opal 0.10

bkutil:
JSON.parse and JSON.load raise ParserError when encountering an invalid JSON string.

When JavaScripts native JSON.parse fails, re-raise the SyntaxError as
JSON::ParserError. Class ancestors for ParserError match MRI
implementation.

Fixes #1545

### DIFF
--- a/spec/opal/stdlib/json/parse_spec.rb
+++ b/spec/opal/stdlib/json/parse_spec.rb
@@ -30,4 +30,8 @@ describe "JSON.parse" do
     JSON.parse('{"a": "b"}').should == {"a" => "b"}
     JSON.parse('{"a": null, "b": 10, "c": [true, false]}').should == {"a" => nil, "b" => 10, "c" => [true, false]}
   end
+
+  it "raises ParserError for invalid JSON" do
+    lambda { JSON.parse("invalid_json") }.should raise_error(JSON::ParserError)
+  end
 end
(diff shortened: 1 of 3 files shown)


spec/opal/stdlib/json/parse_spec.rb
```diff
@@ -30,4 +30,8 @@
     JSON.parse('{"a": "b"}').should == {"a" => "b"}
     JSON.parse('{"a": null, "b": 10, "c": [true, false]}').should == {"a" => nil, "b" => 10, "c" => [true, false]}
   end
+
+  it "raises ParserError for invalid JSON" do
+    lambda { JSON.parse("invalid_json") }.should raise_error(JSON::ParserError, "Unexpected token i in JSON at position 0")
```
iliabylich: Looks like it fails on Travis because of the platform-specific error message. With node `0.10.25` (I know, it's very old) and in chrome I'm getting:
``` js
try {
  JSON.parse('invalid_json')
} catch(e) {
  console.log(e.message)
}
// => Unexpected token i
```

Probably it's easier to leave only `should raise_error(JSON::ParserError)`

bkutil: Thank you, fixed.

iliabylich:
Thank you!

Issue #896, opened 2015-05-28
invalide results when devide


AliOsm:
when I run this lines
```
puts 3/2
puts 3.0/2.0
puts 2/3
puts 2.0/3.0
```
I get this results but this is wrong
```
1.5
1.5
0.6666666666666666
0.6666666666666666
```
I should get this results
```
1
1.5
0
0.6666666666666666
```
like when I run that lines in terminal.

elia:
In opal all numbers are mapped to javascript numbers, so that's the expected behavior.

For a complete list of trade offs see http://opalrb.org/docs/compiled_ruby/

PR #1574, opened 2016-10-16
add Numbers into Unsupported Features

takahashim:
Related to https://github.com/opal/opal.github.io/issues/21 and https://github.com/opal/opal/issues/896


meh:
That sounds a little misleading to me.

Opal does support numbers, it just doesn't support strongly typed numerics like Ruby does, there's a single class for `Float` and `Integer`, but there are classes for `Complex` and `Rational`.

takahashim:
@meh thanks for your feedback!
Should I rename the title as `#### Integer and Float type system ####` instead of `#### Numbers ####`?

elia:
@takahashim yes, maybe something like `Integer / Float difference`. 
Also can you please expand the explanation with what @meh added and remove the `method_added`/`method_removed` as they're now supported?

takahashim:
@elia Thanks!  I fixed it.
And I'll send another PR to remove `method_added` and `method_removed` from the document.

elia:
Thank you!

PR #1575, opened 2016-10-29
Remove method_added and method_removed from Unsupported Features

takahashim:
`Module#method_added` and `Module#method_removed` are supported now.


elia:
Thanks!

PR #1585, opened 2016-11-13
Resolves #1583, Improves File.join on browser

### LINKED ISSUE #1583: File.join removes protocol slash
### Description

Ruby 2.1.5
```ruby
puts File.join('file://', 'home', 'foo', 'bar') # file://home/foo/bar
```
Opal 0.10
```javascript
// file:/home/foo/bar
```

### Implementation
```ruby
def join(*paths)
  paths.join(SEPARATOR).gsub(%r{#{SEPARATOR}+}, SEPARATOR)
end
```
https://github.com/opal/opal/blob/4cb3aed3d8af39b686dcc971619229d56468973e/opal/corelib/file.rb#L257-L259

One solution could be to try to parse the first path with [URI.parse](http://ruby-doc.org/stdlib-2.0.0/libdoc/uri/rdoc/URI.html#method-c-parse) ?
We could also modify the Regex to use "negative lookbehind" and replace separator only if it isn't preceded by `:`

```ruby
def join(*paths)
  paths.join(SEPARATOR).gsub(%r{(?<!:)#{SEPARATOR}+}, SEPARATOR)
end
```

What do you think ?

Mogztter:


Mogztter:
Rebased on master

opal/corelib/string.rb
```diff
@@ -426,6 +426,21 @@ def end_with?(*suffixes)
...
+        if (self.length >= suffix.length &&
+            self.substr(0, suffix.length) == suffix) {
+          return true;
+        }
+      }
+    }
+
+    false
+  end
+
```
elia: I think this can be omitted and `File.join` can use `#start_with?` instead.

http://www.rubydoc.info/stdlib/core/String#start_with%3F-instance_method

spec/filters/bugs/file.rb
```diff
@@ -0,0 +1,12 @@
+opal_filter "File" do
+  fails "File.join returns a duplicate string when given a single argument"
+  #fails "File.join respects the given separator if only one part has a boundary separator"
+  #fails "File.join prefers the separator of the right part if both parts have separators"
```
elia: you can safely delete commented lines

spec/filters/bugs/file.rb
```diff
@@ -0,0 +1,12 @@
...
+  fails "File.join raises a TypeError exception when args are nil"
+  fails "File.join calls #to_str"
+  fails "File.join calls #to_path"
+  fails "File.join flattens nested arrays"
+  fails "File.join raises an ArgumentError if passed a recursive array"
+  fails "File.join inserts the separator in between empty strings and arrays"
+  fails "File.join doesn't mutate the object when calling #to_str"
```
elia: If this spec relies on mutable strings then it should be in `unsupported/file.rb`

Mogztter:
> I left a couple of comments, but overall looks good, who knew File.join was this complicated.

Indeed, I've spent some time to comply with the spec :sweat_smile: 
The code is far from perfect but at least it is a starting point on which we can iterate to improve readability/performance...

I will make the requested changes today or tomorrow

Mogztter:
Rebased on master, commits squashed and code fixed!

opal/corelib/string.rb
Mogztter: You are right, done!

Mogztter:
Excellent, looking forward to next the release 😉

Issue #1584, opened 2016-11-12
Run the tests suite on AppVeyor (Windows)


Mogztter:
AppVeyor is a Continuous Delivery service for Windows: https://www.appveyor.com/
It's relatively easy to setup AppVeyor for a GitHub repository and the build can be configure with an `appveyor.yml` file at the root of the project (see https://github.com/asciidoctor/asciidoctor.js/blob/master/appveyor.yml)

Currently almost all tests are passing except:

```
Failures:

  1) Opal::PathReader#read responds to #path
     Failure/Error: expect(path_reader.read(path)).to eq(contents)

       expected: "require 'opal'\nputs 'hi from opal!'\n"
            got: "require 'opal'\r\nputs 'hi from opal!'\r\n"

       (compared using ==)

       Diff:
     # ./spec/lib/shared/path_reader_shared.rb:24:in `block (3 levels) in <top (required)>'

Finished in 34.8 seconds
344 examples, 1 failure
```

Mogztter:
Done!
Thanks @elia

PR #1587, opened 2016-11-13
Fix test on Windows

Mogztter:
First step to be able to run the tests suite on AppVeyor #1584

### DIFF
--- a/spec/lib/shared/path_reader_shared.rb
+++ b/spec/lib/shared/path_reader_shared.rb
@@ -21,7 +21,8 @@ shared_examples :path_reader do
 
   describe '#read' do
     it 'responds to #path' do
-      expect(path_reader.read(path)).to eq(contents)
+      # Normalizing line endings
+      expect(path_reader.read(path).gsub(/\r\n/, "\n")).to eq(contents)
     end
 
     it 'returns nil if the file is missing' do


Mogztter:
The build failure is `missing SAUCE_ACCESS_KEY env var`

Mogztter:
Rebased on master

elia:
This looks a bit odd as the content of `contents` should come from `File.read` and thus have the same line ending as the contents coming from `path_reader.read`. I'll wait until the AppVeyor build actually runs[1] and see if that error pops out…

_[1]: at this time is blocked at `bundle install` trying to install the `json` gem_

Mogztter:
Ok, could you please give me a link to the most recent build on Appveyor ?

elia:
> Ok, could you please give me a link to the most recent build on Appveyor ?

The PR fixing the Ruby version worked and now it fails properly on specs (because it can't find the phantomjs executable). Here's the link anyway:

https://ci.appveyor.com/project/elia/opal/build/4

Mogztter:
Right, we could use [phantomjs-prebuilt](https://www.npmjs.com/package/phantomjs-prebuilt) npm package to install Phantomjs or create a Rake task to do it ? Which do you prefer ?

elia:
@Mogztter I think I solved the phantomjs problem using choco (https://github.com/opal/opal/commit/8ac3ad23244f808e9dbac40eab3a58d4fbc20d3d) but the build seems to stall at some point now, any help is very appreciated of course 🙂

elia:
Everything's working now, and #1584 it doesn't seem to happen anymore. Maybe it has been fixed by other means, or else it was related to the git crlf configuration. Anyway please close these two issues/pr unless you have any other concerns we can address.

Mogztter:
> Everything's working now, and #1584 it doesn't seem to happen anymore. Maybe it has been fixed by other means, or else it was related to the git crlf configuration.

Yes that's possible.
Nice work by the way! I think we should also run `mspec` wdyt ?

elia:
Now that the basics are there why not? :)

PR #1589, opened 2016-11-14
Improve Regexp class.

iliabylich:
Added support for named captures, extended mode was properly implemented and moved to runtime, all native flags are supported.
``` sh
$ cat test.rb
r = /\$(?<dollars>\d+)\.(?<cents>\d+)/mixne
puts r
md = r.match("$3.67")
p md

$ opal test.rb
(?mix:\$(?<dollars>\d+)\.(?<cents>\d+))
#<MatchData "$3.67" dollars:"3" cents:"67">
```

All regexps are now compiled to `Opal.Regexp.$create(source, flags)` (definitely should be optimized).

Also I've merged `bugs/regexp` with `unsupported/regexp` to simplify the development process. Now I feel like ~50% should be moved to `unsupported` - probably these specs should be reviewed :smile: .

Also I've tested this PR with [these changes to rubyspec](https://gist.github.com/iliabylich/552b46adbf92cf38d2d19192557942fb) - it passes, so hopefully all those filters will be removed after merging.

Bechmarks:
``` sh
%x{
  for (var i = 0; i < 1000000; i++) {
    var r = #{/\w/};
  }
}
node bm_new.js  18,96s user 0,04s system 100% cpu 18,874 total

%x{
  for (var i = 0; i < 1000000; i++) {
    var r = /\w/;
  }
}
node bm_native.js  0,14s user 0,01s system 100% cpu 0,150 total

r = /\w/;
%x{
  for (var i = 0; i < 1000000; i++) {
    #{r.match("w")}
  }
}
node bm_new.js  0,12s user 0,02s system 100% cpu 0,136 total

r = `/\w/`;
%x{
  for (var i = 0; i < 1000000; i++) {
    #{r.match("w")}
  }
}
node bm_native.js  0,13s user 0,01s system 100% cpu 0,138 total
```
Creating regexps through `Opal.Regexp.$create` is very slow, working on simple cases optimization (like `/\w/`), adding WIP label.

With this patch I can compile `uri` with `uri/rfc2396_parser`. It has some string mutation, so after minor changes locally I've got the following:
``` sh
$ cat test.rb
require 'uri'

uri = URI.parse('https://google.com/search?a=b')
puts uri

puts uri.host
puts uri.port
puts uri.path
puts uri.query

$ opal -ssocket -I/home/ilyabylich/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/ test.rb
https://google.com/search?a=b
google.com
443
/search
a=b
```


meh:
For the optimization, the best way to go would be to have a method in the compiler to check if any of the extended Ruby stuff is being used or not, and emitting a raw `/whatever/` when that's the case.

It should also include straightforward conversions like the `\A` and friends.

PR #1591, opened 2016-11-27
Optimize Opal.is_a

jgaskins:
This short-circuits walking the class/module hierarchy to check whether an object is an instance of a particular class/superclass/module. It also removes duplicated code to check the object's meta class in addition to the object's class.

I noticed that one of my apps using `grand_central` for state management (so it uses `case` heavily to match actions on their class) spent a lot of time in `Opal.ancestors` to construct an array of classes and modules. Since most matches occur within the first 1-3 elements of that array, we don't need to construct the array all the way down.

## Before:

|Benchmark|Iterations per sec|
|-----|-------|
|Object#is_a? hit|1,144,293|
|Object#is_a? superclass hit|1,174,402|
|Object#is_a? deep superclass hit|1,208,026|
|Object#is_a? miss|1,177,395|
|Class#=== hit|1,105,455|
|Class#=== superclass hit|1,047,927|
|Class#=== deep superclass hit|1,118,437|
|Class#=== miss|1,051,812|

## After:

|Benchmark|Iterations per sec|
|-----|-------|
|Object#is_a? hit|3,617,699|
|Object#is_a? superclass hit|2,869,255|
|Object#is_a? deep superclass hit|3,170,777|
|Object#is_a? miss|3,088,117|
|Class#=== hit|3,728,483|
|Class#=== superclass hit|3,317,002|
|Class#=== deep superclass hit|2,986,606|
|Class#=== miss|3,126,489|

### Benchmark details:

Performed on release versions of Chrome, Safari, and Firefox. These results are from Chrome, but all three browsers had similar results.

- `Object#is_a?` hit: `object.is_a? Object`
- `Object#is_a?` superclass hit: `string.is_a? Object`
- `Object#is_a?` deep superclass hit: `Class.new(Class.new(Class.new(Class.new))).new.is_a?(Object)` (it doesn't create the subclasses on each iteration, just shown here for clarity).
- `Object#is_a?` miss: `string.is_a? Hash`

The structure is the same for `Class#===`, just using i.e. `Object === object` instead.

You can see in the commit log that I originally tried setting `klass.$$proto['$$class$' + klass.id]`. That _was_ faster than the current live implementation, but it wasn't this fast for some reason — probably due to having to construct the string — and it failed one of the specs.


opal/corelib/runtime.js
```diff
@@ -1241,26 +1241,26 @@
...
+  }
 
-    for (i = 0, length = ancestors.length; i < length; i++) {
-      if (ancestors[i] === klass) {
-        return true;
-      }
-    }
+  Opal.is_a = function(object, klass) {
+    if(check_class_hierarchy(object.$$class, klass)) return true;
+    if(check_class_hierarchy(object.$$meta, klass)) return true;
```
meh: Wouldn't this be clearer?

```javascript
return check_class_hierarchy(object.$$class, klass) ||
       check_class_hierarchy(object.$$meta, klass);
```

jgaskins: Probably :-)

hmdne:
We now cache Opal.ancestors:

https://github.com/opal/opal/blob/v1.2.0/opal/corelib/runtime.js#L1293-L1295

PR #1592, opened 2016-11-28
add ascii_only? method

patrickkettner:


patrickkettner:
updated with requests. It does not fix any of the failing specs

elia:
The build is failing due to the newly added AppVeyor config, the rest was passing

PR #1594, opened 2016-11-29
correct gsub behavior when only line anchors are used

### LINKED ISSUE #1593: gsub's broken with ^ and $
in MRI
```ruby
puts "ab\ncd".gsub(/$/, 'e')
abe
cde
=> nil

puts "ab\ncd".gsub(/^/, 'e')
eab
ecd
=> nil
```

in Opal 0.10.3
```js
"ab\ncd".$gsub(/$/, 'e')
"eab
ecd"
"ab\ncd".$gsub(/^/, 'e')
"eaeb
cd"
```

in vanilla js
```js
"ab\ncd".replace(/$/gm, 'e')
"abe
cde"
"ab\ncd".replace(/^/gm, 'e')
"eab
ecd"
```

patrickkettner:
fixes #1593


opal/corelib/string.rb
```diff
@@ -436,6 +436,7 @@ def gsub(pattern, replacement = undefined, &block)
       }
 
       var result = '', match_data = nil, index = 0, match, _replacement;
+      var onlyAnchor = /\/[\^\$]\//.test(pattern.toString())
```
meh: This looks like a really slow way to check for that.

patrickkettner: happy to rewrite it - though I'm not sure how else youd like to do it. Perhaps checking `pattern.source` instead of stringifying (thats all that browser engines are doing when tostring is being called, really)

elia: `pattern.source` seems the way to go:

screen shot 2016-12-09 at 01 38 46

probably with something like: 

```js
pattern.source === '^' || pattern.source === '$'
```

patrickkettner:
Upon further thought, that won't catch cases like /(^|$)/. I'll need to have a bit of a think on how to properly support it in a correct way

elia:
@patrickkettner even if it's not particularly short or readable it's probably worth having a look at the original implementation to understand what's the rule (or implementation detail) behind that behavior:

https://github.com/ruby/ruby/blob/5ae717f249e2eb40165c91c8ef153adbfc0466c4/string.c#L4809-L4936

elia:
```
opal:master ⤑ be bin/opal -e 'puts "asdf\nqwer".gsub(/(^|$)/, "z")'     ~/C/opal
zazsdf
zqzwer
opal:master ⤑ be ruby -e 'puts "asdf\nqwer".gsub(/(^|$)/, "z")'         ~/C/opal
zasdfz
zqwerz
opal:master ⤑                                                           ~/C/opal
```

here's another case using `/(^|$)/`

patrickkettner:
@elia it _did_ fix that spec. pushed the updated regex, `.source` change @meh requested, and updated the changelog. Let me know if there is anything more I can do

elia:
> Let me know if there is anything more I can do

Just one thing: rebasing on master 🙂

patrickkettner:
done!

elia:
@patrickkettner seeing a failure on windows: https://ci.appveyor.com/project/elia/opal/build/106#L189

PR #1597, opened 2016-11-30
correct anchor to get proper regex source and use supplied flags

patrickkettner:
currently, anchor is grabbing `pattern[1..-2]`, which works fine for most regexes, but pukes when you supply one with flags, defeating the entire purpose of it. Instead, we now grab the `source` directly. 

Also, currently the flags are being dropped, this fixes that as well

### DIFF
--- a/stdlib/strscan.rb
+++ b/stdlib/strscan.rb
@@ -297,6 +297,10 @@ class StringScanner
 
 private
   def anchor(pattern)
-    `new RegExp('^(?:' + pattern.toString().substr(1, pattern.toString().length - 2) + ')')`
+    %x{
+      var flags = pattern.toString().match(/\/([^\/]+)$/);
+      flags = flags ? flags[1] : undefined;
+      return new RegExp('^(?:' + pattern.source + ')', flags);
+    }
   end
 end


stdlib/strscan.rb
```diff
@@ -297,6 +297,10 @@ def unscan
 
 private
   def anchor(pattern)
-    `new RegExp('^(?:' + pattern.toString().substr(1, pattern.toString().length - 2) + ')')`
+    %x{
+      var flags = pattern.toString().match(/\/([^\/]+)$/)
+      flags = flags && flags[1] || undefined
```
elia: @patrickkettner can you replace with the ternary operator `?:` which is slitghly more explicit?

stdlib/strscan.rb
```diff
@@ -297,6 +297,10 @@ def unscan
...
+      return new RegExp('^(?:' + pattern.source + ')', flags)
```
elia: Can you add the final semicolon to all the JS lines?

Trailing semicolons wars aside, it should fix the specs because opal detects its presence and adds a leading return if its missing.

eg.

```rb
def anchor(pattern)
  %x{
    var flags = pattern.toString().match(/\/([^\/]+)$/)
    flags = flags && flags[1] || undefined
    return new RegExp('^(?:' + pattern.source + ')', flags)
  }
end
```

through `| opal -cEO` generates:

```js
(function(Opal) {
  var TMP_anchor_1, self = Opal.top, $scope = Opal, $scopes = [Opal], nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  return (Opal.defn(Opal.Object, '$anchor', TMP_anchor_1 = function $$anchor(pattern) {
    var self = this;

    return 
    var flags = pattern.toString().match(/\/([^\/]+)$/)
    flags = flags && flags[1] || undefined
    return new RegExp('^(?:' + pattern.source + ')', flags)
  
  }, TMP_anchor_1.$$arity = 1), nil) && 'anchor'
})(Opal);
```

patrickkettner:
pushed the requested updates. sorry about the lack of semicolons, the `%x{}` syntax isn't supported in my editor so it missed my linter.

PR #1608, opened 2016-12-17
Moooore tests on AppVeyor

Mogztter:
Looking for :bug: 

:wink:

### DIFF
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -239,7 +239,9 @@ task :cruby_tests do
     js_filename: js_filename,
   )
   sh build_js_command
-  sh "NODE_PATH=stdlib/nodejs/node_modules node #{js_filename}"
+  env = {'NODE_PATH' => 'stdlib/nodejs/node_modules'}
+  cmd = "node #{js_filename}"
+  system(env, cmd)
 end
 
 task :test_nodejs do
@@ -254,7 +256,9 @@ task :test_nodejs do
     js_filename: js_filename,
   )
   sh build_js_command
-  sh "NODE_PATH=stdlib/nodejs/node_modules node #{js_filename}"
+  env = {'NODE_PATH' => 'stdlib/nodejs/node_modules'}
+  cmd = "node #{js_filename}"
+  system(env, cmd)
 end
 
 desc 'Runs opal-rspec tests to augment unit testing/rubyspecs'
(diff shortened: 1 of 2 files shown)


Mogztter:
Hmmm I think the "export variable" `NODE_PATH=stdlib/nodejs/node_modules` is not supported in a Windows/Power-shell environment.

https://github.com/opal/opal/blob/master/tasks/testing.rake#L242

elia:
> @elia Could we add a special case for Windows ? SET NODE_PATH=stdlib/nodejs/node_modules ?

Sure, probably it can be also done in a system agnostic way using the first argument of `Kernel#system` but a check about supported ruby version is needed for that.

Mogztter:
Build success !

tasks/testing.rake
```diff
@@ -214,6 +214,15 @@ module MinitestSuite
   end
 end
 
+def execute_cmd_with_env(env, cmd)
+  if RUBY_VERSION.to_f > 2
```
elia: @Mogztter shouldn't this be `>= 2` to include 2.0 (I'm assuming env support was introduced since 2.0)

If that's true you can safely assume we're targeting 2.0 only, and leaving older releases for `< 2.0` (I'll update the gemspec soon to reflect that).

Mogztter: I didn't find exactly when `Kernel.system` was introduced but the Ruby 2.0.0 documentation shows this method: `system([env,] command... [,options]) → true, false or nil`
https://ruby-doc.org/core-2.0.0/Kernel.html#method-i-system

> shouldn't this be >= 2 to include 2.0 (I'm assuming env support was introduced since 2.0)
If that's true you can safely assume we're targeting 2.0 only, and leaving older releases for < 2.0 (I'll update the gemspec soon to reflect that).

You mean removing the condition and use `Kernel.system` all the time since we're targeting 2+ only ? Or just replace `> 2` with `>= 2` ?

Mogztter: @elia Exactly my first or second proposition ? 😉

elia: @Mogztter LOL sorry! 

I meant the first one, using `Kernel#system` directly. Or better removing the ruby-version check and raising an error in case it fails (which is the default behavior of `FileUtils#sh` that was used originally).

PR #1612, opened 2016-12-27
update Array#zip to properly handle blocks

### LINKED ISSUE #1611: Array#zip inconsistencies
in MRI
```rb
foo = [['A', 'B']]

foo.zip do | a,b |
  puts a
  puts b
end
```

outputs

```
A
B
```

In latest opal stable, outputs 

```
[["A", "B"]]

```

I thought it may be a 2d array issue, but flattening didn't fix it, just resulted in other errors

in MRI
```rb
foo = ['A', 'B']

foo.zip do | a,b |
  puts a
  puts b
end
```

outputs

```
A

B

```

and in Opal

```
["A"]

["B"]

```

(note how the output is arrays of strings rather than just strings)

patrickkettner:
fixes #1611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@ Whitespace conventions:
 - Fixed `Integer("0")` raising `ArgumentError` instead of parsing as 0
 - Fixed `JSON#parse` to raise `JSON::ParserError` for invalid input
 - `Module#append_features` now detects cyclic includes
+- Fixed `Array#zip` to properly handle blocks (#1612)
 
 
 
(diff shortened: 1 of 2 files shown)


elia:
@patrickkettner just a hint: use [`Opal.yield1`](https://github.com/opal/opal/blob/master/opal/corelib/runtime.js#L1406) and [`Opal.yieldX`](https://github.com/opal/opal/blob/master/opal/corelib/runtime.js#L1427) to comply with (the complex) Ruby `yield` behavior 😃

patrickkettner:
Sorry, what would  be bring yielded where?

patrickkettner:
I _think_ I did the right thing, haha. Let me know if I didn't

PR #1613, opened 2016-12-30
Allow emojis in strings

jgaskins:
The comment about high-value characters being returned as `"\uXXXXXXXX"` from `String#inspect` doesn't seem to hold. It worked just fine on MRI 2.1, 2.2, and 2.3. That is, `String#inspect` displayed the string exactly as it was without Unicode escaping.

### DIFF
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -52,7 +52,12 @@ module Opal
       end
 
       def compile
-        push translate_escape_chars(value.inspect)
+        sanitized_value = value.inspect
+        sanitized_value.gsub! /\\u\{[0-9a-f]+\}/ do |char|
+          @compiler.warning("Ignoring unsupported character #{char}", @sexp.line)
+          ''
+        end
+        push translate_escape_chars(sanitized_value)
       end
     end
 
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -52,8 +52,7 @@ module Opal
       end
 
       def compile
-        sanitized_value = value.inspect
-        sanitized_value.gsub! /\\u\{[0-9a-f]+\}/ do |char|
+        sanitized_value = value.inspect.gsub /\\u\{[0-9a-f]+\}/ do |char|
           @compiler.warning("Ignoring unsupported character #{char}", @sexp.line)
           ''
         end
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -52,23 +52,7 @@ module Opal
       end
 
       def compile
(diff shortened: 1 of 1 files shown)


jgaskins:
Well, it looks like it does happen sometimes. All the emoji I tested worked. Maybe it's just undisplayable characters it escapes the Unicode values for?

jgaskins:
Just added a commit that allows characters that MRI doesn't escape to be embedded directly into the string. This allows us to use emoji in our Ruby code without breaking on undisplayable characters.

Makes me wonder, though, does it make sense to ignore invalid characters with a warning or would it make more sense to raise an exception? I think a warning could get lost in the wash of messages in development with dozens of asset requests flying by in the console, but an exception makes it really obvious why your app isn't displaying what you expect.

elia:
@jgaskins can you add back some explaining comment just above the `sanitized_value` assignment? I think it's good to have something there for future readers 👓 

> Makes me wonder, though, does it make sense to ignore invalid characters with a warning or would it make more sense to raise an exception? I think a warning could get lost in the wash of messages in development with dozens of asset requests flying by in the console, but an exception makes it really obvious why your app isn't displaying what you expect.

Agree, although I suspect the ruby spec wouldn't load anymore if we do that. As a solution it's probably worth having a look at this post and see if there's a way to translate any char to JS (I just skimmed it but probably converting `"\u{1F404}"` in Ruby to `"\uD83D\uDC04"` in JS would work).

elia:
(Note: `"\u{24B62}"` is currently skipped and could be useful in a test case for this)

jgaskins:
Sure, I can add the comment. How does that codepoint conversion work, though? That seems to be the ideal solution.

Also, the `"\u{hex}"` string seems to work in browsers and the version of Node.js that I have installed, but some other tools do seem to choke on it (for example, whatever version of Node is running specs on Travis CI, as seen in the spec failure in my first commit on this PR), but eventually this entire workaround may be able to be removed.

elia:
> How does that codepoint conversion work, though? That seems to be the ideal solution.

Not leaving out the link would have helped I gues 😅 – here it is:

http://www.2ality.com/2013/09/javascript-unicode.html

jgaskins:
Awesome, thanks. I'll play with it this week.

elia:
@jgaskins took license of doing it myself in: https://github.com/opal/opal/commit/5756e4a8f275069abea8648673521cc69db24f80

I'll rebase/squash/merge manually after the tests pass.

jgaskins:
Awesome. I ported the code from the JS, too, but forgot to push it. Eh, whatever makes it work. :-)

elia:
merged in cb00d70

PR #1615, opened 2017-01-08
Convert Node.js' IO errors to Ruby's IOError object

Mogztter:
When running Opal in a Node.js environment the following code is not working as expected:

```ruby
def try_read
  ::File.read("missing") # The file "missing" is missing (obviously), so the code will throw an exception...
rescue IOError
  # ... but in a Node.js environment the exception thrown is *not* an IOError.
  puts "As a result this code will not be executed :'("
end
```

The reason is that Node.js errors (for instance `ENOENT` https://nodejs.org/api/errors.html) are not an instance of Ruby's `IOError`.

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -19,6 +19,20 @@
     warnings[string] = true;
     #{warn(`string`)};
   }
+  function executeIOAction(action) {
+    try {
+      return action();
+    } catch (error) {
+      if (error.code === 'EACCES' ||
+          error.code === 'EISDIR' ||
+          error.code === 'EMFILE' ||
+          error.code === 'ENOENT' ||
+          error.code === 'EPERM') {
+        throw Opal.IOError.$new(error.message)
+      }
+      throw error;
+    }
+  }
 }
 
 class File < IO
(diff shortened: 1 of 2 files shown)


elia:
The solution is in the right direction, I'd just ask you to:

- [ ] avoid introducing new constants in the global namespace (`IOAction`); instead if absolutely necessary keep it under the `Nodejs` module
- [ ] consider using JS functions instead of ruby methods, similarly to what's done with [this helper](https://github.com/opal/opal/blob/fab514abff90394d209c831f0125e2192e6450d9/opal/corelib/array.rb#L10-L18) in the `Array` class
- [ ] consider using a check on `error.code` or `error.errno` to interpret the error as an IOError

Mogztter:
@elia  Thanks for your review.

> avoid introducing new constants in the global namespace (IOAction); instead if absolutely necessary keep it under the Nodejs module

I've moved my code in the header of `stdlib/nodejs/file.rb` file.

> consider using JS functions instead of ruby methods, similarly to what's done with this helper in the Array class

Done.

>  consider using a check on error.code or error.errno to interpret the error as an IOError

Now that i'm using JS function, I can use `error.code`.

elia:
Great job! 👍🏼

Mogztter:
:blush:

PR #1624, opened 2017-01-21
Reduce stub method calls on startup

jgaskins:
In an effort to reduce startup time (it's only about 60-80ms in Safari on my machine, which is pretty great, but it's nearly 300ms in Chrome), I noticed that `Opal.add_stubs` monopolized CPU time on loading the runtime.

This PR reduces the number of iterations within this function on startup by almost 4x in real-world apps. A Clearwater app was setting almost 11k stubs. With this patch, it was reduced to just over 2800. Since Opal sets up stubs for methods that are either defined or called within a file, it isn't necessary for all subscribers to receive all stubs. I couldn't find any adverse effects of this patch, so I figured I'd send it your way.

Admittedly, it doesn't have a huge effect on startup time, but it does shave off 20-40ms in Chrome. I'm still investigating ways to improve this further.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1189,19 +1189,23 @@
     var subscriber, subscribers = Opal.stub_subscribers,
         i, ilength = stubs.length,
         j, jlength = subscribers.length,
-        method_name, stub;
+        method_name, stub,
+        opal_stubs = Opal.stubs;
 
     for (i = 0; i < ilength; i++) {
       method_name = stubs[i];
-      // Save method name to populate other subscribers with this stub
-      Opal.stubs[method_name] = true;
-      stub = Opal.stub_for(method_name);
 
-      for (j = 0; j < jlength; j++) {
-        subscriber = subscribers[j];
+      if(!opal_stubs.hasOwnProperty(method_name)) {
+        // Save method name to populate other subscribers with this stub
+        opal_stubs[method_name] = true;
+        stub = Opal.stub_for(method_name);
 
-        if (!(method_name in subscriber)) {
-          subscriber[method_name] = stub;
+        for (j = 0; j < jlength; j++) {
+          subscriber = subscribers[j];
+
+          if (!(method_name in subscriber)) {
+            subscriber[method_name] = stub;
+          }
         }
       }
     }


elia:
That's great, just noticed is failing on gem conflict on jruby that's solved on master. Could you please rebase so I can merge when it passes all checks?

(I'll also do a proper review once I reach a keyboard)

opal/corelib/runtime.js
```diff
@@ -1189,19 +1189,23 @@
...
-      for (j = 0; j < jlength; j++) {
-        subscriber = subscribers[j];
+      if(!opal_stubs.hasOwnProperty(method_name)) {
+        // Save method name to populate other subscribers with this stub
+        opal_stubs[method_name] = true;
+        stub = Opal.stub_for(method_name);
 
-        if (!(method_name in subscriber)) {
-          subscriber[method_name] = stub;
+        for (j = 0; j < jlength; j++) {
```
elia: A subscriber could have been added in the meanwhile, e.g. by:

```rb
class OpalJQuery < `jQuery`
```

so the subscriber block should be either brought out or use some tracking system to check which subscribers are already up-to-date.

Perhaps the stub body could be saved in `Opal.stubs` and resused for new subscribers, that should help both memory and cpu.

@jgaskins thoughts?

elia:
(build is green after a rebase on master: https://travis-ci.org/opal/opal/builds/198659972 btw)

opal/corelib/runtime.js
jgaskins: I tried a bunch of things to make this fail and I couldn't, but you're probably a lot more familiar with the bootstrapping process than I am. Do you know if this scenario fails with this PR?

The flow, as I understand it, would be something like, given the following file:

```ruby
require 'opal-jquery'

# I don't actually know if this API is right anymore, it's been a while since I used opal-jquery.
Element['#element'].not_actually_a_method!
```

When Opal runs `Opal.add_stubs(['not_actually_a_method!'])`, `OpalJQuery` will already be registered as a subscriber by the `require` line at the top, right?

elia: > When Opal runs Opal.add_stubs(['not_actually_a_method!']), OpalJQuery will already be registered as a subscriber by the require line at the top, right?

yes it would, the scenario I had in mind is this:

```rb
%x{
  var Foo = function() {}
  Foo.prototype.bar = function(){return 'bar'}
}

class Foo < `Foo`
  alias_native :bar
end

Object.new.baz rescue p $! # should be MethodMissing
Foo.new.baz rescue p $! # should be MethodMissing
```

where a new _subscriber_ (the bridged class) is added after the stub is created, but turns out it's not a problem, in fact testing the code above against this PR I get the correct behavior:

```
opal:pr/1624 ⤑ pbpaste | be bin/opal                                                                                                                                                                    ~/C/opal
#<NoMethodError: undefined method `baz' for #<Object:0x126>>
#<NoMethodError: undefined method `baz' for #<Foo:0x128>:Foo>
```

elia:
@jgaskins 👍🏼

jgaskins:
Nice, I thought there might've been a scenario I'd missed.

PR #1625, opened 2017-01-22
Add "$$" context to Native docs

jmhmccr:
Since global variables aren't accounted for in yard, add some
extra context to the top level `Native` docs.


elia:
Thanks!!

PR #1627, opened 2017-01-27
Make a couple optimizations in Array

jgaskins:
- `Array#initialize` was reassigning function arguments — Ruby default positional args compiles to reassignment of JS function args. This leads to a deoptimization when also using the `arguments` object.
- `Array#-` called `Array#clone` if the other array is empty, which is an expensive operation. It's cheaper to use `JS.Array.prototype.slice`.

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -21,7 +21,13 @@ class Array < `Array`
     `toArraySubclass(objects, self)`
   end
 
-  def initialize(size = nil, obj = nil, &block)
+  def initialize(*args, &block)
+    # Don't take these in as method args because they were getting reassigned
+    # in the compiled JS (to nil if they weren't passed in). Reassigning
+    # function arguments while also referencing the arguments object in the JS
+    # function results in a deoptimization.
+    size, obj = args[0], args[1]
+
     %x{
       if (size > #{Integer::MAX}) {
         #{raise ArgumentError, "array size too big"}
@@ -161,7 +167,7 @@ class Array < `Array`
     end
 
     return [] if `self.length === 0`
-    return clone.to_a if `other.length === 0`
+    return `self.slice()` if `other.length === 0`
 
     %x{
       var result = [], hash = #{{}}, i, length, item;
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -22,11 +22,12 @@ class Array < `Array`
   end
 
   def initialize(*args, &block)
-    # Don't take these in as method args because they were getting reassigned
-    # in the compiled JS (to nil if they weren't passed in). Reassigning
-    # function arguments while also referencing the arguments object in the JS
-    # function results in a deoptimization.
-    size, obj = args[0], args[1]
+    # Don't take these in as optional method args because they were getting
+    # reassigned in the compiled JS (to nil if they weren't passed in).
+    # Reassigning function arguments while also referencing the arguments object
+    # in the JS function results in a deoptimization.
+    size = `typeof args[0] === 'undefined' ? nil : args[0]`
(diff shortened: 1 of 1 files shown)


Mogztter:
Performance! 🤗
Did you run some benchmarks before/after ?

jgaskins:
@Mogztter Good point, I completely forgot. I found them as bottlenecks in one of my apps with the Chrome profiler. With these changes, they were no longer at the top of that trace.

Running some microbenchmarks shows another story with the `Array.new` thing, though. In iterations per second (higher is better):

### Chrome 55

`master`:

- `Array.new()`: 281,308
- `Array.new(10)`: 101,796
- `Array#-`, empty arg: 3,630

This PR:

- `Array.new()`: 192,434 (-31.5%)
- `Array.new(10)`: 95,083 (-6.6%)
- `Array#-`, empty arg: 690,247 (+19,015%)

### Safari 10.1

`master`:

- `Array.new()`: 458,980
- `Array.new(10)`: 301,081
- `Array#-`, empty arg: 9,975

This PR:

- `Array.new()`: 327,852 (-28.6%)
- `Array.new(10)`: 256,562 (-14.8%)
- `Array#-`, empty arg: 1,805,971 (+18,105%)

I definitely want to test the `Array.new` cases more because they _did_ improve my app's performance in both browsers. The subtracting an empty array thing, though, is … well … even better than I realized. :-)

Mogztter:
Wow that's impressive!
Out of curiosity, how do you setup your microbenchmarks ?

jgaskins:
I have a Clearwater app that runs them and renders the results. I'll push the repo when I get a chance today.

jgaskins:
Some more thoughts on this while they're in my head:

- it's possible that deoptimized functions run slower with the profiler running depending on the reason for the deopt.
- it's also possible that `a, b = c, d` simply compiles to slower code and it should be done on two separate lines if performance is the goal.
- also, I used the Ruby array subscript operator rather than the JS one, which is indeed a fair bit slower due to integer index coercion, bounds checking, negative subscripts, etc. I should've remembered this from the work I did in #1072.
- I probably shouldn't have used splat-args at all.

jgaskins:
@Mogztter Pushed the [benchmarks repo](https://github.com/jgaskins/clearwater_ruby_benchmarks), individual benchmarks are in [`measurement.rb`](https://github.com/jgaskins/clearwater_ruby_benchmarks/blob/e82eda84265e22b7c3e2764edbc3ce83ab9413ea/assets/js/measurement.rb#L4-L71).

After screwing around with it a bit more, I don't think I can get `Array#initialize` to go faster reliably. I'll just remove that part of the diff until I've got some to investigate more. The `Array#-` part is definitely better and was easily the largest part of the trace that got me to look into this.

jgaskins:
@elia Yep, `clone` is meant to be literally a clone, as if you were copying DNA, so all singleton methods would go, too. `dup` is more like `obj.class.allocate` + copy ivars.

However, subtracting an array from another array simply returns a new array with none of the dup/clone magic at all, similar to how a lot of `Enumerable` methods return arrays even if the object isn't an array. So we can cut out all of the extra work involved. :-)

```ruby
[1] pry(main)> a = []
=> []
[2] pry(main)> a.instance_variable_set(:@foo, true)
=> true
[3] pry(main)> (a - []).instance_variables
=> []
[4] pry(main)> a.instance_variables
=> [:@foo]
[5] pry(main)> a.define_singleton_method(:bar) { 1 }
=> :bar
[6] pry(main)> (a - []).bar
NoMethodError: undefined method `bar' for []:Array
from (pry):6:in `__pry__'
```

elia:
Thanks for the clarification!

PR #1628, opened 2017-01-29
Optimize comparison on literal values

Mogztter:
Given the follwing code:

```ruby
twenty_two_int = 22
twenty_three_int = 23
twenty_three_string = '23'
zero_point_one = 0.1
if 23 == twenty_three_int
  puts '23 == twenty_three_int'
end
if twenty_three_int == twenty_three_int
  puts 'twenty_three_int == twenty_three_int'
end
if '23' == twenty_three_string
  puts "'23' == twenty_three_string"
end
if 23 == twenty_two_int
  puts '23 == twenty_two_int'
end
if 0.1 == zero_point_one
  puts '0.1 == zero_point_one'
end
```

**Before**

```javascript
(function(Opal) {
  twenty_two_int = 22;
  twenty_three_int = 23;
  twenty_three_string = "23";

  if ((23)['$=='](twenty_three_int)) {
    self.$puts("23 == twenty_three_int")};

  if (twenty_three_int['$=='](twenty_three_int)) {
    self.$puts("twenty_three_int == twenty_three_int")};

  if ("23"['$=='](twenty_three_string)) {
    self.$puts("'23' == twenty_three_string")};

  if ((23)['$=='](twenty_two_int)) {
    return self.$puts("23 == twenty_two_int")};

  if ((0.1)['$=='](zero_point_one)) {
    return self.$puts("0.1 == zero_point_one")
  } else {
    return nil
  };
})(Opal);
```

**After**

```javascript
(function(Opal) {
  twenty_two_int = 22;
  twenty_three_int = 23;
  twenty_three_string = "23";

  if ((23)==(twenty_three_int)) {
    self.$puts("23 == twenty_three_int")};

  if (twenty_three_int['$=='](twenty_three_int)) {
    self.$puts("twenty_three_int == twenty_three_int")};

  if ("23"==(twenty_three_string)) {
    self.$puts("'23' == twenty_three_string")};

  if ((23)==(twenty_two_int)) {
    return self.$puts("23 == twenty_two_int")};

  if ((0.1)==(zero_point_one)) {
    return self.$puts("0.1 == zero_point_one")
  } else {
    return nil
  };
})(Opal);
```

### DIFF
--- a/lib/opal/nodes/call.rb
+++ b/lib/opal/nodes/call.rb
@@ -17,6 +17,9 @@ module Opal
       OPERATORS = { :+ => :plus, :- => :minus, :* => :times, :/ => :divide,
                     :< => :lt, :<= => :le, :> => :gt, :>= => :ge }
 
+      # JavaScript comparisons operators that get optimized
+      COMPARISON_OPERATORS = [ '==', '!=', '>', '<', '>=', '<=']
+
       def self.add_special(name, options = {}, &handler)
         SPECIALS[name] = options
         define_method("handle_#{name}", &handler)
@@ -143,7 +146,15 @@ module Opal
       end
 
       def compile_simple_call_chain
-        push recv(receiver_sexp), method_jsid, "(", expr(arglist), ")"
+        if is_literal_type(receiver_sexp) && COMPARISON_OPERATORS.include?(meth.to_s)
+          push recv(receiver_sexp), meth.to_s, "(", expr(arglist), ")"
+        else
+          push recv(receiver_sexp), method_jsid, "(", expr(arglist), ")"
+        end
+      end
+
+      def is_literal_type(sexp)
+        sexp.type == 'int' || sexp.type == 'str' || sexp.type == 'float'
       end
 
       def splat?
--- a/lib/opal/nodes/call.rb
+++ b/lib/opal/nodes/call.rb
@@ -18,7 +18,7 @@ module Opal
                     :< => :lt, :<= => :le, :> => :gt, :>= => :ge }
 
       # JavaScript comparisons operators that get optimized
-      COMPARISON_OPERATORS = [ '==', '!=', '>', '<', '>=', '<=']
+      COMPARISON_OPERATORS = [ '==', '!=', '>', '<', '>=', '<=', '===']
 
       def self.add_special(name, options = {}, &handler)
         SPECIALS[name] = options
@@ -154,7 +154,7 @@ module Opal
       end
 
       def is_literal_type(sexp)
-        sexp.type == 'int' || sexp.type == 'str' || sexp.type == 'float'
+        sexp.type == :int || sexp.type == :str || sexp.type == :float
       end
 
       def splat?


Mogztter:
@elia Could you please restart the build on Appveyor ?

elia:
@Mogztter build restarted and passed (also if you sign up I can add you as a collaborator)

Although I'm in favor of optimizing I'm not sure this is a good idea, the problem I see with this approach are:

- using `==` instead of `===` (trivial) 
- no support for method override
- it should live behind a flag/compiler-option

Another possible approach for optimising these operators could be to implement `#==` and other methods directly on `Number` and return early when values are identical (MRI does something similar). 

Let me know your thoughts and if you want to give it a shot.

/cc @opal/core

Mogztter:
> build restarted and passed (also if you sign up I can add you as a collaborator)

I signed up with my GitHub account (Mogztter) on Appveyor. Do you need my email ?

> Although I'm in favor of optimizing I'm not sure this is a good idea, the problem I see with this approach are...

You mean, you are not in favor even if I use `===` and a flag/compiler option ?

> Another possible approach for optimising these operators could be to implement #== and other methods directly on Number...

I think this is already the case [number.rb#L264-L276](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/number.rb#L264-L276) ?

> ...  and return early when values are identical (MRI does something similar).

The issue is that we are doing extra work because the values are *not* identical:

 1. [Kernel#===](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/kernel.rb#L14-L16) computes `(self * 2) + 1` on both Number (object_id)
2. [Number#==](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/number.rb#L266-L268) checks if "other" is a number and wraps it into a `Number`
3. Compares the values

(I need to check again the exact stack trace, I may have skipped some steps)

> Let me know your thoughts and if you want to give it a shot.

Yes I want to give it a shot!
The compilation is really slow since we switch to whitequark/parser and I won't be able to use 0.11 if I'm unable to improve the performance during the compilation.
Having said that, this optimization should also improve performance during runtime because we are removing extra layers but as you mentioned there's no support for method override 😒

elia:
> I signed up with my GitHub account (Mogztter) on Appveyor. Do you need my email ?

I tried but I can't add you to just opal/opal for now I'll keep doing it manually when needed

> You mean, you are not in favor even if I use `===` and a flag/compiler option ?

No, it's that I consider compiler options as the last resort, they make the code worse and it's very difficult to test the project with every flag combination. So I consider worth trying other ways first and see if we reach the same result or a decent compromise.

> I think this is already the case [number.rb#L264-L276](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/number.rb#L264-L276) ?

Missed it, sorry 😁.

> ...  and return early when values are identical (MRI does something similar).

The issue is that we are doing extra work because the values are *not* identical:

> 1. Kernel#===

I think [BasicObject#==](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/basic_object.rb#L5-L7) is the one that's being used, did I miss something?

> 2. [Number#==](https://github.com/opal/opal/blob/a7620e2fce4a9ab030d5e5e9dc4a03ac19d3432d/opal/corelib/number.rb#L266-L268) checks if "other" is a number and wraps it into a `Number`

**Maybe** this is where we can try something else, maybe doing `return self.valueOf === other.valueOf;` is enough faster. Also we should check if we can leverage `typeof(other) === 'number'` instead of `other.$$is_number` to make it faster.

For perf microbenchmarks you can try:
- https://jsbench.github.io
- https://jsperf.com (which seems to be kinda back online)

> Yes I want to give it a shot!

🙇

Mogztter:
I've created a simple benchmark on jsperf.com:

https://jsperf.com/opal-builder-0-10-3 ([v0.10.3](https://github.com/opal/opal/tree/v0.10.3))
https://jsperf.com/opal-builder-master-3392ee2a (https://github.com/opal/opal/commit/3392ee2a35a2d5a42ed5fde47992c564a0981511)

Since Opal is registered as a global variable I can't compare both Opal 0.10.3 and Opal master on the same benchmark without  changing the code.

opal0103

opal-master-3392ee2a

I'll work on my pull request to see if I can improve the performance.

Mogztter:
> I think BasicObject#== is the one that's being used, did I miss something?

@elia I can confirm that the following code is calling `Kernel#===` on master branch. 

```javascript
$case = _lex_from_state_actions['$[]'](self.cs);
if ((80)['$===']($case)) {
  // ...
```

Mogztter:
I think it's too extreme for the performance gain, closing 🚪

PR #1632, opened 2017-02-12
Returns earlier on force_encoding to improve performance

Mogztter:
Before: https://jsperf.com/opal-force-encoding-master-3392ee2a

master-3392

After: https://jsperf.com/opal-force-encoding-master-52c0f3e0

after


Mogztter:
That was quick! Thanks @meh 😉

PR #1633, opened 2017-02-15
Let Hash.new take `Object.create(null)` JS objects

ajjahn:
Since JS objects created with a `null` prototype are essentially a pure bag of keys and values, they should behave the same as JS object literals when given to `Hash.new`.


elia:
Thanks!

PR #1635, opened 2017-02-18
Resolves #1630, Escapes meta characters in RegExp

### LINKED ISSUE #1630: Invalid regular expression in PathReader.expand
### Stack trace
`SyntaxError: Invalid regular expression: /\A\.?\.\/: \ at end of pattern`

### Code

```ruby
    def expand(path)
      if Pathname.new(path).absolute? || path =~ %r{\A\.?\.#{File::SEPARATOR}}
        path
      else
        file_finder.find(path)
      end
    end
```
https://github.com/opal/opal/blob/d2883ad0de65461d6f17e1cf233e9a59b2b1e545/lib/opal/path_reader.rb#L16

### JavaScript

The generated code is:
```javascript
new RegExp("\\A\\.?\\.\\");
```

### Fix

When running Node.js on Windows `File.SEPARATOR` is equals to `\\` ([nodejs/file.rb#L49](https://github.com/opal/opal/blob/d2883ad0de65461d6f17e1cf233e9a59b2b1e545/stdlib/nodejs/file.rb#L49)).
I think we need to escape the double backslashes since we are using the RegExp constructor: `new RegExp("\\A\\.?\\.\\\\");`

Mogztter:
Especially backslashes!
When running on Windows with Node.js, File::SEPARATOR is equals to `path.sep = '\\'`.

Reference: https://nodejs.org/api/path.html#path_path_sep

Fixes #1630

### DIFF
--- a/lib/opal/path_reader.rb
+++ b/lib/opal/path_reader.rb
@@ -3,7 +3,7 @@ require 'opal/hike_path_finder'
 
 module Opal
   class PathReader
-    RELATIVE_PATH_REGEXP = %r{#{Opal::REGEXP_START}\.?\.#{File::SEPARATOR}}
+    RELATIVE_PATH_REGEXP = %r{#{Opal::REGEXP_START}\.?\.#{Regexp.quote File::SEPARATOR}}
 
     def initialize(file_finder = HikePathFinder.new)
       @file_finder = file_finder


lib/opal/path_reader.rb
```diff
@@ -3,7 +3,7 @@
 
 module Opal
   class PathReader
-    RELATIVE_PATH_REGEXP = %r{#{Opal::REGEXP_START}\.?\.#{File::SEPARATOR}}
+    RELATIVE_PATH_REGEXP = %r{#{Opal::REGEXP_START}\.?\.#{Regexp.quote File::SEPARATOR}}
```
Mogztter: Another way to do it is to use `File::SEPARATOR.gsub('\\', '\\\\\\\\')`.
But `Regexp.quote` escapes any characters that would have special meaning in a regular expression (general purpose) + this method is already used in other places in the code.

Mogztter:
👍

PR #1637, opened 2017-02-21
optimize hash

elia:
https://gist.github.com/435d1b7e7d840035d53b91d6583cdee3

before:

```
2017-02-bwl21:master ⤑ opal -I. main.rb                                                                                                                                                    ~/C/o/_/2017-02-bwl21
36.244 ruby
0.071 convert ruby to js
0.362 javascript
result is equal: true
```

after:
```
2017-02-bwl21:master ⤑ opal -I. main.rb                                                                                                                                                    ~/C/o/_/2017-02-bwl21
11.121 ruby
0.081 convert ruby to js
0.361 javascript
result is equal: true
```

cc @bwl21

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -1136,7 +1136,7 @@ class Array < `Array`
 
   def hash
     %x{
-      var top = (Opal.hash_ids == undefined),
+      var top = (Opal.hash_ids === undefined),
           result = ['A'],
           hash_id = self.$object_id(),
           item, i, key;
@@ -1169,7 +1169,7 @@ class Array < `Array`
         return result.join(',');
       } finally {
         if (top) {
-          delete Opal.hash_ids;
+          Opal.hash_ids = undefined;
         }
       }
     }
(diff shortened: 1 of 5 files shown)


opal/corelib/array.rb
```diff
@@ -1136,40 +1136,39 @@ def flatten!(level = undefined)
...
-          result.push(item.$hash());
+          result.push(item.$hash(), ',');
         }
 
-        return result.join(',');
+        return result.join();
       } finally {
         if (top) {
-          delete Opal.hash_ids;
+          Opal.hash_ids = undefined;
```
meh: Is there a reason for this?

elia: @meh yeah, I was looking at the time spent per line in the chrome profiler, I'm not an expert but my guess was that changing the number of keys is more costly than an assignment.

jgaskins: Deleting a property causes the VM to put it in "dictionary" or "hash table" mode. Iterating over it with `for(var foo in bar)` triggers a de-opt for that function. :-\ You'd think that, based on what those words mean, objects created with `Object.create(null)` would be in this mode by default, but I guess they aren't until you `delete` from them.

https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#52-the-object-being-iterated-is-not-a-simple-enumerable

PR #1638, opened 2017-02-25
Optimize Class#new and aliased methods

jgaskins:
This PR isn't necessarily intended to be merged as is, but it wouldn't be a terrible thing if it were (except for the commented-out code I left in there). The main thing is that I noticed a de-opt in both `Class#new` and the aliased-method wrapper. In both cases, I noticed the block was being assigned directly and that `method.apply` was being used. I happened to notice that that's literally what `Opal.send` does and it somehow avoids a de-opt, so I went ahead and rewrote them in terms of `Opal.send`. Sure enough, no longer deoptimized. I have no idea why.

Running a quick benchmark using the Chrome profiler (a Clearwater app rendering about 2,000 components 25x):

| | `new_body` | `Class#new` |
|-|----|----|
| before | 60.8ms | 77.4ms |
| after | 11.3ms | 7.3ms |

So we're looking at about a 5x speedup in using method aliases (in this case, it was mostly `Array#map`) and a 10x speedup in `Class#new`. I think the selective clearing of `$$p` might have helped (the [optimizer can be touchy](https://gist.github.com/Hypercubed/89808f3051101a1a97f3) with function properties), but I haven't tested it exhaustively. If so, it could probably be built into the compiler.

The primary thing I'm wondering is whether using this pattern (`Opal.send` all the things) is preferable to manually setting up blocks and calling `function.apply`.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1817,9 +1817,16 @@
     }
 
     new_body = function() {
-      body.$$p = new_body.$$p;
-      new_body.$$p = null;
-      return body.apply(this, arguments);
+      var block = new_body.$$p;
+      var args = new Array(arguments.length);
+
+      for(var index = 0; index < arguments.length; index++) {
+        args[index] = arguments[index];
+      }
+
+      if(block) { new_body.$$p = null }
+
+      return Opal.send(this, body, args, block);
     };
 
     new_body.length = body.length;
--- a/opal/corelib/class.rb
+++ b/opal/corelib/class.rb
@@ -34,15 +34,28 @@ class Class
   def inherited(cls)
   end
 
-  def new(*args, &block)
-    %x{
-      var obj = #{allocate};
+  %x{
+    Opal.defn(self, '$new', function Class$new() {
+      var self = this;
+      var object = #{allocate};
+      var block = Class$new.$$p;
+      var args = new Array(arguments.length);
+      for(var index = 0; index < arguments.length; index++) {
+        args[index] = arguments[index];
+      }
 
-      obj.$initialize.$$p = block;
-      obj.$initialize.apply(obj, args);
-      return obj;
-    }
-  end
+      if(block) { Class$new.$$p = null }
+
+      Opal.send(object, object.$initialize, args, block);
+
+      return object;
+    });
+  }
+  # def new(*args, &block)
+  #   object = allocate
+  #   `Opal.send(#{object}, #{object}.$initialize, args, block)`
+  #   object
+  # end
 
   def superclass
     `self.$$super || nil`


meh:
Just a heads up, we have a PR in-flight (#1637) that touches that `new_body` stuff.

meh:
I'm okay with `Opal.send` all the things, especially if it makes a difference wrt performance.

I feel less okay with the current *hand compiled* `Class.new` we have here, moving it onto the code generator might be a better idea as you mention, since it's likely to improve performance across the board instead of just for that specific function.

Could you maybe make a diff or point out the differences between the code that's currently generated and the *hand compiled* version?

meh:
Also we should probably merge the `new_body` changes separately so the two pull requests don't have to stomp eachother's toes, since there's clearly a performance win there.

opal/corelib/runtime.js
```diff
@@ -1817,9 +1817,16 @@
     }
 
     new_body = function() {
-      body.$$p = new_body.$$p;
-      new_body.$$p = null;
-      return body.apply(this, arguments);
+      var block = new_body.$$p;
+      var args = new Array(arguments.length);
+
+      for(var index = 0; index < arguments.length; index++) {
```
elia: @jgaskins can you cache the length argument too, e.g. `for (var i = 0, ii = arguments.length; i < ii; i++)`?

Or otherwise do we have a preferred way of dealing with for loops? 
I think this is worth standardizing to let the special cases stand out (e.g. descending index loops). /cc @meh

elia:
@jgaskins great stuff, I was looking at ways to optimize the alias body wrapper and was even ready to go back to using the original body (see 2c4827313) at the cost of being less compliant.

Looking at `Class.new` here's the diff with the generated code:

```rb
class Class
  def new(*args, &block)
    %x{
      var obj = #{allocate};

      obj.$initialize.$$p = block;
      obj.$initialize.apply(obj, args);
      return obj;
    }
  end
end
```

```js
  (function(Opal) {
    var self = Opal.top, $scope = Opal, $scopes = [Opal], nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice, $klass = Opal.klass;

    Opal.add_stubs(['$allocate']);
    return (function($base, $super, $visibility_scopes) {
      function $Class(){};
      var self = $Class = $klass($base, $super, 'Class', $Class);

      var def = self.$$proto, $scope = self.$$scope, $scopes = $visibility_scopes.slice().concat($scope), TMP_Class_new_1;

      return (Opal.defn(self, '$new', TMP_Class_new_1 = function($a_rest) {
        var self = this, args, $iter = TMP_Class_new_1.$$p, block = $iter || nil;

        var $args_len = arguments.length, $rest_len = $args_len - 0;
        if ($rest_len < 0) { $rest_len = 0; }
        args = new Array($rest_len);
        for (var $arg_idx = 0; $arg_idx < $args_len; $arg_idx++) {
          args[$arg_idx - 0] = arguments[$arg_idx];
        }
        TMP_Class_new_1.$$p = null;
      
        var obj = self.$allocate();

        obj.$initialize.$$p = block;
        obj.$initialize.apply(obj, args);
        return obj;
    
      }, TMP_Class_new_1.$$arity = -1, TMP_Class_new_1.$$parameters = [['rest', 'args'], ['block', 'block']]), nil) && 'new'
    })($scope.base, null, $scopes)
  })(Opal);
```

_generated with `pbpaste | be bin/opal -cEOA | pbcopy`_


Except the fact that `Class$new` looks very good seems that the main diff is `Opal.send`, so I'd keep just that change there.

opal/corelib/runtime.js
jgaskins: Sure, that makes sense. This PR was mostly to throw ideas at the wall to see if they stick. :-)

I think I checked the performance tradeoff between allocating another local variable in the stack and calling `arguments.length` a long time ago for a PR. I remember the local variable was sometimes slightly faster and sometimes break-even depending on the browser, but never slower, so it's probably a good idea especially for a method that is likely to be called within a loop.

jgaskins:
@elia Using the original body was something I added in #1204 because I didn't understand why it was wrapped and, based on the comments, nobody else was sure, either. :-) I was surprised to see the wrapper was added back in, but since you mentioned it being less compliant, it sounds like you found the purpose for it?

The other (minor) difference besides `Opal.send` between the current wrapper and the one in this PR is unconditional vs conditional clearing of `$$p`. I'm not sure if that's what was causing the de-opt, but it's possible.

elia:
> it sounds like you found the purpose for it?

Yes, basically using a different body helps with `Method` and `UnboundMethod` returning the correct owner. Also I think it's useful when dealing with super. A way to clone a function would have helped a lot with performance here.

> The other (minor) difference besides Opal.send between the current wrapper and the one in this PR is unconditional vs conditional clearing of $$p. I'm not sure if that's what was causing the de-opt, but it's possible.

Sure sorry, what I really meant is to keep wrap the relevant changes in the usual `def self.` wrap.

@jgaskins ping me when done and I'll merge.

elia:
@jgaskins I added the required changes and a further alias optimization in:

- d4676e2f67 (Avoid Chrome deoptim by conditionally reass. block)
- 49026aab3d (Improve docs and rename method wrapper in .alias())
- fbe1e58b01 (Avoid long wrapper chains in Opal.alias())

PR #1640, opened 2017-03-02
Number#=== is called instead of Kernel#=== on switch/case with Numeric

### LINKED ISSUE #1639: Kernel#=== is called on Numeric comparison (switch/case)
Follow-up: https://github.com/opal/opal/pull/1628

#### How to reproduce

.numeric-switch.rb
```ruby
value = 12

case value 
when 80 then
  p 'Hello!'
else
  p 'Bonjour!'
end
```

Edit `opal/corelib/kernel.rb`

```diff
   def ===(other)
+    p 'Kernel#=== is called'
     self.object_id == other.object_id || self == other
   end
```

Run!

```
bundle exec opal -- numeric-switch.rb 
```
```
"Kernel#=== is called"
"Bonjour!"
```

Mogztter:
Resolves #1639

### DIFF
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -261,6 +261,10 @@ class Number < Numeric
     end
   end
 
+  def ===(other)
+    `self === other`
+  end
+
   def ==(other)
     %x{
       if (other.$$is_number) {


meh:
Guess I was too eager to merging because I was going to do a couple changes to `Numeric` 🐼 

Apparently everything breaks with this.

Mogztter:
Yes I'm looking at Travis right now, sorry about that 😕

Mogztter:
I think it's better to revert my changes

meh:
Yeah, reverted, my bad for being so eager 🐼 

The change just looked so innocent, I think it even brought the specs into an infinite loop or something.

PR #1641, opened 2017-03-04
Resolves #1639, define Number#===

### LINKED ISSUE #1639: Kernel#=== is called on Numeric comparison (switch/case)

Mogztter:
@meh Second try :wink: 

#### Benchmark

.benchmark.rb
```ruby
require 'benchmark'

n = 50000
Benchmark.bm do |x|
  x.report { for i in 1..n; [1, 2, 3].grep(1); end }
end
```

```
bundle exec opal -- benchmark.rb
```

Before | After
------------ | -------------
`user     system      total        real` | `user     system      total        real`
`0.173000   0.173000   0.691999 (  0.173000)` | `0.112000   0.112000   0.448000 (  0.112000)`
`0.168000   0.168000   0.672000 (  0.168000)` | `0.113000   0.113000   0.452000 (  0.113000)`
`0.170000   0.170000   0.680000 (  0.170000)` | `0.113000   0.113000   0.452000 (  0.112000)`
`0.169000   0.169000   0.676000 (  0.168000)` | `0.113000   0.113000   0.452001 (  0.114000)`
`0.169000   0.169000   0.676000 (  0.169000)` | `0.113000   0.113000   0.452001 (  0.114000)`
**Average** |
`0,1698	0,1698	0,6791998	0,1696` | `0,1128	0,1128	0,4512004	0,113`

In this case the code is approximatively 50% faster

### DIFF
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -261,6 +261,19 @@ class Number < Numeric
     end
   end
 
+  def ===(other)
+    %x{
+      if (other.$$is_number) {
+        if (typeof other === 'number') {
+          return self.valueOf() === other;
+        }
+        return self.valueOf() === Number(other);
+      } else {
+        return #{self == other};
+      }
+    }
+  end
+
   def ==(other)
     %x{
       if (other.$$is_number) {
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -264,10 +264,7 @@ class Number < Numeric
   def ===(other)
     %x{
       if (other.$$is_number) {
-        if (typeof other === 'number') {
-          return self.valueOf() === other;
-        }
-        return self.valueOf() === Number(other);
+        return self.valueOf() === other.valueOf();
       } else {
         return #{self == other};
       }


Mogztter:
`mspec_ruby_nodejs` is failing but I don't think this is a "random" failure. I can reproduce this failure with `RANDOM_SEED=65408` and `RANDOM_SEED=28686` but not with `RANDOM_SEED=59980` nor `RANDOM_SEED=30356`.

opal/corelib/number.rb
```diff
@@ -261,6 +261,19 @@ def **(other)
     end
   end
 
+  def ===(other)
+    %x{
+      if (other.$$is_number) {
+        if (typeof other === 'number') {
+          return self.valueOf() === other;
+        }
+        return self.valueOf() === Number(other);
```
meh: Could you benchmark this without the `typeof` check and just `self.valueOf() === other.valueOf()`?

Mogztter:
Before | After (valueOf)
------------ | -------------
`user     system      total        real` | `user     system      total        real`
`0.173000   0.173000   0.691999 (  0.173000)` | `0.090000   0.090000   0.360000 (  0.090000)`
`0.168000   0.168000   0.672000 (  0.168000)` | `0.087000   0.087000   0.348001 (  0.087000)`
`0.170000   0.170000   0.680000 (  0.170000)` | `0.087000   0.087000   0.348001 (  0.087000)`
`0.169000   0.169000   0.676000 (  0.168000)` | `0.097000   0.097000   0.388000 (  0.098000)`
`0.169000   0.169000   0.676000 (  0.169000)` | `0.087000   0.087000   0.348000 (  0.086000)`
**Average** |
`0,1698	0,1698	0,6791998	0,1696` | `0,0896	0,0896	0,3584004	0,0896`

That's even faster with @meh's suggestion, 25% faster than the previous implementation (typeof) and close to 90% compared to the actual code (master).

opal/corelib/number.rb
Mogztter: Good suggestion! See below :100:

opal/corelib/number.rb
```diff
@@ -261,6 +261,16 @@ def **(other)
     end
   end
 
+  def ===(other)
+    %x{
+      if (other.$$is_number) {
+        return self.valueOf() === other.valueOf();
+      } else {
+        return #{self == other};
```
Mogztter: NOTE: This line is mandatory in order to satisfy Ruby specs:
 * ["Fixnum calls 'other == self' if the given argument is not a Fixnum"](https://github.com/ruby/spec/blob/906bd11ebf8d75f8dba8824a986cf3d83ceda850/core/fixnum/shared/equal.rb#L13)
 * ["Float calls 'other == self' if coercion fails"](https://github.com/ruby/spec/blob/906bd11ebf8d75f8dba8824a986cf3d83ceda850/core/float/shared/equal.rb#L8)

meh:
Looks good, thanks!

PR #1642, opened 2017-03-11
Adds a module id since seekmodule is bundled in opal.js

Mogztter:
Plus this avoid conflict when using Require.js, see: https://github.com/asciidoctor/asciidoctor.js/pull/280#issuecomment-285859595


elia:
@Mogztter the CI failures are due to master, this change is fine, but before merge some more docs/comments are needed to help future generations (e.g. me in 5 minutes) understand what's going on 😄. 

Also being the JS minified requires to have a clear way of updating the code + patches (see the current comment for an example).

Mogztter:
@elia You are absolutely right!

> to help future generations (e.g. me in 5 minutes) 

I'm the same kind of future generation :rofl: 
I've created a [commit on the seekmodule repository](https://github.com/Mogztter/seedrandom/commit/e047540c3d81f955cab9a01d17b8141d439fbd7d) and update the documentation on "How to update the code".

PR #1644, opened 2017-03-13
cruby_tests has been renamed to test_cruby

Mogztter:
Follow-up of https://github.com/opal/opal/commit/9c168a5010210490c9a6fb6a72950ada359421f5

This should fix the build on Appveyor.


elia:
@Mogztter Thanks!! I think I just checked travis yesterday and probably I had some file filter active when I did the search and replace

PR #1645, opened 2017-03-13
Rewrite logical operator assignment (or_asgn/and_asgn) to condition before compilation.

iliabylich:
`lhs ||= rhs` becomes `lhs = lhs ||= rhs`
`lhs &&= rhs` becomes `lhs = lhs && rhs`
`lhs += rhs` becomes `lhs = lhs + rhs`


lib/opal/rewriters/logical_operator_assignment.rb
```diff
@@ -0,0 +1,199 @@
...
+        def self.process(lhs, rhs, root_type)
+          s(root_type,
+            get(lhs, rhs),
+            set(lhs, rhs))
+        end
+
+        def self.new_temp
+          @@counter ||= 0
+          @@counter += 1
+          :"$or_equals_lhs_recvr_tmp_#{@@counter}"
```
elia: @iliabylich this seems a bit too long, maybe `$oreql_lhs_#{@@counter}`?

Also I wouldn't trust much Opal implementation of cvars… 😏

lib/opal/rewriters/logical_operator_assignment.rb
```diff
@@ -0,0 +1,199 @@
...
+        send:   SendHandler,
+        csend:  ConditionalSendHandler
+      }
+
+      # lhs ||= rhs
+      def on_or_asgn(node)
+        lhs, rhs = *node
+
+        result = HANDLERS
+          .fetch(lhs.type) { raise NotImplementedError }
```
elia: @iliabylich consider this comment as part of a more general problem, we should have an internal error class here, that should be rescue/raised in `Compiler#compile` updating the message to show as much info as possible, e.g. filename/line/column etc., In the meanwhile this is better than nothing 🙂

spec/lib/rewriters/logical_operator_assignment_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+      rewritten = rewrite(input)
+      expected = parse(to)
+
+      expect(rewritten).to eq(expected)
+    end
+  end
+
+  context 'rewriting or_asgn and and_asgn nodes' do
+    context 'local variable' do
+      include_examples 'it rewrites', from: 'a = local; a ||= 1', to: 'a = local; a || a = 1'
```
elia: @iliabylich I always thought it was `a = a || 1` like all other operators, but probably you know better or there's some perf reason…

lib/opal/rewriters/logical_operator_assignment.rb
```diff
@@ -0,0 +1,199 @@
...
+      # Produces `lhs || lhs = rhs`
+      class LocalVariableHandler < AbstractHandler
+        def self.get(lhs, rhs)
+          lhs.updated(:lvar) # lhs
+        end
+
+        def self.set(lhs, rhs)
+          lhs.updated(:lvasgn, [*lhs, rhs]) # lhs = rhs
+        end
+      end
```
elia: @iliabylich maybe this series of classes can be simplified/shortened with this kind of code:

```ruby
class GetSetHandler < AbstractHandler
  def self.get(lhs, rhs)
    lhs.updated(@get_type) # lhs
  end
  def self.set(lhs, rhs)
    lhs.updated(@set_type, [*lhs, rhs]) # lhs = rhs
  end
end

# Takes    `lhs ||= rhs`
# Produces `lhs || lhs = rhs`
class LocalVariableHandler < AbstractHandler
  @get_type = :lvar
  @set_type = :lvasgn
end
```

or even (maybe too far?)

```ruby
class GetSetHandler < AbstractHandler
  def self.get(lhs, rhs)
    lhs.updated(@get_type) # lhs
  end
  def self.set(lhs, rhs)
    lhs.updated(@set_type, [*lhs, rhs]) # lhs = rhs
  end
  
  def self.handle(set_type, get_type)
    @set_type = set_type
    @get_type = get_type
    HANDLERS[set_type] = self
  end
end

# Takes    `lhs ||= rhs`
# Produces `lhs || lhs = rhs`
class LocalVariableHandler < AbstractHandler
  handle :lvasgn, :lvar
end
```

Anyway this should be both simpler and faster (more optimizable) and take less memory (I'm thinking about the JS version mostly)

spec/lib/rewriters/logical_operator_assignment_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+  end
+
+  around(:each) do |e|
+    Opal::Rewriters::LogicalOperatorAssignment::AbstractHandler.reset_tmp_counter!
+    Opal::Rewriter.disable { e.run }
+  end
+  let(:cache_tmp_name) { :$or_equals_lhs_recvr_tmp_1 }
+  let(:cached) { s(:js_tmp, cache_tmp_name) }
+
+  shared_examples 'it rewrites' do |from:, to:|
```
elia: Sadly we still support ruby 2.0 which won't run this code, a bold move could be to skip it for the next release as it's doing rails I think… the first supporting this syntax should be 2.1

lib/opal/rewriters/logical_operator_assignment.rb
iliabylich: I was implementing cvars in Opal, I trust them :smile: 
What do you think about extracting and reusing the code that generates temporary variables from the compiler? Something simialar to this code somewhere in `Opal::TemporaryVariable.next`? I just feel that I reinvent a wheel here. Moreover, this code can be used later in other rewriters.

We could pass an instance of compiler to a rewriter, but for me it looks like a bad design. AST rewriters exist outside of the compiler/parser. Rewriter is kind of a big stateless function that takes one ast returns another one.

lib/opal/rewriters/logical_operator_assignment.rb
iliabylich: I was thinking about something similar initially, but I just don't like the idea of having abstract class that can't do anything on its own. Having an auto-generated registry is a good idea, but there are only 7 classes. I'm not completely against it, I just don't see any reason to compilate this code. Ping me please if you still want to change, I'm ok with it.

lib/opal/rewriters/logical_operator_assignment.rb
iliabylich: Yeah, I would better raise a `Parser::SyntaxError` here, but there's no real loc that could be pointed by this exception. The error may come *only* from other broken rewriter that could be placed before in the chain of `Opal::Rewriter.list`. If the grammar is broken `parser` shows it. But generally - yes, we have enough custom exceptions in `opal/nodes`. Probably we could use `parser.diagnostic` to properly display those errors.

spec/lib/rewriters/logical_operator_assignment_spec.rb
iliabylich: No, it should be `a || a= 1`. Assignment happens only if `a` is nil. I've found it today when I was implementing a case for a `send` node. Writer gets called only when reader returns `nil`.

spec/lib/rewriters/logical_operator_assignment_spec.rb
iliabylich: That is sad, I'll fix it.

spec/lib/rewriters/logical_operator_assignment_spec.rb
elia: > Writer gets called only when reader returns nil.

👍🏼 good to know, maybe I got confused by the fact that for a new lvar it defines the var anyway without giving `NameError: undefined local variable or method ‘a’ for main:Object` (which is the case for `a || a = 1`.

spec/lib/rewriters/logical_operator_assignment_spec.rb
elia: so, I tried to poll the opal/opal gitter room, no one replied so far, so it's probably ok to bump the version requirement for ruby to 2.1. I can do it on master and then you can rebase the PR.

@opal/core any thoughts on it?

lib/opal/rewriters/logical_operator_assignment.rb
elia: Don't do nothing about this yet, but on the same topic we should start adding the to-be-compiled files to the backtrace as they generally do in templating engines like haml (not sure if it's in haml itself or rails).

lib/opal/rewriters/logical_operator_assignment.rb
elia: > I was implementing cvars in Opal, I trust them 😄

Clearly I had old info 😄👍🏼

> What do you think about extracting and reusing the code that generates temporary variables from the compiler?

Can it be a node type (IIRC there's already `:js_tmp`) with a progressing value as child that acts as an id, then the compiler can have its table of tmpvars and return the same jsid for the same numeric id. eg. 

```
node = s(:temporary_var, 23132)
compiler.get_temp_name(node) # => "TMP_12"
```

similarly to what (I think) happens with symbols in MRI

---

Anyway I'm in favor of the extraction

lib/opal/rewriters/logical_operator_assignment.rb
elia: Let's forget the registry for now, I think I still like the more concise first version that uses `ivars`:

```rb
class GetSetHandler < AbstractHandler
  def self.get(lhs, rhs)
    lhs.updated(@get_type) # lhs
  end
  def self.set(lhs, rhs)
    lhs.updated(@set_type, [*lhs, rhs]) # lhs = rhs
  end
end

# Takes    `lhs ||= rhs`
# Produces `lhs || lhs = rhs`
class LocalVariableHandler < GetSetHandler
  @get_type = :lvar
  @set_type = :lvasgn
end
```

spec/lib/rewriters/logical_operator_assignment_spec.rb
iliabylich: Hmmm, I think you are right, it must be `lhs = lhs || rhs` for all kinds of variables, but `lhs || lhs = rhs` for all kinds of method calls. For instance/class/global variables there's no difference, but it definitely must be fixed for local variables.

elia: @iliabylich also consider that `a = 1 unless a; a` actually defines the lvar

iliabylich:
Alright, ready for merging.

PR #1647, opened 2017-03-19
resolves #1646 File.file?, File.directory? and File.size should never return nil

### LINKED ISSUE #1646: Certain nodejs File operations not to spec
If the path does not exist, File.file? and File.directory? should return false, not nil.

If the path does not exist, File.size should throw an exception rather than return nil.

mojavelinux:


elia:
Thanks!

PR #1652, opened 2017-03-29
use keys variable in loop in Opal.hash_clone

mojavelinux:
The keys variable was defined in the loop but never used. This patch uses it. It also stores the smap in a variable to avoid redundant lookups.


elia:
Good catch! Thanks!

Issue #1607, opened 2016-12-16
Failing module specs with some rand seeds


elia:
https://travis-ci.org/opal/opal/jobs/209959078

```
export RANDOM_SEED=25311
bundle exec rake mspec_ruby_nodejs   
```

results:

```
1)
An exception occurred during: loading ruby/core/module/class_eval_spec.rb ERROR
TypeError: wrong argument type Number (expected Module)
TypeError: wrong argument type Number (expected Module)
    at module_constructor.$$include [as $include] (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:3056:16)
    at /tmp/opal-nodejs-runner-20170311-4925-jd4bh7:141096:12
    at /tmp/opal-nodejs-runner-20170311-4925-jd4bh7:141109:7
    at Opal.modules.ruby/core/module/fixtures/classes (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:141990:5)
    at Opal.load (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:2242:7)
    at Object_alloc.Opal.require [as $require] (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:2266:17)
    at Opal.modules.ruby/core/module/class_eval_spec (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:143400:8)
    at Opal.load (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:2242:7)
    at module_constructor.$$load [as $load] (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:4860:19)
    at Object_alloc.TMP_4 (/tmp/opal-nodejs-runner-20170311-4925-jd4bh7:51080:62)
2)
An exception occurred during: before :each
Module#alias_method preserves original super call after alias redefine ERROR
NameError: uninitialized constant ModuleSpecs::Aliasing
Aliasing: uninitialized constant ModuleSpecs::Aliasing
3)
An exception occurred during: before :each
Module#alias_method preserves the arguments information of the original methods ERROR
NameError: uninitialized constant ModuleSpecs::Aliasing
Aliasing: uninitialized constant ModuleSpecs::Aliasing
4)
An exception occurred during: before :each
Module#alias_method works in module ERROR
NameError: uninitialized constant ModuleSpecs::Aliasing
Aliasing: uninitialized constant ModuleSpecs::Aliasing
5)
An exception occurred during: before :each
[...]

Issue #548, opened 2014-06-15
[edge-case] constant lookup diverge from MRI


elia:
```
➋➀ opal:master> irb
>> module B; end
=> nil
>> class B::A < BasicObject; include B; end
NameError: uninitialized constant B::A::B
    from (irb):2:in `<class:A>'
    from (irb):2
    from /Users/elia/.rvm/rubies/ruby-2.1.2/bin/irb:11:in `<main>'
>> 
```

The same is done in `stdlib/native.rb`, in opal doesn't raise any error

I'm not even sure it's a feature in MRI, we should probably look into @rubyspec or ask matz :P

Issue #1581, opened 2016-11-08
autoload lookup differs from MRI


patrickkettner:
A gem I am using has a custom module called `Array`, which is an extension of another class called `Generator`.
When compiled with Opal, anything that extends the custom `Array` is broken because the builtin `Array` is always returned from `Opal.get`, even when the full namespace is given (e.g. `Opal.get('Foo').$$scope.get('Bar').$$scope.get('Generator').$$scope.get('Array')`). 

The `$$parent` is `Enumerable`, rather than the expected `Generator`.

elia:
This code:

```rb
module Generator
end

module Foo
  module Bar
    class Array < Generator
    end
  end
end
p ::Foo::Bar::Array.JS['$$parent'].JS['$$name']
```

returns `Generator` on both master and v0.10.3, can you provide an example demonstrating the :bug:?

patrickkettner:
apologises, haven't been able to create a trimmed down repro.

I believe the issue is because they are in separate files, so they stubbing process is different. But thats a very poorly informed hunch. I'll try to get a repro quickly

patrickkettner:
I still haven't been able to create a reduced test case because of my lack of ruby knowledge, however I did [create a gist](https://gist.github.com/patrickkettner/1c102194d978d6c2a940ce49f4b3d883) that shows how to reproduce this in the hope that someone might be able to figure out the root cause. Full steps are included within

elia:
@patrickkettner the gist was perfect! here's the reduced test case:

https://gist.github.com/elia/ad02a7713706f6ba9fc7dd41a736d456

main.rb
```rb
module Foo
  autoload :Array, 'foo'
end

puts Foo::Array.superclass
```

foo.rb
```rb
class Bar
end
module Foo
  class Array < Bar
  end
end
```

and here's the output from ruby and opal:

```
⤑ ruby -I. main.rb
Bar
⤑ opal -I. main.rb
Object
```

which means that the problem is with autoload/const_missing accepting `::Array` instead of `Foo::Array`

elia:
Now working correctly since v1

Issue #1577, opened 2016-11-05
Include nested module with same name


fazibear:
With this snippet ruby mri will throw exception, opal don't.
```ruby
module NotExist
end

class Terefere
  include NotExist::Terefere
end
```

PR #1653, opened 2017-04-03
constants system rewrite (in two parts)

elia:
**🙅‍♂️ DO NOT MERGE YET**

_a complete rewrite of the constants system to actually mimic CRuby behavior_

- fixes https://github.com/opal/opal/issues/548
- fixes https://github.com/opal/opal/issues/1607
- related: https://github.com/opal/opal/issues/1581 https://github.com/asciidoctor/asciidoctor.js/pull/280 https://github.com/opal/opal/issues/1577

**To do before merging:**

- [x] remove debug code and debug/stale comments
- [x] update API documentation
- [x] update guides (if needed)

---

## Part 1: Make it correct ✅ 

This is more or less done:
- the lookup correctly treats relative and qualified constant names
- nesting is passed around for each scope in the form of an array with the nearest nesting first
- `Module.nesting` and `Module.constants` correctly work
- `BasicObject` behaves properly (see #548)

## Part 2: Make it fast ⏳ 

This is still to be done, some considerations:
- a constant lookup cache should probably be created
- consider re-using Adam's very smart idea of a prototype chain for constants (can work for both inherited and nesting lookups)
- remember that any cache must be properly invalidated or updated
- consider using a property on `$nesting`
- consider using shorter names
- consider restoring some form of `Opal.get`


lib/opal/nodes/call.rb
```diff
@@ -261,38 +248,64 @@ def compile_default?
...
 
-      add_special :__OPAL_COMPILER_CONFIG__ do
+      add_special :__OPAL_COMPILER_CONFIG__ do |compile_default|
         push fragment "Opal.hash({ arity_check: #{compiler.arity_check?} })"
       end
 
+      add_special :nesting do |compile_default|
+        recv, method, *args = children
+        push_nesting = push_nesting?(children)
+        push '(Opal.Module.$$nesting = $nesting, ' if push_nesting
```
iliabylich: Is it necessary to populate `Opal.Module.$$nesting` and call the default behavior? It's a special case for `CallNode` anyway.

lib/opal/nodes/call.rb
```diff
@@ -261,38 +248,64 @@ def compile_default?
...
+        compile_default.call
+        push ')' if push_nesting
+      end
+
+      def push_nesting?(recv)
+        recv = children.first
+
+        children.size == 2 && (           # only receiver and method
+          recv.nil? || (                  # and no receiver
```
iliabylich: I think it could be replaced with `recv.nil? || recv == s(:const, nil, :Module)`. Is it written this way to cover both `Module` and `::Module` (with `nil` and `s(:cbase)` as constant scopes)?

opal/corelib/basic_object.rb
```diff
@@ -48,21 +48,21 @@ def !=(other)
 
   def instance_eval(*args, &block)
     if block.nil? && `!!Opal.compile`
-      Kernel.raise ArgumentError, "wrong number of arguments (0 for 1..3)" unless (1..3).cover? args.size
+      ::Kernel.raise ::ArgumentError, "wrong number of arguments (0 for 1..3)" unless (1..3).cover? args.size
```
iliabylich: Is it required to put `::` everywhere now to avoid calling local constants? (Like potentially generated on the fly `BasicObject::Kernel`)

lib/opal/nodes/call.rb
elia: I thought that overriding `Module.nesting` should be supported and I'd be very curious to see an app that has this as a bottleneck 😄 – Also could be a different `Module` than `::Module`.

opal/corelib/basic_object.rb
elia: no, it's because from inside BasicObject you can't access top level constants… except by making the lookup _qualified_ with the leading double-dots and ignore `Module.nesting`.

See also the changes in `native.rb`

lib/opal/nodes/call.rb
elia: Yes, I think I meant to cover both `nil` and `s(:cbase)`. This part could probably use a test.

elia:
> Are there any intermediate benchmark results?

No, still need to start on that side

> And did you try to run the rubyspec test suite with those failing random seeds?

yes, and there was no error 😸, even before adding all those specs

opal/corelib/runtime.js
```diff
@@ -2288,10 +2143,16 @@
   Opal.Module      = Module      = Opal.setup_class_object('Module',      Module_alloc,      'Object',      _Object.constructor);
   Opal.Class       = Class       = Opal.setup_class_object('Class',       Class_alloc,       'Module',      Module.constructor);
 
-  Opal.constants.push("BasicObject");
-  Opal.constants.push("Object");
-  Opal.constants.push("Module");
-  Opal.constants.push("Class");
+  // BasicObject can reach itself, avoid const_set to skup the
```
iliabylich: `skip`

stdlib/json.rb
```diff
@@ -54,7 +54,7 @@ class ParserError < JSONError
             }
 
             if (!options.parse && (klass = #{`hash`[JSON.create_id]}) != nil) {
-              klass = Opal.get(klass);
+              klass = Opal.const_get_qualified('::', klass);
```
iliabylich: Is it possible to rewrite it using `const_get` from Ruby? If it's a global constant probably `Object.const_get` may  work here. For me `"::"` looks like a private api from `runtime.js` that shouldn't be used explicitly in corelib/stdlib

elia: @iliabylich I think the important point here is that we should define the public runtime API before 1.0

meh:
I'm going to review it this weekend, I did read it already when you pushed the first stuff, all I could think of is the diff is unreadable.

elia:
> EDIT: wait, it was unreadable, or I'm thinking of something else, or whatever, what are you still doing awake?

@meh I did a bunch of side stuff that was initially in the PR but then I separated and pushed to master, that's probably why now it's more readable 📖 

Anyway I'd like to hear some thoughts on this from you before proceeding in trying to optimize for performance, so looking forward to your review ✏️ 👓

opal/corelib/module.rb
```diff
@@ -615,9 +641,16 @@ def name
...
+        var old = self.$$const[name];
+        delete self.$$const[name];
+        return old;
+      } else if (self.$$autoload != null && self.$$autoload[name] != null) {
+        delete self.$$autoload[name];
+        return nil;
+      } else {
+        #{raise NameError, "constant #{self}::#{name} not defined"}
+      }
     }
```
meh: All of this would probably look better if you removed the `else if { } else { }` since you're returning from each of them.

lib/opal/nodes/top.rb
```diff
@@ -19,11 +19,11 @@ def compile
           body_code = [body_code] unless body_code.is_a?(Array)
 
           if compiler.eval?
-            add_temp '$scope = (self.$$scope || self.$$class.$$scope)'
+            add_temp '$nesting = self.$$is_a_module ? [self] : [self.$$class]'
```
elia: @meh there's `$$is_module` and `$$is_class` that XOR on modules and classes, but there's also `$$is_a_module` that is `true` for both and replaces `$$is_module || $$is_class` checks. 

That said I agree that a better name is desirable.

meh: Maybe `$$is_namespace`? Since that's what the check is looking for.

elia:
current status of perf vs. master:

```
opal:master ⤑ be bin/opal -g benchmark-ips -ropal/platform -rbenchmark/ips -e 'module A; module B; Benchmark.ips do |x|; x.report("Kernel") { Kernel }; x.report("::Kernel"){ ::Kernel };x.compare!;end; end; end'
              Kernel    28.156k i/100ms
            ::Kernel    29.693k i/100ms
              Kernel      2.815M (± 5.8%) i/s -     14.022M in   5.002232s
            ::Kernel      3.879M (± 6.0%) i/s -     19.300M in   4.997887s

Comparison:
            ::Kernel:  3879237.1 i/s
              Kernel:  2814554.1 i/s - 1.38x  slower

opal:elia/constants ⤑ be bin/opal -g benchmark-ips -ropal/platform -rbenchmark/ips -e 'module A; module B; Benchmark.ips do |x|; x.report("Kernel") { Kernel }; x.report("::Kernel"){ ::Kernel };x.compare!;end; end; end'
              Kernel     8.394k i/100ms
            ::Kernel    10.546k i/100ms
              Kernel    606.318k (± 5.9%) i/s -      3.022M in   5.004157s
            ::Kernel      6.897M (± 7.9%) i/s -     33.937M in   4.962153s

Comparison:
            ::Kernel:  6896933.5 i/s
              Kernel:   606317.8 i/s - 11.38x  slower
```

lib/opal/nodes/top.rb
elia: @meh wrt the `$$is_a_module` prop I'll open a separate issue/PR

elia:
After adding the cache perf go back more than acceptable:

```
bundle exec opal -ropal/platform -gbenchmark-ips -rbenchmark/ips -A /Users/elia/Code/opal/benchmark-ips/bm_constants_lookup.rb
              Kernel    99.137k i/100ms
            ::Kernel    85.107k i/100ms
              Kernel      7.586M (± 6.6%) i/s -     37.771M in   5.004006s
            ::Kernel      4.149M (± 7.2%) i/s -     20.681M in   5.015473s

Comparison:
              Kernel:  7585546.3 i/s
            ::Kernel:  4148983.1 i/s - 1.83x  slower

```

_bm_constants_lookup.rb_

```rb
module A
  module B
    module C
      Benchmark.ips do |x|
        x.report("Kernel") { Kernel }
        x.report("::Kernel"){ ::Kernel }
        x.compare!
      end
    end
  end
end
```

elia:
Checking performance at each commit shows an overall speed improvement from 50% to almost 80% ([details in this gist](https://gist.github.com/elia/66341dae29adf153c912f672e0eff354)). 🏎

meh:
Isn't that going to slow down startup considerably, or am I reading it wrong?

I can already see @jgaskins yelling at us.

jgaskins:
Some things might be a little slower, but I'd be interested in seeing what this looks like in a real-world app. Because some constants are referenced over and over just during bootstrapping of an app, the extra cost might be amortized. I'll run a benchmark tomorrow and see what I can find out.

elia:
@meh @jgaskins tbh I think the opposite to be true, the caching price is paid only the first time a const is referenced and only *if* it is referenced, previously it was upfront for any constant, even if never used. Also the cache is invalidated globally for any const change but it's done locally and only for referenced constants. Unless an app is going to keep modifying constants at runtime it will be much better. 😄

jgaskins:
Okay, quick analysis before bedtime (that's a lie, I actually should've been in bed an hour ago).

This is how I usually run load-time benchmarks:

```html
<div>
  <span>Load time: </span>
  <span id="benchmark-result"></span>
</div>
<script>var started_at = performance.now()</script>
<script src="app.js"></script>
<script>document.getElementById('benchmark-result').textContent = (performance.now() - started_at).toString()</script>
```

Then I load the page a bunch of times to get a feel for where the number converges. The app is compiled/minified/gzipped just as you would in production to remove the overhead of multiple requests.

The app I chose to test this with was 1543 LoC (not counting blank lines and comments). There are 139 top-level constants. It uses Clearwater, GrandCentral, and Opal::Pusher, which have a few nested constants, but not a whole lot. Maybe 30 across all of them. The benchmark values below only include load time without rendering (Clearwater renders with `requestAnimationFrame`, so the finish-time check occurs while the render is still in the queue). Rendering does access a bunch of constants, as well, and it's important to capture as much as we can, but I think that introduces a bunch of other variables Opal can't control for. I just wanted to have a bunch of constants declared at load time.

On Chrome, the numbers converged around 148ms for this PR vs 157ms on the `master` branch (5.7% faster).

On Safari, it reduced it to 72ms from 78ms (7.7% faster).

On Firefox, this was a little slower than `master`, but I have a feeling it takes Firefox longer to warm up its JIT because the numbers were 242ms vs 253ms. At that point, a few milliseconds in either direction may not mean much. I also have no idea what's causing this slowdown.

This PR also had a lot less fluctuation in the load-time benchmarks than the `master` branch. The numbers still fluctuated and there were still outliers, but it really felt like the range was smaller. I don't have hard data to back that up, though, because I wasn't recording individual times.

It seems like a load-time performance win, tbh. Slower in Firefox, but the vast majority of mobile users (where load-time performance matters most) are on Chrome or Safari. I'll try to test a bit more on mobile devices this weekend if I get a chance.

PR #1658, opened 2017-04-22
Add aliases for `$$` from 'stdlib/English'

AlexWayfer:
`$PID` and `$PROCESS_ID` are equivalent for `$$`.

RuboCop by default warns that they are preferable to `$$`:

*   [RuboCop cop](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/SpecialGlobalVars)
*   [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#no-cryptic-perlisms)
*   [`stdlib/English`](https://ruby-doc.org/stdlib/libdoc/English/rdoc/English.html).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Whitespace conventions:
 
 ### Added
 
+- Added aliases for `$$` from [`stdlib/English`](https://ruby-doc.org/stdlib/libdoc/English/rdoc/English.html)
 - Added support for complex (`0b1110i`) and rational (`0b1111r`) number literals. (#1487)
 - Added 2.3.0 methods:
     * `Array#bsearch_index`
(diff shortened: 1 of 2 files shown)


meh:
`$$` here is abused, it's not actually the PID, I don't know how meaningful this change would be.

AlexWayfer:
Oh, okay, I assumed this.

So, maybe we should remove `$$` and leave `$global`? Or rename it.

meh:
There are some things in Ruby which make absolutely no sense at all in Opal, i.e. the PID, since there isn't one.

And it's a shortcut used often.

AlexWayfer:
It's sad :pensive: 

Okay, I'm closing PR…

PR #1662, opened 2017-05-09
Use truth helpers instead of inlining

elia:
Perf on nodejs looks roughly the same (possibly in favor of the helper), probably thanks to JIT.

The core is much more readable and shorter.

I think there are some more places that can be replaced, but I'll wait for an ok from @opal/core.

### DIFF
--- a/lib/opal/nodes/helpers.rb
+++ b/lib/opal/nodes/helpers.rb
@@ -47,9 +47,8 @@ module Opal
           return optimize
         end
 
-        with_temp do |tmp|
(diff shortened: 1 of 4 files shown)


meh:
I'm ok with this.

iliabylich:
hmm there are some broken tests in `spec/lib` that do assertions on compiled code. that's why I push an idea of moving as much logic as we can to rewriters to simplify JS generating 😄

elia:
@iliabylich yeah I saw that, I'll fix them as I check for other places that can be simplified

PR #1664, opened 2017-05-16
Check that process.stderr and process.stdout are defined

Mogztter:
`process` object is defined in a Webpack environment but `process.stderr` and `process.stdout` are not.


elia:
Thanks!

PR #1665, opened 2017-05-16
Fixed __END__ section parsing.

iliabylich:


elia:
Just out of curiosity (since _no one uses DATA_) have you tried something like the following?

```rb
"
__END__
just a string
"

__END__
data stuff
```

iliabylich:
Of course it doesn't work 😄. Don't merge it yet please, I have an idea how to fix it.

elia:
@iliabylich possibly related: https://github.com/whitequark/parser/blob/7818d36672366d9cd23dd62baae65d9892410484/test/test_lexer.rb#L2363-L2372

iliabylich:
@elia Yes, `parser` ignores a valid `__END__` section because it's not a part of the ruby code. I've found a solution (thanks to rubocop source), your example should work now.

PR #1668, opened 2017-05-22
Use shorter helpers for constant lookups

elia:
Conciseness should improve both readability and perf (less lookups on the `Opal` object).


Mogztter:
Dollar sign is already used on function name, what about `##` and `###` ?

elia:
> Dollar sign is already used on function name, what about ## and ### ?

@Mogztter sadly `$` is the only _unconventional_ symbol that can be used in var names, everything else lives in remote utf-8 regions 🌋

Mogztter:
💲 💲 💲 is fine but what about emoji ? `🔎(this, 'Asciidoctor')` 

(spoiler alert, I'm just kidding 😆)

PR #1676, opened 2017-06-06
Improve marshalling

iliabylich:
+ Added support for `Hash` with default value
+ Improved loading of User-defined classes (with `self._load` and `_dump` methods)
+ Improved loading of User-Marshal classes (with `marshal_dump/marshal_load` methods)
+ Renamed `Range` ivar responsible for `..`/`...` (`exclude_end?`) marker to support Ranges encoded by MRI.
+ Removed `_load` from several corelib classes (MRI doesn't have them, they were added to Opal as a temporary workaround)
+ Documented `ReadBuffer`

Also fixed an inverted test mode, it seems to work now.


elia:
great stuff

PR #1683, opened 2017-07-07
Bignum support for Marshal.dump

youchan:
I modified `Marshal.dump` to correspond to a large integer value.

`Marshal.dump(value)` will be overflow when in -0x100000000 > `value` >= 0x100000000
MRI treats the value as `Bignum` in this case.

Opal
```
>> Marshal.dump(0x100000000)
=>"\u0004\bi\u0001\u0000"
```

MRI
```
irb> Marshal.dump(0x100000000)
=> "\x04\bl+\b\x00\x00\x00\x00\x01\x00"
```

### DIFF
--- a/opal/corelib/marshal/write_buffer.rb
+++ b/opal/corelib/marshal/write_buffer.rb
@@ -16,8 +16,13 @@ end
 
 class Fixnum
   def __marshal__(buffer)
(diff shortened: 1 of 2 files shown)


iliabylich:
Thank you!

Issue #1696, opened 2017-08-20
Problem if inline javascript ends with a semicolon


bwl21:
I wanted to update to 0.11.rc1 and found a difference in the compilateion result:

```ruby
    abc          = `localStorage.setItem('abc_data', #{abc});`
  end
```

transpiles to (0.10.5)

```javascript
      return abc = localStorage.setItem('abc_data', abc);;
```
in 0.11.rc1 it compiles to

```javascript
      return (abc = localStorage.setItem('abc_data', abc););
```
Chrome complains the semicolon before the closing paranthesis as error.

Of course the trailing ';' ist not required. Nevertheless it is an issue which might break existing sources.

@elia considered this as a bug which needs to be fixed

iliabylich:
@elia Why do you think that it should be fixed? I just feel that JS manipulation may introduce some weak rules that are good for some cases, but are bad for some other corner cases. Maybe I'm wrong, but I think that it's better to always return JS that was initially written in `xstr`, without any manipulations.

To be fair, I think that it's better to not touch `xstr` at all.

For example https://github.com/opal/opal/pull/1698 fixes it, but it introduces another issue:
``` ruby
def m
  %x{
    var a = 1;
    return a;}
end
```

becomes

``` js
Opal.defn(Opal.Object, '$m', TMP_m_1 = function $$m() {
  var self = this;
  var a = 1;
  return a
}, TMP_m_1.$$arity = 0)
```

(No trailing `;` after return)

Yes, we don't have such code in Opal corelib/stdlib, but developers that use Opal can write it. For me this corner case is as real as the one provided by @bwl21 

Introducing such xstr-rewrites makes Opal something similar to JS linter, I don't think that we should do it. For me one simple rule `we don't touch xstr-s` is enough.

bwl21:
I agree that the xstr shall not be changed ore rewritten. In my case it was such that the xstr was embedded in a way that is no valid javascript. And it is a backwards compatibility issue because it worked in Opal 0.9.

Of course it is easy to find and to fix on the ruby side. So IMHO this issue can also be fixed by a proper note in the release notes. I agree, that a fix in the code might introduce more problems than it fixes.

The question is, do we have a rule how to write xstr such that it safely works in all cases?
In this case the rule for opal users might be: "do not terminate xstr with semicolon.

iliabylich:
@bwl21 The rule is `do not terminate one-line xstr with semicolon`.

[...]

PR #1698, opened 2017-08-29
remove trailing semicolon when generating xstring code

CicholGricenchos:
PR for https://github.com/opal/opal/issues/1696.


CicholGricenchos:
Codeclimate is complaining about the complexity, but I have no idea how to reduce it...

lib/opal/nodes/literal.rb
```diff
@@ -241,6 +245,15 @@ def compile
 
         wrap '(', ')' if recv?
       end
+
+      def detect_and_remove_trailing_semicolon value, node
+        if value.match(/;\s*$/)
+          compiler.warning 'Xstr ends with semicolon', node.line
+          value.gsub(/;\s*$/, '')
```
elia: should be `sub` with `REGEXP_END` instead of `$` so it becomes `\z`

see https://github.com/opal/opal/blob/66849b5da5c54236eea5a0e5eaf95ba06a744fe3/lib/opal/regexp_anchors.rb

spec/lib/compiler_spec.rb
```diff
@@ -307,6 +307,14 @@
...
   end
 
+  describe 'x-strings' do
+    it 'removes the trailing semicolon' do
+      compiler = Opal::Compiler.new('a = `1;`')
+      expect(compiler).to receive(:warning).once
+      expect(compiler.compile).not_to include(";)")
+    end
+  end
+
```
elia: should also test for multiline xstr, possibly with multiple semicolons within them, especially before newlines

lib/opal/nodes/literal.rb
```diff
@@ -216,10 +216,14 @@ class XStringNode < Base
...
       def compile
-        children.each do |child|
+        last_index = children.size - 1
+        children.each_with_index do |child, index|
           case child.type
           when :str
             value = child.loc.expression.source
+            if last_index == index
+              value = detect_and_remove_trailing_semicolon(value, child)
+            end
```
elia: instead of checking for `last_index` why not check for `expr?` and `recvr?` ([example](https://github.com/opal/opal/blob/66849b5da5c54236eea5a0e5eaf95ba06a744fe3/lib/opal/nodes/if.rb#L46))

that way we can ignore the problem if the line is not used for something else

lib/opal/nodes/literal.rb
elia: Also this should just issue the warning as suggested by @iliabylich it's better not to touch xstr and let them be a tool that's both powerful and dangerous 😄

CicholGricenchos:
Multiline xstr
```ruby
a = `1;
2;`
```
which yields
```javascript
return (a = 1;
2;)
```
could not be valid. So we could just ignore the multiline case.

And the recv case
```ruby
`a;`.method
```
should not be valid too.

Thus we only remove and give warning for one-line expr case.

lib/opal/nodes/literal.rb
```diff
@@ -241,6 +244,15 @@ def compile
 
         wrap '(', ')' if recv?
       end
+
+      def detect_and_remove_trailing_semicolon value, node
```
iliabylich: Could you add parentheses around arguments? `def detect_and_remove_trailing_semicolon(value, node)`

spec/lib/compiler_spec.rb
```diff
@@ -307,6 +307,25 @@
     end
   end
 
+  describe 'x-strings' do
+    it 'removes the trailing semicolon for one-line expr' do
+      compiler = Opal::Compiler.new('a = `1;`')
+      expect(compiler).to receive(:warning).once
+      expect(compiler.compile).not_to include(";)")
+
+      compiler = Opal::Compiler.new <<-EOF
```
iliabylich: Could you add parentheses around arguments:
``` ruby
Opal::Compiler.new(<<-EOF)
  a = `1;
  2;`
EOF
```

spec/lib/compiler_spec.rb
```diff
@@ -307,6 +307,25 @@
...
+        a = `1;
+        2;`
+      EOF
+      expect(compiler).not_to receive(:warning)
+      compiler.compile
+
+      compiler = Opal::Compiler.new '`1;`'
```
iliabylich: Could you add parentheses around arguments?
``` ruby
Opal::Compiler.new('`1;`')
```

lib/opal/nodes/literal.rb
CicholGricenchos: Sure, added.

iliabylich:
Thank you!

PR #1700, opened 2017-09-09
File.expand_path with actual pwd

### LINKED ISSUE #1699: File.expand_path 'relative/path' should append the current working directory
.MRI
```ruby
::File.expand_path 'chapter-01/index.adoc'
# "/home/guillaume/workspace/opensource/nodebook/chapter-01/index.adoc"
```

.Opal
```js
Opal.const_get_qualified('::', 'File').$expand_path('chapter-01/index.adoc');
// "chapter-01/index.adoc"
```

elia:
Fixes #1699

### DIFF
--- a/opal/corelib/file.rb
+++ b/opal/corelib/file.rb
@@ -7,20 +7,48 @@ class File < IO
 
   class << self
     def expand_path(path, basedir = nil)
-      path = [basedir, path].compact.join(SEPARATOR)
-      parts = path.split(SEPARATOR)
+      sep = SEPARATOR
       new_parts = []
-      parts[0] = Dir.home if parts.first == '~'
-      parts[0] = Dir.pwd if parts.first == '.'
+
+      if path.start_with?('~') || (basedir && basedir.start_with?('~'))
+        home = Dir.home
+        raise(ArgumentError, "couldn't find HOME environment -- expanding `~'") unless home
+        raise(ArgumentError, "non-absolute home") unless home.start_with?(sep)
+        home += sep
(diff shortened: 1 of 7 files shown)


opal/corelib/file.rb
```diff
@@ -7,20 +7,28 @@ class File < IO
...
+      absolute = path.start_with?(sep)
       new_parts = []
-      parts[0] = Dir.home if parts.first == '~'
-      parts[0] = Dir.pwd if parts.first == '.'
+
+      unless absolute
+        if parts.first == '~'
+          new_parts << Dir.home
+        else
+          new_parts << Dir.pwd
```
Mogztter: As far as I understand C code, I think the `Dir.pwd` is only prepend if the `basedir` argument is `nil`: https://github.com/ruby/ruby/blob/0471781b2a7a40681ccdc051f0239eb19719d79a/file.c#L3414-L3425

And I think `basedir` can be a non absolute path... ?

elia: Thanks! I jotted down the code last knight without actually checking the original C 😄

Mogztter: > ... without actually checking the original C 

"Great minds think alike." :grinning: 

> I jotted down the code last knight...

"last knight" ?! :horse: :crossed_swords: I think you need some sleep :stuck_out_tongue_winking_eye:

elia: > "last knight" ?! 🐴 ⚔️ I think you need some sleep 😜

definitely! 😆

Mogztter:
Thanks @elia 👍

PR #1709, opened 2017-10-19
Use magic encoding comment to encode compiled strings.

iliabylich:
``` ruby
# encoding: ascii-8bit
'λ'
```

Now compiles to:
``` js
"λ".$force_encoding("ASCII-8BIT")
```

When no magic comment specified all strings are treated as UTF-8 and have no re-encoding.

Also the parser now raises an error for invalid UTF-8 byte sequences like:
``` ruby
"\xFF"
```

or

``` ruby
# encoding: utf-8
"\xFF"
```

If you want to make an ascii-string with 255 char, use `encoding` comment and specify any encoding that supports it. You can always check it by running `String#valid_encoding?` method.

Also now we probably can correctly implement `String#bytes` and friends.

Two related modifications:
1. `Object#to_s` now converts its return value from String object to a plain string. [Demo of the bug](http://opalrb.com/try/?code:obj%20%3D%20Object.new%0A%0Adef%20obj.to_s%0A%20%20%22string%22.force_encoding(%27ASCII-8BIT%27)%0Aend%0A%0Aputs%20%22%23%7Bobj%7D%22). 
2. Since `String#force_encoding` (which is now used in many places in specs) returns a String object, it becomes mutable. There were two specs for `Marshal` module that were relying on setting instance variables on the `String`. Before this patch the string was immutable and `instance_variable_set` was no-op. Now it's created using `.force_encoding` and thus is mutable and can have ivars.

About performance:
1. For no magic comment there's absolutely no difference.
2. For `encoding: utf-8` comment there's absolutely no difference.
3. For `encoding: any-non-utf-8-encoding` there's a difference for inline string allocation.

### DIFF
--- a/lib/opal/ast/builder.rb
+++ b/lib/opal/ast/builder.rb
@@ -5,10 +5,6 @@ require 'parser/ruby23'
 module Opal
   module AST
     class Builder < ::Parser::Builders::Default
-      def string_value(token)
-        token[0]
-      end
-
       def n(type, children, location)
         ::Opal::AST::Node.new(type, children, location: location)
       end
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -116,7 +116,7 @@ module Opal
 
     def show_sexp
       evals_or_file do |contents, filename|
-        buffer = ::Parser::Source::Buffer.new(filename)
+        buffer = ::Opal::Source::Buffer.new(filename)
         buffer.source = contents
         sexp = Opal::Parser.default_parser.parse(buffer)
         puts sexp.inspect
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -171,7 +171,7 @@ module Opal
     end
 
     def parse
-      @buffer = ::Parser::Source::Buffer.new(file, 1)
+      @buffer = ::Opal::Source::Buffer.new(file, 1)
       @buffer.source = @source
 
       @parser = Opal::Parser.default_parser
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -61,13 +61,13 @@ module Opal
           string_value = string_value.force_encoding('UTF-8')
         end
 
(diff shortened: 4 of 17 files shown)


iliabylich:
```
base64/test_base64:29:44: error: literal contains escape sequences incompatible with UTF-8
base64/test_base64:29:     assert_equal("/w==\n", Base64.encode64("\377"))
base64/test_base64:29:                                            ^~~~~~  
```

This file starts with the `# coding: US-ASCII` line, `"\377" -> 255`, which is obviously invalid for US-ASCII that is ranged from 0 to 127.

``` ruby
# coding: us-ascii

s = "\377"
p s
# => "\xFF"
p s.encoding
# => #<Encoding:ASCII-8BIT>
p s.valid_encoding?
# => true
p __ENCODING__
# => #<Encoding:US-ASCII>
```

Of course it's valid for ruby, because it somehow changes the string encoding.

For a string that is initially valid MRI doesn't change it:
``` ruby
# coding: us-ascii

s = "\37"
p s
# => "\x1F"
p s.encoding
# => #<Encoding:US-ASCII>
p s.valid_encoding?
# => true
p __ENCODING__
# => #<Encoding:US-ASCII>
```

@whitequark @opal/core Does anyone have any ideas what is going in this case?

iliabylich:
Changing encoding to `ascii-8bit` solves the problem. @elia Can we just move this spec under opal suite and fix encoding locally?

whitequark:
@iliabylich Ruby MRI is cheating, basically. I looked into implementing its extremely complicated rules for parser and came with no very good solution.

elia:
@iliabylich duplicating the spec on our side is fine by me but maybe it's worth proposing the change to upstream ruby…

_btw I find hilarious that this PR triggered the build n. 377 on appveyor 😂_

iliabylich:
Finally green! @opal/core Any comments?

elia:
@iliabylich this merge broke master, I suspect that it's related to some kind of change in how travis run the spec suite but didn't find anything obvious

iliabylich:
All errors are related to chrome, so yes, it's not about this change, it's about travis.

PR #1711, opened 2017-11-03
WIP: Add JS.async and JS.await

xavriley:
These commits work toward adding support for `JS.async` and `JS.await`
to allow these ES7 keywords to be used. For background on these
functions http://rossboucher.com/await/#/ is a good introduction.

My test case for code up to now has been

```
require 'js'

class Foo
  include JS

  def self.bar(&block)
    self.JS.async block.call
  end
end

Foo.bar do
  self.JS.await sleep(2)
end
```

My motivation for adding these is to support an async/await workflow for
`sleep` to allow a concurrent, non-window-blocking sleep to be used in multiple "threads".
For example in javascript this would look like this:

```
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

async function kick() {
  console.log("boom")
  await sleep(1000);
}

async function hihats() {
  console.log("tick")
  await sleep(500);
  console.log("tick")
  await sleep(500);
}

// viewing the console shows these execute concurrently
kick();
hihats();
```

Still todo:

- [ ] verify that this approach (modifying sexp) is viable
- [ ] add specs
- [ ] make sure that `async` and `await` are output in compiled js
- [ ] decide how to handle issues around scoping the async calls
- [ ] check performance impact for compiler and runtime


hmdne:
Superseeded with #2221 which we are most probably merging soon

PR #1716, opened 2017-11-25
kwargs are not js identifiers

elia:
Before this it was using `deafult$` instead of `default` as kwarg key

### DIFF
--- a/lib/opal/nodes/args/kwarg.rb
+++ b/lib/opal/nodes/args/kwarg.rb
@@ -13,16 +13,18 @@ module Opal
       children :name
 
       def compile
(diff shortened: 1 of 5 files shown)


iliabylich:
I'm not sure that changing the format of AST is a good idea. How about moving rewritten name to the `meta` field of the node? Or making a class-method `fix_variable_name` on rewriter and reusing it in the handler for `kwarg/kwoptarg` nodes?

elia:
I thought about the latter (class method) but I tried to stay within the rewriting system, didn't know (or remember) about the `meta` field, I'll redo that way putting the _fixed name_ in the meta

PR #1722, opened 2017-12-21
dummy location for rewriters

elia:
@iliabylich picking up where we left on gitter, feel free to comment and add commits direclty to the branch

### DIFF
--- a/lib/opal/rewriters/base.rb
+++ b/lib/opal/rewriters/base.rb
@@ -5,12 +5,46 @@ require 'opal/ast/node'
 module Opal
   module Rewriters
     class Base < ::Parser::AST::Processor
+      class DummyLocation
+        def node=(*)
+        end
+
+        def expression
+          self
+        end
+
+        def begin_pos
+          0
+        end
+
+        def end_pos
+          0
+        end
+
+        def source
+          ''
+        end
+
+        def line
+          0
+        end
+
+        def column
+          0
+        end
+
+        def last_line
+          Float::INFINITY
+        end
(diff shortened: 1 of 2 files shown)


lib/opal/cli_runners/server.rb
```diff
@@ -30,8 +37,8 @@ def exit_status
         nil
       end
 
-      def app(source)
-        lambda do |env|
+      def build_app(source)
+        app = lambda do |env|
```
iliabylich: Probably it makes sense to extract this lambda to a separate callable object:
``` ruby
class App
  def initialize(source)
    @source = source
  end

  BODY = <<-HTML.freeze
    <!doctype html>
    <html>
      <head>
        <meta charset="utf-8"/>
        <script src="/cli_runner.js"></script>
      </head>
    </html>
  HTML

  def call(env)
    case env['PATH_INFO']
    when '/'
      [200, {}, [BODY]]
    when '/cli_runner.js'
      [200, {'Content-Type' => 'text/javascript'}, [source]]
    else
      [404, {}, [body]]
    end
  end
end

def build_app(source)
  app = App.new(source)

  if static_folder
    app = Rack::Cascade.new(
      Rack::Static.new(:urls => [''], root: static_folder),
      app,
    )
  end

  app
end
```

elia: The server stuff slipped in by error, anyway I'm following your advice. 
Someday I'd like to rewrite the cli runners and make them pluggable from outside…

PR #1726, opened 2017-12-27
Implement File#symlink? for Node.js

### LINKED ISSUE #1725: Implement File#symlink? for Node.js
https://ruby-doc.org/core-2.2.0/File.html#method-c-symlink

We can implement this method using `fs.lstatSync(path)` and `stats.isSymbolicLink()`: https://nodejs.org/api/fs.html#fs_class_fs_stats

Mogztter:
Resolves #1725

### DIFF
--- a/test/nodejs/test_file.rb
+++ b/test/nodejs/test_file.rb
@@ -101,6 +101,13 @@ class TestNodejsFile < Test::Unit::TestCase
     assert(File.file?('test/nodejs/fixtures/symlink-to-file'),       'test/nodejs/fixtures/symlink-to-file should be a file')
   end unless windows_platform?
 
+  def test_file_symlink?
+    assert(File.symlink?('test/nodejs/fixtures/symlink-to-file'))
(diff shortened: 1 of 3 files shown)


test/nodejs/test_file.rb
```diff
@@ -101,6 +101,13 @@ def test_file_check_with_symlinks
     assert(File.file?('test/nodejs/fixtures/symlink-to-file'),       'test/nodejs/fixtures/symlink-to-file should be a file')
   end unless windows_platform?
 
+  def test_file_symlink?
+    assert(File.symlink?('test/nodejs/fixtures/symlink-to-file'))
+    assert(File.symlink?('test/nodejs/fixtures/symlink-to-directory'))
+    refute(File.symlink?('test/nodejs/fixtures'))
+    refute(File.symlink?('test/nodejs/fixtures/hello.rb'))
+  end
```
iliabylich: `end unless windows_platform?` 😄

Mogztter: Good catch, thanks, this is fixed.

Mogztter:
🎉

PR #1728, opened 2017-12-27
Implement Dir#glob for Node.js

### LINKED ISSUE #1727: Implement Dir#glob for Node.js
Reference implementation: https://ruby-doc.org/core-2.2.0/Dir.html#method-c-glob

We could use the Node.js module `glob` since we already using it ?

Mogztter:
Resolves #1727

### DIFF
--- a/stdlib/nodejs/dir.rb
+++ b/stdlib/nodejs/dir.rb
@@ -28,6 +28,15 @@ class Dir
       }
     end
 
+    def glob(pattern)
+      pattern = [pattern] unless pattern.respond_to? :each
+      pattern.flat_map do |subpattern|
+        subpattern = subpattern.to_path if subpattern.respond_to? :to_path
+        subpattern = Opal.coerce_to!(subpattern, String, :to_str)
+        `__glob__.sync(subpattern)`
+      end
+    end
+
     alias getwd pwd
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Whitespace conventions:
 - Added `Array#pack` (supports only `C, S, L, Q, c, s, l, q, A, a` formats). (#1723)
 - Added `String#unpack` (supports only `C, S, L, Q, S>, L>, Q>, c, s, l, q, n, N, v, V, U, w, A, a, Z, B, b, H, h, u, M, m` formats). (#1723)
 - Added `File#symlink?` for Node.js. (#1725)
+- Added `Dir#glob` for Node.js (does not support flags). (#1727)
 
 
 
--- a/test/nodejs/test_dir.rb
+++ b/test/nodejs/test_dir.rb
@@ -10,13 +10,30 @@ class TestNodejsDir < Test::Unit::TestCase
   end
 
   def test_dir_entries
-    path = tmpdir + "/testing_nodejs_dir_implementation_#{Time.now.to_i}"
+    path = tmpdir + "/testing_nodejs_dir_entries_implementation_#{Time.now.to_i}"
     Dir.mkdir(path)
     result = Dir.entries(path)
     assert_equal(result.length, 0)
     File.open(path + '/bar', "w") {}
     File.open(path + '/baz', "w") {}
     result = Dir.entries(path)
+    assert_equal(result.length, 2)
+  end
+  
+  def test_dir_glob
+    path = tmpdir + "/testing_nodejs_dir_glob_implementation_#{Time.now.to_i}"
+    Dir.mkdir(path)
+    result = Dir.glob(path + '/*')
+    assert_equal(result.length, 0)
+    File.open(path + '/bar', "w") {}
+    File.open(path + '/baz', "w") {}
+    result = Dir.glob(path + '/ba*')
+    assert_equal(result.length, 2)
+    result = Dir.glob(path + '/*z')
+    assert_equal(result.length, 1)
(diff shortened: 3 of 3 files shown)


Mogztter:
Actually `Dir#glob` can be used with an `Array` or a `Pathname` as an argument.
So we should probably use duck typing (ie. use `respond_to` to iterate over the `Array` or convert the `Pathname` to a `String`) ?

stdlib/nodejs/dir.rb
```diff
@@ -28,6 +30,16 @@ def entries(dirname)
       }
     end
 
+    def glob(pattern)
+      pattern = [pattern] unless pattern.respond_to? :each
+      result = Set.new
+      pattern.each do |value|
```
Mogztter: @iliabylich I'm trying to write idiomatic Ruby but I don't know how to simplify this part. Maybe I should use `map`, `flatten` and `uniq` ?

```rb
pattern.map {|value| `__glob__.sync(#{value})`}.flatten.uniq
```

iliabylich: What is the nesting level of `__glob__.sync(pattern)`? If it's a plain array you can use `flat_map` (which does `.map.flatten(1)`):
``` ruby
pattern.flat_map { |value| `__glob__.sync(#{value})` }.uniq
```

Also it seems that MRI doesn't call `.uniq`:
``` ruby
2.4.0 :001 > Dir.glob(['test/*', 'test/*'])
 => ["test/nodejs", "test/opal", "test/cruby", "test/nodejs", "test/opal", "test/cruby"]
2.4.0 :002 > Dir.glob(['test', 'test'])
 => ["test", "test"]
```
The interpolation is not required.
So I guess it's just a
``` ruby
pattern.flat_map { |value| `__glob__.sync(value)` }
```

Mogztter: Perfect, then I will use `flat_map`. Thanks for the quick reply 👍

Mogztter: > The interpolation is not required.

The String interpolation is not required ? 
As far as I know, the following code is mandatory:
```ruby
value = value.to_s if value.respond_to? :to_s
```

Otherwise an exception is thrown when `Dir.glob` is used with a `Pathname`:

```ruby
Dir.glob(Pathname.new('/foo/bar'))
```
```
TestNodejsDir#test_dir_glob:
Exception: glob pattern string required
    TypeError: glob pattern string required
        at new Minimatch (/home/dev/opal/stdlib/nodejs/node_modules/glob/node_modules/minimatch/minimatch.js:108:11)
```

iliabylich: `Dir.glob` calls `to_str`, string interpolation calls `to_s`:
```
$ cat test.rb
o = Object.new
def o.to_str; p 'to_str'; end
def o.to_s;   p 'to_s';   end

Dir.glob(o)
"#{o}"

$ ruby test.rb
"to_str"
"to_s"
```

`value` variable is available in both Ruby and JS code, so you can use it as is without `#{}` in backticks (that's what I meant).

Of course, you still need to call `Opal.coerce_to!(value, String, :to_str)` to match MRI behavior, so the final should like the following (NOT TESTED):
``` ruby
def glob(pattern)
  pattern = [pattern] unless pattern.respond_to?(:each)
  pattern.flat_map do |subpattern|
    subpattern = Opal.coerce_to!(subpattern, String, :to_str)
    `__glob__.sync(subpattern)`
  end
end
```

But even this implementation is not correct, for example, for `Pathname` argument it calls `to_path` method, so I think `Opal.coerce_to!(subpattern, String, :to_s)` is better for a minimal implementation.

stdlib/nodejs/dir.rb
```diff
@@ -1,3 +1,5 @@
+require 'set'
```
iliabylich: Unused `require`

stdlib/nodejs/dir.rb
Mogztter: Alright, I was confused because I didn't know that the `value` variable was available in both Ruby and JS code.

> But even this implementation is not correct, for example, for Pathname argument it calls to_path method, so I think Opal.coerce_to!(subpattern, String, :to_s) is better for a minimal implementation.

I could add a condition to use `to_path` if the method is available ?

```ruby
subpattern = subpattern.to_path if subpattern.respond_to? :to_path
```

stdlib/nodejs/dir.rb
Mogztter: Ooops 😳

stdlib/nodejs/dir.rb
iliabylich: Yes, from what I see it matches the logic of MRI:
``` ruby
subpattern = subpatter.to_path if subpattern.respond_to?(:to_path)
subpattern = Opal.coerce_to!(subpattern, String, :to_str)
`__glob__.sync(subpattern)`
```

``` ruby
$o1 = BasicObject.new
$o2 = BasicObject.new

def $o1.respond_to?(*args)
  Kernel.p "$o1.respond_to?(#{args})"
  true
end

def $o1.to_path
  Kernel.p "$o1.to_path => $o2"
  $o2
end

def $o1.to_s
  '/var' # ignored
end

def $o2.respond_to?(*args)
  Kernel.p "$o2.respond_to?(#{args})"
  true
end

def $o2.to_str
  '/tmp' # return value
end

p Dir.glob([$o1])
```

Hope it helps.

PR #1732, opened 2017-12-31
Change the CLI runners API to just be callables

elia:
Now runners consist of just an object that can respond to #call and 
return the exit code.

Old runners are still called with the legacy API but rewriting them with
the new one is encouraged.

### DIFF
--- a/spec/lib/cli_spec.rb
+++ b/spec/lib/cli_spec.rb
@@ -1,9 +1,9 @@
 require 'lib/spec_helper'
 require 'opal/cli'
 require 'stringio'
+require 'tmpdir'
 
 RSpec.describe Opal::CLI do
-  let(:fake_stdout) { StringIO.new }
   let(:file)    { File.expand_path('../fixtures/opal_file.rb', __FILE__) }
   let(:options) { {} }
   subject(:cli) { described_class.new(options) }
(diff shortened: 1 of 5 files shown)


spec/lib/cli_spec.rb
```diff
@@ -161,13 +161,29 @@
...
-      expect_output_of{ subject.run }.to include(".$puts(2342)")
-      expect_output_of{ subject.run }.not_to include("2342\n")
+      it 'outputs the compiled javascript' do
+        expect_output_of{ subject.run }.to include(".$puts(2342)")
+        expect_output_of{ subject.run }.not_to include("2342\n")
+      end
+
+      context 'with the :map_file runner option' do
+        let(:map_file) { "#{Dir.mktmpdir 'opal-map'}/file.map" }
+        let(:runner_options)  { {map_file: map_file, evals: ['123']} }
```
iliabylich: Is `:eval` field necessary here? From what I understand the builder doesn't use it

elia:
@opal/core using JSON for the runner options is the best I came up with, but I'm open to suggestions

spec/lib/cli_spec.rb
elia: @iliabylich good catch, fixed

PR #1735, opened 2018-01-05
Implement OpenURI module

Mogztter:
First draft of the OpenURI module (copied from https://raw.githubusercontent.com/ruby/ruby/373babeaac8c3e663e1ded74a9f06ac94a671ed9/lib/open-uri.rb)

For now I only support a subset of features but I want to check with you that I'm on the right track 😉


test/opal/test_openuri.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: false
+require "test/unit"
+require "open-uri"
+
+class TestOpenURI < Test::Unit::TestCase
+  def test_open_https
+    open("https://opalrb.com/") {|f|
+      assert_match(/Opal: Ruby to JavaScript Compiler/, f.read)
```
iliabylich: So we can't change this text on the website? 😄

Mogztter: lol! 😆 
We should probably start a local http server to not rely on network but I was sure how to do it

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+# URI objects can be read directly. The returned string is also extended by
+# OpenURI::Meta.
+#
+#   str = uri.read
+#   p str.base_uri
+#
+# Author:: Tanaka Akira <akr@m17n.org>
+
+module OpenURI
+  Options = {
```
Mogztter: I think I should remove the code related to the options and add a warning message to state that options are unsupported. Wdyt ?

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+    def initialize(message, io)
+      super(message)
+      @io = io
+    end
+    attr_reader :io
+  end
+
+  # Raised on redirection,
+  # only occurs when +redirect+ option for HTTP is +false+.
+  class HTTPRedirect < HTTPError
```
Mogztter: Since we do not support redirect, I should probably remove this class

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+  def self.request(uri)
+    %x{
+      var xhr = new XMLHttpRequest();
+      xhr.open('GET', uri, false);
+      xhr.send();
+      return xhr;
+    }
+  end
+
+  class HTTPError < StandardError
```
Mogztter: If the return code is not 200, 301, 302, 303 or 307, we should throw this exception. Also if something bad happens with XHR (for instance an invalid URI), we should also throw this exception.

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+  def self.open_loop(uri, options) # :nodoc:
+    req = request(uri)
+    # FIXME Handle exception and status code ?
+    data = `req.responseText`
+    buf = Buffer.new
+    buf << data
+    io = buf.io
+    # FIXME Generate a URI object from the uri String
+    #io.base_uri = uri
+    status = `req.status`
```
Mogztter: This part is ugly, I want to assign multiples attributes with the request object... do you have a better idea/syntax ?

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+  #
+  # Otherwise, the original Kernel#open is called.
+  #
+  # OpenURI::OpenRead#open provides URI::HTTP#open, URI::HTTPS#open and
+  # URI::FTP#open, Kernel#open.
+  #
+  # We can accept URIs and strings that begin with http://, https:// and
+  # ftp://. In these cases, the opened file object is extended by OpenURI::Meta.
+  def open(name, *rest, &block) # :doc:
+    if name.respond_to?(:to_str) && %r{\A[A-Za-z][A-Za-z0-9+\-\.]*://} =~ name
```
Mogztter: The original code was:
```ruby
 if name.respond_to?(:to_str) 
  && %r{\A[A-Za-z][A-Za-z0-9+\-\.]*://} =~ name 
  && (uri = URI.parse(name)).respond_to?(:open)
```

But `URI` is not (yet) implemented. Do you think we should implement URI first ?

test/opal/test_openuri.rb
Mogztter: @iliabylich The original test file is using `webrick`, do you think we should add this development dependency ?

iliabylich: Yes. Otherwise the test becomes unreliable. For example, `opal-browser` uses [this](https://github.com/opal/opal-browser/blob/master/config.ru#L12) server to test HTTP module.

stdlib/open-uri.rb
iliabylich: It's not that easy. Regular expressions in URI are very complex.

stdlib/open-uri.rb
iliabylich: A simply mass-assignment should work, but I'm not sure that it's a good idea
``` ruby
opal-repl
>> a, b = `[1, 2]`
=> [1, 2]
>> a
=> 1
>> b
=> 2
```

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+
+    def meta_add_field(name, value) # :nodoc:
+      meta_add_field2(name, [value])
+    end
+
+    # returns a Time that represents the Last-Modified field.
+    def last_modified
+      if vs = @metas['last-modified']
+        v = vs.join(', ')
+        Time.httpdate(v)
```
iliabylich: There's no such method in Opal, so I guess that's a dead code?

stdlib/open-uri.rb
```diff
@@ -0,0 +1,515 @@
...
+    :read_timeout => true,
+    :open_timeout => true,
+    :ssl_ca_cert => nil,
+    :ssl_verify_mode => nil,
+    :ftp_active_mode => false,
+    :redirect => true,
+    :encoding => nil,
+  }
+
+  def OpenURI.check_options(options) # :nodoc:
```
iliabylich: This method (and the one below) are not used anywhere in the code. Do we need them?

test/opal/test_openuri.rb
Mogztter: Thanks for the pointer, we are now using Sinatra :tada:

stdlib/open-uri.rb
Mogztter: Ok I will stick with this syntax, now that we have tests we can improve the code

stdlib/open-uri.rb
Mogztter: Indeed, I have nightmares about regular expressions :rofl:

stdlib/open-uri.rb
Mogztter: I've implemented this method, I will add a test

stdlib/open-uri.rb
Mogztter: Yes I've removed all the code about options since we do not support them

tasks/testing.rake
```diff
@@ -227,6 +227,20 @@ module Testing
     end
   end
 
+  class HTTPServer
+    def with_server
+      begin
+        server = Process.spawn 'ruby test/opal/http_server.rb'
+        puts 'Waiting for server…'
+        sleep 0.1 until system "curl -s 'http://localhost:4567/' > /dev/null"
```
iliabylich: Do we run this suite on AppVeyor? If so this `curl` should be something platform-agnostic.

iliabylich: You can check the source code of the chrome runner, it spawns a server in the background. We don't run it on windows, but the code uses something like MRI's `TCPSocket` to check that the server is running.

Mogztter: Done! We should probably use it in `SauceLabs.with_server`

Mogztter:
Code Climate is not happy about `xmlhttprequest/lib/XMLHttpRequest.js`. Is there a way to disable inspection on `node_modules` ?

iliabylich:
@Mogztter Yes: https://docs.codeclimate.com/v1.0/docs/excluding-files-and-folders 😄

iliabylich:
Hmmm,

> Auto-generated file and folder exclusions
If you do not have a .codeclimate.yml committed, Code Climate will use a default config containing the following default exclude patterns:
config/
db/
dist/
features/
**/node_modules/
script/
**/spec/
**/test/
**/tests/
**/vendor/
**/*.d.ts

But we have one and it doesn't include `**/node_modules`. Could you add it please? It just doesn't make sense to track external libraries.

Mogztter:
I don't know why code climate want me to fix `stdlib/forwardable.rb` and `lib/opal/cli_runners` but I did some refactoring in https://github.com/opal/opal/pull/1745 and https://github.com/opal/opal/pull/1746

iliabylich:
> I don't know why code climate want me to fix stdlib/forwardable.rb and lib/opal/cli_runners but I did some refactoring in #1745 and #1746

As I know @elia can tell CodeClimate to ignore some issues manually 😄 
Anyway, thank you.

elia:
looking into it right now…

> As I know @elia can tell CodeClimate to ignore some issues manually 😄

I expected you was able too after signing in with github 🤷‍♂️

Mogztter:
Green bars :green_heart:

Mogztter:
@elia Please let me know if you want me to make any changes or additions :wink:

PR #1736, opened 2018-01-06
WIP 👷‍♂️ Remove symlinks in built gem, fixes installing on Windows

blacktm:
Unfortunately, symlinks cannot be written on Windows with MSYS2/MinGW during gem installation without elevated privileges, and so it fails. This PR restores the `bin/` directory and removes the symlink added in 66849b5 and removes another symlink in a `node_modules` dir (symlinks found using `find . -type l -ls`).

Before this patch, the install would fail with:

```
black@DESKTOP-9BINLQT MINGW64 /c/Users/black/Desktop
$ gem install opal-0.11.0.gem
ERROR:  While executing gem ... (Errno::EACCES)
    Permission denied @ rb_file_s_symlink - (./exe, C:/Ruby25-x64/lib/ruby/gems/2.5.0/gems/opal-0.11.0/bin)
```

After, gem installs successfully:

```
black@DESKTOP-9BINLQT MINGW64 /c/Users/black/Desktop
$ gem install opal-0.11.0-blacktm.gem
unknown encoding name ""+e)" for lib/opal/cli_runners/node_modules/chrome-remote-interface/chrome-remote-interface.js, skipping
Successfully installed opal-0.11.0
Parsing documentation for opal-0.11.0
Installing ri documentation for opal-0.11.0
Done installing documentation for opal after 20 seconds
1 gem installed

black@DESKTOP-9BINLQT MINGW64 /c/Users/black/Desktop
$ opal -v
Opal v0.11.0
```

Since this install error was not caught in the AppVeyor CI (perhaps because the commands are run in administrator mode, similar to `sudo`), I might suggest adding a manual step to the release process to test installing the built gem on Windows in the meantime (maybe other OSs as well).


elia:
Thanks! 

Is this mainly a revert of the commit that introduced the symlinks? I’d prefer not to revert the whole thing and just maybe add a placeholder in bin/opal that will instruct the user to use exe/opal. 

I think that installing the gem as part of the ci is a great suggestion, and it’s probably doable using bundler’s gem tasks: `rake install`. Can you add it to the PR?

blacktm:
@elia Oh, sorry, I misunderstood. Yes, I'll change it back to how it was, try to just remove the symlinks. Just curious, why are you moving away from the `bin/` convention and over to `exe/`? Also, this does seem to be a [ruby gems issue](https://github.com/rubygems/rubygems/issues/1623) — so frustrating. 😖 I'll try adding `rake install` to the CI config, although I worry they're running everything in administrator mode (I did some experimentation, [tried to get the build to fail](https://ci.appveyor.com/project/elia/opal/build/505#L398) with a simple `gem install opal`, but still passed). Oh, computers... 💻🔨

elia:
@blacktm FYI https://github.com/opal/opal/pull/1874, the analysis in this PR was very helpful 👍🏼 

the updated gemspec has been released as v0.11.2

Issue #1734, opened 2018-01-05
Pseudo-tree-shaking


jgaskins:
I had a look at the unused code in the Opal corelib:

<img width="1265" alt="Unused code report" src="https://user-images.githubusercontent.com/108205/34594045-b5c992da-f19c-11e7-976f-4ace19f297f8.png">

It's a lot.

This was for a Clearwater app on 0.10, but 0.11 is probably along the same lines. I was wondering if it might make sense to do a little bit of tree-shaking to filter out methods with names that are never called on any class. It wouldn't be perfect because of duck typing (for example, `MyClass#call` would make it into the build even if it were never called because `Proc#call` is used everywhere), but if it's possible to filter out method definitions and shrink the payload, total bundle size could probably be cut in half in a significant number of apps.

I'm not sure how this could work, though. I was thinking something like `compiler.method_calls`, but on a larger scale. It also means (and this is probably the hard part) that an _entire application_ would need to be parsed for method calls before any code was generated. I have a feeling that would require significant changes to the compiler — IIRC, the compiler generates the code as it parses, right?

If this can't be done, would a post-processor that grabs the `add_stubs` calls from a bundle and removes all calls to `Opal.defn` for methods that aren't in that set make sense?

jgaskins:
Took a shot at post-processing my concatenated/minified app bundle. Looks like this might take a bit more than just basing it on on `compiler.method_calls` / `Opal.add_stubs` to have much effect. Consider the following code:

```ruby
class Foo
  def bar
    baz
  end

  def baz
  end
end
```

In this example, `baz` is not removed because it's called by `bar`, but if nothing ever calls `bar`, `bar` gets removed, but `baz` is never actually invoked and should be removed.

In a quick attempt at doing this locally on my app bundle, I get: `{:stubs=>1018, :method_defs=>438, :filtered=>27}` — only about 6% of methods that would be filtered out. I don't know how many bytes that is, though, because I haven't parsed functions yet. This is just a quick scan with regexes so far.

I'm assuming the high ratio of stubs to defs is due to metaprogramming, but I'm not certain yet. That means it shouldn't get too many false positives, though, which is good. Should be able to whitelist any false positives that do occur, though, due to dynamic calls via `send` or whatever.

All this is to say that any tree shaking may need to make multiple passes and not be based on `add_stubs`.

jgaskins:
[...]

PR #1737, opened 2018-01-06
Close parentheses in error message

jgaskins:
Selfish reasons for doing this, tbh — my script mentioned in #1734 can't parse JS functions correctly without this change because it thinks it's still inside a parenthetical expression — it doesn't know about strings. :-) Adding a closing parenthesis here fixes it.


elia:
Thanks! this kind of selfishness is more than appreciated 🙂

PR #1740, opened 2018-01-10
Add a short $$ helper to navigate constants manually

### LINKED ISSUE #1554: Accessing nested modules?
I can't figure out how to access a nested module after loading an Opal-compiled file. Here's the ruby code:

``` ruby
module Foo
  module Bar
    class Baz
      def initialize
        puts "Init, wahoo"
      end
    end
  end
end
```

I tried to load the code in a node REPL and access the `Baz` class:

``` javascript
$> var Opal = require('/path/to/test.js');
$> Opal.Opal.Foo.Bar  // => undefined
```

So obviously I have no idea what I'm doing. A few questions:
1. How do I load and use opal-generated code in node and/or the browser? Opal doesn't appear to have exported any modules - what's the story here?
2. Why can't I access the `Baz` class? The few examples I've seen online seem to do it by simply dotting their way down the module hierarchy.

Thanks in advance for helping out an opal n00b.

elia:
The documentation should read better now

E.g.

```js
  Opal.$$.Regexp.$$.IGNORECASE
```

Fixes #1554 cc @camertron @radanskoric

### DIFF
--- a/docs/compiled_ruby.md
+++ b/docs/compiled_ruby.md
@@ -478,26 +478,107 @@ If you want to integrate a JavaScript library with Opal, so that you can make Ru
 ## Ruby from JavaScript
 
 Accessing classes and methods defined in Opal from the JavaScript runtime is
-possible via the `Opal` js object. The following class:
+possible via the `Opal` JS object. Consider following code:
 
 ```ruby
+module Bar
+  BAR = 123
+end
+
 class Foo
-  def bar
-    puts "called bar on class Foo defined in Ruby code"
+  include Bar
+  FOO = 456
+
+  def foo
+    puts "called #foo on class ::Foo defined in Ruby code"
   end
 end
+
+BAZ = 789
 ```
 
-Can be accessed from JavaScript like this:
+### Navigating constants
+
+Top level constants can be accessed as properties of `Opal`:
+
 
 ```javascript
-Opal.Foo.$new().$bar();
-// => "called bar on class Foo defined in Ruby code"
+Opal.Object;       // => Object
+Opal.Kernel;       // => Kernel
+Opal.Array;        // => Array
+Opal.RUBY_VERSION; // => "2.3"
+Opal.Foo;          // => Foo
+Opal.BAZ;          // => 789
 ```
 
-Remember that all Ruby methods are prefixed with a `$`.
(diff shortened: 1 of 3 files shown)


docs/compiled_ruby.md
```diff
@@ -478,26 +478,107 @@ If you want to integrate a JavaScript library with Opal, so that you can make Ru
...
+### Calling methods
+
+**Methods** are always prefixed by a single `$`:
 
 ```javascript
-Opal.Foo.$new().$bar();
-// => "called bar on class Foo defined in Ruby code"
+// Equivalent to:
+//   Foo.new.foo
+Opal.Foo.$new().$foo(); // => "called bar on class Foo defined in Ruby code"
```
iliabylich: Should be `called baz on class Foo defined in Ruby code`.

docs/compiled_ruby.md
```diff
@@ -478,26 +478,107 @@ If you want to integrate a JavaScript library with Opal, so that you can make Ru
...
+//   1 == 2
+(1)['$=='](2); // => false
+```
+
+**Passing blocks** from JS is a bit more tricky, so the suggested solution is to use `Opal.send(…)`:
+
+```javascript
+// Equivalent to:
+//   [1,2,3].map {|n| n * 2}
+Opal.send([1,2,3], 'map', [], function(n) {return n * 2}); // => [2,4,6]
```
iliabylich: Extra paren in the end of this line.

docs/compiled_ruby.md
```diff
@@ -478,26 +478,107 @@ If you want to integrate a JavaScript library with Opal, so that you can make Ru
...
+```
+
+Alternatively blocks can also be called *the hard way* by assigning the block-function to the property `$$p` of the method:
+
+```javascript
+// Equivalent to:
+//   [1,2,3].map {|n| n * 2}
+array = [1,2,3]
+array.$map.$$p = function(n) {return n * 2}); // "$$p" stands for "proc"
```
iliabylich: > Extra paren in the end of this line.
Sorry, my bad, this one.

radanskoric:
Thank you for improving this @elia! ❤️  

P.S. Sorry for not responding sooner, I was traveling without internet.

PR #1745, opened 2018-01-15
Remove code duplication in runners

Mogztter:
Code Climate is unhappy about the code duplication in runners.

We should probably use composition over inheritance but I didn't want to introduce too much changes.
Also `applescript.rb` has pretty much the same logic so we could refactor this file to use `cmd.rb`.

### DIFF
--- a/lib/opal/cli_runners.rb
+++ b/lib/opal/cli_runners.rb
@@ -66,6 +66,8 @@ module Opal
       names.each { |name| self[name] = runner }
     end
 
+    require 'opal/cli_runners/cmd'
+
     autoload :Applescript, 'opal/cli_runners/applescript'
     autoload :Chrome,      'opal/cli_runners/chrome'
     autoload :Nashorn,     'opal/cli_runners/nashorn'
--- a/lib/opal/cli_runners.rb
+++ b/lib/opal/cli_runners.rb
@@ -52,8 +52,6 @@ module Opal
       0
     }
 
-
-
     # Legacy runners
 
     def self.register_legacy_runner(klass_name, *names)
@@ -79,5 +77,26 @@ module Opal
     register_legacy_runner :Nashorn,     :nashorn
     register_legacy_runner :Nodejs,      :nodejs, :node
     register_legacy_runner :Server,      :server
+
+    # @elia Is this what you had in mind ?
+=begin
+    node_runner = -> data {
+
+      NODE_PATH = File.expand_path('../stdlib/nodejs/node_modules', ::Opal.gem_dir)
+      paths = ENV['NODE_PATH'].to_s.split(':')
+      paths << NODE_PATH unless paths.include? NODE_PATH
+      node_modules = paths.join(':')
+
+      command_options = {:name => 'nodejs', :env => {'NODE_PATH' => node_modules}, :cmd => 'node'}
+      command_options[:options] = data[:options] || {}
+      cmd = Cmd.new(command_options)
+
+      cmd.run(data[:builder].to_s, data[:argv])
+      cmd.exit_status
+    }
+    register_runner :nodejs, node_runner
+    register_runner :node, node_runner
+=end
+
   end
 end
--- a/lib/opal/cli_runners/applescript.rb
+++ b/lib/opal/cli_runners/applescript.rb
@@ -7,52 +7,28 @@ module Opal
         unless system('which osalang > /dev/null')
           raise MissingJavaScriptSupport, 'JavaScript Automation is only supported by OS X Yosemite and above.'
         end
-
-        @output = options.fetch(:output, $stdout)
+        command_options = {:name => 'applescript', :cmd => ['osascript', '-l', 'JavaScript']}
+        command_options[:options] = options
+        @cmd = Cmd.new(command_options)
       end
-      attr_reader :output, :exit_status
 
       def puts(*args)
-        output.puts(*args)
+        @cmd.puts(args)
       end
 
       def run(code, argv)
-        require 'tempfile'
-        tempfile = Tempfile.new('opal-applescript-runner-')
-        # tempfile = File.new('opal-applescript-runner.js', 'w') # for debugging
-        tempfile.puts code
-        tempfile.puts "'';" # OSAScript will output the last thing
-        tempfile.close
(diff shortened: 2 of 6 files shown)


lib/opal/cli_runners/nodejs.rb
```diff
@@ -1,66 +1,29 @@
...
 
       def puts(*args)
         output.puts(*args)
       end
 
       def node_modules
-        paths = ENV['NODE_PATH'].to_s.split(':')
-        paths << NODE_PATH unless paths.include? NODE_PATH
-        paths.join(':')
+        NODE_PATH
```
iliabylich: Previous version included `ENV['NODE_PATH']`. Even we nobody passes it it still doesn't make sense to have a method that returns a constant.

lib/opal/cli_runners/cmd.rb
```diff
@@ -0,0 +1,56 @@
...
+require 'shellwords'
+
+module Opal
+  module CliRunners
+    class Cmd
+      def initialize(options, name, env, *cmd)
+        @output = options.fetch(:output, $stdout)
+        @name = name
+        @env = env
+        @cmd = *cmd
```
iliabylich: Is `*` required here?

iliabylich:
> We should probably use composition over inheritance but I didn't want to introduce too much changes

Absolutely agree with you.

lib/opal/cli_runners/nodejs.rb
Mogztter: Oops, I missed my rebase 😞

lib/opal/cli_runners/cmd.rb
Mogztter: probably not, I'm not familiar with the splat operator

Mogztter:
> We should probably use composition over inheritance but I didn't want to introduce too much changes
Absolutely agree with you.

Ok, in this case, I will change the code to use composition

iliabylich:
Yeah, the change looks trivial, `super(command_options)` -> `Cmd.new(command_options).run`

elia:
@Mogztter @iliabylich composition should be easy also because the api is moving away from full-blown classes to "call-ables", like here:

https://github.com/opal/opal/blob/7ecfd4bb04cedab39339e01d1b674e44345bbceb/lib/opal/cli_runners.rb#L43-L53

Instead of migrating everything right away I added support for the legacy classes via this method:

https://github.com/opal/opal/blob/7ecfd4bb04cedab39339e01d1b674e44345bbceb/lib/opal/cli_runners.rb#L59-L67

@Mogztter if you think you can do as part of this PR good, otherwise we'll do it later I guess.

lib/opal/cli_runners.rb
```diff
@@ -77,5 +77,26 @@ def self.register_legacy_runner(klass_name, *names)
     register_legacy_runner :Nashorn,     :nashorn
     register_legacy_runner :Nodejs,      :nodejs, :node
     register_legacy_runner :Server,      :server
+
+    # @elia Is this what you had in mind ?
```
Mogztter: @elia ping :thinking:

lib/opal/cli_runners/applescript.rb
```diff
@@ -7,52 +7,28 @@ def initialize(options)
...
         end
-
-        @output = options.fetch(:output, $stdout)
+        command_options = {:name => 'applescript', :cmd => ['osascript', '-l', 'JavaScript']}
+        command_options[:options] = options
+        @cmd = Cmd.new(command_options)
       end
-      attr_reader :output, :exit_status
 
       def puts(*args)
```
Mogztter: I wasn't able to determine if this method is used or not...

lib/opal/cli_runners/applescript.rb
```diff
@@ -7,52 +7,28 @@ def initialize(options)
...
 
       def run(code, argv)
-        require 'tempfile'
-        tempfile = Tempfile.new('opal-applescript-runner-')
-        # tempfile = File.new('opal-applescript-runner.js', 'w') # for debugging
-        tempfile.puts code
-        tempfile.puts "'';" # OSAScript will output the last thing
-        tempfile.close
-        _successful = system_with_output('osascript', '-l', 'JavaScript', tempfile.path , *argv)
+        osascript_code = "#{code}\n'';" # OSAScript will output the last thing
```
Mogztter: I guess the following code are equivalent (but I may be wrong):

```ruby
tempfile.puts code
tempfile.puts "'';" # OSAScript will output the last thing
```

```ruby
tempfile.puts "#{code}\n'';"
```

lib/opal/cli_runners.rb
elia: Yeah, keep in mind that it can be done with modules/classes too:

```rb
# cli_runners/nodejs.rb

  module Opal
    module CliRunners
      class Nodejs
        def self.call(data)
          # …
        end
      end
    end
  end
```

```rb
# cli_runners.rb
register_runner :nodejs, Nodejs
```

lib/opal/cli_runners/applescript.rb
elia: `exit_status` is surely used here: https://github.com/opal/opal/blob/319a38463530a0da527fb190a3fb361608bc303e/lib/opal/cli_runners.rb#L59-L67

But of course is rewritten with the new API it's not needed

lib/opal/cli_runners.rb
Mogztter: > Yeah, keep in mind that it can be done with modules/classes too:

I will do that

lib/opal/cli_runners/applescript.rb
Mogztter: Speaking of `exit_status`, on the master branch the return code is `0` even when a spec fails... do you reproduce this issue ?

1. edit `spec/opal/core/array/intersection_spec.rb` to create a failure
1. run `bundle exec rake mspec_opal_nodejs PATTERN=spec/opal/core/array/intersection_spec.rb`
1. `echo $?` is `0`

Mogztter:
Reading my code again... I don't think it's really worth it.
It introduces a lot of code changes but without clear-cut improvement, sorry :disappointed:

PR #1746, opened 2018-01-15
Remove duplication in forwardable

Mogztter:


stdlib/forwardable.rb
```diff
@@ -1,16 +1,23 @@
+def forwardable_delegate(hash)
```
iliabylich: This code produces a global method that is available on every object. Extracting it to the module and including into `Forwardable/SingleForwardable` pollutes ancestors. How about making an inline JS function that takes a `hash` and a `callback` and invokes it with for every `[accessor, method]` pair?

iliabylich:
Also I don't understand why, but CodeClimate thinks that this PR introduces 2 new issues 😄

stdlib/forwardable.rb
Mogztter: > How about making an inline JS function that takes a hash and a callback and invokes it with for every [accessor, method] pair?

It's a good idea 👍

Mogztter:
> Also I don't understand why, but CodeClimate thinks that this PR introduces 2 new issues 😄

99 issues in the code, fix two issues, 99 issues in the code 😱🤣

elia:
I just approved this PR on codeclimate, it was the same two files… 🙄

stdlib/forwardable.rb
```diff
@@ -1,16 +1,30 @@
...
+      for (var j = 0; j < methods.length; j++) {
+        delegator(accessor, methods[j]);
       }
     }
+  }
+}
+
+module Forwardable
+  def instance_delegate(hash)
+    `forwardToDelegator(hash, def_instance_delegator)`
```
Mogztter: @iliabylich The code just feels wrong but somehow the tests suite passed... function `def_instance_delegator` should be undefined in the JavaScript environment ?

stdlib/forwardable.rb
```diff
@@ -1,16 +1,30 @@
-module Forwardable
-  def instance_delegate(hash)
-    hash.each {|methods, accessor|
-      methods = [methods] unless methods.respond_to? :each
-
-      methods.each {|method|
-        def_instance_delegator(accessor, method)
+%x{
+  function forwardToDelegator(hash, delegator) {
+    for (var i = 0, keys = hash.$$keys, smap = hash.$$smap, len = keys.length, key, value; i < len; i++) {
```
Mogztter: `hash` argument is an `Opal.Hash` object right ?

stdlib/forwardable.rb
```diff
@@ -1,16 +1,30 @@
...
+      methods = keys[i];
```
Mogztter: I took the following code from `opal/corelib/hash.rb`

iliabylich:
Oh, sorry, I've just realized that it was copied with minimal changes from MRI - https://github.com/ruby/ruby/blob/trunk/lib/forwardable.rb (I thought MRI has a C version).

Now I think that it's better to ignore this and close this PR. Having a code that is as similar to MRI as possible helps maintaining it.

Sorry again 😄 

@elia What do you think?

stdlib/forwardable.rb
iliabylich: Initially I thought that it could be possible because we compile `def def_instance_delegator` to `Opal.def(self, function def_instance_delegator ...` and so the function becomes accessible because of hoisting. But we have a `$$` prefix for generated function names, so it shouldn't work. 

The reason whey the build is green is simple: there are no tests for `Forwardable` in the rubyspec test suite.

PR #1754, opened 2018-01-19
Add String#prepend to corelib/unsupported

Mogztter:
String are immutable in JavaScript and thus String#prepend cannot be implemented.
The method now throws a "not implemented" exception.


elia:
@Mogztter I have the serious concern that implementing this could mislead both developers and libraries, you have a specific use-case in mind?

iliabylich:
Agree, I'd better add it to `corelib/unsupported`

Mogztter:
> I have the serious concern that implementing this could mislead both developers and libraries

I agree, I've changed the code in the Reveal.js converter:

```ruby
attrs_str = attrs.empty? ? '' : attrs.join(' ').prepend(' ')
```

```ruby
attrs_str = attrs.empty? ? '' : ' ' + attrs.join(' ')
```

> Agree, I'd better add it to corelib/unsupported

:+1: 

"String#prepend not supported. Mutable String methods are not supported in Opal."

PR #1756, opened 2018-02-04
Adds specs on Array.dup optimization

### LINKED ISSUE #1755: When using Array.dup, attribute $$id is added to the string representation
Consider the following code:

```js
const sourceLines = ["== Test", "This is the first paragraph.", "", "This is a second paragraph."];
const result = sourcesLine.$dup();
console.log('result', result);
// prints:

// [ '== Test',
// 'This is the first paragraph.',
// '',
// 'This is a second paragraph.',
// '$$id': 630 ]
```

Please note that the size of the array is right.
I don't know if it's intended or not but this is a new behavior introduced in the 0.11.0 release.

Mogztter:
Resolves #1755

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -869,12 +869,11 @@ class Array < `Array`
 
   def dup
     %x{
-      if (
-        self.$$class === Opal.Array &&
-        self.$$class.$allocate.$$pristine &&
-        self.$copy_instance_variables.$$pristine &&
-        self.$initialize_dup.$$pristine
-      ) return self.slice(0);
+      if (self.$$class.$allocate.$$pristine &&
+          self.$copy_instance_variables.$$pristine &&
+          self.$initialize_dup.$$pristine) {
+        return self.slice(0);
+      }
     }
 
     super
--- a/spec/opal/core/array/dup_spec.rb
+++ b/spec/opal/core/array/dup_spec.rb
@@ -1,0 +1,23 @@
+describe "Array#dup" do
+  it "should use slice optimization" do
+    a = Array.new
+    `a.slice = function() { return ['sliced'] }`
+    lambda { a.dup }.should_not raise_error
+    a.dup.should == ['sliced']
+  end
+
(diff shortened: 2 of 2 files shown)


opal/corelib/array.rb
```diff
@@ -869,12 +869,12 @@ def drop(number)
 
   def dup
     %x{
-      if (
-        self.$$class === Opal.Array &&
-        self.$$class.$allocate.$$pristine &&
-        self.$copy_instance_variables.$$pristine &&
-        self.$initialize_dup.$$pristine
-      ) return self.slice(0);
+      if (typeof self.slice === 'function' &&
```
iliabylich: I think it doesn't make sense to check that `.slice` is a function. This method gets called on the instance of Array / Array subclass, so it's always there.

Mogztter: @iliabylich It's an edge case but what about `delete Array.slice` ? 🤔

Mogztter: Ok, I've removed the check

opal/corelib/array.rb
```diff
@@ -869,12 +869,11 @@ def drop(number)
 
   def dup
     %x{
-      if (
-        self.$$class === Opal.Array &&
-        self.$$class.$allocate.$$pristine &&
-        self.$copy_instance_variables.$$pristine &&
-        self.$initialize_dup.$$pristine
-      ) return self.slice(0);
+      if (self.$$class.$allocate.$$pristine &&
```
iliabylich: `self.$$class === Opal.Array` should stay because slice produces an array:
``` ruby
>> class A < Array; end; `#{A.new}.slice().$class()`
=> Array
>> class A < Array; end; A.new.dup.class
=> A
```

So the method itself doesn't need any changes.
Sorry, missed that initially 😄

iliabylich: Wait, different classes have different allocators, ignore my previous comment please, everything is OK. There's even a test in the rubyspec's suite that passes.

iliabylich:
Thanks!

PR #1758, opened 2018-02-06
Add Date#next_year and Date#prev_year

jgaskins:
See http://ruby-doc.org/stdlib-2.5.0/libdoc/date/rdoc/Date.html#method-i-next_year

### DIFF
--- a/spec/opal/core/date_spec.rb
+++ b/spec/opal/core/date_spec.rb
@@ -112,11 +112,25 @@ describe Date do
     end
   end
 
+  describe "#next_year" do
+    it "returns the date with the next calendar year from self" do
+      Date.new(2013, 2, 5).next_year.should == Date.new(2014, 2, 5)
+      Date.new(2013, 2, 5).next_year(3).should == Date.new(2016, 2, 5)
+    end
+  end
+
   describe "#prev_month" do
     it "returns the date with the previous calendar month" do
       Date.new(2013, 2, 9).prev_month.should == Date.new(2013, 1, 9)
       Date.new(2013, 7, 31).prev_month.should == Date.new(2013, 6, 30)
       Date.new(2013, 1, 3).prev_month.should == Date.new(2012, 12, 3)
+    end
+  end
+
+  describe "#prev_year" do
+    it "returns the date with the previous calendar year from self" do
+      Date.new(2013, 2, 5).prev_year.should == Date.new(2012, 2, 5)
+      Date.new(2013, 2, 5).prev_year(3).should == Date.new(2010, 2, 5)
     end
   end
 
--- a/stdlib/date.rb
+++ b/stdlib/date.rb
@@ -547,6 +547,10 @@ class Date
     }
   end
 
+  def next_year(years=1)
+    self.class.new(year + years, month, day)
+  end
+
   def prev_day(n=1)
     self - n
   end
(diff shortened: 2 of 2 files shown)


spec/opal/core/date_spec.rb
```diff
@@ -112,6 +112,13 @@
     end
   end
 
+  describe "#next_year" do
```
iliabylich: Could you try to remove filters for `Date#next_year` and `Date#prev_year` in `spec/filters/bugs/date.rb`? If they pass, remove your specs please.
You can run all `Date` specs using `env RUBYSPECS=true PATTERN=spec/ruby/library/date/*_spec.rb bundle exec rake mspec_ruby_nodejs`

stdlib/date.rb
```diff
@@ -547,6 +547,10 @@ def next_month(n = 1)
     }
   end
 
+  def next_year(years=1)
```
iliabylich: Add spaces around `=` please

stdlib/date.rb
```diff
@@ -561,6 +565,10 @@ def prev_month(n = 1)
     }
   end
 
+  def prev_year(years=1)
```
iliabylich: Add spaces around `=` please

spec/opal/core/date_spec.rb
jgaskins: I followed existing 4-year-old conventions for the specs for the other `Date#prev_*` and `Date#next_*` methods. If those conventions are wrong, I put these specs right next to the others so you can delete them all at once.

iliabylich: > I followed existing 4-year-old conventions for the specs for the other Date#prev_* and Date#next_* methods

There's no such convention.

There are a lot of specs under `spec/opal/` that are just a copy of rubyspec's test suite. Initially they were copied (with small changes) because self-written parser (or compiler) was unable to parse/compile rubyspec specs. 

Ideally `spec/opal` directory should include only Opal-specific specs (like specs for `.JS` syntax or some edge case optimizations). We should review them and drop duplicates (definitely not in this PR).

stdlib/date.rb
jgaskins: There aren't consistent conventions for [spaces](https://github.com/opal/opal/blob/0374448b034e88662b329e95af0aa77cba9d40d3/stdlib/observer.rb#L126) [around](https://github.com/opal/opal/blob/0374448b034e88662b329e95af0aa77cba9d40d3/stdlib/observer.rb#L167) [default](https://github.com/opal/opal/blob/0374448b034e88662b329e95af0aa77cba9d40d3/stdlib/pp.rb#L22) args. Those 3 are just what I noticed in `stdlib`. There are probably more.

It's reasonable to ask contributors to use existing conventions, but it'd be courteous to them to fix the conventions used in your project before asking those contributors to do more work to conform to them. When they're not written down anywhere (whether it's a line in the [contribution guidelines](https://github.com/opal/opal/blob/master/CONTRIBUTING.md#submitting-a-pull-request) or a linter either in CI or a GitHub PR bot), the only way people know what your conventions are is by looking at existing code. If existing conventions are not what you're looking for, it's not the responsibility of the contributor to fix it.

spec/opal/core/date_spec.rb
jgaskins: You just denied a convention exists and then explained with more words why that convention exists.

PR #1763, opened 2018-02-07
Removed redundant spec from spec/opal that were initially copied from RubySpec.

iliabylich:
Before: 688 examples
After: 324 examples


elia:
Nice!

PR #1765, opened 2018-02-11
whitequark/parser requires String#unpack

Mogztter:
Therefore `opal-builder.js` needs to require 'corelib/string/unpack'

See: https://github.com/whitequark/parser/blob/v2.4.0.2/lib/parser/lexer.rl#L186


Mogztter:
Thanks @iliabylich for the quick review.
On a related note, I'm investigating an issue on Asciidoctor.js. The compilation phase is now failing, I think it's related to the `parser/ruby25` upgrade, I will let you know 😉

PR #1766, opened 2018-02-12
Optimize String#unpack

Mogztter:
- 'C*' returns the buffer
- 'c*', 'a*', 'A*' reads all

### DIFF
--- a/opal/corelib/string/unpack.rb
+++ b/opal/corelib/string/unpack.rb
@@ -686,6 +686,15 @@ class String
       'p': null
     }
 
+    var optimized = {
+      'C*': identityFunction,
+      'c*': toNByteSigned(1, identityFunction),
+      'A*': wrapIntoArray(joinChars(bytesToAsciiChars(filterTrailingZerosAndSpaces(identityFunction)))),
+      'a*': wrapIntoArray(joinChars(bytesToAsciiChars(identityFunction))),
+      'M*': qpdecode(joinChars(bytesToAsciiChars(identityFunction))),
+      'm*': base64Decode(joinChars(bytesToAsciiChars(identityFunction)))
+    }
+
     function alias(existingDirective, newDirective) {
       readChunk[newDirective] = readChunk[existingDirective];
       handlers[newDirective] = handlers[existingDirective];
@@ -706,6 +715,12 @@ class String
       var output = [];
 
       var buffer = utf16LEToBytes(self);
+
+      // optimization
+      var optimizedHandler = optimized[format];
+      if (optimizedHandler) {
+        return optimizedHandler(buffer);
+      }
 
       function autocomplete(array, size) {
         while (array.length < size) {
@@ -740,7 +755,7 @@ class String
         var part = processChunk(directive, count);
 
         if (count !== Infinity) {
-          var shouldAutocomplete = autocompletion[directive]
+          var shouldAutocomplete = autocompletion[directive];
 
           if (shouldAutocomplete == null) {
             #{raise "Unsupported unpack directive #{`directive`.inspect} (no autocompletion rule defined)"}
--- a/opal/corelib/string/unpack.rb
+++ b/opal/corelib/string/unpack.rb
@@ -686,19 +686,13 @@ class String
       'p': null
     }
 
-    function toArray(callback) {
-      return function(data) {
-        return [callback(data)];
-      }
-    }
-
     var optimized = {
       'C*': identityFunction,
       'c*': toNByteSigned(1, identityFunction),
       'A*': wrapIntoArray(joinChars(bytesToAsciiChars(filterTrailingZerosAndSpaces(identityFunction)))),
       'a*': wrapIntoArray(joinChars(bytesToAsciiChars(identityFunction))),
-      'M*': toArray(qpdecode(joinChars(bytesToAsciiChars(identityFunction)))),
-      'm*': toArray(base64Decode(joinChars(bytesToAsciiChars(identityFunction)))),
+      'M*': wrapIntoArray(qpdecode(joinChars(bytesToAsciiChars(identityFunction)))),
+      'm*': wrapIntoArray(base64Decode(joinChars(bytesToAsciiChars(identityFunction)))),
       'S*': mapChunksToWords(chunkBy(2, identityFunction)),
       's*': toNByteSigned(2, mapChunksToWords(chunkBy(2, identityFunction))),
       'L*': mapChunksToWords(chunkBy(4, identityFunction)),
--- a/opal/corelib/string/unpack.rb
+++ b/opal/corelib/string/unpack.rb
@@ -687,24 +687,24 @@ class String
     }
 
     var optimized = {
-      'C*': identityFunction,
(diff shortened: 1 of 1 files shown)


iliabylich:
I think we can optimize it even more by having
``` js
var readOptimized = {
  'C*': identityFunction,
  'c*': ...,
  'A*': ...,
  'a*': ...
}

// later in `unpack`
var buffer = utf16LEToBytes(self);

var optimized = readOptimized[format]
if (optimized) {
  return optimized(buffer);
}
// else use buffers and other non-optimized code
```

All cases with <char><star> format patterns can be handled without buffers or chunks.

opal/corelib/string/unpack.rb
```diff
@@ -686,6 +686,15 @@ class String
       'p': null
     }
 
+    var optimized = {
+      'C*': identityFunction,
```
iliabylich: Hmm, didn't expect it to be that composable 😄 
Now I start thinking that probably it makes sense to use references to handlers here (because what you've got is a true code duplication).
```
{
  'C*': handlers['C'],
  'c*': handlers['c'],
  ...
}
```

But the way, how fast is it? It can be optimized even more with inline functions:
``` js
'A*': function(buffer) {
  var string = String.fromCharCode.apply(null, buffer);
  string = string.replace(/[\u0000 ]+$/, '');
  return [string];
},
'a*': function(buffer) {
  return String.fromCharCode.apply(null, buffer);
} 
```

iliabylich: Moreover, I guess case for `CHAR*` can be generalized for **all** chars. I'm checking it locally

opal/corelib/string/unpack.rb
```diff
@@ -686,6 +686,33 @@ class String
...
+      'a*': wrapIntoArray(joinChars(bytesToAsciiChars(identityFunction))),
+      'M*': toArray(qpdecode(joinChars(bytesToAsciiChars(identityFunction)))),
+      'm*': toArray(base64Decode(joinChars(bytesToAsciiChars(identityFunction)))),
+      'S*': mapChunksToWords(chunkBy(2, identityFunction)),
+      's*': toNByteSigned(2, mapChunksToWords(chunkBy(2, identityFunction))),
+      'L*': mapChunksToWords(chunkBy(4, identityFunction)),
+      'l*': toNByteSigned(4, mapChunksToWords(chunkBy(4, identityFunction))),
+      'Q*': mapChunksToWords(chunkBy(8, identityFunction)),
+      'q*': toNByteSigned(8, mapChunksToWords(chunkBy(8, identityFunction))),
+      'S>*': mapChunksToWords(invertChunks(chunkBy(2, identityFunction))),
```
Mogztter: Not sure about this notation `S>*` ?

opal/corelib/string/unpack.rb
```diff
@@ -686,6 +686,33 @@ class String
...
+      }
+    }
+
+    var optimized = {
+      'C*': identityFunction,
+      'c*': toNByteSigned(1, identityFunction),
+      'A*': wrapIntoArray(joinChars(bytesToAsciiChars(filterTrailingZerosAndSpaces(identityFunction)))),
+      'a*': wrapIntoArray(joinChars(bytesToAsciiChars(identityFunction))),
+      'M*': toArray(qpdecode(joinChars(bytesToAsciiChars(identityFunction)))),
+      'm*': toArray(base64Decode(joinChars(bytesToAsciiChars(identityFunction)))),
```
Mogztter: I've added a `toArray` function because the specs on `M*` and `m*` expect an Array.

opal/corelib/string/unpack.rb
iliabylich: No, some of them wrap result into array (you just pushed a function `toArray`):
``` fails "String#unpack with format 'B' decodes all the bits when passed the '*' modifier" # Expected [""] to be computed by "".unpack from "B*" (computed "" instead)
fails "String#unpack with format 'H' decodes all the nibbles when passed the '*' modifier" # Expected [""] to be computed by "".unpack from "H*" (computed "" instead)
fails "String#unpack with format 'M' ignores the count or '*' modifier and decodes the entire string" # Expected ["abc"] to be computed by "a=\nb=\nc=\n".unpack from "M*" (computed "abc" instead)
fails "String#unpack with format 'U' raises ArgumentError on a malformed byte sequence and doesn't continue when used with the * modifier" # Expected ArgumentError but no exception was raised ([227] was returned)
fails "String#unpack with format 'Z' decodes past whitespace bytes when passed the '*' modifier" # Expected ["a b c"] to be computed by "a b c".unpack from "Z*" (computed "a b c" instead)
fails "String#unpack with format 'b' decodes all the bits when passed the '*' modifier" # Expected [""] to be computed by "".unpack from "b*" (computed "" instead)
fails "String#unpack with format 'h' decodes all the nibbles when passed the '*' modifier" # Expected [""] to be computed by "".unpack from "h*" (computed "" instead)
fails "String#unpack with format 'm' ignores the count or '*' modifier and decodes the entire string" # Expected ["abcDEF"] to be computed by "YWJj\nREVG\n".unpack from "m*" (computed "abcDEF" instead)
fails "String#unpack with format 'u' ignores the count or '*' modifier and decodes the entire string" # Expected ["abcDEF"] to be computed by "#86)C\n#1$5&\n".unpack from "u*" (computed "abcDEF" instead)
```

Mogztter: > Now I start thinking that probably it makes sense to use references to handlers here (because what you've got is a true code duplication).

True 😐 

> But the way, how fast is it? It can be optimized even more with inline functions:

Hmmm, interesting... 
> Moreover, I guess case for CHAR* can be generalized for all chars. I'm checking it locally

I think that's what I did, if the buffer remains intact then we can use directly the handler function.

opal/corelib/string/unpack.rb
Mogztter: Oh! I've missed the `wrapIntoArray` function... composition FTW! 🤣🤣🤣

opal/corelib/string/unpack.rb
iliabylich: Why?
``` ruby
('a'..'z').to_a.join.unpack("S>*")
 => [24930, 25444, 25958, 26472, 26986, 27500, 28014, 28528, 29042, 29556, 30070, 30584, 31098]

File: opal/corelib/string/unpack.rb (hunk 1 above)
Comment (iliabylich) [reply opal/corelib/string/unpack.rb]: I was wrong, there's also an autocompletion rule 😄 I love this method, guys from MRI completely ignore SRP.

> True 😐

so, how about
```
{
  'C*': handlers['C'],
  'c*': handlers['c'],
  ...
  'a*': wrapIntoArray(handlers['a'])
}
```

opal/corelib/string/unpack.rb
Mogztter: I wasn't sure that `S>*` was a valid format

opal/corelib/string/unpack.rb
Mogztter: > so, how about...

Done!

PR #1767, opened 2018-02-17
Builder fix and allow for .rb.erb

janbiedermann:
.rb.erb is required to make ruby-hyperloop work with electron-opal
ruby-hyperloop cant use .opalerb because the files are required server (matz ruby) and client side (opal ruby) the same

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -153,9 +153,20 @@ module Opal
     end
 
     def processor_for(source, filename, path, options)
+      if path
+        first_extname = File.extname(path)
+        second_path = path[0...(-first_extname.length)]
+        second_extname = File.extname(second_path)
+      end
       processor = processors.find { |p| p.match? path } or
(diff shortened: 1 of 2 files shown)


janbiedermann:
oops, dont know why the fix multiline is still in there, tell me if it hurts and to what branch i shall provide the PR

janbiedermann:
this needs more work

elia:
Have you tried working on the server side to support `opalerb` as an extension? I have nothing against improving support for this things but maybe it's easier (probably depends on the cruby framework).


An additional consideration:

I'd say that we need a better API for extension/compiler support in builder, but that's also something that brings in the direction of a full blown building system like Sprockets or Webpack etc. So my hope is that we'll find the sweet spot between the two extremes 😄

janbiedermann:
so actually thats what i am doing here https://github.com/janbiedermann/opal-webpack-loader
ruby-hyperloop eliminates .erb files, so this is no longer a issue
moving to webpack seems to provider improved options for building assets
Thats why I close this

PR #1768, opened 2018-02-17
fix multiline detection

janbiedermann:
from #1752


elia:
fixed in https://github.com/opal/opal/pull/1776

PR #1770, opened 2018-02-20
Run rubocop on stdlib

iliabylich:
Last part!


stdlib/nodejs/require.rb
```diff
@@ -16,15 +16,15 @@ def __prepare_require__(path)
     end
 
     name
-  rescue => e
-    raise [path, name, full_path].inspect+e.message
+  rescue StandardError => e
```
elia: looked up in the style guide and didn't find it enforced, let alone a good reason to do that, can the rule be disabled if present?

stdlib/nodejs/rubygems.rb
```diff
@@ -56,11 +56,9 @@ def curl url
...
 when 'install'
   name = ARGV.shift
-  if ARGV.include? '-v'
-    version = ARGV[ ARGV.index('-v')+1 ]
-  else
-    version = nil
-  end
+  version = if ARGV.include? '-v'
+              ARGV[ARGV.index('-v') + 1]
+            end
```
elia: For this code the author clearly wanted to highlight the fact that version can end up intentionally nil, I'm ok with the original version, another good one would be:

```rb
version = nil
version = ARGV[ARGV.index('-v') + 1] if ARGV.include? '-v'
```

_note 1: I didn't actually check who the author was (could be me of course)_
_note 2: I suspect this file could be removed entirely_ 😄

stdlib/opal-platform.rb
```diff
@@ -3,13 +3,10 @@
...
-else # possibly browser, which is the primary target
-  OPAL_PLATFORM = nil
-end
+OPAL_PLATFORM = if nashorn
+                  'nashorn'
+                elsif node
+                  'nodejs'
+                elsif headless_chrome
+                  'headless-chrome'
+                end
```
elia: I find the original form much better, also the comment disappeared

Since I also don't like very much all the other changes of this kind (mostly `case/when`) can we ditch this rule altogether?

stdlib/ostruct.rb
```diff
@@ -15,26 +17,33 @@ def []=(name, value)
...
-        raise ArgumentError.new "wrong number of arguments (0 for 1)"
+        raise ArgumentError, 'wrong number of arguments (0 for 1)'
       end
-      @table[new_ostruct_member(name[0 .. -2])] = args[0]
+      @table[new_ostruct_member(mid[0..-2])] = args[0]
+    elsif args.empty?
+      @table[mid.to_sym]
     else
-      @table[name.to_sym]
+      super
```
elia: Is this an unrelated fix? If so maybe it's worth its own PR or going straight to master

stdlib/ostruct.rb
```diff
@@ -15,26 +17,33 @@ def []=(name, value)
...
     end
   end
 
+  def respond_to_missing?(mid, include_private = false)
+    mname = mid.to_s.chomp('=').to_sym
+    @table&.key?(mname) || super
+  end
+
```
elia: same as above

stdlib/promise.rb
```diff
@@ -146,7 +146,7 @@ def exception?
   end
 
   def realized?
-    !!@realized
+    @realized != false
```
elia: This is not equivalent, did you check if the new code is safe?

stdlib/nodejs/irb.rb
```diff
@@ -2,44 +2,39 @@
...
+                               prompt_interrupted = false
+                               $repl.prompt = prompt.call(self)
+                               begin
+                                 result = `OpalNode.run(cmd, filename)`
+                                 result = nil if `#{result} == nil`
+                                 callback.call('=> ' + result.inspect)
+                               rescue StandardError => e
+                                 callback.call(e.backtrace.join("\n"))
+                               end
+                             }
```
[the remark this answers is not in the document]
iliabylich: Agree, I'll fix it

stdlib/nodejs/rubygems.rb
iliabylich: Yeah, it contains some totally weird code (there was an interpolation in the single-quoted string and it invokes `system` in a couple of places)

stdlib/ostruct.rb
iliabylich: That's what MRI does. Would you like me to revert it, add to filters for now and send it in a separate PR?

stdlib/promise.rb
iliabylich: Yes, `@realized` can be `false` or a symbol, also all specs pass (there are some in `spec/opal/stdlib/promise/*.rb` I guess)

stdlib/ostruct.rb
elia: @iliabylich commit straight to master, the diff in this PR should become clear after that

iliabylich:
@elia Ping

stdlib/nodejs/irb.rb
```diff
@@ -2,44 +2,42 @@
...
+          prompt_interrupted = false
+          $repl.prompt = prompt.call(self)
+          begin
+            result = `OpalNode.run(cmd, filename)`
+            result = nil if `#{result} == nil`
+            callback.call('=> ' + result.inspect)
+          rescue => e
+            callback.call(e.backtrace.join("\n"))
+          end
+        },
```
elia: It's ok anyway but I think the body of this proc should be just one indentation level more than `eval:`

stdlib/opal-platform.rb
```diff
@@ -3,13 +3,11 @@
...
-  OPAL_PLATFORM = nil
-end
+OPAL_PLATFORM = if nashorn
+                  'nashorn'
+                elsif node
+                  'nodejs'
+                elsif headless_chrome
+                  'headless-chrome'
+                else # possibly browser, which is the primary target
+                end
```
elia: still prefer the earlier version with its regular repetition, also the missing nil bothers me

stdlib/pathname.rb
```diff
@@ -87,23 +87,23 @@ def plus(path1, path2) # -> path # :nodoc:
...
-    while r2 = chop_basename(prefix2)
+    while (r2 = chop_basename(prefix2))
       prefix2, basename2 = r2
       index_list2.unshift prefix2.length
       basename_list2.unshift basename2
     end
     return path2 if prefix2 != ''
     prefix1 = path1
-    while true
+    loop do
```
elia: I'll guess 🤖 👮 has a good reason for doing this but in opal while true looks faster:

```
          while true   407.000  i/100ms
                loop   184.000  i/100ms
          while true      7.088k (± 2.6%) i/s -     35.409k in   4.999414s
                loop      1.970k (± 2.6%) i/s -      9.936k in   5.046192s

Comparison:
          while true:     7087.7 i/s
                loop:     1970.5 i/s - 3.60x  slower
```

code:

```rb
Benchmark.ips do |x|
  x.report('while true') do
    n = 0
    while true
      n += 1
      break if n == 10000
    end
  end

  x.report('loop') do
    n = 0
    loop do
      n += 1
      break if n == 10000
    end
  end

  x.compare!
end
```

iliabylich:
@elia I've rebased this branch on the master, the build is green. Any comments?

iliabylich:
@elia Thanks for patience!

PR #1773, opened 2018-02-21
Raise LoadError directly from the runtime

elia:
This is potentially a breaking change for people that were relying on the bug but I think it's worth fixing it.

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -63,12 +63,13 @@ module Opal
         public_send("#{k}=", v)
       end
 
-      @stubs             ||= []
-      @preload           ||= []
-      @processors        ||= ::Opal::Builder.processors
-      @path_reader       ||= PathReader.new(Opal.paths, extensions.map { |e| [".#{e}", ".js.#{e}"] }.flatten)
-      @prerequired       ||= []
-      @compiler_options  ||= Opal::Config.compiler_options
+      @stubs                    ||= []
+      @preload                  ||= []
+      @processors               ||= ::Opal::Builder.processors
+      @path_reader              ||= PathReader.new(Opal.paths, extensions.map { |e| [".#{e}", ".js.#{e}"] }.flatten)
+      @prerequired              ||= []
+      @compiler_options         ||= Opal::Config.compiler_options
+      @missing_require_severity ||= Opal::Config.missing_require_severity
 
       @processed = []
     end
@@ -83,6 +84,7 @@ module Opal
     end
 
     def build_str(source, filename, options = {})
+      return if source.nil?
       path = path_from_filename(filename)
       asset = processor_for(source, filename, path, options)
       requires = preload + asset.requires + tree_requires(asset, path)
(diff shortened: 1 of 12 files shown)


lib/opal/builder.rb
```diff
@@ -176,10 +177,11 @@ def read(path)
         case compiler_options[:dynamic_require_severity]
         when :raise   then raise MissingRequire, message
         when :warning then warn message
-        else # noop
+        else # noop (:ignore, :load_error)
         end
 
-        return "raise LoadError, #{message.inspect}"
+        # return "raise LoadError, #{message.inspect}"
```
iliabylich: Do we still need this line?

iliabylich:
@elia So if I understand correctly,
+ for `dynamic_require_severity = :raise` compiler throws an error for any static require, and later potentially throws a `LoadError` for dynamic requires (like via `Opal.load`), this part is basically the fix itself.
+ for `dynamic_require_severity = :warning` compiler prints a message during compilation and in runtime.

Could you clarify it please?
Also CLI doesn't support `-Draise`, so I'm confused a bit about this PR.

elia:
@iliabylich ok, first thing that `:raise` is actually a typo, it should be `:error` instead. Second, it should have worked this way:

`Opal::Config.dynamic_require_severity` controls compile time, while runtime is controlled by `Opal.config.missing_require_severity`. 

Before the fix the runtime part wasn't working (it was missing a `throw`) except by that hack in builder that was returning `throw LoadError` as the source for missing files.

It might be the case that there's still no way to ignore missing files altogether, or maybe there is.

I surely need to add some specs for the whole thing so to better define the final behavior.

lib/opal/builder.rb
elia: It's a leftover, I'll remove it

iliabylich:
@elia Have you finished with this PR? (Asking because you removed a WIP label but the comment that I mentioned is still there 😅)

elia:
@iliabylich yeah, I removed the WIP because the bulk of the work is done, of course I since removed the commented line. But the CI is still red thanks to our good 🤖👮 , I think we should move it from traivis/appveyor to codeclimate in order to have the CI check for correctness and codeclimate for cosmetics 💅 (which are important, don't get me wrong 😄)

iliabylich:
@elia Absolutely agree

elia:
Just wrote to codeclimate asking why it's not reporting the same "offenses" as travis

elia:
I'll close and reopen to try out houndci and see if it does a better job

iliabylich:
HoundCI doesn't use `.rubocop.yml`, right? (at least in the basic configuration) I see a lot of warnings in the files that are excluded.

elia:
@iliabylich as soon as I saw how houndci operates I removed the integration, it's too noisy. I'm also a bit confused about how these code analyzers read `.rubocop.yml` because codeclimate too seems to ignore it to some degree (but at least it's not that noisy!).

Anyway for now we'll stick to rubocop in travis.

PR #1774, opened 2018-02-22
Optimize Opal.is_a (0.10.x backport)

jgaskins:
Previously, this call would construct a full list of ancestors on each
call. This was a great abstraction, but this optimizes that check by
walking the ancestor chain manually and bailing out early.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1241,28 +1241,25 @@
     return null;
   };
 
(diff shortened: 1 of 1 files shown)


jgaskins:
These spec failures seem to happen on the `0-10-stable` branch, too. I'm not sure I understand why.

elia:
I'm closing this too, let me know if you still need this

PR #1776, opened 2018-02-28
Fix xstring management

elia:
Reimplement x-string compilation
- avoid fancy stuff when the x-str is passed to Compiler#returns, just
  wrap it and let the node do its job
- ignore empty lines at the beginning and end of the x-str
- assume multi-line x-str will properly use `;` and `return`
- assume single-line x-str can accidentally have trailing `;` and 
  leading `return`

For single line x-strings:
- remove a trailing `;`
- detect an embedded `return`
- prepend a `return` when needed
- append a `;` when needed
- warn the user not to use the semicolon in single-line x-strings


lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
+          # - detect an embedded `return`
+          # - prepend a `return` when needed
+          # - append a `;` when needed
+          # - warn the user not to use the semicolon in single-line x-strings
+
+          first_child = stripped_children.shift
+
+          if first_child.type == :str
+            first_value = first_child.loc.expression.source.strip
+            has_embeded_return = first_value =~ /^return\b/
```
iliabylich: `\b` is not supported by Opal, so it breaks `opal-parser`

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
+      #   # this will generate two additional empty
+      #   # children before and after `foo()`
+      #   %x{
+      #     foo()
+      #   }
+      def strip_empty_children(children)
+        start_index = 0
+        end_index = children.size - 1
+
+        while start_index <= end_index &&
```
iliabylich: At least the first loop definitely could be replaced to

``` ruby
children.drop_while { |child| <while condition> }
```

But there's no `pop_while` or any similar method. Maybe something like:
``` ruby
_, *children_to_keep, _ = children.chunk { |child| <while condition> }
```

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
       def compile
-        children.each do |child|
-          case child.type
-          when :str
-            value = child.loc.expression.source
-            if expr? && children.size == 1
-              value = detect_and_remove_trailing_semicolon(value, child)
+        should_add_semicolon = false
+        has_embeded_return   = false
+        returning, children  = unpack_return(self.children)
```
iliabylich: What do you think about moving children dispatching to constructor? (like we do in `def/iter/call` nodes)

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
+          end
+
+          if returning && !has_embeded_return
+            push('return ')
+          end
+
+          last_child = stripped_children.pop || first_child
+          if last_child.type == :str
+            last_value = last_child.loc.expression.source.rstrip
+            if (returning || expr?) && last_value[-1] == ';'
```
iliabylich: Maybe `last_value.end_with?(';')` ?

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
-          value
+          raise "Unsupported xstr part: #{child.type}"
+        end
+      end
+
+      # Check if there's only one child or if they're all part of
+      # the same line (e.g. because of interpolations)
+      def single_line?(children)
+        (children.size == 1) ||
+          children.none? do |c|
```
iliabylich:
``` ruby
children
  .select { |c| c.type == :str }
  .none? { |c| c.loc.expression.source.end_with?("\n") }
```

(Not sure if it's better though 😄 )

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
+      def unpack_return(children)
+        first_child = children.first
+        returning   = false
+
+        if first_child.type == :js_return
+          returning = true
+          children = first_child.children
         end
+
+        return returning, children
```
iliabylich: CI is green?! There's an explicit return on the last line, `[return, children]` should be enough.

iliabylich: Oh, my bad, I was looking into the log of the last commit pushed into maser.

lib/opal/nodes/literal.rb
```diff
@@ -213,42 +213,132 @@ class XStringNode < Base
...
+
+          last_child = stripped_children.pop || first_child
+          if last_child.type == :str
+            last_value = last_child.loc.expression.source.rstrip
+            if (returning || expr?) && last_value[-1] == ';'
+              compiler.warning(
+                'Removed semicolon ending x-string expression, interpreted as unintentional',
+                last_child.line,
+              )
+              last_value = last_value[0...-1]
```
iliabylich: Also maybe it's better to rewrite it to `[0..-2]`. Initially I thought that there's a bug (or a weird self-wriiten `dup` 😄 )

spec/lib/compiler_spec.rb
```diff
@@ -308,20 +308,106 @@
...
+      it 'does not warn for multiline x-strings' do
+        expect_compiler(%{a = `1;\n2;`})          { |c| c.not_to receive(:warning) }
+        expect_compiler(%{a = `1;\n2;3; `})       { |c| c.not_to receive(:warning) }
+        expect_compiler(%{def foo;` 1;\n  `;end}) { |c| c.to receive(:warning).once }
+      end
+    end
 
-      compiler = Opal::Compiler.new('`1;`')
-      expect(compiler).not_to receive(:warning)
+    def expect_compiler(code)
```
iliabylich: That's a very, very complex construction.

I like how it's written in the parser gem. There's a field on the parser called `diagnostics_consumer` (or something similar, can't remember) that gets called on every error or warning. In the "test" mode it simply collects all arguments (adapter pattern) and puts them to some static array (like `ScratchPad` in rubyspec). Later in the spec any matcher can directly use this array.

elia: I see, basically it would be like 

```rb
warnings = []
allow_any_instance_of(Opal::Compiler).to receive(:warning) { |*args| warnings << args }
expect(warnings.size).to eq(1)
# etc.
```

---

The reason I used this odd construct is that it worked well with backtrace being the expectation at the same exact line that generated it. Let me see if I can come up with anything better

elia:
I'll extract xstr from literal once we're good with the diff

---

OT: but this is the best approach to rubocop I've seen so far, allowing to take everything with a grain of salt and to break the rules where it makes sense, although I wonder if "wontfix" comments would pop up in later pull-requests…

screen shot 2018-03-07 at 00 15 06

_I marked this as "wontfix" just for the sake of trying it out_

lib/opal/nodes/literal.rb
iliabylich: My bad, I thought JS doesn't support `\b`

lib/opal/nodes/literal.rb
iliabylich: That's true for Opal, but what about MRI? `opal-parser` is slow anyway 😄 . And from what I understand `opal-parser` is just a proof-of-concept, everyone uses Ruby-based version of Opal.

iliabylich: Flip-flop solves it very well by they way :)

Mogztter:
@elia I've upgraded to the latest version `48bac38` and in some cases I have an extra semicolon:

```diff
       if ($truthy(path['$respond_to?']("path"))) {
         path = path.$path()};
-      return executeIOAction(function(){return __fs__.existsSync(path)});
+      return executeIOAction(function(){return __fs__.existsSync(path)});;
     }, TMP_File_exist$q_3.$$arity = 1);
```

```diff
-    Opal.defn(self, '$===', TMP_Module_$eq$eq$eq_3 = function(object) {
+    Opal.def(self, '$===', TMP_Module_$eq$eq$eq_3 = function(object) {
       var self = this;
 
       
       if ($truthy(object == null)) {
         return false};
-      return Opal.is_a(object, self);
+      return Opal.is_a(object, self);;
     }, TMP_Module_$eq$eq$eq_3.$$arity = 1);

```

elia:
@Mogztter thanks, working on a fix, should be ready soon

elia:
@Mogztter see https://github.com/opal/opal/commit/7944860ae4473af5cf0ef4d9763e3d06c9668e38

Mogztter:
Nice :+1:

Mogztter:
@elia I've just upgraded and the extra semicolon is gone but now I have lines without semicolon (but maybe it's expected):

https://github.com/opal/opal/blob/fc20415fe76bc103ae3b3d3564863c8890140b7c/stdlib/nodejs/file.rb#L62-L64

```diff
-      return executeIOAction(function(){return __fs__.readFileSync(path).toString()});
+      return executeIOAction(function(){return __fs__.readFileSync(path).toString()})
```

https://github.com/opal/opal/blob/fc20415fe76bc103ae3b3d3564863c8890140b7c/stdlib/nodejs/file.rb#L71-L74

```diff
-      return executeIOAction(function(){return __fs__.existsSync(path)});;
+      return executeIOAction(function(){return __fs__.existsSync(path)});
```

iliabylich:
@Mogztter That's expected. For single-line x-strings there must be either `return value;` or just `value`

Mogztter:
Alright then it's OK, thanks for the clarification @iliabylich

PR #1779, opened 2018-03-03
Add Date#to_n (in native.rb)

sfcgeorge:
I expected Date to have a `to_n` method as it has an obvious JS equivalent, so I added it on @elia's suggestion. Time already has a `to_n` so this seems reasonable.

I grepped the codebase for `def to_n` and they're all in the native.rb file so that's where I put this for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Whitespace conventions:
 
 ### Added
 
+- Added `Date#to_n` that returns the JavaScript Date object (in native.rb). (#1779)
 - Added `Array#pack` (supports only `C, S, L, Q, c, s, l, q, A, a` formats). (#1723)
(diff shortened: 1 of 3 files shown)


sfcgeorge:
Only thought is that every other `to_n` is on corelib classes, this is a stdlib class. But even if Date hasn't been required it should still be fine. I could add `if const_defined?(:Date)` around the class def if you think that would be better though, so you don't end up with a Date class that has nothing but the `to_n` method.

elia:
@sfcgeorge it's better to move `Date#to_n` directly to _stdlib/date.rb_

See for example [_stdlib/ostruct.rb_](https://github.com/opal/opal/blob/a968b95350f0a42b2a5eaabfe9055132d53cdb93/stdlib/ostruct.rb#L62-L64)

sfcgeorge:
Moved to_n directly into Date class as requested and rebased.

The spec may or may not be in the right place now. It's with all the other to_n specs. OpenStruct doesn't have a corresponding spec to copy (in the Opal suite) and Date doesn't have a (Opal) spec to add to.

PR #1781, opened 2018-03-12
correct and optimize File constructor in nodejs environment

mojavelinux:
- remove b flag silently (this already matches how Node reads files)
- detect the encoding option using match?
- use sub instead of gsub to remove encoding option
- only sub if the encoding option is detected
- rename encoding flag to encoding option (per Ruby docs)

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -160,14 +160,14 @@ class File < IO
   # Instance Methods
 
   def initialize(path, flags = 'r')
-    binary_flag_regexp = /b/
-    encoding_flag_regexp = /:(.*)/
-    # binary flag is unsupported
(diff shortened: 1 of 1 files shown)


mojavelinux:
There's absolutely no reason to warn about the 'b' flag. All that flag does in the Ruby side is suppresses EOL <-> CRLF conversion on Windows, which is the default in Node anyway.

I also optimized how the encoding option is stripped.

mojavelinux:
Technically, we could support the encoding option. We'd just have to store the setting, then apply it on read.

mojavelinux:
> Technically, we could support the encoding option.

Scratch that. Node's idea of encoding is very different than Ruby's. First of all, Node automatically handles UTF-8 and UTF-16 variants...so you don't even have to mess with them. And Node is missing most of the encodings from which you might want to convert. So in the end, it's kind of pointless and encourages the use of UTF.

iliabylich:
Thank you!

PR #1782, opened 2018-03-12
Upgrade parser, modernize parser/builder to emit lambda as a :lambda node.

iliabylich:
This PR:
1. Bumps `parser` dependency.
2. Partially enables "modernized" mode, so `->{}` is now `s(:lambda)`
3. Updates a rubocop (because of the new `parser` version that is incompatible with the previous version)
4. Adds a `String#b` (used by `parser`)
5. Disables `--fail-fast` rubocop option as it's currently broken.
6. Includes minor refactoring caused by new cops from the latest version of rubocop


opal.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'sourcemap', '~> 0.1.0'
   spec.add_dependency 'hike', '~> 1.2'
   spec.add_dependency 'ast', '>= 2.3.0'
-  spec.add_dependency 'parser', '= 2.4.0.2'
+  spec.add_dependency 'parser', '= 2.5.0.4'
```
elia: rubocop should be added as development dependency with the exact version we're using, also the version in codeclimate should be updated https://github.com/iliabylich/opal/blob/c97062039d3b54cc1385e4e70d658591e3e5dd26/.codeclimate.yml#L6

lib/opal/parser.rb
```diff
@@ -3,6 +3,7 @@
 require 'opal/ast/builder'
 require 'opal/rewriter'
 
+# rubocop:disable Naming/UncommunicativeMethodParamName
```
elia: Is there a way to avoid these directives? they're so ugly 🙀

iliabylich: Probably we could extract it to a separate file and disable it for this cop using `.rubocop.yml`

iliabylich:
@elia Latest rubocop requires ruby >= 2.1.0 - https://github.com/bbatsov/rubocop/blob/master/rubocop.gemspec#L12. Should we drop it from the travis config?

elia:
> Should we drop it from the travis config?

it's fine, we're still supporting more versions than the MRI team (which as of now I think is 2.2+)

iliabylich:
I've already pushed corresponding changes

elia:
yeah, saw that, merging

PR #1786, opened 2018-03-19
Try other names for chrome executable on linux

graywolf:
Not all distributions use `google-chrome-stable`, Archlinux for example packages `chromium`, which is basically chrom without branding. So let's try other names too so it works out-of-the-box on more systems.


iliabylich:
Could you run `rubocop` please? There are two minor warnings for your code, `-a` option should fix it. We have a job on CI for this, but Travis ignores us for some reason

graywolf:
Done

iliabylich:
Thanks!

PR #1787, opened 2018-03-23
implement StringIO#each_line

mojavelinux:


mojavelinux:
This is necessary in order for each_line to be used on the result of open when the path is a URI. Otherwise, it tries to read the URI from the filesystem (due to the IO#each_line implementation).

Please advise where to add a test for this method.

mojavelinux:
It appears that there are no tests for StringIO#each_line in https://github.com/ruby/spec.git. Does that mean we need to add a test upstream?

iliabylich:
There's a `library` directory in RubySpec that has tests for stdlib classes, so your specs are here: https://github.com/ruby/spec/blob/master/library/stringio/each_line_spec.rb

We don't run any StringIO specs from RubySpec yet, so you can try enabling them by adding `ruby/library/stringio/each_line_spec` line to the `spec/ruby_specs` file.

You can run it later using `env RUBYSPECS=true PATTERN=spec/ruby/library/stringio/each_line_spec.rb bundle exec rake mspec_ruby_nodejs` command (all example fail for me locally, but at least I can start all of them, so all startup dependencies of this file are supported by Opal).

Hope it helps.

mojavelinux:
Actually, it does look like there are tests. They are linked to StringIO#each, which I guess is an alias for StringIO#each_line.

How do we get these tests to run?

mojavelinux:
I aliased these two methods together.

mojavelinux:
> We don't run any StringIO specs from RubySpec yet, so you can try enabling them by adding ruby/library/stringio/each_line_spec line to the spec/ruby_specs file.

That's what I was looking for.

mojavelinux:
I'll follow-up a bit later.

mojavelinux:
I enabled the specs for StringIO#each and StringIO#each_line (which are the same) and adjusted the implementation to comply. The implementation still has some gray areas, but it passes all the specs.

PR #1789, opened 2018-03-26
check for chromium-browser binary in chrome cli runner

mojavelinux:
On Fedora, the name of the Chromium binary is chromium-browser.


iliabylich:
Thanks!

Also just in case, there's a `ENV['GOOGLE_CHROME_BINARY']` variable that can be set to override all this logic. But I'm still sure that we should handle default cases without any extra configuration.

The failure on appveyor is not related to your changes, the version of the `parser` that we are using was yanked, we need to bump it to `2.5.0.5`

Issue #1788, opened 2018-03-24
spec/opal/stdlib/native/date_spec.rb fails under certain conditions/TZ


elia:
```
⤑ date                                                                                                                     ~/C/opal
Sat Mar 24 10:39:31 CET 2018
⤑ be rake mspec_opal_nodejs PATTERN=spec/opal/stdlib/native/date_spec.rb                                                   ~/C/opal
mkdir -p tmp
Adding   1 files (PATTERN=spec/opal/stdlib/native/date_spec.rb)
export RANDOM_SEED=58477 # to re-use the same randomization
ruby -w -rbundler/setup -r/Users/elia/Code/opal/tasks/testing/mspec_special_calls exe/opal -Ispec/mspec/lib -Ispec -Ilib -smspec/helpers/tmp -smspec/helpers/environment -smspec/guards/block_device -sa_file -slib/spec_helper -smspec/commands/mspec-run -setc -Rnodejs -Dwarning -A --enable-source-location tmp/mspec_nodejs.rb
/Users/elia/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/parser-2.5.0.4/lib/parser/lexer.rb:10839: warning: assigned but unused variable - testEof
Object freezing is not supported by Opal
opal 0.11.1.dev (2018-03-06 revision 0)
F
Failures:

  1. Date #to_n returns native JS date object

    Expected "1984-01-23T23:00:00.000Z" to equal "1984-01-24T00:00:00.000Z"
    SpecExpectationNotMetError: Expected "1984-01-23T23:00:00.000Z" to equal "1984-01-24T00:00:00.000Z"
    at Opal.send (/private/var/folders/w0/yjfqr9n94ld7ft4j3hlz_fsm0000gn/T/opal-nodejs-runner-20180324-19501-1gcjglb:1610:19)
    at #exception [as $exception] (/private/var/folders/w0/yjfqr9n94ld7ft4j3hlz_fsm0000gn/T/opal-nodejs-runner-20180324-19501-1gcjglb:5478:14)
    at Kernel#raise [as $raise] (/private/var/folders/w0/yjfqr9n94ld7ft4j3hlz_fsm0000gn/T/opal-nodejs-runner-20180324-19501-1gcjglb:5158:31)
[...]

PR #1792, opened 2018-03-26
Fix Date#to_n spec.

iliabylich:
For more information - https://github.com/opal/opal/issues/1788#issuecomment-376011523
Closes https://github.com/opal/opal/issues/1788

`Date.prototype.toISOString` relies on the timezone, so sometimes the spec was failing.
`Date.prototype.toDateString` doesn't care about TZ, that's exactly what we need.


elia:
@iliabylich thanks! 

_Feel free to push this kind of stuff and minor fixes directly to master. To check if the specs pass before pushing I use the `elia/ci-check` branch: `git push -f origin HEAD:elia/ci-check` which is built upon push._

PR #1795, opened 2018-03-28
optimize Regexp#match and Regexp#match? when pos is undefined

mojavelinux:
- use original regexp when pos is not defined

### DIFF
--- a/opal/corelib/regexp.rb
+++ b/opal/corelib/regexp.rb
@@ -161,10 +161,17 @@ class Regexp < `RegExp`
       }
 
       if (pos === undefined) {
-        pos = 0;
-      } else {
-        pos = #{Opal.coerce_to(pos, Integer, :to_int)};
+        if (string === nil) return #{$~ = nil};
+        var m = self.exec(#{Opal.coerce_to(string, String, :to_str)});
+        if (m) {
+          #{$~ = MatchData.new(`self`, `m`)};
+          return block === nil ? #{$~} : #{yield $~};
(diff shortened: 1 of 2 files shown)


mojavelinux:
When pos is not defined, there's no reason to build the regexp again as we know we're looking for the first match from the start of the string.

iliabylich:
Looks good! Could you add some tests for this functionality somewhere in `spec/opal/core/regexp/match_spec.rb`? From what I understand now `Regexp#match` calls `exec` only once when `pos` is not specified. Maybe the test could be something like:
``` ruby
describe 'Regexp#match' do
  describe 'when pos is not specified' do
    it 'calls .exec only once on the current object' do
      regexp = /test/
      count = 0
      `regexp.exec = function(s) { count++ }`
      regexp.match('test test')
      count.should == 1
    end
  end
end
```
And you can run it with `env PATTERN=spec/opal/core/regexp/match_spec.rb bundle exec rake mspec_opal_nodejs`

mojavelinux:
Yep, I can add a test.

mojavelinux:
Tests added. I think you'll like them.

While writing the tests, I realized that pos is not even supported by Opal's Regexp#match?. That will need to be a follow-up issue. There's a stub assertion in there once it is implemented.

iliabylich:
Thank you for this high quality PR!

mojavelinux:
Thank you for being welcoming and making it easy to contribute!

PR #1796, opened 2018-03-28
don't substitute . in regexp source

mojavelinux:
It's dangerous to blindly replace `.` (dot) with `[\s\S]` in the regexp source. For example, if a dot appears inside a character group (e.g., `[.]`), Opal will cause the regular expression to break.

The onus should be on the library author to use `[\s\S]` in place of `.` when writing multi-line regular expressions as this syntax is already portable between Ruby and JavaScript. Opal should not try to do surgery on the regular expression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ Whitespace conventions:
 
 ### Changed
 
+- **BREAKING** The dot (`.`) character is no longer replaced with [\s\S] in a multiline regexp passed to Regexp#match and Regexp#match?
+  * You're advised to always use [\s\S] instead of . in a multiline regexp, which is portable between Ruby and JavaScript
 - The internal API for CLI runners has changed, now it's just a callable object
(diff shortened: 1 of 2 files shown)


mojavelinux:
cc: @Mogztter

iliabylich:
That's a breaking change:
``` ruby
$ cat test.rb
p /a.b/m.match("a\nb")

$ ruby test.rb
#<MatchData "a\nb">

(master)
$ opal test.rb
#<MatchData "a\nb">

(patch)
$ opal test.rb
nil
```

I'm fine with this change as I don't like implicit behavior. But if we are going to merge it definitely needs a changelog entry. @elia What do you think?

mojavelinux:
I agree it's a breaking change, but I'd say not breaking the original regular expression is more important than breaking a hack that should never have been added in the first place.

I'll gladly add a CHANGELOG entry to the PR.

mojavelinux:
CHANGELOG entry added.

mojavelinux:
:tada:

Issue #1797, opened 2018-03-31
Method argument is wrong when using the splat operator


Mogztter:
Consider the following code:

```rb
def method(type, one, two, three, four, five)
  p "type: #{type}, one: #{one}, two: #{two}, three: #{three}, four: #{four}, five: #{five}"
end

method(:while, ['a', 'b'], *['c', ['d', 'e']], 'f', ['g', 'h'])
```

**MRI**
```
ruby test.rb
"type: while, one: [\"a\", \"b\"], two: c, three: [\"d\", \"e\"], four: f, five: [\"g\", \"h\"]"
```

**Opal**
```
bundle exec opal test.rb
"type: while, one: a,b, two: c, three: d,e, four: f, five: g"
```

The fifth argument should be `["g", "h"]`.

iliabylich:
Minimized case:
``` ruby
def m(*args)
  p args
end

m(*[], [1])
```

``` sh
$ ruby test.rb && opal test.rb
[1]
1
```

iliabylich:
ok, any post-rest argument that is an array is affected by this issue.
the reason is that arglist `(*one, two)` compiles to `Opal.to_ary(one).concat(two)` which is ok in most cases. But in JS `[1].concat([2])` is `[1, 2]`, so it works incorrectly when `two` is an array.

Mogztter:
Thanks for the fix @iliabylich 
And sorry for the "over-complicated" code to describe the issue 😳

PR #1798, opened 2018-04-02
Wrap array in the arglist into one more array.

iliabylich:
Closes https://github.com/opal/opal/issues/1797

### DIFF
--- a/lib/opal/nodes/arglist.rb
+++ b/lib/opal/nodes/arglist.rb
@@ -45,7 +45,7 @@ module Opal
           if code.empty?
             code = join
           else
-            code << fragment('.concat(') << join << fragment(')')
+            code << fragment('.concat([') << join << fragment('])')
           end
         end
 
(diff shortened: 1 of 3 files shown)


lib/opal/nodes/arglist.rb
```diff
@@ -45,7 +45,7 @@ def compile
           if code.empty?
             code = join
           else
-            code << fragment('.concat(') << join << fragment(')')
+            code << fragment('.concat([') << join << fragment('])')
```
Mogztter:
```js
["a"].concat("b") // ["a", "b"]
["a"].concat(["b"]) // ["a", "b"]
["a"].concat([["b"]]) // ["a", ["b"]]
```
JavaScript 😱

`push` is more predictable but the return value is the new array length (and also mutate the array).

Mogztter:
@iliabylich I can confirm that this fix solve our compilation issue in Asciidoctor.js

Mogztter:
Thank you! Going to update to the latest version :tada: :tada: :tada:

PR #1800, opened 2018-04-09
Move IO methods that rely on path to File

Mogztter:
`read` and `each_line` rely on the `@path` attribute (which is only define on a `File` class). When working with a `StringIO` in a Node.js environment, if Opal try to use `read` or `each_file` an exception will be thrown because `@path` is undefined (`Opal.nil`).

In this pull request, I've moved these two methods on the `File` class.

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -175,6 +175,47 @@ class File < IO
 
   attr_reader :path
 
+  def read
+    if @eof
+      ''
+    else
+      res = `executeIOAction(function(){return __fs__.readFileSync(#{@path}).toString()})`
+      @eof = true
+      @lineno = res.size
+      res
+    end
+  end
+
+  def each_line(separator = $/, &block)
+    if @eof
+      return block_given? ? self : [].to_enum
+    end
+
+    if block_given?
+      lines = File.read(@path)
+      %x{
+        self.eof = false;
+        self.lineno = 0;
+        var chomped  = #{lines.chomp},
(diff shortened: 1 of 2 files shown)


Mogztter:
> I'm not sure why this method was initially in IO class (since it's actually a File#read method), is it a bug?

Yes I think so. We did not catch this bug because at the time we were not using `StringIO.read` on a Node.js environment.

> If so, what happens with existing implementation (StringIO#read is there and it doesn't use any paths, but StringIO#each_line is missing). I'm totally confused smile

Dan added `StringIO#each_line` in https://github.com/opal/opal/pull/1787
So we should be fine since we don't need a dedicated Node.js implementation for `StringIO` :relieved: 

> But before this PR StringIO#each_line was inherited from IO#each_line (and broken, as it uses @path), your PR simply removes it, right?

Actually there's two issues:

1. The first one (resolved by this PR) is that `IO#each_line` and `IO#read` should not use `@path` and looking at the implementation these methods should definitely be defined in `File` (not in `IO`)
2. The second one (might be related to our environment) is that we are calling `StringIO.read` on a Node.js environment where `nodejs/io` is loaded but Opal is calling `IO.read` (Node.js flavor) and thus fails (because `IO.read` uses `path`).

This PR fixes the first issue.
For the second issue I would expect Opal to call `StringIO.read` (defined in `stdlib`) but it could be the expected behavior or it could be a wrong usage in Asciidoctor.js :thinking: 

Does it make sense ? :wink:

PR #1802, opened 2018-04-12
resolves #1801 add logger impl to stdlib

### LINKED ISSUE #1801: Add basic support for Logger
In order to log messages, many libraries (including Asciidoctor) depend on at least having the basic interface / base class provided by the Logger class. I think that Opal should provide enough of the Logger so that it can be used for writing messages to stdout/stderr and extended. I don't think we need to get into file support right now (esp since that's not portable in Opal).

Here's the basic API that I have in mind.

```rb
class Logger
  module Severity
    DEBUG = 0
    INFO = 1
    WARN = 2
    ERROR = 3
    FATAL = 4
    UNKNOWN = 5
  end
  include Severity

  attr_reader :level
  attr_accessor :progname
  attr_accessor :formatter

  def initialize pipe
    @pipe = pipe
    @level = DEBUG
  end

  def info(progname = nil, &block); ...; end
  def debug(progname = nil, &block); ...; end
  def warn(progname = nil, &block); ...; end
  def error(progname = nil, &block); ...; end
  def fatal(progname = nil, &block); ...; end
  def unknown(progname = nil, &block); ...; end
  def info?; @level <= INFO; end
  def debug?; @level <= DEBUG; end
  def warn?; @level <= WARN; end
  def error?; @level <= ERROR; end
  def fatal?; @level <= FATAL; end
  def level=(level); ... end
  def add(severity, message = nil, progname = nil, &block)
    severity ||= UNKNOWN
    return true if severity < @level
    progname ||= @progname
    unless message
      if block_given?
        message = yield
      else
        message = progname
        progname = @progname
      end
    end
    @pipe.write(format_message(severity, Time.now, progname, message))
  end
  def format_message severity, datetime, progname, message; ...; end
end
```

The Asciidoctor project can help with getting this implemented. cc: @Mogztter 

(I'll fill in the rest of the implementation in a PR).

mojavelinux:
- add StringIO-based Logger implementation w/ default formatter
- enable ruby/library/logger/severity_spec

### DIFF
--- a/spec/opal/stdlib/logger/logger_spec.rb
+++ b/spec/opal/stdlib/logger/logger_spec.rb
@@ -1,0 +1,308 @@
+require 'logger'
+require 'stringio'
+
+# Our implementation of Logger only supports StringIO pipes.
+# Since we can't portably write a log file, nothing in rubyspec will run.
+describe Logger do
+  before do
+    @pipe = StringIO.new
+    @logger = Logger.new @pipe
+  end
+
+  describe "level" do
+    it "should set the level to DEBUG by default" do
+      @logger.level.should == Logger::Severity::DEBUG
+    end
+
+    it "should allow the level to be changed using an integer of a known level" do
+      @logger.level = Logger::Severity::INFO
+      @logger.level.should == Logger::Severity::INFO
+    end
+
+    it "should allow the level to be changed to an arbitrary integer value" do
+      @logger.level = 1000
+      @logger.level.should == 1000
+    end
+
+    it "should allow the level to be set using a case-insensitive level name" do
+      @logger.level = 'WaRn'
+      @logger.level.should == Logger::Severity::WARN
+    end
+
+    it "should raise an ArgumentError if the level is set to an unknown name" do
+      lambda { @logger.level = 'foobar' }.should raise_error(ArgumentError)
+    end
+
+    it "should report when debug level is enabled" do
+      @logger.debug?.should == true
+    end
+
+    it "should report when debug level is not enabled" do
+      @logger.level = Logger::Severity::DEBUG + 1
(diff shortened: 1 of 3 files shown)


mojavelinux:
@elia @iliabylich I'm unable to activate any of the other logger tests from ruby spec as they all rely on writing a log file. Is it possible to enable file IO in this part of the test suite? If so, then I'll be able to update the implementation to add basic support for writing to a log file. If that's not possible, should we add tests somewhere else? Please advise.

mojavelinux:
cc: @Mogztter

iliabylich:
In general looks good! Could you run `rubocop -a stdlib/logger.rb` please? There are some warnings.

On tests: rubyspec has some tests, but it uses tmp files for testing (it can be supported by node suite, but not on chrome). Could you write some tests using `StringIO` and put them under `spec/opa/stdlib`?

By they way, I'm not sure if it's necessary to test `Formatter` and `Logger` together. By testing them separately you could use some dummy formatter in `Logger` tests, I guess that would simplify things a lot.

mojavelinux:
> Could you run rubocop -a stdlib/logger.rb please?

Absolutely. I'll be sure to do that once the tests are all written.

> Could you write some tests using StringIO and put them under spec/opal/stdlib?

Yep, I can do that.

> I'm not sure if it's necessary to test Formatter and Logger together.

I'll actually need to test both since the formatter can be changed.

(I didn't implement the custom datetime for the default formatter because I really don't see how that is useful when you can simply change the formatter. It seems like that feature was overdesigned in stdlib).

mojavelinux:
Done!

50 tests added!

iliabylich:
Sorry for delay

mojavelinux:
Thanks @iliabylich!

PR #1803, opened 2018-04-17
Resolve #1793, cache global Regexp

### LINKED ISSUE #1793: Cache dynamic regular expressions
Certain methods such as String#gsub and String#match dynamically generate a JavaScript regular expression from a Ruby regular expression in order to prepend additional flags (specifically the 'g' flag). For programs that hit these methods thousands of times using complex regular expressions, this can have a measurable impact on performance.

I'd like to suggest that Opal cache these generated regular expressions so they only have to be compiled once.

Here's an example that demonstrates how a regexp cache could be used in the String#gsub method:

```rb
const GSUB_RX_CACHE = new Map();
const GSUB_RX_CACHE_I = new Map();

...

if (pattern.$$is_regexp) { 
  const ignoreCase = pattern.ignoreCase;
  const rxCache = ignoreCase ? GSUB_RX_CACHE_I : GSUB_RX_CACHE;
  const patternSource = pattern.source;
  pattern = rxCache.get(patternSource);
  if (pattern) {
    pattern.lastIndex = 0;
  } else {
    pattern = new RegExp(patternSource, ignoreCase ? 'gmi' : 'gm');
    rxCache.set(patternSource, pattern);
  }
} else {
  ...
}
```

(I use two different caches because it's faster than generating a dynamic key).

Here are the methods that could benefit from this cache:

* String#gsub
* String#scan
* String#split
* String#match
* String#match? *

\* For String#match?, it may not even be necessary to recreate the regular expression since the 'g' flag isn't actually need. And I'm not sure the 'g' flag is need for String#match either (or perhaps there are just cases when it isn't needed).

Mogztter:


iliabylich:
Could you also check if it's possible to optimize `Regexp#match` and `Regexp#match?`?

Mogztter:
> Could you also check if it's possible to optimize Regexp#match and Regexp#match??

Ok!
Some tests are failing... investigating :eyeglasses:

iliabylich:
Shouldn't there also be something like `pattern.$$g.lastIndex = null` before returning it?

Mogztter:
Another benefit is that there's less code duplication :wink: 

> Shouldn't there also be something like pattern.$$g.lastIndex = null before returning it?

I think you're right, I still have one failing test though:

```
fails "String#index with Regexp returns the index of the first match of regexp" # Expected 9 to equal 1
```

Mogztter:
Weirdly it's working on Travis... I'm using `bundle exec rake` on my local copy.
Could someone try to run the tests suite to confirm if it's just me or if there's really an issue ?

iliabylich:
> Weirdly it's working on Travis

It doesn't. Exit code is 0, but there's still a failing spec - https://travis-ci.org/opal/opal/jobs/367786469. That's a bug in the `Kernel#exit` method.

``` ruby
  def exit(status = true)
    $__at_exit__ ||= []

    until $__at_exit__.empty?
      block = $__at_exit__.pop
      block.call
    end

    %x{
      if (status == null) {
        status = 0
      } else if (status.$$is_boolean) {
        status = status ? 0 : 1;
      } else if (status.$$is_numeric) {
        status = status.$to_i();
      } else {
        status = 0
      }

      Opal.exit(status);
    }
    nil
  end
```

Could you change `$$is_numeric` to `$$is_number`? It should fix the issue.

iliabylich:
Basically on master `opal -e 'exit(1)'` actually exits with 0 code. With a small fixed mentioned above it should exit with 1.

Mogztter:
Nice catch, I will fix the `exit` method.

Mogztter:
Given the error message and looking at the code I think this spec shouldn't be run ?

https://github.com/ruby/spec/blob/4e665c86ad56122a0e47445446ca057297756f92/core/string/index_spec.rb#L196

There's a special condition on the Ruby spec file: `not_supported_on :opal do`

iliabylich:
The spec should, but only partially. `not_supported_on` has a special behavior when used by the RubySpec - https://github.com/opal/opal/blob/master/tasks/testing/mspec_special_calls.rb#L20-L24.
You can check it yourself:
```
$ ruby -rbundler/setup -ropal -r ./tasks/testing/mspec_special_calls.rb bin/opal -c --no-opal --no-exit -e '1; not_supported_on :opal do; 2; end; 3'
/* Generated by Opal 0.11.1.dev */
(function(Opal) {
  var self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.const_get_qualified, $$ = Opal.const_get_relative, $breaker = Opal.breaker, $slice = Opal.slice;


  1;
  ;
  return 3;
})(Opal);
```

iliabylich:
@Mogztter https://github.com/ruby/spec/blob/4e665c86ad56122a0e47445446ca057297756f92/core/string/index_spec.rb#L206 <- this expectation fails. I'll take a look

iliabylich:
Ok, the problem here is that in somes cases regexes that are multiline if original regex is multiline (that's what `Opal.global_regexp` does), but in some cases generated regexes are **always** multiline.

For the failing test case (I've added `console.log(regex)` to the `String#index` locally:
```
# master
$ opal -e 'p "b\nablabla".index(/$/)'
/$/gm
1

# your branch
$ opal -e 'p "b\nablabla".index(/$/)'
/$/g
9
```

I guess we need another method like `Opal.global_multiline_regexp`. An extra parameter for `Opal.global_regexp` that forces multiline flag is also an option, but tbh I don't like such polymorphism of interfaces. And we definitely need to set such regex on another prop (maybe on the same if the source is also multiline). @elia what do you think?

Mogztter:
> but in some cases generated regexes are always multiline.

On master, the following methods are always multiline:

* split
* scan
* rpartition
* rindex
* index
* gsub

Is that correct ?

> I guess we need another method like Opal.global_multiline_regexp

I agree.

> And we definitely need to set such regex on another prop

`$$gm` ?

> (maybe on the same if the source is also multiline). 

Why not, it will avoid to create an extra RegExp if the regexp is already multiline:


````js
Opal.global_multiline_regexp= function(pattern) {
  if (pattern.multiline) {
    if (pattern.$$g == null) {
      pattern.$$g = new RegExp(pattern.source, 'gm' + (pattern.ignoreCase ? 'i' : ''));
    } else {
      pattern.$$g.lastIndex = null; // reset lastIndex property
    }
    return pattern.$$g;
  }
  if (pattern.$$gm == null) {
    pattern.$$gm = new RegExp(pattern.source, 'gm' + (pattern.ignoreCase ? 'i' : ''));
  } else {
    pattern.$$gm.lastIndex = null; // reset lastIndex property
  }
  return pattern.$$gm;
}
`````

elia:
@Mogztter can you add a check to see if the `pattern` already has the `g` flag? I'm not sure if an Opal generated regexp can have `/g` but it's certainly possible for a regexp to be coming from JS

Mogztter:
Sure, I'm on it! :grinning:

Mogztter:
> Could you change $$is_numeric to $$is_number? It should fix the issue.

@iliabylich Done https://github.com/opal/opal/pull/1808

> can you add a check to see if the pattern already has the g flag?

@elia Done

iliabylich:
@elia any comments from your side?

mojavelinux:
:+1:

PR #1804, opened 2018-04-17
better errors, better backtraces, filenames with extensions

elia:
### Backtrace

    echo "def foo\n endz" > tmp/foo.rb

#### before:

```
⤑ bin/opal tmp/foo.rb                                                                                                      ~/C/opal
tmp/foo:1:8: error: bare backslash only allowed before newline
tmp/foo:1: def foo\n endz
tmp/foo:1:        ^      
…
```

#### after:

```
opal:elia/errors-and-filenames-review ⤑ bin/opal tmp/foo.rb                                                                            ~/C/opal
tmp/foo.rb:1:8: error: bare backslash only allowed before newline
tmp/foo.rb:1: def foo\n endz
tmp/foo.rb:1:        ^      
```

### Module names

are now discarding everything after the first dot in the basename: `foo/bar.js.opalerb => Opal.modules["foo/bar"]`

### Source maps

Untested, but now if possible the original node location is used instead of DUMMY_LOCATION in rewriters, this is a potential fix for sourcemaps that erroneously pointed at line 1 col 0


lib/opal/rewriters/base.rb
```diff
@@ -94,6 +95,22 @@ def append_to_body(body, node)
...
       end
+
+      # Store the current node for reporting.
+      attr_accessor :current_node
+
+      # Intercept the main call and assign current node.
+      def process(node)
+        self.current_node = node
+        super
+      end
```
iliabylich: Probably it makes sense to add `ensure; self.current_node = nil` to prevent confusing error locations when someone calls `on_*` manually.

lib/opal/errors.rb
```diff
@@ -12,4 +16,17 @@ def initialize(gem_name)
       super("can't find gem #{gem_name}")
     end
   end
+
+  class CompilationError < Error
+    attr_accessor :location
```
iliabylich: I think we could add a constructor for this attribute and use it with a `raise CompilationError, msg, location` or `raise CompilationError, msg, location: location`.

lib/opal/compiler.rb
```diff
@@ -184,8 +185,16 @@ def parse
 
       begin
         sexp, comments, tokens = @parser.tokenize(@buffer)
+      rescue ::Opal::Error => error
+        backtrace = error.backtrace
+        if error.respond_to? :location
+          line = error.location.line
+          backtrace << "#{file}:#{line}:in #{error.location.expression.source_line}"
```
iliabylich: It looks procedural. Maybe we could override a `backtrace` method on our error classes to include this custom line by default? Or it can be another method that does `backtrace + custom_line`

iliabylich: I guess the first option doesn't work well - https://bugs.ruby-lang.org/issues/14693 😄

iliabylich:
Some jobs on CI are failing, but the code LGTM

elia:
@iliabylich specs are now green, the only one whining is codeclimate, I'd leave the method as is, but open to suggestions on how to improve it

iliabylich:
Maybe an abstraction of `ErrorLocation (error -> string)` and `BacktraceLIne (error -> backtrace)`? We can use as many objects as we want because that's the last place before throwing an error. It doesn't have to be optimized.

lib/opal/errors.rb
elia: > The optional second parameter sets the message associated with the exception, and the third parameter is an array of callback information. 

Can't be done AFAIK

iliabylich:
Also I'm not sure about renaming `Opal::Parser::WithRubyLexer` to `Opal::Parser::RubyLexer`. It's not a lexer, it's a parser that uses some specific lexer. Same for the `CLexer`

lib/opal/errors.rb
iliabylich: Let's keep it as is then

elia:
Reverted the removal of the `With` prefix

PR #1806, opened 2018-04-18
Add C lexer from opal/c_lexer.

iliabylich:


iliabylich:
ok. Travis seems to be working with this patched config. This PR includes a c_lexer gem from https://github.com/opal/c_lexer. The gem itself is not released yet, that's why I've temporary changed `.travis.yml` to build it on-the-fly (I'll revert it before merging).

@opal/core any comments? I'm going to review c_lexer a few more times before releasing it, so I need 1-2 days. If everything ok with this PR then the plan is:
1. release `c_lexer` (I think we should stick to the version of the `parser` to avoid confusions)
2. revert changes in the `.travis.yml`
3. add `c_lexer` to the gemfile (or maybe gemspec)
4. make sure that it works on appveyor
5. merge it!

iliabylich:
I've added an extra layer for selecting the best parser (from the performance perspective). `c_lexer` doesn't change any constant from the `parser` gem. Instead, it introduces another class (`Parser::Ruby25WithCLexer`) that uses `CLexer` by default. I'd like to get a review if possible. If everything is ok I think I'm ready to release `c_lexer`.

iliabylich:
All green! Only JRuby builds and opal-parser use pure ruby lexer.

elia:
@iliabylich the :cop: got me, missing `# frozen_string_literal: true`, although I'm not sure why codeclimate is happy …maybe a previews review 🤷‍♂️?

PR #1809, opened 2018-04-22
Fix guides, docs and tutorials

eiko:
Various code samples from guides, docs, tutorials, etc have fallen out of date and no longer work. I've tweaked the samples to make them work. 

The goal was to get the code to a point where new users can get samples to work by copy / pasting, but I am very open to reviewing and refactoring this code to make it more proper.

Changes:
- [x] Make the dist task build an "auto-loadable" version of Opal, so CDN links work properly again. This is how Opal Sprockets used to work before we switched to Opal Builder.
- [x] Update JQuery guide to load dependencies properly
- [x] Remove -j use from Rails guides (Rails dropped support for that option)
- [x] Fix index example in "Using Sprockets" (I feel like there's an easier way)
- [x] Update Sinatra doc to use Opal-Sprockets (This was messy because Opal-Sprockets doesn't make source maps readily available)

### DIFF
--- a/docs/jquery.md
+++ b/docs/jquery.md
@@ -30,26 +30,53 @@ gem 'opal-jquery'
 
 #### Usage
 
-`opal-jquery` can now be easily added to your opal application sources using a
-standard require:
+`opal-jquery` can now be easily added to your opal application sources:
 
 ```ruby
 # app/application.rb
-require 'opal'
-require 'jquery'
-require 'opal-jquery'
+require 'opal/jquery'
 
 alert "Hello from jquery + opal"
 ```
 
-NOTE: this file requires two important dependencies, `jquery` and `opal-jquery`.
-You need to bring your own `jquery.js` file as the gem does not include one. If
-you are using the asset pipeline with rails, then this should be available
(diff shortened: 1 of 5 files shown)


docs/jquery.md
```diff
@@ -30,27 +30,54 @@ gem 'opal-jquery'
...
-`opal-jquery` can now be easily added to your opal application sources using a
-standard require:
+`opal-jquery` can now be easily added to your opal application sources:
 
 ```ruby
 # app/application.rb
-require 'opal'
-require 'jquery'
-require 'opal-jquery'
+require 'opal/jquery'
```
elia: ~~requiring opal and jquery separately is still needed, any reason for this change and the switch to `/`?~~

I see, it's stated below, can you add a comment or indication on the lines of "see below for a full example"?

eiko: Comment added! When I first made this pull request, using `require 'opal/jquery'` was the only way I could make this example work. Now that I understand Opal better, I see that `require 'opal-jquery'` can work if the user compiles their app with the added flag: `--gem=opal-jquery`.

I personally prefer `require 'opal/jquery'` because it always works, no matter how you build your dependencies.

PR #1818, opened 2018-05-06
Add a test on module name with front slash

Mogztter:
This test should fail on Appveyor (Windows)

### DIFF
--- a/spec/lib/builder_spec.rb
+++ b/spec/lib/builder_spec.rb
@@ -124,4 +124,10 @@ RSpec.describe Opal::Builder do
       expect(builder.build('stringio', requirable: true).to_s).to include(%{Opal.modules["stringio"]})
     end
   end
+
+  describe ':requirable' do
+    it 'it uses front slash as module name' do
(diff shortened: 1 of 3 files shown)


Mogztter:
I can reproduce this issue when using the builder in JavaScript: https://ci.appveyor.com/project/Mogztter/opal-node-compiler/build/3

Code:
https://github.com/Mogztter/opal-node-compiler/blob/fc1e2a64aeed71288ff012d7d0288658764fc697/spec/main.spec.js#L29-L35

I will investigate :wink:

iliabylich:
Yes, I think you are right - https://github.com/ruby/ruby/blob/trunk/file.c#L4641. Also it uses a constant `"/"` for joining instead of a `File::SEPARATOR` (i.e. overriding a constant doesn't change the return value of `File.join('a', 'b')`)

Mogztter:
@iliabylich Thanks for the confirmation

I've changed the Node.js implementation to use `path.posix.join` instead of `path.join`. It should fix the new test.

Issue #1684, opened 2017-07-10
Don't touch Object.prototype


jods4:
Opal does modify some built-in objects.
For example it adds `$require` on `Object.prototype`.

That's a terrible idea that breaks all kind of things.
For example, it's impossible to load `nodegit` after Asciidoctor, which uses Opal.

>Sidenote: because `nodegit` uses `promisify-node`, which iterates exports to promisify everything, which finds `$require`, which then goes down the rabbit hole and eventually crashes on some objects.

Admittedly, `promisify-node` could be better and probably ignore non-own properties. 
But modifying `Object` is still a terrible idea and guaranteed to run into trouble with one library or another.

You should avoid it.
Installing your properties as non-enumerable would be a first layer of safety that hides them from `for..in`, which would at least avoid some crashes (including the one with `promisify-node`).

iliabylich:
@jods4 Do you mean something like
``` js
Object.defineProperty(Array.prototype, '$flatten', {
  value: function() {
    return []
  },
  enumerable: false
})
```

I'm sure we can use it (since it's [widely supported](http://kangax.github.io/compat-table/es5/#test-Object.defineProperty)) if there's no difference in performance.

meh:
@iliabylich the downside is it would increase code size and clarity by a lot.

iliabylich:
@meh Not really, since we have `Opal.def`/`Opal.defn`. `Module#define_method/define_singleton_method` also uses it under the hood.

iliabylich:
``` js
var Benchmark = require('benchmark');

Object.defineProperty(Object.prototype, 'throughDefineProperty', {
  value: function() {
    return null;
  },
  enumerable: false
});

Object.prototype.directlyOnPrototype = function() {
  return null;
};

var object = {};

var suite = new Benchmark.Suite;

suite.add('throughDefineProperty', function() {
  object.throughDefineProperty();
}).add('directlyOnPrototype', function() {
  object.directlyOnPrototype();
}).on('cycle', function(event) {
  console.log(String(event.target));
}).run({ 'async': true })
```

``` sh
$ node --version
v6.10.1
$ node test.js
throughDefineProperty x 108,281,919 ops/sec ±1.50% (81 runs sampled)
directlyOnPrototype x 112,527,379 ops/sec ±1.64% (84 runs sampled)
```

It seems that there's no difference, I'll try to fix it today. @elia @meh any comments?

meh:
[...]

PR #1821, opened 2018-05-08
Use define property for methods and props

iliabylich:
Closes https://github.com/opal/opal/issues/1684

This PR defines all props using `Object.defineProperty`.
It also introduces `$$methods` and `$$ancestors` props.
Both are always up-to-date in runtime (yes, `prepend` would be much easier to implement with this patch).

There's a test under `spec/opal` suite that verifies that there are no iterable props on core JS objects.
Also I guess we need a better random generator library. `seedrandom` that is currently used pollutes `Math` module.


iliabylich:
Of course it **must** be squashed before merging 😄 There's too much garbage in intermediate commits.

iliabylich:
```
$ repeat 5 time node master.js
node master.js  0.22s user 0.03s system 98% cpu 0.261 total
node master.js  0.23s user 0.03s system 98% cpu 0.265 total
node master.js  0.23s user 0.03s system 98% cpu 0.262 total
node master.js  0.23s user 0.03s system 98% cpu 0.272 total
node master.js  0.22s user 0.03s system 98% cpu 0.259 total
```

```
$ repeat 5 time node patch.js
node patch.js  0.24s user 0.04s system 97% cpu 0.285 total
node patch.js  0.24s user 0.04s system 98% cpu 0.282 total
node patch.js  0.24s user 0.04s system 98% cpu 0.284 total
node patch.js  0.24s user 0.03s system 97% cpu 0.285 total
node patch.js  0.25s user 0.03s system 97% cpu 0.287 total
```

I can also benchmark other things, just let me know.

elia:
@iliabylich I think we should consider for a couple of minutes if ivars should be defined as non-enumerable, probably not, but it's better to be sure

PS. I'll deal with code climate tomorrow as it's having issues with the CDN 🌐 right now 🙂, it's all good anyway

iliabylich:
@elia Fixed. Also added docs for new props. Could you take a look please?

Mogztter:
This is awesome, thanks @iliabylich ! :tada:

PR #1822, opened 2018-05-11
Regexp names

janbiedermann:


iliabylich:
Could you check if there are any specs in the RubySpec suite that are green now? `env INVERT_RUNNING_MODE=true bundle exec rake mspec_ruby_nodejs` shows them, filters are in the `spec/filters/bugs/regexp.rb` directory.

janbiedermann:
yes, named captures tests work, except one:
```
fails "Regexp#names returns each capture name only once" # Exception: Invalid regular expression: /n(?<cap>ee)d(?<cap>le)/: Duplicate capture group name
```
This exception comes from js. Not sure what to do? Catch in the initializer? But then the regexp is useless anyway within js.

janbiedermann:
hit the wrong button

hmdne:
Superseded by #2272

PR #1824, opened 2018-05-12
Implement binread (mimic ASCII-8BIT encoding)

Mogztter:
Credits to: https://github.com/mathiasbynens/jsesc

### DIFF
--- a/stdlib/nodejs/io.rb
+++ b/stdlib/nodejs/io.rb
@@ -13,44 +13,6 @@
       throw error;
     }
   }
-
-  var regexWhitelist = /[ -~]/;
-  var regexSingleEscape = /["'\\\b\f\n\r\t]/;
-  var singleEscapes = {
-    '"': '\\"',
-    '\'': '\\\'',
-    '\\': '\\\\',
-    '\b': '\\b',
-    '\f': '\\f',
-    '\n': '\\n',
-    '\r': '\\r',
-    '\t': '\\t'
-    // `\v` is omitted intentionally, because in IE < 9, '\v' == 'v'.
-    // '\v': '\\x0B'
-  };
- 
-  function toASCII(buffer) {
-    var array = Array.from(buffer);
-    var length = array.length;
-    var index = -1;
-    var result = '';
-    while (++index < length) {
-      var character = String.fromCharCode(array[index]);
-      if (regexWhitelist.test(character)) {
-        result += character;
-        continue;
-      }
-      if (regexSingleEscape.test(character)) {
-        result += singleEscapes[character];
-        continue;
-      }
-      var hexadecimal = array[index].toString(16).toUpperCase();
-      var longhand = hexadecimal.length > 2;
-      var escaped = '\\' + (longhand ? 'u' : 'x') + ('0000' + hexadecimal).slice(longhand ? -4 : -2);
-      result += escaped;
-    }
-    return result;
-  }
 }
 
 class IO
@@ -74,7 +36,7 @@ class IO
   end
 
   def self.binread(path)
-    `return executeIOAction(function(){return toASCII(__fs__.readFileSync(#{path}))})`
+    `return executeIOAction(function(){return __fs__.readFileSync(#{path}).toString('utf-8')})`
   end
 end
 
--- a/stdlib/nodejs/io.rb
+++ b/stdlib/nodejs/io.rb
@@ -13,6 +13,44 @@
       throw error;
     }
   }
+
+  var regexWhitelist = /[ -~]/;
+  var regexSingleEscape = /["'\\\b\f\n\r\t]/;
+  var singleEscapes = {
+    '"': '\\"',
+    '\'': '\\\'',
+    '\\': '\\\\',
+    '\b': '\\b',
+    '\f': '\\f',
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t'
+    // `\v` is omitted intentionally, because in IE < 9, '\v' == 'v'.
+    // '\v': '\\x0B'
+  };
+ 
+  function toASCII(buffer) {
+    var array = Array.from(buffer);
+    var length = array.length;
+    var index = -1;
+    var result = '';
+    while (++index < length) {
+      var character = String.fromCharCode(array[index]);
+      if (regexWhitelist.test(character)) {
+        result += character;
+        continue;
+      }
+      if (regexSingleEscape.test(character)) {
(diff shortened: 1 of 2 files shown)


stdlib/nodejs/io.rb
```diff
@@ -13,6 +13,44 @@
       throw error;
     }
   }
+
+  var regexWhitelist = /[ -~]/;
```
Mogztter: Can be generated using:

```js
const regenerate = require('regenerate');
const set = regenerate()
        .addRange(0x20, 0x7E) // printable ASCII symbols
```

stdlib/nodejs/io.rb
```diff
@@ -13,6 +13,44 @@
...
+  var singleEscapes = {
+    '"': '\\"',
+    '\'': '\\\'',
+    '\\': '\\\\',
+    '\b': '\\b',
+    '\f': '\\f',
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t'
+    // `\v` is omitted intentionally, because in IE < 9, '\v' == 'v'.
```
Mogztter: Not sure about this one but I was a bit afraid to activate it :confused:

iliabylich:
And maybe converting a file content to a utf8 string is easier:
```
> fs.readFileSync("tmp/foo").toString('utf-8')
'Le français c\'est compliqué :)\\n'
> "Le fran\xC3\xA7ais c'est compliqu\xC3\xA9 :)\\n"
'Le franÃ§ais c\'est compliquÃ© :)\\n'
```

(The output of the first command looks correct)

test/nodejs/test_io.rb
```diff
@@ -15,4 +15,9 @@ def test_binread_noexistent_should_raise_io_error
       IO.binread('tmp/nonexistent')
     end
   end
+
+  def test_binread_encoding
+    File.write('tmp/foo', 'Le français c\'est compliqué :)\n')
+    assert_equal("Le fran\\xC3\\xA7ais c'est compliqu\\xC3\\xA9 :)\\n", IO.binread('tmp/foo'))
```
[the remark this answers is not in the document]
Mogztter: Indeed I'm using the string representation of `"\xC3"`... but I don't know how to get back the character `\xC3`

Mogztter:
> And maybe converting a file content to a utf8 string is easier:

That was my first thought but I wasn't sure that the expected output was identical with MRI.
I'm now using `toString('utf-8')`

test/nodejs/test_io.rb
```diff
@@ -15,4 +15,10 @@ def test_binread_noexistent_should_raise_io_error
       IO.binread('tmp/nonexistent')
     end
   end
+
+  def test_binread_encoding
+    File.write('tmp/foo', 'Le français c\'est compliqué :)\n')
+    assert_equal("Le fran\xC3\xA7ais c'est compliqu\xC3\xA9 :)\\n", IO.binread('tmp/foo'))
```
iliabylich: I'm pretty sure that this expectation fails because Ruby strings become JS strings (usually) without any changes:
```
$ node
> "Le fran\xC3\xA7ais c'est compliqu\xC3\xA9 :)\\n"
'Le franÃ§ais c\'est compliquÃ© :)\\n'
> "Le français c'est compliqué :)\\n"
'Le français c\'est compliqué :)\\n'
```

iliabylich: Maybe something like:
``` ruby
content = 'Le français c\'est compliqué :)\n'
File.write('tmp/foo', content)
written = IO.binread('tmp/foo')
assert_equal(written, content)
assert_equal(written.encoding, Encoding::ASCII_8BIT)
```

(requires adding `.force_encoding('ascii-8bit')` to the `IO.binread` method).

iliabylich: My bad, `parser` handled it for us:
```
$ cat sample.rb
assert_equal("Le fran\xC3\xA7ais c'est compliqu\xC3\xA9 :)\\n", IO.binread('tmp/foo'))
$ ruby-parse sample.rb
(send nil :assert_equal
  (str "Le français c'est compliqué :)\\n")
  (send
    (const nil :IO) :binread
    (str "tmp/foo")))
```
These two expectations are just equal.

Mogztter: Yes I added both `"Le fran\xC3\xA7ais c'est compliqu\xC3\xA9 :)\\n"` and `"Le français c'est compliqué :)\\n"`

iliabylich:
Thanks!

Mogztter:
Damn I found a bug when using `binread` (on an image) with `Base64.strict_encode64`... I will open a new pull request, sorry :disappointed:

Mogztter:
```
ruby -e "p ::IO.binread 'test/cruby/test/cgi/testdata/small.png'" 
"\x89PNG\n\x1A\n\x00\x00\x00\nIHDR\x00\x00\x00\x11\x00\x00\x00\x11\b\x02\x00\x00\x00\xB4\x0F\xD0\xAD\x00\x00\x00\x1AIDAT(\x91c\xFC\xFF\xFF?\x03\x89\x80\x89T\n\xA3zF\xF5\x8C\xEA\x81\x03\x00\x8Ah\x03\x1F\xEE,\x99\x00\x00\x00\x00\x00IEND\xAEB`\x82"
```
```
node -e "console.log(require('fs').readFileSync('test/cruby/test/cgi/testdata/small.png').toString('utf-8'))"
�PNG


IHDR�ЭIDAT(�c���?���T
�zF����h�,�IEND�B`�
```

**encode in base64**
```
ruby -e "require 'base64'; p ::Base64.strict_encode64(::IO.binread 'test/cruby/test/cgi/testdata/small.png')"
"iVBORwoaCgAAAApJSERSAAAAEQAAABEIAgAAALQP0K0AAAAaSURBVCiRY/z//z8DiYCJVAqjekb1jOqBAwCKaAMf7iyZAAAAAABJRU5ErkJggg=="
```

When doing the same using Node.js, the following exception is thrown:
```
ArgumentError: invalid character (failed: The string to be encoded contains characters outside of the Latin1 range.)
```

Mogztter:
I gave it more thoughts and I think it's a bad idea. Ruby and Node.js use two fundamentally different approaches. Now `binread` produces a valid stream of characters when reading a text file (encoded in utf-8) but will produce an invalid stream of characters when reading a binary file.

I think we should stick with `toString('binary')` as it's the closest we can get to the Ruby implementation.

@iliabylich Unless you have a better idea, could you please revert this commit ?

PR #1827, opened 2018-05-15
Revert "Implement binread (mimic ASCII-8BIT encoding)"

iliabylich:
Reverts opal/opal#1824

@elia Or maybe force push HEAD~1 to the master? That's a bad idea if somebody is using Opal directly from the master branch, so I think reverting is better.

### DIFF
--- a/stdlib/nodejs/io.rb
+++ b/stdlib/nodejs/io.rb
@@ -36,7 +36,7 @@ class IO
   end
 
   def self.binread(path)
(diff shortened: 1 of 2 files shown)


elia:
I occasionally force pushed and it's probably harmless for anyone using opal from git, but a revert is also fine, merging 😉

Mogztter:
Thank you both :wink:

PR #1832, opened 2018-05-24
ES6 import export, webpack based build process

janbiedermann:
This PR implements wrapping opal modules as ES6 importable/exportable modules.
Together with:
https://github.com/janbiedermann/opal-webpack-compile-server
https://github.com/janbiedermann/opal-webpack-loader
https://github.com/janbiedermann/opal-webpack-resolver-plugin

allows for a completely webpack based build process.
Builds are superfast, asynchronous and depending on webpack handling imports also in parallel.
Ina a real world app, complete build generating 8,5MB of js from opal code:
sprockets building all code:  34172ms
opal-webpack-loader: Time: 13799ms

The webpack based build process and ES6 import/export modules allow for using all the webpack features, like:
- https://webpack.js.org/guides/hot-module-replacement/
- https://webpack.js.org/guides/tree-shaking/
- https://webpack.js.org/guides/code-splitting/
- https://webpack.js.org/guides/lazy-loading/

At the same time, integration of opal code in the build chain of other projects is now much better possible, for example:
- electron
- react native

Instructions for getting this running are in the readme here https://github.com/janbiedermann/opal-webpack-loader
there is a example app available here: https://github.com/janbiedermann/opal-webpack-loader-example-app

This is used in 2 real life apps.

This code introduces a new compiler option: es_six_imexable
it takes care of the wrapping opal modules as ES6 module

I used duck tape to make it all stick together, so the compiler of this PR doesn't work in node or the browser yet, also it may fail without the es_six_imexable option.

This is not so, that you can import a ruby module in ES6 like 'import 'my_rubymodule', instead, the imports are generated automatically. See comments in code.

Its a start, and it works :)

Comments much appreciated :)


janbiedermann:
Cleaned it up, guarded es_six_imexable specific code, should function ok now

elia:
@janbiedermann quick question: what's the meaning of "imexable"? is it some ES6 specific term?

janbiedermann:
imexable is just short for (import|export)-able.

janbiedermann:
Now that source maps are fixed, those thousands of requests to load source maps when you open the dev console in the browser are a thing of the past, its possible to use all webpack devtool variations, from simple source maps, which load waaaayyyyy faster to inline source maps, see:
https://webpack.js.org/configuration/devtool/

janbiedermann:
We have success with lazy loading opal code using webpack dynamic imports.
we have success building cordova/react-native apps from webapp code, using webpack generated bundles. Works surprisingly well.
See https://github.com/janbiedermann/opal-webpack-loader-example-app

lib/opal/nodes/top.rb
```diff
@@ -41,11 +47,66 @@ def compile
...
+          if compiler.required_trees.size > 0
+            base_dir = Pathname.new(compiler.file).dirname
+
+            compiler.required_trees.each do |module_path|
+              # javascript import doesn't allow for import of directories, to support require_tree
+              # the compiler must import each file in the tree separately
+              import_child_paths(import_lines, base_dir, module_path)
+            end
+          end
+          if import_lines.size > 0
```
iliabylich: `.any?` instead of `size > 0`

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
       def version_comment
         "/* Generated by Opal #{Opal::VERSION} */"
       end
+
+
+      def self.module_name(filenamepath, original_filepath = nil, original_filename = nil)
```
iliabylich: `filenamepath` is a very confusing name. Is it a `name` or a ` path`?

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+        original_filename = filenamepath unless original_filename
+        original_filepath = Pathname.new(filenamepath).expand_path unless original_filepath
+        o_s = original_filepath.to_s
```
iliabylich: Not sure if it's worth starting this method with `Pathname` since you use it mostly as a string. Maybe convert it to the string in the beginning and wrap it with a `Pathname` in the end?

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+        path, _ = Pathname.new(filenamepath).expand_path.split
```
iliabylich: Please, add an argument to the `split` method call. For example I don't remember what's the default value (and I think I'm not the only one)

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+        if Opal.paths.include?(path.expand_path.to_s)
+          e = if o_s.end_with?('.js.rb')
+                -7
```
iliabylich: magical constants? What's the intention of this code? To remove an extension suffix? `String#sub` with `/\.js\.rb\z/` can probably do it nicer. (Ruby 2.5 has a `String.delete_suffix` method, but we support older versions)

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+                -4
+              else
+                -1
+              end
+          return o_s[(path.expand_path.to_s.size+1)..e]
+        end
+        if path.root?
+          pwd = Dir.pwd
+          if o_s.start_with?(pwd)
+            e = if o_s.end_with?('.js.rb')
```
iliabylich: Same here, `String#sub` + there seems to be a code duplication. Probably it would be easier to extract to a separate method (in a separate class that probably should be created)

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+          else
+            return o_s
+          end
+        end
+        module_name(path, original_filepath, original_filename)
+      end
+
+      private
+
+      def import_name(m_name)
```
iliabylich: `module_name`

lib/opal/nodes/top.rb
```diff
@@ -108,6 +172,66 @@ def compile_end_construct
...
+        m_name.gsub('.','o_').gsub('-','_').gsub('/','_').gsub('@','_at_') + rand(36**8).to_s(36)
```
iliabylich: Please, add a space after comma. I think `rake rubocop` would give you the same suggestion.

opal/corelib/runtime.js
```diff
@@ -1177,6 +1177,8 @@
 
   // The Array of ancestors for a given module/class
   Opal.ancestors = function(module) {
+    if (!module) { return []; }
```
iliabylich: Why is it necessary?

iliabylich:
Yes. Please, revert changes to mspec and rubyspec submodules. CI fails because of that, there are some new tests in the upstream that we can't pass atm.

lib/opal/cli.rb
```diff
@@ -131,6 +131,7 @@ def compiler_option_names
         inline_operators
         enable_source_location
         parse_comments
+        es_six_imexable
```
elia: can you also change `es-six-imexable` / `es_six_imexable` to `es6-modules` / `es6_modules` everywhere?

lib/opal/nodes/top.rb
janbiedermann: Thats Pathname#split and doesn't accept arguments. I added comments in the code.

opal/corelib/runtime.js
janbiedermann: This is guarding a condition where module parameter is passed as nil.
This may be related to a bug, where the wrong context object passed around. Did not yet have time to investigate further. This seems save 'fix' for now and shouldn't hurt anything else.

janbiedermann:
moved the module_name method from top to the compiler and renamed it to module_name_from_paths, properly named all vars, added comments. All methods in  top.rb generate (partial) js.

the spec/(mspec|ruby) commit is a accident. I will revert once i figure out how to exactly achieve this.

janbiedermann:
ok, i hope i got the reverting of spec/(mspec|ruby) right

I hope everything is ok now

iliabylich:
1. rubocop has some complains on CI. Could you run `rubocop -a` to fix them? (Most probably it should be enough, unless there are some issues that can't be fixed automatically).
2. Could you maybe setup a simple test that compiles a single empty file and asserts its content?

lib/opal/compiler.rb
```diff
@@ -57,6 +57,45 @@ def self.module_name(path)
...
+        if original_path_s.start_with?(pwd)
+          # got a match in current dir
+          # remove current dir at the beginning and the filename extension to get the module name like 'some/ruby'
+          return original_path_s[(pwd.size + 1)..-1].sub(/\.(js|rb|js\.rb)\z/, '')
+        else
+          # no match at all, return original path
+          return original_path_s
+        end
+      end
+      module_name_from_paths(nil, path, original_path)
```
iliabylich: Usually passing `nil` as a first argument means that something's wrong. Maybe change it to keyword arguments?

lib/opal/nodes/call.rb
```diff
@@ -226,11 +226,17 @@ def handle_special(&compile_default)
         file = compiler.file
         if arg.type == :str
           dir = File.dirname(file)
-          compiler.requires << Pathname(dir).join(arg.children[0]).cleanpath.to_s
+          mod_filename = Pathname(dir).join(arg.children[0]).cleanpath.to_s
+          compiler.requires << mod_filename
+        end
+        if compiler.es6_modules?
+          mod_filename = mod_filename.end_with?('.rb') ? mod_filename : mod_filename + '.rb'
+          push fragment("self.$require(#{Opal::Nodes::TopNode.module_name(mod_filename).inspect})")
```
iliabylich: I don't see this method (`module_name`) in the `TopNode`. Was it moved to the compiler?

lib/opal/nodes/top.rb
```diff
@@ -13,9 +13,15 @@ class TopNode < ScopeNode
       children :body
 
       def compile
-        push version_comment
+        push version_comment unless compiler.es6_modules?
+
+        if compiler.es6_modules?
+          module_name = self.class.module_name(compiler.file)
```
iliabylich: Same here, I don't see this method.

lib/opal/nodes/top.rb
```diff
@@ -41,11 +47,74 @@ def compile
...
+            # or the opal-webpack-loader don't mix them up with javascript imports
+            # that is just a sensible convention for example, a require 'react', without ending:
+            #   import 'react';  --> resolves to react.js in javascript space, may be resolved by webpack otherwise
+            # vs. with ending:
+            #   import 'react.rb';  --> resolves to react.rb in the opal/ruby space which gets transpiled to js by the loader
+            # if a javascript file gets required as:
+            #   require 'runtime'
+            # it gets imported like so:
+            #   import 'runtime.rb'
+            # the opal-webpack-resolver-plugin will then check for a runtime.rb, but also for a runtime.js if the runtime.rb is not found.
```
iliabylich: There are too many comments and LOC that are not related to the `TopNode` *directly*. It definitely should be extracted to a separate utility class, `TopNode` is a utility node responsible for compiling a JS code that wraps a single ruby file.

Your code *mostly* does a different thing: it handles paths, does some recursive walking etc. All these calculations should be extract to some module like `ES6Paths` that must be used by this class.

lib/opal/nodes/top.rb
```diff
@@ -108,6 +180,34 @@ def compile_end_construct
...
       end
+
+      private
+
+      def generate_import_name(module_name)
+        # generate random import name for a ruby module_name. Also replaces some characters that are illegal in JS import names.
+        module_name.gsub('.', 'o_').gsub('-', '_').gsub('/', '_').gsub('@', '_at_') + rand(36**8).to_s(36)
+      end
+
+      def import_child_paths(import_lines, base_dir, module_path)
```
iliabylich: Same for this method: it's a part of the compilation node, but it doesn't even use `self`. It should also be extracted @elia wdyt?

lib/opal/nodes/top.rb
```diff
@@ -108,6 +180,34 @@ def compile_end_construct
...
+        # recursively walk a directory and generate import lines for all .rb/.js files
+        directory_path = base_dir + module_path
+        directory_path.each_child do |child_path|
+          if child_path.directory?
+            import_child_paths(import_lines, base_dir, child_path.expand_path.to_s[(base_dir.expand_path.to_s.length+1)..-1])
+          elsif child_path.file?
+            path_s = child_path.basename.to_s
+            if path_s.end_with?('.rb') || path_s.end_with?('.js')
+              module_path = child_path.expand_path.to_s[(base_dir.expand_path.to_s.length+1)..-4]
```
iliabylich: `-4` is a magical constant

janbiedermann:
Alright, i introduced ES6ModulesHelpers and moved most things there and refactored a little. TopNode and Compiler look much nicer now 👍 

Also added two small tests, testing for import line and export line.
Ok?

lib/opal/nodes/top.rb
janbiedermann: all requested changes are meanwhile marked outdated. So i resolve this conversion for now.
Ill work on that branch soon again, merging master, etc.

janbiedermann:
as all is green for this now, ill continue further in https://github.com/opal/opal/pull/1969

janbiedermann:
superseded by PR#1970

PR #1833, opened 2018-05-27
fix corelib/string

janbiedermann:
with @length generated code tries to assign nil to read only property length, `self.length`works correct


iliabylich:
Thanks, good catch!

PR #1837, opened 2018-06-01
source map line number 0 is illegal and fails webpack source map veri…

janbiedermann:
opal sometimes creates source maps with indices that point to line 0 of the original source.
spec says, lines count from 1, so 0 line is illegal and fails webpack source map verification.
fragments seem to have no line number if they don't belong to a sexp.
Thats ok, so if line is nil or 0, this just puts a 1 instead as fix.
So all generated line indices are legal, webpack is happy.

### DIFF
--- a/lib/opal/source_map.rb
+++ b/lib/opal/source_map.rb
@@ -19,7 +19,7 @@ module Opal
 
         mappings = @fragments.map do |fragment|
           mapping = nil
-          source_line   = fragment.line
+          source_line   = (fragment.line && fragment.line > 0) ? fragment.line : 1
           source_column = fragment.column
           source_code   = fragment.code
 
--- a/lib/opal/source_map.rb
+++ b/lib/opal/source_map.rb
@@ -19,7 +19,10 @@ module Opal
 
         mappings = @fragments.map do |fragment|
           mapping = nil
-          source_line   = (fragment.line && fragment.line > 0) ? fragment.line : 1
+          source_line   = fragment.line
+          # source map validators enforce line numbers starting with 1
+          source_line   = 1 if source_line.nil? || source_line.zero?
+
           source_column = fragment.column
           source_code   = fragment.code
 
@@ -49,7 +52,8 @@ module Opal
 
         # Ensure mappings isn't empty: https://github.com/maccman/sourcemap/issues/11
         unless mappings.any?
-          zero_offset = ::SourceMap::Offset.new(0, 0)
+          # source map validators enforce line numbers starting with 1
+          zero_offset = ::SourceMap::Offset.new(1, 0)
           mappings = [::SourceMap::Mapping.new(source_file, zero_offset, zero_offset)]
         end
 


janbiedermann:
this bug probably affects use of source maps in browsers too

janbiedermann:
for reference:
https://github.com/mozilla/source-map
there is a link to the spec `This is a library to generate and consume the source map format described here.`, but that sentence is not true then.

janbiedermann:
https://github.com/mozilla/source-map/issues/344

lib/opal/source_map.rb
```diff
@@ -19,7 +19,7 @@ def map
 
         mappings = @fragments.map do |fragment|
           mapping = nil
-          source_line   = fragment.line
+          source_line   = (fragment.line && fragment.line > 0) ? fragment.line : 1
           source_column = fragment.column
           source_code   = fragment.code
```
elia: Instead of checking `fragment.line` here, it would be better to check `source_line` inside the `if` in the next line, so to save some method calls and make rubocop happy, e.g. `source_line = 1 if source_line.zero?`.

elia:
@janbiedermann I think https://github.com/mozilla/source-map#sourcemapconsumerprototypeoriginalpositionforgeneratedposition clears that the library is 1-based while the spec is 0-based, and from what I see the library we use is 0-based too https://github.com/maccman/sourcemap.

---

As the saying goes, the **two** great problems in computer science are:

1. naming things
2. invalidating caches
3. _off-by-one errors!_ 😄

janbiedermann:
i missed that, thanks, ill adjust the check in the next days

janbiedermann:
it seems maccman/sourcemap is 1 based too:
https://github.com/maccman/sourcemap/blob/b65550a30ae3216e2d9afa19dee3593c2282a1e3/lib/source_map/map.rb#L35
https://github.com/maccman/sourcemap/blob/b65550a30ae3216e2d9afa19dee3593c2282a1e3/lib/source_map/map.rb#L191
https://github.com/maccman/sourcemap/blob/b65550a30ae3216e2d9afa19dee3593c2282a1e3/lib/source_map/map.rb#L200

It just doesn't enforce line numbers to valid and > 1.

The root problem seems to come from parser25, producing nil or 0 line numbers.
I adjusted the fix to a if in next line

lib/opal/source_map.rb
janbiedermann: done

iliabylich:
If there are nils in the parser's sourcemaps then it's definitely a bug in the parser gem. But Opal generates a lot of AST nodes in runtime (both in rewriters and node handlers), so it's also can be our bug.

And by the way, I'm not familiar with how sourcemaps work, but shouldn't we add +1 to **every** line number?

janbiedermann:
it seems line numbers start at 1 already, no need to add +1 to every line.
this problem with nil or 0 happens only for a few files, i will check again if you say it could be opal, maybe i can identify the responsible ast nodes

elia:
@janbiedermann @iliabylich could it be that the zeros come from ast rewriters loosing the original node position? I remember improving the situation but maybe there are still some corner cases.

elia:
Thanks! Please send more PRs if you find more about source maps subtleties!

PR #1838, opened 2018-06-03
const_set before inherited

janbiedermann:
set const before calling inherited otherwise class.name leads to incorrect results and method_missing depending things misbehave, get called in wrong context


janbiedermann:
added your test :)

iliabylich:
Thanks!

PR #1839, opened 2018-06-05
guard ancestors

janbiedermann:
otherwise `Opal.is_a({}, Opal.Array)` will fail
this was previously automatically guarded by a `while (module) {`

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1177,6 +1177,8 @@
 
   // The Array of ancestors for a given module/class
   Opal.ancestors = function(module) {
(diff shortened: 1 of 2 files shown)


iliabylich:
LGTM. Could you add a simple test for this please to avoid potential regression in the future? `spec/opal/core/runtime/is_a_spec.rb` is a good place for this:
``` ruby
it 'works for non-Opal objects' do
  `Opal.is_a({}, Opal.Array)`.should == false
end
```

janbiedermann:
added your test

iliabylich:
Thank you!

PR #1842, opened 2018-06-08
String improvements

iliabylich:
+ `String#to_r`
+ `String#to_c`
+ `String#match?`
+ `String#unicode_normalize` (returns `self`)
+ `String#unicode_normalized?` (returns `true`)


iliabylich:
@elia ping

elia:
@iliabylich thanks, sorry for the delay

PR #1844, opened 2018-06-14
Simplify args compilation

iliabylich:
This is mostly a refactoring, I've attempted it multiple times but had to give up every time. Finally it's done.

1. Added `MlhsArgs` rewriter that converts `def m( (a, b) )` to `def m($mlhs1); a, b = $mlhs1`. It removes all ugly stuff like `mlhs_mapping` from the `DefNode` and `NodeWithArgs`
2. Added `InlineArgs` rewriter that creates virtual AST nodes for arguments initialization and insert them in the method body. I've added some comments, hope it's clear enough. It removes ugly stuff like `scope.working_arguments` and `scope.kwargs_initialized` and a horrible `PostArgs` node.
3. Extracted arity checking to `arity_check` and `iter_arity_check` nodes.
4. Extracted code responsible for `Method#parameters` to a separate class.


iliabylich:
@elia Sorry for delay, anything else?

elia:
@iliabylich sorry, this one went out of the radar 😅

PR #1848, opened 2018-07-01
Fixed Kernel#public_methods to return instance methods if argument is false

ysakasin:
`Kernel#public_methods` is a alias of `Kernel#methods` in opal (#909). But their behavior is not same in MRI. This PR fixed it.

Here is an example.
``` ruby
class Child
  def foo
    "foo"
  end
end

c = Child.new

def c.bar
  "bar"
end


p c.public_methods(false) #=> [:bar, :foo]
p c.methods(false)        #=> [:foo]
```

`methods(false)` returns only singular methods, but`public_methods(false)` returns singular methods and instance methods.

### ruby-doc
- [Object#public_methods](https://ruby-doc.org/core-2.5.1/Object.html#method-i-public_methods)
- [Object#methods](https://ruby-doc.org/core-2.5.1/Object.html#method-i-methods)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@ Whitespace conventions:
 - Fixed `pattern` argument handling for `Enumerable#grep` and `Enumerable#grep_v`. (#1757)
 - Raise `ArgumentError` instead of `TypeError` from `Numeric#step` when step is not a number. (#1757)
 - At run-time `LoadError` wasn't being raised even with `Opal.config.missing_require_severity;` set to `'error'`.
+- Fixed `Kernel#public_methods` to return instance methods if the argument is set to false.
 
 
 ## [0.11.0] - 2017-12-08
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -52,7 +52,15 @@ module Kernel
     }
   end
 
-  alias public_methods methods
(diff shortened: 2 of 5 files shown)


ysakasin:
CI seems to have error due to installing jruby... Test cases doesn't fail.

Could someone restart build?

elia:
@ysakasin great!

Can you also remove [these lines](https://github.com/opal/opal/blob/0-11-stable/spec/filters/bugs/kernel.rb#L56-L59) to enable the related Ruby Spec examples?

elia:
@ysakasin I restarted the build and tried to upgrade the jruby version, both without success, I'll merge anyway after the Ruby Spec examples are activated/checked. 👍🏼

ysakasin:
@elia Thanks for reviewing.

I enabled a Ruby Spec example and pushed additional fix.
We can't enable these lines because opal can't distinguish between public and private method.
- `Kernel#public_methods when passed false returns a list of public methods in without its ancestors`
- `Kernel#public_methods when passed nil returns a list of public methods in without its ancestors`

Also this line already removed in master: `Kernel#public_methods returns a list of the names of publicly accessible methods in the object and its ancestors and mixed-in modules`

elia:
👍🏼

ysakasin:
Thank you merging!

PR #1863, opened 2018-07-15
fix String.encoding for "use strict"

janbiedermann:
ES5/"use strict"; make properties of primitives read only, assigning to encoding fails with a TypeError, assignment to read only property.

This patch works around this restriction by using a simple proxy object for encoding.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -54,6 +54,9 @@
   // The Opal object that is exposed globally
   var Opal = this.Opal = {};
 
+  // in strict mode eval doesn't introduce a new variable into the outer context.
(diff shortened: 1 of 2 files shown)


janbiedermann:
ES6 modules use "use strict" by default.

janbiedermann:
So this doesn't work for now, as the encoding is not set on the instance, but on the prototype instead.
Silly.

janbiedermann:
New fix is to just comment the assignment. Tests pass, though force_encoding now is mostly useless.

Mogztter:
@janbiedermann What about:

```js
var result = new String(self.toString())
Object.defineProperty(result, 'encoding', {
  value: encoding,
  writable: true
})
return result
```

janbiedermann:
superseded by PR#1970

PR #1885, opened 2018-09-11
Add Date#next_year and Date#prev_year

elia:
See ruby-doc.org/stdlib-2.5.0/libdoc/date/rdoc/Date.html#method-i-next_year

Completes and closes #1758


stdlib/date.rb
```diff
@@ -347,6 +347,11 @@ def gregorian_leap?(year)
   end
 
   def initialize(year = -4712, month = 1, day = 1, start = ITALY)
+    %x{
+      if (year === 1582 && month === 10 && day > 4 && day < 15) {
+        day = 4;
```
iliabylich: Is it necessary? If so maybe it makes sense to add a comment?

PR #1886, opened 2018-09-15
Resolves #1883, use require directly instead of the require_node method

### LINKED ISSUE #1883: require function is used in a way in which dependencies cannot be statically extracted
This warning message is thrown by Webpack when it processes this file: https://github.com/Mogztter/opal-node-runtime/blob/master/src/nodejs.js

```
WARNING in ./node_modules/opal-runtime/src/nodejs.js
28:48-55 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./node_modules/opal-runtime/src/nodejs.js
 @ ./node_modules/opal-runtime/src/index.js
 @ ./node_modules/asciidoctor.js/dist/browser/asciidoctor.js
 @ ./index.jsx
 @ multi ./index.jsx
```

I think that Webpack is confused because `require` is registered as a constant and we are using a function to require Node modules in the Kernel class: https://github.com/opal/opal/blob/d77bd3ce19e101988578cdb6314d659491a77844/stdlib/nodejs/kernel.rb#L3-L22

I'm wondering if we could generate code that we will be easier to parse for static analysis ? I don't really see the value of `@__fs__ = node_require :fs` (but I'm might be missing something ?)
Maybe we should use:

```js
@__fs__ = `require('fs')`
```

Original issue: https://github.com/asciidoctor/asciidoctor.js/issues/508

Mogztter:
Static tools should now be able to resolve dependencies.

### DIFF
--- a/stdlib/nodejs/kernel.rb
+++ b/stdlib/nodejs/kernel.rb
@@ -18,7 +18,7 @@ module Kernel
 
   # @deprecated Please use `require('module')` instead
   def node_require(path)
-    warn '[DEPRECATION] node_require is deprecated.  Please use `require(\'module\')` instead.'
+    warn '[DEPRECATION] node_require is deprecated. Please use `require(\'module\')` instead.'
     `#{NODE_REQUIRE}(#{path.to_str})`
   end
 end
--- a/stdlib/nodejs/kernel.rb
+++ b/stdlib/nodejs/kernel.rb
@@ -16,6 +16,7 @@ module Kernel
     }
   end
 
+  # @deprecated Please use `require('module')` instead
   def node_require(path)
     `#{NODE_REQUIRE}(#{path.to_str})`
   end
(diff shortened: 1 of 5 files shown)


stdlib/nodejs/kernel.rb
```diff
@@ -16,6 +16,7 @@ def caller(*args)
     }
   end
 
+  # @deprecated Please use `require('module')` instead
   def node_require(path)
```
elia: Can you add a `warn` inside the method too? 
Shouldn't bother users too much since it happens only once when files are initially required?

Mogztter: I added the `warn` message using the GitHub UI 😨

Mogztter: @elia Does the deprecation message look good ?

elia: @Mogztter it looks good, only a nitpick, you should remove the backticks around `node_require` or make clear `require('module')` needs to be wrapped in them.

Mogztter: Indeed, I will remove the backticks around `node_require`, my bad

Mogztter:
🎉 🎉 🎉

PR #1888, opened 2018-10-02
improve String#to_proc perf

### LINKED ISSUE #1659: Passing a symbol proc to an enumerable method instead of a block is considerably slower
Passing a symbol proc to an iteration method of an enumerable is considerable slower than passing a block.

Consider the following case:

```ruby
lines.map {|line| line.strip }
```

If we rewrite this as:

```ruby
lines.map(&:strip)
```

The code takes almost twice as long to execute when running in Opal. (whereas MRI executes it slightly faster).

I would expect for Opal to produce the same JavaScript in both cases, effectively expanding the symbol proc to long form. Instead, it does the resolution of the proc from the symbol at runtime.

elia:
Builds on https://github.com/opal/opal/commit/6ce0619e0e1acdb2ba8ca406263bd8a2a329a91f

Fixes #1659 

reaching 1.06x vs the inlined block

cc @mojavelinux @Mogztter

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1654,7 +1654,9 @@
     var body = (typeof(method) === 'string') ? recv['$'+method] : method;
 
     if (body != null) {
-      body.$$p = block;
+      if (typeof block === 'function') {
+        body.$$p = block;
+      }
       return body.apply(recv, args);
     }
 
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -1313,16 +1313,33 @@ class String < `String`
   end
 
   def to_proc
-    sym = `self.valueOf()`
+    method_name = '$' + `self.valueOf()`
 
     proc do |*args, &block|
       %x{
         if (args.length === 0) {
           #{raise ArgumentError, 'no receiver given'}
         }
-        var obj = args.shift();
-        if (obj == null) obj = nil;
-        return Opal.send(obj, sym, args, block);
+
+        var recv = args[0];
+
+        if (recv == null) recv = nil;
+
+        var body = recv[#{method_name}];
+
+        if (!body) {
+          return recv.$method_missing.apply(recv, args);
+        }
+
+        if (typeof block === 'function') {
+          body.$$p = block;
+        }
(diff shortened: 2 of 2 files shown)


elia:
IPS benchmark results (see https://github.com/opal/opal/pull/1888/commits/3868d5e0aaeb27c92d195fe7d5e55d249f1521d9)

```
BEFORE

opal:elia/string-to-proc-perf ⤑ bundle exec rake bench:ips FILE=benchmark-ips/bm_symbol_to_proc.rb                                     ~/C/opal
=== Files: benchmark-ips/bm_symbol_to_proc.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_symbol_to_proc.rb
           map block     1.090k i/100ms
          map symbol   869.000  i/100ms
           map block     11.302k (± 1.0%) i/s -     56.680k in   5.015370s
          map symbol      8.878k (± 2.3%) i/s -     45.188k in   5.092622s

Comparison:
           map block:    11302.5 i/s
          map symbol:     8878.1 i/s - 1.27x  slower



AFTER

opal:master ⤑ bundle exec rake bench:ips FILE=benchmark-ips/bm_symbol_to_proc.rb                                                       ~/C/opal
=== Files: benchmark-ips/bm_symbol_to_proc.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_symbol_to_proc.rb
           map block   980.000  i/100ms
          map symbol   977.000  i/100ms
           map block     10.430k (± 1.1%) i/s -     52.920k in   5.074384s
          map symbol      9.804k (± 2.6%) i/s -     49.827k in   5.086232s

Comparison:
           map block:    10430.1 i/s
          map symbol:     9803.7 i/s - 1.06x  slower
```

mojavelinux:
Nice work. Every day Opal gets faster is a good day for me ;)

elia:
oops, the second commit had no change attached, fixed now

Issue #727, opened 2015-03-05
Browser support


elia:
> On an unrelated subject (well, maybe not so unrelated), I believe the claim that Opal runs on IE 6 is no longer true. I have seen things like Object.keys used in corelib (I think in the Array or Hash implementation), and that only became available in IE 9 I think. Not sure what if anything needs to be done about that, but wanted to hear your thoughts on what browsers should (ideally vs practically) be supported and how to make sure they are as work on Opal continues.
> — @vais in #726 

cc @adambeynon @meh

elia:
Maybe it's time to wire up browserstack on opal/opal too…

vais:
[hash.rb](https://github.com/opal/opal/blob/master/opal/corelib/hash.rb) leans pretty heavily on Object.keys, [which is IE 9 and up](http://kangax.github.io/compat-table/es5/#Object.keys) (click on "Show obsolete browsers?" checkbox to see IE 7 and 8, but IE 6 does not show up even with that checkbox checked, which is telling).

If the decision is that we want to support IE below 9, then for Object.keys we should probably use a polyfill like [the one shown on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys). Or go a full shim library route?

The bigger issue is automated testing. If that is in place, getting back to IE 6 compatibility would be tedious, but trivial. I think Opal is in a unique position to be able to support ancient browsers, because we only have to write the ugly code once :)

meh:
@elia yeah, I'll look into that soon enough, always wanted to do it but dreaded the experience.

@vais I'd go with something like `Opal.keys` in the runtime or an `Opal` helper like the stuff we have already in `helpers.rb`.

vais:
@meh `helpers.rb` does look like a good place for this kind of stuff. But so I understand better, are there specific reasons that you prefer this approach vs. using a "shim" library? Performance? Dependency?

meh:
@vais dependency mostly.

adambeynon:
@vais, @elia, @meh  We don't actually use `Object.keys()` inside hash.rb, the `.keys` property is added to the hash instance during initialisation. I have had a quick check inside the other core lib files as well, and I don't think we use `Object.keys()` anywhere else (anymore?)

vais:
@adambeynon in that case, sorry for creating confusion! I need to actually try to run the spec suite on IE and post here with any results. Until that is done, it's all just empty speculation on my part.

[...]

PR #1890, opened 2018-10-05
Remove polyfills for atob/btoa

elia:
It's supported by IE10+ and all other browsers, it's easy to
add a polyfill in case support for IE9- is required.

https://caniuse.com/#feat=atob-btoa

(ref https://github.com/opal/opal/issues/727#issuecomment-354538720 by @Mogztter)

### DIFF
--- a/stdlib/base64.rb
+++ b/stdlib/base64.rb
@@ -1,59 +1,7 @@
 module Base64
-  # FROM https://github.com/davidchambers/Base64.js/blob/69262ec7e1fa4541de5700a1b0b03b0de0e3f5aa/base64.js
   %x{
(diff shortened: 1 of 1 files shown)


Mogztter:
I think it's worth it 😁
It makes the code easier to maintain and as you mentioned it's relatively easy to add a polyfill (and... who is still using IE9 anyway ?!)

Indeed for Node.js we should use an implementation that relies on `Buffer` (native).

takaram:
Node.js implements `atob` / `btoa` since v16.0.
https://developer.mozilla.org/en-US/docs/Web/API/atob#browser_compatibility

Probably we can merge this now?

PR #657, opened 2014-11-24
added random.rb with mersenne twister. Kernel#rand now uses this. Implem...

daveheitzman:
...enter Kernel#srand

PR #1891, opened 2018-10-08
random with mersenne twister

elia:
Fixes https://github.com/opal/opal/pull/657

@iliabylich any idea on how to deal with requires, I thought about adding `/opal/opal/with_mersenne_twister` but I don't like it much.

I'll add an entry in Travis to test with Mersenne Twister random generator once figured out the best route for requiring it.


opal/corelib/random/seedrandom.js.rb
```diff
@@ -1,4 +1,6 @@
 class Random
+  SEEDRANDOM_GENERATOR = Object.new
+
```
elia: ✂️

elia:
I don't know tbh, the mersenne twister is bigger, maybe the default choice should be closer to MRI allowing anyone who needs lightness to pick securerandom or maybe a non _pseudo_ generator based on `Math.random`.

iliabylich:
Looks like there was a spec that expected seedrandom to be used 😄 . Also the new randomization code should be wrapped with `lint:disable` comments

PR #1895, opened 2018-10-27
Resolves #1794, File.expand_path on Windows root should include trailing slash

### LINKED ISSUE #1794: File.expand_path on Windows root should include trailing slash
When File.expand_path is run on a Windows root (i.e., bare drive letter), it should include a trailing slash (to be consistent with MRI). However, it does not. It also has difficulty expanding a raw drive letter.

Here's what MRI does:

```rb
Dir.pwd
#=> C:/Users/Usename/directory

File.expand_path 'C:'
#=> C:/Users/Username/directory

File.expand_path 'C:/'
#=> C:/

File.expand_path 'D:'
#=> D:/

File.expand_path 'D:/'
#=> D:/
```

Here's what Opal does (with nodejs support enabled):

```rb
Dir.pwd
#=> C:\Users\Usename\directory

File.expand_path 'C:'
#=> C:/Users/Username/directory/C:

File.expand_path 'C:/'
#=> C:

File.expand_path 'D:'
#=> C:/Users/Username/directory/D:

File.expand_path 'D:/'
#=> D:
```

The correct behavior for `C:` is debatable.

(the value of `Dir.pwd` is show here for context, but note that the value it returns is also incorrect).

Mogztter:
Resolves #1794


Mogztter:
As it turns out, Node.js `resolve` and Ruby `expand_path` are very similar!

Ruby is using posix separator even on Windows so I'm using `.split(path.sep).join(path.posix.sep)` to convert OS dependent separator to posix separator.

I've also fixed a bug in `Pathname`.
The `absolute?` and `relative?` methods in `stdlib/pathname.rb` does not work properly with Windows paths. In Node.js the `path` module is OS dependent so I'm using `path.isAbsolute?` to determine if the path is absolute or not. Luckily `Pathname.absolute?` and `path.isAbsolute?` work the same.

stdlib/nodejs/dir.rb
```diff
@@ -1,16 +1,18 @@
...
+  `var __path__ = #{@__path__}`
 
   class << self
     def [](glob)
       `__glob__.sync(#{glob})`
     end
 
     def pwd
-      `process.cwd()`
+      `__path__.normalize(process.cwd()).split(__path__.sep).join(__path__.posix.sep)`
```
Mogztter: I don't think `normalize` is mandatory here

stdlib/nodejs/pathname.rb
```diff
@@ -0,0 +1,31 @@
...
+    else
+      raise TypeError, "no implicit conversion of #{path.class} into String"
+    end
+    raise ArgumentError if @path == "\0"
+  end
+
+  attr_reader :path
+
+  def absolute?
+    `__path__.isAbsolute(#{@path.to_str})`
```
Mogztter: Use `isAbsolute` otherwise `c:/absolute/path` is not considered as an absolute path by the `stdlib/pathname` implementation.

stdlib/nodejs/file.rb
```diff
@@ -59,102 +59,111 @@ class File < IO
     ALT_SEPARATOR = `__path__.sep`
   end
 
-  def self.read(path)
-    `return executeIOAction(function(){return __fs__.readFileSync(#{path}).toString()})`
-  end
+  class << self
```
Mogztter: I'm using `class << self`` to remove `def.self` redundancy but I can revert this change if neededed

stdlib/nodejs/pathname.rb
```diff
@@ -0,0 +1,31 @@
+class Pathname
+  include Comparable
+
+  @__path__ = `require('path')`
+  `var __path__ = #{@__path__}`
+
+  def initialize(path)
```
Mogztter: @elia Am I forced to duplicate this code ? Is there a better way to do it ?

test/nodejs/test_file.rb
```diff
@@ -147,15 +147,21 @@ def test_windows_separators
...
+    assert_equal(Dir.pwd + '/foo/bar.js', File.expand_path('./foo/bar.js'))
+    assert_equal(drive_letter + '/foo/bar.js', File.expand_path('/foo/bar.js'))
     assert_equal('c:/foo/bar.js', File.expand_path('c:/foo/bar.js'))
     assert_equal('c:/foo/bar.js', File.expand_path('c:\\foo\\bar.js'))
-    assert_equal('c:/foo/bar.js', File.expand_path( 'bar.js', 'c:\\foo'))
-    assert_equal('c:/foo/baz/bar.js', File.expand_path('\\baz\\bar.js', 'c:\\foo'))
-    assert_equal('c:/baz/bar.js', File.expand_path( '\\baz\\bar.js', 'c:\\foo\\..'))
-    assert_equal('c:/foo/bar.js', File.expand_path( '\\..\\bar.js', 'c:\\foo\\baz'))
-    assert_equal('c:/foo/bar.js', File.expand_path( '\\..\\bar.js', 'c:\\foo\\baz\\'))
+    assert_equal('c:/foo/bar.js', File.expand_path('bar.js', 'c:\\foo'))
```
Mogztter: Some assertions were wrong.
For instance `\\baz\\bar.js` is an absolute path, so `File.expand_path( '\\baz\\bar.js', 'c:\\foo')` should be equal to `c:/baz/bar.js` (not `c:/foo/baz/bar.js`).

Mogztter:
Build is green, I just need to fix the code duplication that I introduced by copying/pasting the `Pathname` constructor :thinking:

stdlib/nodejs/file.rb
elia: Yes please, keep the explicit `self.`, it's easier to recognize singleton methods in long method lists this way, also keeps the diff cleaner 😉

stdlib/nodejs/pathname.rb
elia: I think would be savvy to `require 'pathname'` at the top of the file, and only overwrite where needed, if an extractable pattern arises then we can evaluate a pluggable system for different platforms (like it's done with `Opal .exit()`), but for now it's probably unnecessary.

test/nodejs/test_file.rb
elia: 👍 maybe check with ruby spec's code, if you didn't already

stdlib/nodejs/dir.rb
elia: remove it 👍, from [node's docs]() seems it just removes double dots, and [we don't know if cwd() can have them](https://nodejs.org/api/process.html#process_process_cwd), so it shouldn't make a difference in correctness, on the contrary it might be that `.cwd()` is more close to `.pwd`.

elia:
@Mogztter you prefer to cleanup the commits before the merge or it's better to just squash-and-merge?

test/nodejs/test_file.rb
Mogztter: I did check and they are also losing their sanity: `# FIXME: these are insane!`

https://github.com/ruby/ruby/blob/62ebf35592717655a34620301fd45dfc522af1e2/spec/ruby/core/file/expand_path_spec.rb#L94-L101

We could probably add (carefully) more tests cases but I think for now we are good :sweat_smile:

stdlib/nodejs/file.rb
Mogztter: Ok

elia:
Merged! 😄👍

PR #1904, opened 2018-11-30
Bump parser to 2.5.3 for Ruby 2.3.8, 2.4.5 and 2.5.3 support

rene-demonsters:
Parser has added support for Ruby 2.3.7, Ruby 2.4.5 and Ruby 2.5.3. Currently the version of Parser that's required by Opal support only up to Ruby 2.4.4, this should take care of that.

See: https://github.com/whitequark/parser/pull/528

Also do note that the latest Opal 0.11.4 still has `spec.add_dependency 'parser', '= 2.3.3.1'` even though the bump to a 2.5.x version was made 8 months ago, and 0.11.4 was released less than a month ago.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,12 +63,6 @@ matrix:
     - rvm: 2.3.7
       env: RUN=rspec
 
-    - rvm: 2.2.10
-      env: RUN=rspec TILT_VERSION=2.0.1 SPROCKETS_VERSION='~> 3.7'
-
-    - rvm: 2.1.10
-      env: RUN=rspec RACK_VERSION='< 2.0'
-
     - rvm: jruby-9.2.0.0
       env: RUN=rspec
 
--- a/Gemfile
+++ b/Gemfile
@@ -51,5 +51,5 @@ group :doc do
 end
 
 platforms :ruby, :mswin, :mswin64, :mingw, :x64_mingw do
-  gem 'c_lexer', '2.5.1.0.pre2' unless RUBY_ENGINE == 'truffleruby'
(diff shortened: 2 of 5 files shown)


iliabylich:
@rene-demonsters Done - https://rubygems.org/gems/c_lexer/versions/2.5.3.0.0. Could you bump it in the Gemfile please?

rene-demonsters:
@iliabylich I bumped it, but it fails to build native extensions, see: https://travis-ci.org/opal/opal/jobs/461790306

The new commit is just because I made a small mistake with bumping, forgot to add a .0 so it'll still fail on that once it gets to that point.

iliabylich:
2.1.1 is dead - https://www.ruby-lang.org/en/downloads/branches/ could you remove it from the build matrix please? I don't want to adapt c_lexer to dead version of ruby 😄 (same for 2.2.x btw)

rene-demonsters:
I've dropped both from the matrix. If you want 2.2 back I can add it back.

rene-demonsters:
Actually shouldn't I bump the other versions in Travis to the latest ones as well?

iliabylich:
well, we can do it, but I think that's not necessary atm. I can't remember any important changes in minor versions.

iliabylich:
> Parser only supports the last minor version of each ruby
Parser **parses** as the last minor version of Ruby would parse, but it supports all of them. So `parser/ruby25` follows parsing rules of the latest Ruby 2.5.x, but it works with pretty much any version of Ruby

iliabylich:
Feel free to send another PR that bumps Ruby versions in `.travis.yml`, generally speaking that's not necessary right now. Thanks a lot!

Issue #1907, opened 2018-11-30
Splat operator as block parameter causes incorrect block length


iamprich:
Found using Opal v0.11.4
```
def eachy(&block)
  yield [123456, true]
end

eachy { |key, *value| puts "#{key} - #{value}" }
```
When passing a splat operator as a block parameter, Opal renames the parameter to `$a_rest`. This variable is unused, and is re-evaluated within the block under it's original name.

```
return $send(self, 'eachy', [], (TMP_2 = function(key, $a_rest){var self = TMP_2.$$s || this, value;
```

Upon minification, the `$a_rest` parameter is removed, which causes the length to be incorrect when the block is returned. `TMP_2.length == 1` when it should equal `2`.

I have a fix for this issue and will be submitting a pull request.

elia:
This works on master (`2.0.0dev`). The reproduction from the report now matches MRI:

```ruby
def eachy(&block)
  yield [123456, true]
end

eachy { |key, *value| puts "#{key} - #{value}" }
# => 123456 - [true]   (same as MRI)
```

Blocks no longer rely on `length` alone to decide auto-splatting; #2800 ("Don't rely on block.length alone for auto-splat") switched that to an explicit arity, which also removes the minification hazard described here, since arity no longer depends on the compiled parameter list surviving a minifier.

Looks safe to close.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1908, opened 2018-11-30
Use arity for block length if available

iamprich:
This pull request aims to solve issue #1907. By checking to see if `block.$$arity` is available, we ensure that any block parameters containing a splat operator will have the correct length. JavaScript functions will continue to use `block.length`.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1430,11 +1430,13 @@
     var has_mlhs = block.$$has_top_level_mlhs_arg,
         has_trailing_comma = block.$$has_trailing_comma_in_args;
 
-    if (block.length > 1 || ((has_mlhs || has_trailing_comma) && block.length === 1)) {
+    var block_length = (block.$$arity ? Math.abs(block.$$arity) : block.length)
+
+    if (block_length > 1 || ((has_mlhs || has_trailing_comma) && block_length === 1)) {
       arg = Opal.to_ary(arg);
     }
 
-    if ((block.length > 1 || (has_trailing_comma && block.length === 1)) && arg.$$is_array) {
+    if ((block_length > 1 || (has_trailing_comma && block_length === 1)) && arg.$$is_array) {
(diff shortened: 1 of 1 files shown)


iamprich:
I'm not completely understanding the failure and can't find it in the files. I manually patched and tested this pull request in try opal as well as a production rails app and everything seemed to work fine.

elia:
> I'm not completely understanding the failure and can't find it in the files. I manually patched and tested this pull request in try opal as well as a production rails app and everything seemed to work fine.

That's fine, this is one of the trickiest part of Ruby, and the Ruby Spec catches a number of corner cases that are very unlikely in most codebases. I'll look into that and see if I can give you any hint on how to fix that test, but maybe you'll get there first 🙂. 

Meanwhile, thanks for your contribution! Feel free to ask if you bump into anything that's too obscure or for any guidance 👍

iamprich:
I managed to find the cause of the failure, which is related to how `arity` deals with optional arguments (splat or keyword).

A solution would be to remove `Math.abs` and instead deconstruct if `block_length` has a value of `< 0` or `> 1`

So a (not super pretty) fix would look like
```
var block_length = (block.$$arity ? block.$$arity : block.length)

...

if (((block_length > 1 || block_length < 0) || (has_trailing_comma && block_length === 1)) && arg.$$is_array) {
      return block.apply(null, arg);
    }
}
```
When reproducing the failing test in try opal, this solution returned the expected value and passed my production rails app test. I'm attempting to refactor the code to read a little better before pushing my changes.

Thanks for making yourself available for any questions! I'm sure I'll come up with a few.

iamprich:
Well, the one failing test is now passing.... but I seemed to have spawned 18 new failing test :disappointed_relieved:

I just set up the test environment locally, which I should have done in the first place. Sorry about that. 

Hopefully I can get this all green!

hmdne:
I rebased this to the current master with a new PR #2250. I would rather force push it... but the repo vanished.

Issue #1846, opened 2018-06-23
Fundamentals JavaScript objects are altered


Mogztter:
```js
require('opal-runtime');
const https = require('https');
```

```
_http_outgoing.js:55
const hasOwnProperty = Function.call.bind(Object.prototype.hasOwnProperty);
                                     ^

TypeError: Cannot read property 'bind' of undefined
    at _http_outgoing.js:55:38
    at NativeModule.compile (internal/bootstrap/loaders.js:235:7)
    at NativeModule.require (internal/bootstrap/loaders.js:155:18)
    at _http_server.js:39:29
    at NativeModule.compile (internal/bootstrap/loaders.js:235:7)
    at NativeModule.require (internal/bootstrap/loaders.js:155:18)
    at https.js:34:5
    at NativeModule.compile (internal/bootstrap/loaders.js:235:7)
    at Function.NativeModule.require (internal/bootstrap/loaders.js:155:18)
    at Function.Module._load (internal/modules/cjs/loader.js:530:25)
```

The line `const hasOwnProperty = Function.call.bind(Object.prototype.hasOwnProperty);` was added 2 months ago in Node.js 10+

As you can see below, `Function.call` is `undefined` once `opal-runtime` is required:

```js
console.log('Function.call', Function.call); // Function.call function call() { [native code] }
require('opal-runtime');
console.log('Function.call', Function.call); // Function.call undefined
```

In `proc.rb` we define properties on the `Function` object: https://github.com/opal/opal/blob/master/opal/corelib/proc.rb#L2-L3

Mogztter:
For reference, here's the code added in Node.js: https://github.com/nodejs/node/blob/8ab7ea6eed76d069dfd82684e2157e7d88badebf/lib/_http_outgoing.js#L55

Mogztter:
`Function.call` is lost after this call: https://github.com/opal/opal/blob/3c8d93ea8b3ed12b4c49c51cc96d092570c9f64c/opal/corelib/runtime.js#L1161

As a workaround it's possible to define a static method in `proc`:

```rb
def self.call
  `Function.call`
end
```

I don't know why it's needed since `Proc` already extends `Function` :thinking:

iliabylich:
You are absolutely correct, that's a bug. This line initializes a singleton class of the bridged class to make the code like this working:
``` ruby
class Proc < `Function`
end

class Class
  def m
  end
end

Proc.m
```
(Because all singleton classes are inherited from `Class`)

[...]

PR #1909, opened 2018-12-09
Allocate a class object for bridged classes

elia:
Fixes https://github.com/opal/opal/issues/1846

Basically creates class objects for bridged classes instead of using the native constructor.

_I'll refine the commits once the PR is working properly and the solution is confirmed_

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+IFS=$'\n\t'
+set -vx
+
+bundle update
+
+git submodule update --init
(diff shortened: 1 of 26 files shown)


iliabylich:
WOW! I was thinking for such a long time how to fix it 😄 
```
$ opal-repl
>> `String.call(null, "test")`
=> "test"
```

All specs from RubySpec are passing. Want me to help fixing the rest?

Mogztter:
🎉🎉🎉

@elia Any performance impact that you are aware of ?

elia:
> @elia Any performance impact that you are aware of ?

I don't think there's any, this generally can only affect startup time, and a few more objects (probably less than 10) won't make any difference 👍

PR #1916, opened 2018-12-26
Fix specs in elia/allocate a class object for bridged classes

iliabylich:


iliabylich:
@elia I guess we don't run CI if PR is not for the master branch, right?

PR #1917, opened 2018-12-26
Fix specs in elia/allocate a class object for bridged classes

iliabylich:
Sorry for the spam 😄


elia:
👍 thanks, I had some of this stuff done locally, I embedded a couple of commits in https://github.com/opal/opal/pull/1909 🚀

PR #1921, opened 2018-12-31
Read encoded files and force encode to utf-8 (binary mode)

Mogztter:


elia:
@Mogztter if this is targeted to Node.js I think would be less hacky to use TextEncoder, e.g.:

```js
> var uint8array = new TextEncoder().encode("😍");
undefined
> new TextDecoder('utf8').decode(uint8array);
'😍'
> new TextDecoder('ascii').decode(uint8array);
'ðŸ˜'
> process.version
'v11.0.0'
> 
```

The polyfill repo has a decent explanation + examples: https://github.com/inexorabletash/text-encoding

Otherwise, here's the MDN page for it: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder

Mogztter:
>  if this is targeted to Node.js I think would be less hacky to use TextEncoder, e.g.:

OH! I didn't know that `TextEncoder` and `TextDecoder` are now available natively in Node.js using the `util` module:

```js
const TextEncoder = require('util').TextEncoder
const TextDecoder = require('util').TextDecoder

var uint8array = new TextEncoder().encode("😍")
console.log(new TextDecoder('utf8').decode(uint8array))
```

The support is also pretty good on browsers except Edge but I expect that to change once Edge runs on Chromium (if we want to implement something similar).

I will update my pull request, thanks for the heads-up :+1:

Mogztter:
> Added in: v8.3.0

I don't think it's a problem since we are now supporting only active LTS ?

Mogztter:
The following code will work for non-utf8 files:

```js
const TextEncoder = require('util').TextEncoder
const TextDecoder = require('util').TextDecoder
const textDecoder = new TextDecoder()
const utf8Encoder = new TextEncoder('utf-8')

// iso-8859-1 file as utf8
const isobuf = require('fs').readFileSync('./iso88591.txt')
textDecoder.decode(utf8Encoder.encode(isobuf.toString('binary'))) // çéà

// win1258 file as utf8
const winbuf = require('fs').readFileSync('./win1258.txt')
textDecoder.decode(utf8Encoder.encode(winbuf.toString('binary'))) // çéà
```

But it won't work for utf8 encoded files:

```js
// utf8 file as utf8
const utf8text = require('fs').readFileSync('./utf8.txt')
textDecoder.decode(utf8Encoder.encode(utf8text.toString('binary'))) // Ã§Ã©Ã
```
 I was using the `try/catch` block to coerce to utf8 only when needed. I guess we could do that using something like: https://github.com/wayfind/is-utf8/blob/master/is-utf8.js ? But maybe I'm missing something ?

stdlib/nodejs/file.rb
```diff
@@ -26,6 +26,79 @@
     warnings[string] = true;
     #{warn(`string`)};
   }
+  function is_utf8(bytes) {
```
Mogztter: Code climate is unhappy: "File file.rb has 308 lines of code (exceeds 250 allowed). Consider refactoring."
Should I move this function somewhere else ? Do you have a better idea on how to handle both binary and utf8 encoded data without using this method ?

elia: CodeClimate "errors" should be taken as just suggestions, I regularly go and silence that stuff, no worry 👍

PR #1924, opened 2019-01-07
File#expand_path should expand ~ (tilde)

Mogztter:


test/nodejs/test_file.rb
```diff
@@ -153,6 +153,7 @@ def test_windows_separators
   def test_windows_file_expand_path
     drive_letter = Dir.pwd.slice(0, 2)
     assert_equal(Dir.pwd + '/foo/bar.js', File.expand_path('./foo/bar.js'))
+    assert_equal(Dir.home + '/foo/bar.js', File.expand_path('~/foo/bar.js'))
```
Mogztter:
```suggestion
```

elia:
If it's not too complicated can you embed these changes inside #1925 ?

Mogztter:
No not at all.
I actually did this on purpose to not mix a "bug fix" (adding a missing functionality) and a new feature (adding a new method).
But if it's easier for you then I will merge both.

elia:
> But if it's easier for you then I will merge both.

Yeah, thanks, they're related, this way we can have the whole picture

Mogztter:
Superseded by #1925

PR #1925, opened 2019-01-07
Implement File#absolute_path

### LINKED ISSUE #1922: Implement File#absolute_path
`File.absolute_path` is the same as `File.expand_path`, except it doesn't expand `~`.
Speaking of `File.expand_path`, in Node.js `~` should be expanded using `require('os').homedir()`.

Mogztter:
Fixes #1922

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -164,11 +164,6 @@ class File < IO
     `return __path__.normalize(__path__.resolve(#{basedir.to_str}, #{path.to_str})).split(__path__.sep).join(__path__.posix.sep)`
   end
 
-  def self.expand_path(path, basedir = nil)
-    # we should resolve ~ before calling absolute_path! see https://github.com/opal/opal/pull/1924
-    absolute_path(path, basedir)
-  end
-
   # Instance Methods
 
   def initialize(path, flags = 'r')
@@ -247,10 +242,6 @@ class File < IO
 
   def mtime
     `return executeIOAction(function(){return __fs__.statSync(#{@path}).mtime})`
-  end
-
-  def to_path
-    @path
   end
 end
 
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -158,10 +158,15 @@ class File < IO
     `return executeIOAction(function(){return __fs__.lstatSync(#{path}).isSymbolicLink()})`
   end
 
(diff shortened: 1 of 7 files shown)


Mogztter:
@elia Done :+1:

Mogztter:
I've also added `ruby/core/file/expand_path_spec` and there's a few errors.
I've fixed a few and added the rest in `bugs/file.rb`.

I think we should be able to fix a few more (related to the leading separator) but I'm not familiar with the code. I will give it a try though :wink:

stdlib/nodejs/file.rb
```diff
@@ -158,10 +158,15 @@ def self.symlink?(path)
...
-    path = path.to_str
+  def self.absolute_path(path, basedir = nil)
+    path = path.respond_to?(:to_path) ? path.to_path : path
     basedir ||= Dir.pwd
-    `return __path__.normalize(__path__.resolve(#{basedir.to_str}, #{path})).split(__path__.sep).join(__path__.posix.sep)`
+    `return __path__.normalize(__path__.resolve(#{basedir.to_str}, #{path.to_str})).split(__path__.sep).join(__path__.posix.sep)`
+  end
+
+  def self.expand_path(path, basedir = nil)
+    # we should resolve ~ before calling absolute_path! see https://github.com/opal/opal/pull/1924
```
elia: Have you tried removing `expand_path` from the node version and relying on the corelib implementation? the platform-specific code seems to be all inside `absolute_path`

Mogztter: No but that's a good idea 👍

Mogztter:
@elia The `platform_is` and `platform_is_not` methods and associated blocks are not executed (`before :each` is called though):

```rb
describe "File.expand_path" do
  before :each do
    platform_is :windows do # not executed
      @base = `cd`.chomp.tr '\\', '/'
      @tmpdir = "c:/tmp"
      @rootdir = "c:/"
    end

    platform_is_not :windows do # not executed
      @base = Dir.pwd
      @tmpdir = "/tmp"
      @rootdir = "/"
    end
  end
  # ...
end
```

So `@base` is "undefined" that's why some specs are failing. I'm not sure how we should fix this issue since the code is a submodule... :thinking:

Mogztter:
Follow-up pull request to execute `platform_is` and `platform_is_not` blocks: https://github.com/opal/opal/pull/1931

PR #1926, opened 2019-01-07
Fix JRuby build on master

Mogztter:


Mogztter:
@elia bundler issue is gone with this fix but the build is still failing on some tests:


```
Integer#to_f bignum returns self converted to a Float FAILED
Expected 13835069737789293000 to have same value and type as 13835069737789300000
SpecExpectationNotMetError: Expected 13835069737789293000 to have same value and type as 13835069737789300000
```

```
Integer#to_f bignum returns self converted to a Float FAILED
Expected 13835069737789293000 to have same value and type as 13835069737789300000
SpecExpectationNotMetError: Expected 13835069737789293000 to have same value and type as 13835069737789300000
```

iliabylich:
@Mogztter Could you add these 2 specs into `filters/bugs` please?

Mogztter:
@iliabylich Sure, I will do it tomorrow 😀 

> I don't understand why they pass on other rubies.

The value is rounded ? 🤔

.travis.yml
```diff
@@ -70,10 +70,8 @@ matrix:
       env: RUN=rspec
 
     - rvm: jruby-9.2.4.1
-      env: RUN=rspec REINSTALL_BUNDLER=true
```
elia: You should just remove `REINSTALL_BUNDLER=true` here, that's why the failures on jruby popped up. That doesn't explain the error, but explains why we didn't have it before.

.travis.yml
```diff
@@ -70,10 +70,8 @@ matrix:
...
 
     - rvm: jruby-head
-      env: RUN=rspec REINSTALL_BUNDLER=true
```
elia: same as above

.travis.yml
Mogztter: Ooops, nice catch :flushed:

.travis.yml
Mogztter: fixed

.travis.yml
Mogztter: fixed

Mogztter:
Build is green again :smile:

PR #1928, opened 2019-01-08
Resolves #1923 Implement ENV.key? in Node.js

### LINKED ISSUE #1923: ENV missing key? method in nodejs module
In the nodejs module for Opal, the global ENV object is missing the key? method.

The following code throws an error, though it should work:

```ruby
require 'nodejs'

ENV.key? 'HOME'
```

Mogztter:
Do not merge, I will add tests (and probably other methods like `keys`, `length`, `has_key?`... i an another pull request)


elia:
@Mogztter good! 👍 

Here's a list of all the methods defined on ENV on MRI. At this point I think it makes sense to extract ENV in its own file and leave it out of the default Opal corelib list.

Mogztter:
> Here's a list of all the methods defined on ENV on MRI.

:eyes: 

> At this point I think it makes sense to extract ENV in its own file and leave it out of the default Opal corelib list.

I've moved all methods to a file named `env.rb` but I think we should require this file in `stdlib/nodejs.rb` (to be backward compatible).
Also the task `bundle exec rake dist` only generates `nodejs.js`.

Mogztter:
> Go for it, actually in Node.js the bundle size it's not a big problem (as it is for browsers), so we should err on the side of more stuff.

Indeed size does not matter that much in Node.js but we still have to be careful because the initialization might take longer.

> Feel free to tweak it as you see fit, btw I think we should better/officially support node & webpack, moving some of the more specific stuff to a separate repo (e.g. node_modules), but that's not for this PR of course.

That would be great, and we could maybe merge with https://github.com/mogztter/opal-node-runtime to produce a package on npmjs.

I don't want to implement all the methods available on MRI for now. The main reason is that I want to keep the implementation as simple as that and arguably I've already implemented the most common methods :grinning: 

If you think we should implement a few more or all, please just let me know.

test/nodejs/test_env.rb
```diff
@@ -0,0 +1,62 @@
...
+  def test_keys
+    ENV['foo'] = 'bar'
+    assert_includes(ENV.keys, 'foo')
+    ENV.delete 'foo'
+  end
+
+  def test_to_s
+    assert_equal('ENV', ENV.to_s)
+  end
+end
```
elia: Missing newline here

stdlib/nodejs/env.rb
```diff
@@ -0,0 +1,46 @@
...
+  %x{
+    var value = process.env[#{name}] || nil;
+    delete process.env[#{name}];
+    return value;
+  }
+end
+
+def ENV.to_s
+  'ENV'
+end
```
elia: The whole file can be changed to use `class << ENV`, I think that would be more clean:

```rb
class << ENV
  def [](name)
    `process.env[#{name}] || nil`
  end
  # …
end
```

It also makes easier to do `alias` (e.g. for `key?` / `has_key?`)

Mogztter: This is the syntax I was looking for, thanks :100:

Mogztter: Done

elia:
@Mogztter can you squash the commits (keeping "Resolves #1923" in the commit description)?
Apart from that it's ready to be merged.

Mogztter:
@elia Done, I've also added a changelog in `UNRELEASED.md`.

PR #1930, opened 2019-01-10
Move Kernel#format (and sprintf alias) to a dedicated file

### LINKED ISSUE #1927: Move Kernel#format to a dedicated module (reduce runtime size)
This method does not depend on other method and is quite huge (500 lines of code for 21 KB).
So if an application does not use `format` or `String#%` then an advanced user could save some bits by not requiring this module.

The idea is to still require this module in `base.rb`:

```rb
require 'corelib/kernel'
require 'corelib/kernel/format'
```

Mogztter:
Reduce compiled size (if you don't need this module)

Resolves #1927


Mogztter:
I don't think the Appveyor failure is related:

```
1) Failure:
TestBenchmark#test_realtime_output [Assertion: Expected 1 to be < 0.9875479999999999.]:
Expected 1 to be < 0.9875479999999999.
125 runs, 549 assertions, 1 failures, 0 errors, 13 skips
You have skipped tests. Run with --verbose for details.
```

And Codeclimate complains about the size of the `format.rb` file and the size of the `format` method :laughing:

opal/opal/base.rb
```diff
@@ -4,6 +4,7 @@
 require 'corelib/class'
 require 'corelib/basic_object'
 require 'corelib/kernel'
+require 'corelib/kernel/format'
```
elia: The require should go into `opal.rb`, `opal/base.rb` is meant to require the bare minimum for opal to function.

UNRELEASED.md
```diff
@@ -84,6 +84,7 @@ Whitespace conventions:
 - Exceptions raised during the compilation now add to the backtrace the current location of the opal file if available (#1814).
 - Better use of `displayName` on functions and methods and more readable temp variable names (#1910)
 - Source-maps are now inlined and already contain sources, incredibly more stable and precise (#1856)
+- `Kernel#format` (and `sprintf` alias) are now in a dedicated module but still required by default (#1930)
```
elia: ❤️

opal/opal/base.rb
Mogztter: But then it's a breaking change, since `Kernel#format` and `Kernel#sprintf` were part of `corelib/kernel` (which is required in `opal/base.rb`).

Mogztter:
> Can you also reformat the commit message to look like this?

Sure I will do it tomorrow :wink:

opal/opal/base.rb
meh: Seems like an acceptable breaking change to me.

Mogztter: Ok done :+1:

UNRELEASED.md
Mogztter: I've updated the changelog to explain the breaking change

Mogztter:
@elia this pull request should be OK. I've addressed all the comments. I will force push to trigger another build :wink:

PR #1931, opened 2019-01-14
Implement PlatformGuard for mspec

Mogztter:
Uses the environment variable `OPAL_PLATFORM_NAME` and `OPAL_POINTER_SIZE`.


spec/support/guard_platform.rb
```diff
@@ -0,0 +1,83 @@
+class PlatformGuard
```
iliabylich: I guess the code has been taken from MSpec - https://github.com/ruby/mspec/blob/master/lib/mspec/guards/platform.rb. Did you make any changes to make it working? If so could you add a comment for each patch please?

How many changes did you make? Is it possible to require original implementation here and monkey-patch it on the fly?

Also I've noticed that in MSpec it inherits from the `SpecGuard`, but here it's a subclass of `Object`. Is there any reason for doing that?

Mogztter: You're correct it's a simplified version of this file.

> Did you make any changes to make it working?

Yes, I've removed unused methods and I'm using environment variable instead of `RbConfig`.

>  If so could you add a comment for each patch please?

Sure

> How many changes did you make? Is it possible to require original implementation here and monkey-patch it on the fly?

Maybe but I believe some parts are "static", for instance the following code will be executed right away ?

```rb
  POINTER_SIZE = begin
    require 'rbconfig/sizeof'
    RbConfig::SIZEOF["void*"] * 8
  rescue LoadError
    WORD_SIZE
  end
```

> Also I've noticed that in MSpec it inherits from the SpecGuard, but here it's a subclass of Object. Is there any reason for doing that?

`SpecGuard` was used in the orignal implementation but it's not needed here.
In the original implementation, `run_if` and `run_unless` are called but in this simplified implementation I just do everything in this file.

iliabylich: > the following code will be executed right away

Yes, but I think we can handle `LoadError` like MRI would do and fallback to `WORD_SIZE` (that's what you do manually, right?)

spec/filters/bugs/language.rb
```diff
@@ -148,6 +148,15 @@
   fails "Magic comments in the main file do not cause bytes to be mangled by passing them through the wrong encoding" # NoMethodError: undefined method `tmp' for #<MSpecEnv:0x8aa2e>
   fails "Magic comments in the main file must be at the first line" # NoMethodError: undefined method `tmp' for #<MSpecEnv:0x8aa2e>
   fails "Magic comments in the main file must be the first token of the line" # NoMethodError: undefined method `tmp' for #<MSpecEnv:0x8aa2e>
+  fails "Magic comments in an -e argument can take Emacs style" # ArgumentError: unknown encoding name - locale
```
Mogztter: I've added a few more specs in `filters/bugs` because more specs are now executed. 
For instance the following spec were previously ignored:

```rb
  platform_is_not wordsize: 32 do
    it "raises a RangeError when passed a big negative value" do
      lambda { 42.round(fixnum_min) }.should raise_error(RangeError)
    end
  end
```

```rb
  platform_is wordsize: 64 do
    describe "Array#pack with format 'L' with modifier '_'" do
      it_behaves_like :array_pack_64bit_be, 'L_'
    end

    describe "Array#pack with format 'L' with modifier '!'" do
      it_behaves_like :array_pack_64bit_be, 'L!'
    end

    describe "Array#pack with format 'l' with modifier '_'" do
      it_behaves_like :array_pack_64bit_be, 'l_'
    end

    describe "Array#pack with format 'l' with modifier '!'" do
      it_behaves_like :array_pack_64bit_be, 'l!'
    end
  end
```

spec/support/guard_platform.rb
Mogztter: Yes, I will try to remove as much code as I can :+1:

Mogztter: It seems to be working if I just redefine `HOST_OS` and `POINTER_SIZE`

tasks/testing/mspec_special_calls.rb
```diff
@@ -86,5 +70,5 @@ def string_value(token)
   end
 end
 
-Dir[File.expand_path('../../../spec/filters/unsupported/**/*.rb', __FILE__)].each { |f| require f }
-Dir[File.expand_path('../../../spec/filters/bugs/**/*.rb', __FILE__)].each { |f| require f }
+Dir[File.expand_path('../../../spec/filters/unsupported/**/*.rb', __FILE__)].each {|f| require f}
+Dir[File.expand_path('../../../spec/filters/bugs/**/*.rb', __FILE__)].each {|f| require f}
```
elia: Can you remove this change?

Mogztter: Sure

Mogztter:
```suggestion
Dir[File.expand_path('../../../spec/filters/bugs/**/*.rb', __FILE__)].each { |f| require f }
```

tasks/testing/mspec_special_calls.rb
```diff
@@ -86,5 +70,5 @@ def string_value(token)
   end
 end
 
-Dir[File.expand_path('../../../spec/filters/unsupported/**/*.rb', __FILE__)].each { |f| require f }
-Dir[File.expand_path('../../../spec/filters/bugs/**/*.rb', __FILE__)].each { |f| require f }
+Dir[File.expand_path('../../../spec/filters/unsupported/**/*.rb', __FILE__)].each {|f| require f}
```
Mogztter:
```suggestion
Dir[File.expand_path('../../../spec/filters/unsupported/**/*.rb', __FILE__)].each { |f| require f }
```

Mogztter:
@elia I've also rebased on master and added a few more specs to `bugs/file.rb`. The reason is that `Dir.home` with an argument is not implemented. As far as I know there's no API in Node.js to determine the home directory of a given user. What we could do is test if the user argument is equals to the current user? That would cover another case but otherwise we should raise a "not implemented" error.
Anyway I don't want to implement this solution in this pull request.

Mogztter:
@elia I've ingored Windows specific failures in 7acbb581

Mogztter:
Rebased on master.

**master**
```
1504 files, 13266 examples, 54666 expectations, 0 failures, 0 errors, 2100 tagged
11273 examples, 0 failures (time taken: 11.33400011062622)
```

**patch**
```
1503 files, 13602 examples, 54892 expectations, 0 failures, 0 errors, 2188 tagged
11521 examples, 0 failures (time taken: 11.473000049591064)
```

elia:
@Mogztter can you rebase/fixup one more time please? After that I think we're good 👍

Mogztter:
@elia Branch rebased. I didn't squash the commit to preserve the authors but maybe we could squash the first two and the last two.

elia:
> @elia Branch rebased. I didn't squash the commit to preserve the authors but maybe we could squash the first two and the last two.

@Mogztter I'm fine, go ahead and squash it (I usually do it with `git rebase -i --autosquash origin/master`), I have enough commits on this repo that I can spare one 😄

PR #1939, opened 2019-01-30
Replace eval() calls with Function()

elia:
As suggested in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval.


iliabylich:
From I understand `Function(code)` works like `TOPLEVEL_BINDING.eval(code)`, i.e. it prevents potential usage of local variables. Is that correct? 👍 if so

stdlib/opal-parser.rb
```diff
@@ -12,9 +12,11 @@ def eval(str)
     compiling_options = __OPAL_COMPILER_CONFIG__.merge(default_eval_options)
     code = Opal.compile str, compiling_options
     %x{
-      return (function(self) {
-        return eval(#{code});
-      })(self)
+      return (
+        Function(
```
hmdne, line 16: You can add parameters to Function by declaring them in the first argument like so: "self,other,anything".

janbiedermann:
esbuild prints warnings about eval() when bundling opal code. They suggest Function too. They suggest 2 alternatives:

https://esbuild.github.io/content-types/#direct-eval

PR #1941, opened 2019-02-11
gmt_offset (alias utc_offset) should return 0 if the date is utc

### LINKED ISSUE #1940: Time.utc should set utc_offset to 0
#### Given

Using a UTC+1 locale:
```rb
source_date_epoch = Time.at 1549743934
p source_date_epoch.utc_offset()
p source_date_epoch.utc().utc_offset()
```
#### MRI

```
3600
0
```

#### Opal

```
3600
3600
```

Mogztter:
resolves #1940

### DIFF
--- a/opal/corelib/time.rb
+++ b/opal/corelib/time.rb
@@ -427,7 +427,7 @@ class Time < `Date`
   end
 
   def gmt_offset
-    `-self.getTimezoneOffset() * 60`
+    `self.is_utc ? 0 : -self.getTimezoneOffset() * 60`
   end
 
   def strftime(format)
--- a/spec/opal/core/time_spec.rb
+++ b/spec/opal/core/time_spec.rb
@@ -59,4 +59,10 @@ describe Time do
       end
     end
   end
+
+  describe '#utc_offset' do
+    context 'returns 0 if the date is UTC' do
+      Time.new.utc.utc_offset.should == 0
+    end
+  end
 end


elia:
@Mogztter is there any rubyspec we can enable along with the fix?

Mogztter:
We could enable https://github.com/ruby/spec/blob/master/core/time/shared/gmt_offset.rb but I'm not sure `with_timezone` will work in Opal (and all tests will fail).
I've only fixed the offset when the date is UTC. When the date is not UTC and has a different timezone than the local date then `gmt_offset` is still wrong.

So I guess we could add our own spec, I believe an UTC date should return 0 as `utc_offset`.

We could also add a spec in https://github.com/ruby/spec/blob/master/core/time/shared/gmt_offset.rb to add:

```rb
Time.new.utc.send(@method).should == 0
```

elia:
> I've only fixed the offset when the date is UTC. When the date is not UTC and has a different timezone than the local date then `gmt_offset` is still wrong.

Have you seen [`getUTCHours`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCHours)? looks like it does the right thing

> So I guess we could add our own spec

That's acceptable 👍 

> We could also add a spec…

That would begreat 👍👍👍

Mogztter:
> Have you seen [`getUTCHours`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCHours)? looks like it does the right thing

Looks promising but it does not exactly what we want.

```js
var date1 = new Date('December 31, 1975, 23:15:30 GMT+11:00');
var date2 = new Date('December 31, 1975, 23:15:30 GMT-11:00');

console.log(date1.getUTCHours());
// expected output: 12

console.log(date2.getUTCHours());
// expected output: 10
```
In our case we want the UTC offset:
* `date1` UTC offset is equal to `+11`
* `date2` UTC offset is equal to  `-11`

We could do something like:

```js
console.log(date1.getHours() - date1.getUTCHours());
// expected output: 11

console.log(date2.getHours() - date2.getUTCHours());
// expected output: 13
```

But as you can see the second example is wrong because the offset is negative, but we don't know that... we could then check if the `getUTCDate` is the same as `getDate`, if not that would indicate that the date changed:


```js
console.log(date1.getHours() - date1.getUTCHours());
// expected output: 11

if (date2.getDate() !== date2.getUTCDate()) {
  console.log(-((24 - date2.getHours()) + date2.getUTCHours()));
  // expected output: -11
}
```

But we should also make sure that it works when the date decreased, for instance when the date is `December 31, 1975, 09:15:30 GMT+11:00`.
So it's not pretty but potentially not wrong :wink: 

Hopefully tests will save us all :laughing:

Mogztter:
It's not working because `getHours` is not the date hours but the local date hours. For instance I'm currently UTC+2:

```js
new Date('2012/04/10 10:10:30 +0300').getUTCHours()
// 7

new Date('2012/04/10 10:10:30 +0300').getHours()
// 9
```

I would expect `getHours` to return `10` but since I'm UTC+2, the local date hours is `10 + 2 - 3 = 9`.

Mogztter:
I've added a spec in Opal to make sure that an UTC date returns 0 as UTC offset: https://github.com/ruby/spec/pull/660

I think we should merge this quick fix and open a new issue to properly handle timezones.

PR #1944, opened 2019-02-21
Implement String#each_codepoint and String#codepoints

### LINKED ISSUE #1943: Implement String#each_codepoint
https://ruby-doc.org/core-2.5.3/String.html#method-i-each_codepoint
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/codePointAt

Mogztter:
Probably a naive implementation but I was able to enable quite a few specs :wink:  

Resolves #1943

//cc @mojavelinux

### DIFF
--- a/opal/corelib/string/encoding.rb
+++ b/opal/corelib/string/encoding.rb
@@ -181,6 +181,23 @@ class String
     self
   end
 
+  def each_codepoint(&block)
+    return enum_for :each_codepoint unless block_given?
+    %x{
+      for (var i = 0, length = self.length; i < length; i++) {
+        var code = self.codePointAt(i);
+        #{yield `code`};
+      }
+    }
+    self
+  end
+
+  def codepoints(&block)
+    # If a block is given, which is a deprecated form, works the same as each_codepoint.
+    return each_codepoint(&block) if block_given?
+    each_codepoint.to_a
+  end
+
   def encode(encoding)
     dup.force_encoding(encoding)
   end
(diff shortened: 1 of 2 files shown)


opal/corelib/string/encoding.rb
```diff
@@ -181,6 +181,23 @@ def each_byte(&block)
     self
   end
 
+  def each_codepoint(&block)
+    return enum_for :each_codepoint unless block_given?
+    %x{
+      for (var i = 0, length = self.length; i < length; i++) {
+        var code = self.codePointAt(i);
```
mojavelinux: The local variable assignment isn't necessary here.

Mogztter: you mean we should `yield` the result of `codePointAt` ?

```
#{yield `self.codePointAt(i)`}
```

mojavelinux: Exactly.

elia:
I accidentally merged this one from an iOS  application that made it all too easy 😅. I’ll leave it merged, the outstanding change was a small one and I can fix it directly on master 👍

Thanks for the steady flow of PRs and fixes @mogztter!! 💪🙏

Mogztter:
> I accidentally merged this one from an iOS  application that made it all too easy sweat_smile. 

:ship: :ship: :ship: 

> Thanks for the steady flow of PRs and fixes @Mogztter!! 

You're welcome :hugs: 

> I’ll leave it merged, the outstanding change was a small one and I can fix it directly on master

I'll take care of it :wink:

PR #1949, opened 2019-03-06
Opal runtime is strict

Mogztter:
#### Tasks

- [x] `Uncaught TypeError: Cannot assign to read only property 'length' of function 'function() ...`
- [ ] `Uncaught SyntaxError: Duplicate parameter name not allowed in this context`
The following code is not strict-compliant:

```js
function hello(_, _) {
  console.log('hello')
}
```
Exception thrown by `spec/ruby/language/fixtures/super.rb` when the module `ZSuperWithUnderscores` defines the following methods:

```rb
def m(_, _)
  super
end

def m_modified(_, _)
  _ = 14
  super
end
```

I don't know exactly what the code is supposed to do... does it make sense in Ruby to do something like this ?
We could fix this issue by making sure that the generated code does not contain two or more parameters with the same name (ie. for instance by adding a counter to the parameter name).

EDIT: in `m_modified`, the value 14 will be assigned to the first parameter. So the code is equivalent to:

```rb
def m_modified(_1, _2)
  _1 = 14
  super
end
```
EDIT2: the current implementation is not working :disappointed: 

* [ ] `undefined : Cannot create property 'encoding' on string ')'`
* [ ] `TypeError: Cannot create property '$$meta' on string 'abc'`
```
$defineProperty
Object.Opal.build_object_singleton_class
```

### DIFF
--- a/lib/opal/cli_runners/chrome.js
+++ b/lib/opal/cli_runners/chrome.js
@@ -91,7 +91,7 @@ CDP(options, function(client) {
     Runtime.exceptionThrown(function(exception) {
       var exceptionDetails = exception.exceptionDetails,
           properties = exceptionDetails.exception.preview.properties,
-          stackTrace = exceptionDetails.stackTrace.callFrames,
+          stackTrace = exceptionDetails.stackTrace ? exceptionDetails.stackTrace.callFrames : [],
           name, message, trace = [], i;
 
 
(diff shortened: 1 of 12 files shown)


Mogztter:
The main challenge is probably to find a way to define properties on string primitive. Technically creating a new String object and using `Object.defineProperty` is working but:

- `$defineProperty` in `runtime.js` is expected to add a property on the instance passed as argument
- Performance wise it's likely a bad idea

Mogztter:
Superseded by #1952

PR #1951, opened 2019-03-09
Make ascii_only a little less wrong

Mogztter:
While working on #1949 I found an issue in the `ascii_only?` implementation.
This implementation is a little less wrong but still failing on:

```rb
"hello".force_encoding("UTF-16LE").ascii_only? # should be false
```

The main issue is that the default encoding in JavaScript is `UTF_16LE` and in Ruby this encoding is considered "non-ASCII-compatible".
Unfortunately Opal can't tell the difference between:

- `"hello".force_encoding("UTF-16LE")`
- `"hello"`

So we cannot return false in the first case and true in the second case... send help!

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -691,7 +691,18 @@ class String < `String`
   end
 
   def ascii_only?
-    `self.match(/[ -~\n]*/)[0] === self`
+    # non-ASCII-compatible encoding must return false
+    # NOTE: Encoding::UTF_16LE is also non-ASCII-compatible encoding,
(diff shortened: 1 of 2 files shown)


Mogztter:
Also I don't get why Ruby returns false in the following case:

```rb
"".force_encoding("UTF-16LE").ascii_only? # false
```

It feels wrong no ?
I mean it does not contain any non ASCII (it not not even contain anything).
Even the above is weird:

```rb
"hello".force_encoding("UTF-16LE").ascii_only? # false
```

"hello" is ASCII only, the encoding might not be "non-ASCII-compatible" but that's not really what we asked no ?

PR #1954, opened 2019-03-11
this is not used (and will be undefined in strict mode)

Mogztter:


Mogztter:
Actually I might be wrong, because `this` in the closure must resolve to the global scope `window`...

PR #1959, opened 2019-03-19
Use this only if the scope is null (not falsy)

### LINKED ISSUE #1958: 0.times.size returns null in strict mode
Opal generates the following code for `times`:

```js
return $send(self, 'enum_for', ['times'], ($$61 = function () {
  var self = $$61.$$s || this

  if ($$61.$$is_lambda || $$61.$$define_meth) {
    var $arity = arguments.length
    if ($arity !== 0) { Opal.block_ac($arity, 0, 'times') }
  }
  return self
}, $$61.$$s = self, $$61.$$arity = 0, $$61.$$parameters = [], $$61))
```

The problem is that `$$61.$$s` is falsy (the value is 0) so we are using `this` instead. But `this` in strict mode is `null`.

Thus, the following returns null instead of 0:

```js
(0).$times().$size() // returns null
```

Mogztter:
Resolves #1958


tasks/testing.rake
```diff
@@ -335,7 +336,7 @@ platforms.each do |platform|
         includes = "-Itest -Ilib -Ivendored-minitest #{includes}"
 
         sh "ruby -rbundler/setup "\
-         "exe/opal #{includes} #{stubs} -R#{platform} -Dwarning -A --enable-source-location #{filename}"
+         "exe/opal #{includes} #{stubs} -R#{platform} -Dwarning -A --enable-strict-mode --enable-source-location #{filename}"
```
Mogztter: Enable strict mode in minitest

lib/opal/nodes/iter.rb
```diff
@@ -17,7 +17,7 @@ def compile
 
         in_scope do
           identity = scope.identify!
-          add_temp "self = #{identity}.$$s || this"
+          add_temp "self = #{identity}.$$s == null ? this : #{identity}.$$s"
```
Mogztter: The actual fix

test/opal/test_strict.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: false
+require "test/unit"
+
+class TestStrict < Test::Unit::TestCase
+  def test_strict_mode
+    assert_equal(0, 0.times.size)
```
Mogztter: If you revert the oneline fix in `lib/opal/nodes/iter.rb`, `minitest` should fail.

lib/opal/cli_options.rb
```diff
@@ -160,6 +160,10 @@ def initialize
         options[:enable_source_location] = true
       end
 
+      on('--enable-strict-mode', '--no-method-missing', 'Enable/Disable method missing') do
```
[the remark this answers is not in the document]
Mogztter: Fixed

test/opal/test_strict.rb
```diff
@@ -0,0 +1,8 @@
...
+  end
+end
```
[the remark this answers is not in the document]
Mogztter: Fixed

opal/corelib/strict.js
```diff
@@ -0,0 +1 @@
+'use strict';
```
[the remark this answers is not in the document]
Mogztter: Fixed

opal/corelib/string/encoding.rb
```diff
@@ -210,6 +210,10 @@ def force_encoding(encoding)
 
       if (encoding === self.encoding) { return self; }
 
+      if (typeof self === 'string') {
```
Mogztter: Same here but I'm not sure about this fix.
The error is `TypeError: Cannot create property 'encoding' on string ')'`.

Mogztter:
@elia I've tried to do as little change as possible but I think we have to address the `String#force_encoding` first.
I'm pretty sure my hack is not accurate but at the same time encoding in Ruby and JavaScript does not work in the same way and I can't think of any other solution to this problem...

elia:
Merged (except for the CLI option) in #2023

PR #1970, opened 2019-05-07
ES6 modules

janbiedermann:
Final PR for ES6 modules and strict mode, closing PR#1832 and PR#1969


janbiedermann:
there is a misbehaviour happening with the strictnodejs runner, executing loaded modules not in strict mode, fix is on the way

janbiedermann:
What? Codeclimate shows Cognitive Complexity of 18? for a few if's? How come?

janbiedermann:
Cognitive Complexity still at 12, for basically 3 if's, i don't know how to fix. How to fix?

janbiedermann:
I am happy with this so far.
The only way i see to support ruby String semantics in "use strict" mode is to use new String instead of primitives, see PR#1973.

janbiedermann:
Oh my, how is this:
```
Opal::Compiler.module_name_from_paths(module_path.sub(/(?<!\.rb)\z/, '.rb'))
```
better than this:
```
module_path = module_path + '.rb' unless module_path.end_with?('.rb')
Opal::Compiler.module_name_from_paths(module_path)
```
in terms of cognitive complexity? Who will understand `/(?<!\.rb)\z/` at first glance?

janbiedermann:
i accidentially opened a can of worms. how come suddenly there are more problems now?

Mogztter:
I think we should split this pull request to ease the review.

In my opinion, the first thing to do is to make the Opal runtime strict compliant `opal/corelib/runtime.js` (or at least provide an option to make it strict compliant). 
Then we should make sure that the code produced by Opal can be strict compliant.
And then add ES6 modules.

What do you think ?

janbiedermann:
I think i should explain more my thinking and process.
By employing webpack more and more and using its features and learning about it, webpack is very cool and may even help with DCE (but also maybe not, not sure yet), it became more and more clear, that it is also the most challanging platform to get things running correctly.
Webpack is rewriting and inserting code a lot.

So my reasoning to get strict mode right, is to get it running in webpack right, basically using webpack as test platform for opal strict mode.

Now the code here in 1970 implements towards the goal:
- ES6 modules, to get things running in webpack in a very modular fashion, to employ hot reloading etc.
- strictnodejs, a opal runner which wraps everything in "use strict", intended for testing
- several strict patches from several PRs
- and (upcoming) a runner, which runs tests in a webpack bundle build by owl (upcoming 0.8.0) in a browser

and then use that to get `use strict` mode and `new String()` right for all environments (node, webpack in browser, webpack npm bundle of opal (for node and browser).

Having said that, either way is fine for me.

janbiedermann:
A Note:
Another thing webpack does, is that each import has its own context, that's what i suspect may cause problems. Its handled in a way, because imports use `global.Opal`, but there may be cases where in code somewhere just `Opal` is used and may not be resolved properly.

janbiedermann:
This branch stays on 1.0 to provide a stable base users of opal-webpack-loader.
Work on this topic for 1.1 continues in #1976.

janbiedermann:
probably obsolete

PR #1976, opened 2019-05-14
ES6 modules and strict mode for 1.1

janbiedermann:
2790e59 + faa60d4 + e826532 may go away once everything is running fine in strict and non strict mode.
I suggest making strict mode the default then for generated code and the runtime.js, so no strictnodejs runner and no --use-strict compiler option are needed anymore.


janbiedermann:
just noticed an issue:
when using --es6-modules with opal, like: `bundle exec opal --es6-modules -OcEe 'require 'socket'; puts "test"'`

then the compiler can't determine the module correctly and the import for socket, in above example, is not generated. result:
```
const opal_code = function() {
  global.Opal.modules["-e"] = function(Opal) {
```
not the "-e" as module name. Of course, some context is missing. maybe needs an additional option for module name or the compiler should use a default module name.

janbiedermann:
I think its best to remove the --es6-modules command line option from the cli. Its generated code makes in fact only sense if used from the loader within webpack. When using this option from the cli on a file, muliple export default lines are generated, useless.

janbiedermann:
In progress getting the tests running, compiled as ES6 modules, several challenges:
in erb_spec: `require 'bla'` where bla resolves to a file bla.opalerb, this can't currently be handled by the webpack resolver. Will need to fix that.
`require_tree '../bla'` also makes problems, need to get the paths for files right. Above fix in f805284 is partial.

janbiedermann:
With the relative require fixed, a new challenge:
Conditional requires don't work as expected. Webpack resolves ALL requires at COMPILE time and wants to include corresponding files in the bundle, for example the nodejs files for a browser bundle. It also evaluates the nodejs require in sdtlib/nodejs/io.rb. And so on, etc., pp. 
opal/platform.rb is the problem. But that file is required by mspec. Thinking.

janbiedermann:
owl 0.8.1 (not released yet) will have a filter, that allows to exclude opal/platform.rb (and others too) from resolving so that a substitute can be supplied. That almost makes mspec work. Still a problem with require_tree.
Also the dynamic require_spec_file doesn't work. With webpack its either import or don't import.

janbiedermann:
I proudly present:
The first mspec test run, bundled by webpack with owl 0.8.1, minified, executed in chrome:
```
core:opal jan$ rake mspec_opal_owl
mkdir -p tmp/webpack_app/opal
Adding 103 files (spec/opal)
export RANDOM_SEED=53120 # to re-use the same randomization
stub: mspec/helpers/tmp
stub: mspec/helpers/environment
stub: mspec/guards/block_device
stub: a_file
stub: lib/spec_helper
stub: mspec/commands/mspec-run
stub: etc
env -i PATH=$PATH ruby -w -rbundler/setup -r/Users/jan/workspace/opal/tasks/testing/mspec_special_calls runner.rb
/Users/jan/.rvm/rubies/ruby-2.6.3/lib/ruby/gems/2.6.0/gems/parser-2.5.3.0/lib/parser/lexer.rb:10836: warning: assigned but unused variable - testEof

yarn exec v1.15.2

Hash: 97d2f2b09686de81dfa7
Version: webpack 4.31.0
Child
    Hash: 97d2f2b09686de81dfa7
    Time: 28440ms
    Built at: 05/17/2019 1:12:35 AM
             Asset      Size  Chunks             Chunk Names
    application.js  2.01 MiB       0  [emitted]  application
    Entrypoint application = application.js
      [3] /Users/jan/workspace/opal/spec/spec_helper.rb 4.45 KiB {0} [built]
     [30] /Users/jan/workspace/opal/opal/opal.rb 4.79 KiB {0} [built]
     [31] /Users/jan/workspace/opal/opal/opal/base.rb 2.27 KiB {0} [built]
     [35] ./opal/platform.rb 879 bytes {0} [built]
     [84] /Users/jan/workspace/opal/opal/opal/mini.rb 4.46 KiB {0} [built]
     [95] /Users/jan/workspace/opal/opal/corelib/kernel/format.rb 20.6 KiB {0} [built]
     [96] /Users/jan/workspace/opal/opal/corelib/string/encoding.rb 14.7 KiB {0} [built]
     [97] /Users/jan/workspace/opal/opal/corelib/math.rb 16.1 KiB {0} [built]
     [98] /Users/jan/workspace/opal/opal/corelib/complex.rb 19.5 KiB {0} [built]
     [99] /Users/jan/workspace/opal/opal/corelib/rational.rb 17.9 KiB {0} [built]
    [100] /Users/jan/workspace/opal/opal/corelib/time.rb 27.1 KiB {0} [built]
    [110] ./spec_owl.rb 51.5 KiB {0} [built]
    [449] ./opal_loader.rb 781 bytes {0} [built]
    [450] multi ../javascripts/application.js 28 bytes {0} [built]
    [451] ../javascripts/application.js 181 bytes {0} [built]
        + 442 hidden modules
Done in 31.38s.
Connecting to localhost:9222...
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --headless --disable-web-security --disable-gpu --remote-debugging-port=9222 
Connecting to localhost:9222...
[0517/011235.995639:ERROR:xattr.cc(63)] setxattr org.chromium.crashpad.database.initialized on file /var/folders/89/9yvnz1nn1431fgggqpfbyp200000gn/T/: Operation not permitted (1)
[0517/011235.998311:ERROR:file_io.cc(89)] ReadExactly: expected 8, observed 0
[0517/011235.999766:ERROR:xattr.cc(63)] setxattr org.chromium.crashpad.database.initialized on file /var/folders/89/9yvnz1nn1431fgggqpfbyp200000gn/T/: Operation not permitted (1)
Connecting to localhost:9222...

DevTools listening on ws://127.0.0.1:9222/devtools/browser/4541d432-fc87-49c6-a5c9-0b8569bb37b0
Connecting to localhost:9222...
GVA info: Successfully connected to the Intel plugin, offline Gen9 
..........................................................................................................................................................................................................................................................................................................................................................................................................opal 1.0.0 (2019-05-12 revision 0)
.........................................................................................................................................................FF........F..EE..E....EE.........................................................................................................................................................................................................................

[ ... error messages cut ... ]

Finished in 0.578000 seconds

103 files, 788 examples, 1577 expectations, 3 failures, 5 errors, 0 tagged
```
(The webpack build time of 28s is due to minification, without its just 7225ms.)

janbiedermann:
first result for mspec_ruby_owl:
```
1504 files, 32221 examples, 108503 expectations, 259 failures, 175 errors, 9869 tagged
```
😄 

(Need to fix some things before i push that.)

janbiedermann:
Note:
when building with source maps it fails with this message:
Error: Invalid mapping: {"generated":{"line":182,"column":0},"source":"/Users/jan/.rvm/rubies/ruby-2.6.3/lib/ruby/gems/2.6.0/gems/ast-2.4.0/lib/ast/node.rb","original":{"line":0,"column":0},"name":null}
(will fix later on)

janbiedermann:
Made the test filtering work with owl. Results:
-
mspec_opal_chrome: `103 files, 397 examples, 797 expectations, 0 failures, 0 errors, 0 tagged`
mspec_opal_owl: `103 files, 393 examples, 793 expectations, 0 failures, 0 errors, 0 tagged` (excluded erb_spec.rb, as owl currently doesn't support .opalerb)
-
mspec_ruby_chrome: `1504 files, 13679 examples, 55167 expectations, 0 failures, 0 errors, 2189 tagged`
mspec_ruby_owl: `1504 files, 13679 examples, 54987 expectations, 132 failures, 87 errors, 2189 tagged`
-
So 132 failures, 87 errors for ruby specs when bundled with webpack and run in chrome vs. just being run in chrome without being bundled in webpack.

janbiedermann:
Minitest results:
```
minitest_cruby_chrome: 125 runs, 549 assertions, 0 failures, 0 errors, 13 skips
minitest_cruby_owl:    125 runs, 542 assertions, 2 failures, 0 errors, 13 skips
```

janbiedermann:
Lazy load ruby code, split up assets using webpack dynamic import(), usage:
```ruby
require_lazy 'module_name' do
   # block here is executed after module has been loaded
end

# with webpack prefetch feature:
require_lazy 'module_name', :prefetch do
   # block here is executed after module has been loaded
end

# with webpack preload feature:
require_lazy 'module_name', :preload do
   # block here is executed after module has been loaded
end

# with webpack prefetch and preload feature:
require_lazy 'module_name', :prefetch, :preload do
   # block here is executed after module has been loaded
end
```

See: https://webpack.js.org/guides/code-splitting#dynamic-imports

Needs specs and testing.

@elia, @Mogztter: is that cool? is that ok?
Ill test and write specs later on.

janbiedermann:
Things to do:
[] fix source maps, they crash chrome and the webpack build process*
[] fix import names: currently they have a random component which confuses webpack checksums. Move `generate import name` to the top_node instance, maybe even all es6 methods, include as module.
[] fix bugs for those failing ruby mspecs.
[] https://github.com/isomorfeus/opal-webpack-loader/issues/5, for minispec_cruby
[] provide additional webpack configs:
- one for mspec, one for minispec, make them use source maps to catch *
- one with webpacks 'library' target, can compile opal for npm as module probably
[] add `require_lazy` specs and make it return a ruby promise

janbiedermann:
Running mspec_ruby_owl seems to crash on travis. I tried to reproduce locally with the exact same node and ruby versions, but it works on my machine without any problem. So for now its in allow failures.

janbiedermann:
Building of node package works, usage for now like this:
**for the browser**
```
import 'opal';
import 'opal/parser'; // if thats wanted
```
**for node:**
require('opal/opal_node'); // includes opal and the node specific things
require(opal/parser_node'); // packaged for node
```
in nodejs:
`> Opal.Parser.$eval('puts File.read("my_file")')` does the right things and prints the files content.

**Comment by janbiedermann** [comment]
node package looks like this:
```
npm notice 📦  opal@1.1.0
npm notice === Tarball Contents === 
npm notice 442B    package.json  
npm notice 6.1kB   .DS_Store     
npm notice 379.2kB opal_node.js  
npm notice 325.4kB opal.js       
npm notice 1.1MB   parser_node.js
npm notice 1.2MB   parser.js     
npm notice === Tarball Details === 
npm notice name:          opal                                    
npm notice version:       1.1.0                                   
npm notice filename:      opal-1.1.0.tgz                          
npm notice package size:  527.6 kB                                
npm notice unpacked size: 3.0 MB 
```
Maybe better to have a separate opal-node package?

**Comment by janbiedermann** [comment]
building the npm package:
```
bundle install
cd npm
npm i
npm run library_build
npm pack
```
and there it is.

**Comment by janbiedermann** [comment]
now with source map support:
```
npm notice 📦  opal@1.1.0
npm notice === Tarball Contents === 
npm notice 442B    package.json      
npm notice 6.1kB   .DS_Store         
npm notice 379.2kB opal_node.js      
npm notice 963.3kB opal_node.js.map  
npm notice 325.4kB opal.js           
npm notice 771.3kB opal.js.map       
npm notice 1.1MB   parser_node.js    
npm notice 2.7MB   parser_node.js.map
npm notice 1.2MB   parser.js         
npm notice 2.8MB   parser.js.map     
npm notice === Tarball Details === 
npm notice name:          opal                                    
npm notice version:       1.1.0                                   
npm notice filename:      opal-1.1.0.tgz                          
npm notice package size:  1.8 MB                                  
npm notice unpacked size: 10.3 MB
``` 
Probably good idea to have 2 packages, one for the browser, one for node?

**Comment by janbiedermann** [comment]
Usage for require_lazy, manually tested and working:
```
require 'promise' # required

require_lazy 'forwardable' do
  if defined? Forwardable
    puts "Yeah, it works!"
  end
end.then do
  puts "And the promise has been resolved too"
end
```
With that code, webpack creates a lazy loaded chunk:
```
03:15:11 webpack.1 |                  Asset      Size       Chunks             Chunk Names
03:15:11 webpack.1 |                   0.js  8.83 KiB            0  [emitted]  <---- this one, 'forwardable' compiled
03:15:11 webpack.1 |         application.js  1.19 MiB  application  [emitted]  application
```
which is then, once the browser hits it, lazy loaded, executed and then the blocks or promises get executed.
Preferred usage is to bundle application code in files that list a lot of requires, and load those files with require_lazy based on user input/application path. This way the initial application size to load can be very nicely reduced. Also the options :prefetch and :preload are available, see links further above, which may make the lazy loading barely noticeable.
No specs yet. Not sure yet how to test properly.

File: lib/opal/nodes/call.rb [hunk 1]
```diff
@@ -273,25 +273,40 @@ def handle_special(&compile_default)
...
-        push fragment '/* webpackPreload: true */' if preload
-        # append '.rb' to the module name so webpack passes the request to the owl resolver and loader
-        push fragment " '#{module_name}.rb'"
-        push fragment ').then(function() {'
-        push expr(iter)
-        push fragment '});'
+        push fragment "(function() {"
+        indent do
+          push line
+          push "var $$promise = $send($$($nesting, 'Promise').$new(), 'then', [],"
```
janbiedermann: not so sure about this line above, the hardcoded `$nesting, 'Promise'`, can i use something from the compiler here?

janbiedermann:
Working on a instance browser/debugger to be used with a soon to be released version of https://github.com/isomorfeus/opal-devtools. Instance registry can be activated with webpack var injection. Example config for webpack:
```
...
    plugins: [
        new webpack.DefinePlugin({
            OPAL_DEVTOOLS_OBJECT_REGISTRY: true
        })
    ],
...
```
Setting OPAL_DEVTOOLS_OBJECT_REGISTRY to `true` enables the registry.
If its set to `false` or the variable is not injected, everything behaves like before.

janbiedermann:
This patch decouples require_tree module inclusion from runtime evalaution by adding a option to require tree:
```
require_tree 'mod_dir', :autoload
```
will on Object put $$autoload for constants to look up, constant name for a file:
`mod_dir/some_thing.rb`
will be:
`ModDir::SomeThing`

for 
```
require_tree 'mod_dir', :autoload_dirskip
```
will on Object put $$autoload for constants to look up, constant name for a file:
`mod_dir/some_thing.rb`
will be:
`SomeThing`

janbiedermann:
Its mostly a functional proof of concept for now, so a small todo for this patch:
- polish
- maybe inflectors for MyConstAPI and so on.

janbiedermann:
To clarify: require_lazy needs a relative file path as option.
There was a bug where it did not determine the module path correctly from the file path, that is now fixed.

lib/opal/nodes/top.rb
```diff
@@ -41,11 +49,19 @@ def compile
...
 
-      def opening
-        if compiler.requirable?
+      def opening(module_name = nil)
+        if compiler.es6_modules?
+          # to enable some webpack features for opal-webpack-loader it has to refer to the opal code from within webpack later on
+          # so we give it a handle 'opal_code'
+          line 'const opal_code = function() {'
+          # global makes sure we get the webpack global context and its Opal.modules, and not a locally shielded Opal.modules
+          line "  global.Opal.modules[#{module_name.inspect}] = function(Opal) {"
```
ahmadine: Is `global` defined in a browser context? Maybe using globalThis would be a better idea? Maybe it doesn't matter, I don't know, it just caught my eye. See #2042

npm/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "opal",
+  "version": "1.1.0",
```
ahmadine: That would be another place where the Opal version is defined. Best note that here, for the future releasers: https://github.com/opal/opal/blob/master/lib/opal/version.rb

By the way, 8ef6d934f01a93a70d83559888239d600ab48d87 adds a development version in a root directory. This may need to be reconciled I think.

lib/opal/es6_modules_helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+      def generate_module_import(ruby_module_path)
+        # module_path is empty for: require File.expand_path(o), can't generate a import, let it be resolved at runtime
+        return [''] if ruby_module_path.empty?
+        import_lines = []
+        ruby_module_path = determine_real_module_path(ruby_module_path)
+        import_module_name, ruby_module_name = Opal::ES6ModulesHelpers.module_names_from_paths(Pathname.new(ruby_module_path), ruby_module_path)
+        import_name = generate_import_name(import_module_name)
+        import_lines << "import #{import_name} from '#{import_module_name}';\n"
+        unless ruby_module_name == 'corelib/runtime'
+          import_lines << "if (typeof global.Opal.modules['#{ruby_module_name}'] === 'undefined') {\n"
```
ahmadine: Ditto about the use of `global`

lib/opal/nodes/top.rb
janbiedermann: This global here is from webpack. global has special meaning within webpack. Webpack usually gives modules their own context. Referring to global makes sure webpack provides the global context. So using `Opal.modules...` here would provide a local, incomplete version of `Opal`, using `global.Opal` makes sure the real Opal is used.

So within imported modules:
`Opal.modules != global.Opal.modules`

Think of `global` here as a compiler directive maybe.

npm/package.json
janbiedermann: These package.json have different purposes.
One is for installing things to run tests, develop opal.
The other is for building opal packages for npm. (And i will have to update these)

lib/opal/es6_modules_helpers.rb
janbiedermann: Ditto about the use of `global`

janbiedermann:
The "Fix global for Node VMs" may test for node directly instead.

lib/opal/nodes/call.rb
hmdne: Opal.Promise in general

janbiedermann:
continued in #2266

PR #1977, opened 2019-05-16
Update ruby/spec and MSpec

eregon:
Here is a PR to update to the latest ruby/spec.
The only change in Opal itself is 665af60, please review it.

Recently, I removed `with_feature` in MSpec (https://github.com/ruby/spec/issues/667) and wanted to check how it impacts Opal.
Notably, `with_feature :encoding` was removed as it was not maintained for recent specs.

It seems nowadays Opal supports a fair bit of Encoding and so in general the update seemed fine, there are just a few failures for new specs.

For the workflow, I used `bundle exec rake mspec_nodejs` which outputs a list `fails` lines to copy/paste under `spec/bugs` or `spec/unsupported` and merge in the sorted lines, easy enough and intuitive.

Somewhat related, I recently introduced [a small tool](https://github.com/ruby/mspec/blob/c25d63dc7e3ad1b07cab7776ac3a4e4db9bd0da8/tool/tag_from_output.rb) in MSpec to add tags based on the output from backtraces. This could potentially be useful for Opal as well, feel free to adapt/make your own based on it as needed. However, this needs to know which spec file was running for failed specs, which currently doesn't seem available in Opal (e.g., `at ::Object#$guard (/tmp/opal-nodejs-runner-20190516-29709-19hfirv:62777:27)`).

Congrats on the 1.0 release!


elia:
> easy enough and intuitive.

That's a big win! 🥳

>  feel free to adapt/make your own based on it as needed

I'll have a look, although my gut tells me it would require runtime sourcemap parsing to know the exact locations (which is something I wanted to do many times, but was not that straightforward).

> Congrats on the 1.0 release!

Thanks man! You deserve part of the credit! ✨ thanks for keeping up with the ruby spec 👍👍👍

PR #1984, opened 2019-06-07
Fix Inclusion of modules ...

janbiedermann:
... which are included by other modules which are included.

At Line 961 below module_iclass was null and a exception was thrown.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -950,7 +950,7 @@
       var proto = includer.$$prototype, parent = proto, module_iclass = Object.getPrototypeOf(parent);
 
       while (module_iclass != null) {
-        if (isRoot(module_iclass) && module_iclass.$$module === module) {
+        if (module_iclass.$$module === module && isRoot(module_iclass)) {
           break;
         }
 
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -949,6 +949,13 @@
 
       var proto = includer.$$prototype, parent = proto, module_iclass = Object.getPrototypeOf(parent);
 
+      var next_ancestor = Object.getPrototypeOf(module_iclass);
+
+      // skip non-root iclasses (that were recursively included)
+      while (next_ancestor.hasOwnProperty('$$iclass') && !isRoot(next_ancestor)) {
+        next_ancestor = Object.getPrototypeOf(next_ancestor);
+      }
+
       while (module_iclass != null) {
         if (isRoot(module_iclass) && module_iclass.$$module === module) {
           break;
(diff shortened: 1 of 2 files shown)


janbiedermann:
That works now, but not sure if it is the correct fix.

elia:
@janbiedermann thanks! can you add some code that reproduces the problem solved by this change or explain that?

I'd like to add a spec example along with the change.

janbiedermann:
Original situation was something like:
```
class A
  include ModB
  include ModC

module ModB
  include ModC

module ModC
  def self.included(base)
    # do somehting
```

suddenly stopped working. null Exceptions was thrown.

Looking at the code its obvious that its somehow wrong, so 6578189 tried to reorder.
That made the issue go away, but failed specs.
Recently the reordered code failed for another inclusion chain, so i came up with fb2b213.

The other question is, why it suddenly stopped working in the first place.
Thats my problem for adding a spec, i could not create a spec that could reproduce the problem.

janbiedermann:
Order may be important, maybe it was:
```
class A
  include ModC
  include ModB
  
module ModB
  include ModC

module ModC
  def self.included(base)
    # do somehting
```

janbiedermann:
I remember, maybe i twas like this:
```
class A
  include ModB
  include ModC

module ModB
  def self.included(base)
    base.include ModC

module ModC
  def self.included(base)
    # do somehting
```

Maybe it can reproduced then in the most simple way:
```
class A
   include ModC
   include ModC
```
i will try later

elia:
I tried this, but seems to be working fine:

```rb
module ModB
  def self.included(base)
    base.include ModC
  end
end

module ModC
  def self.included(base)
    # do somehting
    # Maybe it can reproduced then in the most simple way:
    p included: base
  end
end

class A
  include ModB
  include ModC
end

class A
  include ModC
  include ModC
end

p A.ancestors
```

output:

```
Opal v1.0.0
{"included"=>A}
{"included"=>A}
{"included"=>A}
{"included"=>A}
[A, ModC, ModB, Object, Kernel, BasicObject]
```

janbiedermann:
Right, that was my problem too, i could not recreate the error in the first place. But still, it happened.

elia:
I'll leave this open then, so we can add the necessary information if it happens again 👍

janbiedermann:
Hit it again!

Its like this:
```ruby
module ModuleB
end

module ModuleA
  include ModuleB
end

class ClassC
  include ModuleA
  include ModuleB
end
```
ClassC, when including ModuleA has ModuleB in its ancestor chain but without $$root flag.
When including ModuleB directly the check if ModuleB was included just checks if its in the array of ancestors, but the rebuilding of the ancestor chain, checks for $$root if the module has already been included and so cant find the module as its own. Its the first direct inclusion of the module.

janbiedermann:
Superseeded by #2053

PR #1986, opened 2019-06-08
Fix Enumerable#zip

kachick:


opal/corelib/helpers.rb
```diff
@@ -11,18 +11,22 @@ def self.type_error(object, type, method = nil, coerced = nil)
     end
   end
 
-  def self.coerce_to(object, type, method)
+  def self.coerce_to(object, type, method, argument = undefined)
```
elia: Although there's probably no real case for this I think at this point this one should be `args` and used as `object.__send__ method, *args`

like in:

```
obj.to_enum(method = :each, *args) → enum
obj.to_enum(method = :each, *args) {|*args| block} → enum
```

opal/corelib/array.rb
```diff
@@ -2300,7 +2300,7 @@ def zip(*others, &block)
         }
         others[j] = #{(
           Opal.coerce_to?(`o`, Array, :to_ary) ||
-          Opal.coerce_to!(`o`, Enumerator, :each)
+          Opal.coerce_to!(`o`, Enumerator, :to_enum, :each)
```
elia: I'm not sure why this can't be just `Opal.coerce_to!(`o`, Enumerator, :to_enum)`, can you explain?

kachick: `ruby/spec` checks the given argument as https://github.com/ruby/spec/blob/e489657352f3b9ec3b64a4d20160bd12f336bd3e/core/enumerable/fixtures/classes.rb#L145-L149 🤔

opal/corelib/helpers.rb
kachick: It's right! Updated as https://github.com/opal/opal/pull/1986/commits/dcdfbb28503c6bad6b341c968abd4bbe725dac66

opal/corelib/array.rb
elia: Ok, that's fine, passing an enumerator to `Array#zip` should be pretty rare anyway

elia:
Can you please rebase so that CI can re-run? 🙏

kachick:
Sure. Rebased!

elia:
Thanks!

kachick:
Thank you!

PR #2003, opened 2019-08-08
Fix Module#attr with multiple arguments

kachick:


elia:
Thanks for the contribution! I didn't know about this the _writable_ switch!

---

Checking the MRI implementation I saw that passing `true/false` as second argument is deprecated, here's the C code:

```C
VALUE
rb_mod_attr(int argc, VALUE *argv, VALUE klass)
{
    if (argc == 2 && (argv[1] == Qtrue || argv[1] == Qfalse)) {
        rb_warning("optional boolean argument is obsoleted");
        rb_attr(klass, id_for_attr(klass, argv[0]), 1, RTEST(argv[1]), TRUE);
        return Qnil;
    }
    return rb_mod_attr_reader(argc, argv, klass);
}
```

I'm wondering what's best here… if incorporating the obsolete form with a warning or keep the alias and send a PR to ruby/spec. If you prefer adding the warning please consider staying more close to the C implementations that has a single `if` only for the alternative version.

elia:
@kachick I went on and merged after adding a commit that uses an implementation more similar to the C source, thanks for catching this one!

kachick:
Sorry to late reply 🙇 
Sure~ Thanks to review!

PR #2004, opened 2019-08-09
opal 1.0 fix: raises "warning: Removed semicolon ending x-string expression,...

bwl21:
interpreted as unintentional -- nodejs/fileutils.rb:14"


elia:
Thanks!

PR #2008, opened 2019-08-17
Replace Travis and Appveyor with CircleCI

elia:
Pros:
- all in one place
- sparated results in github checks

Cons:
- more complex configuration, should investigate Orbs to simplify it


elia:
since moved to GitHub actions

PR #2017, opened 2019-09-28
benchmarking.rake - bench:ips - correct a path to benchmarks

ahmadine:
This instruction was probably originally in a Rakefile and so it referenced `__dir__` correctly. After moving it to tasks/, the `__dir__` reference changed. This commit corrects the path.


elia:
Thanks!

PR #2038, opened 2019-12-16
Reduce size of compiled code skipping unused helpers

elia:
- Use magic-comments to require helpers, skip them otherwise (break, slice)
- Reduce the size of the constant getters Opal property name

I attempted to make `nil` into an optional helper, but it's so ubiquitous that I eventually gave up.

### DIFF
--- a/lib/opal/magic_comments.rb
+++ b/lib/opal/magic_comments.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module Opal::MagicComments
-  MAGIC_COMMENT_RE = /\A# *(\w+) *: *(\w+) *$/.freeze
-  EMACS_MAGIC_COMMENT_RE = /\A# *-\*- *(\w+) *: *(\w+) *-\*- *$/.freeze
+  MAGIC_COMMENT_RE = /\A# *(\w+) *: *(\S+.*?) *$/.freeze
+  EMACS_MAGIC_COMMENT_RE = /\A# *-\*- *(\w+) *: *(\S+.*?) *-\*- *$/.freeze
 
   def self.parse(sexp, comments)
     flags = {}
--- a/lib/opal/nodes/logic.rb
+++ b/lib/opal/nodes/logic.rb
@@ -85,6 +85,7 @@ module Opal
       end
 
       def compile_iter
+        helper :slice
         push "return #{scope.identity}.apply(null, $slice.call(arguments))"
       end
     end
--- a/lib/opal/nodes/top.rb
+++ b/lib/opal/nodes/top.rb
@@ -29,8 +29,8 @@ module Opal
             add_temp '$nesting = []'
           end
           add_temp 'nil = Opal.nil'
-          add_temp '$$$ = Opal.const_get_qualified'
-          add_temp '$$ = Opal.const_get_relative'
+          add_temp '$$$ = Opal.$$$'
+          add_temp '$$ = Opal.$$'
 
           add_used_helpers
           add_used_operators
--- a/lib/opal/nodes/top.rb
+++ b/lib/opal/nodes/top.rb
@@ -77,8 +77,7 @@ module Opal
       end
 
       def add_used_helpers
-        helpers = compiler.helpers.to_a
-        helpers.to_a.each { |h| add_temp "$#{h} = Opal.#{h}" }
+        compiler.helpers.to_a.each { |h| add_temp "$#{h} = Opal.#{h}" }
       end
 
       def add_used_operators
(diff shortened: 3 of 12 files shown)


lib/opal/nodes/top.rb
```diff
@@ -76,8 +76,13 @@ def compile_irb_vars
...
       end
 
+      def magic_comments_helpers
+        @magic_comments_helpers ||=
+          compiler.magic_comments[:helpers].to_s.split(',').map(&:strip)
+      end
+
       def add_used_helpers
-        helpers = compiler.helpers.to_a
+        helpers = compiler.helpers.to_a + magic_comments_helpers
```
iliabylich: what happens if the file declares the helper and the code implicitly declares it via `helper` DSL in the compiler node? It worked well before (because `compiler.helpers` is a set), but I guess now it can declare them twice.

lib/opal/magic_comments.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module Opal::MagicComments
-  MAGIC_COMMENT_RE = /\A# *(\w+) *: *(\w+) *$/.freeze
-  EMACS_MAGIC_COMMENT_RE = /\A# *-\*- *(\w+) *: *(\w+) *-\*- *$/.freeze
+  MAGIC_COMMENT_RE = /\A# *(\w+) *: *(\S+.*?) *$/.freeze
```
iliabylich: ignore this comment if I'm overthinking, but the idea of using magic comments feels cryptic to me. Comments shouldn't be used to change behavior of your code. However Ruby does it, so I don't know if it's ok or not.

Also I think I'd personally expect these comments (if they are valuable at all) to be used as some kind of meta-information about the file (like by some external analyzer, rubocop is an example), and so I'd expect that I can safely remove them.

Is it that much worse writing `var <helper> = Opal.<helper>` at the beginning of the file?

lib/opal/nodes/logic.rb
```diff
@@ -85,6 +85,7 @@ def compile_while
       end
 
       def compile_iter
+        helper :slice
```
iliabylich: huh, just got the idea that `shortcut` is probably a better name

lib/opal/nodes/top.rb
```diff
@@ -29,8 +29,8 @@ def compile
             add_temp '$nesting = []'
           end
           add_temp 'nil = Opal.nil'
-          add_temp '$$$ = Opal.const_get_qualified'
-          add_temp '$$ = Opal.const_get_relative'
+          add_temp '$$$ = Opal.$$$'
+          add_temp '$$ = Opal.$$'
```
ahmadine: This will make the newly generated code incompatible with the 1.0.0 runtime. I would suggest bumping the version to like 1.0.1.

elia: @ahmadine you're right, actually master is targeted toward 1.1, while the 1.0 branch is `1-0-stable` and that will receive minor patches and bugfixes (actually I already released 1.0.2).

That said, it should still be compatible, as the local variable names won't change, except if someone was using stuff like `Opal.$$.Regexp.$$.IGNORECASE`.

_I'll update master to be v1.1.0.alpha shortly_

ahmadine: I didn't mean "API" incompatibility, but "ABI" incompatibility. Otherwise, yeah, good to know, I missed the 1.0-stable branch.

elia: I see, yet, old compiled code should keep working (except for the `Opal.$$.Regexp…` example, which was only meant to aid debugging). Both `Opal.const_get_relative` and `Opal.const_get_qualified` are still available. Sorry if I ask again, it's because I want to be sure to really understand the implications of these changes 🙂

ahmadine: New code uses `$$ = Opal.$$` which, when combined with the old runtime, would be equal to `$$ = undefined`. And that isn't going to work, because `Opal.$$ = Opal.const_get_qualified` is just defined in this changeset.

I think most people wouldn't be affected, but if someone combined the prebuilt CDN "binaries" with the master code which says "1.0.0", he could be surprised.

elia: Oh I see now, and yes I agree it shouldn't happen and it will be even less likely when I change the dev version to 1.1.0.alpha 👍

lib/opal/magic_comments.rb
elia: The way I see these helpers is that they're mostly for core stuff and very unlikely to be used elsewhere (although it wouldn't be a problem per se). Broadening the conversation to include also the `use_strict` comment I expect it to be transient and probably change it's default from `false` to `true` in `v1.2`. I personally dislike having to mark _every_ file with _frozen string  literal_ although I like the effect.

In summary `helpers` should only (mostly?) live within opal core, and `use_strict` is transient. Hope in this perspective it becomes less of a concern.

PR #2039, opened 2019-12-16
Fix TruffleRuby on travis

elia:


Mogztter:
I guess we should close this pull request since we moved away from Travis

hmdne:
Let's try to re-enable those tests but for GHA for a change.

Issue #2042, opened 2019-12-20
opal-repl doesn't work in master


ahmadine:
opal-repl doesn't work in master due to this error:

```
Traceback (most recent call last):
        11: from bin/opal-repl:4:in `<main>'
        10: from bin/opal-repl:4:in `load'
         9: from /home/user/opal/exe/opal-repl:86:in `<top (required)>'
         8: from /home/user/opal/exe/opal-repl:23:in `run'
         7: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:198:in `eval'
         6: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:198:in `synchronize'
         5: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:200:in `block in eval'
         4: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:286:in `timeout'
         3: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:201:in `block (2 levels) in eval'
         2: from /usr/local/share/gems/gems/mini_racer-0.2.6/lib/mini_racer.rb:201:in `eval_unsafe'
         1: from JavaScript at <anonymous>:2449:4
JavaScript at <anonymous>:25:28: TypeError: Cannot read property 'console' of undefined (MiniRacer::RuntimeError)
```

I did a git bisect to note what started this behaviour and it's this commit:

```
390f302b8e2a9d8334f9c990009b7b1257ece0aa is the first bad commit
commit 390f302b8e2a9d8334f9c990009b7b1257ece0aa
Author: Elia Schito <elia@schito.me>
Date:   Sun Nov 10 23:53:06 2019 +0100

    Typo, was supposed to be "use strict";

 opal/corelib/runtime.js | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)
```

ahmadine:
390f302b8e2a9d8334f9c990009b7b1257ece0aa

ahmadine:
This is the code that tries to trigger the issue on the smallest surface possible:

```ruby
#!/usr/bin/env ruby
require 'mini_racer'

JS = <<endjs
  (function() {
    "use strict";
    var global_object = this;
    // MiniRacer defines neither window nor global and this is undefined in the strict mode
    typeof (global_object.console); // Whoops
  }());
endjs

[...]

PR #2043, opened 2019-12-20
Restore .call(this) to restore minimal JS environment compatibility

ahmadine:
Please read https://github.com/opal/opal/issues/2042 for description of the issue.


elia:
Thanks, the issue was fixed in #2047, this was used as the starting point for it! 👍

PR #2045, opened 2019-12-22
Use source-maps for Node.js stack-traces 

elia:
Requiring nodejs/stacktrace has been deprecated, source-maps are now supported by the default Node.js runner or by requiring https://github.com/evanw/node-source-map-support before loading code compiled by Opal.

A typical stacktrace will now include files ending in `.rb` and accurate line/col references:

```
⤑ bin/opal -e"puts 5; class A;def a;raise;end;end;  A.new.a"                                                                          ~/C/opal
5
RuntimeError: 
    at corelib/kernel.rb:521:23:in `new'
    at -e:1:23:in `raise'
    at -e:1:39:in `a'
    at /private/var/folders/w0/yjfqr9n94ld7ft4j3hlz_fsm0000gn/T/opal-system-runner-20191222-27867-sozm2m:23569:3:in `null'
    at internal/modules/cjs/loader.js:759:30:in `Module._compile'
    at internal/modules/cjs/loader.js:770:10:in `Module._extensions..js'
    at internal/modules/cjs/loader.js:628:32:in `Module.load'
    at internal/modules/cjs/loader.js:555:12:in `Module._load'
    at internal/modules/cjs/loader.js:822:10:in `Module.runMain'
    at internal/main/run_main_module.js:17:11:in `undefined'
```

_Not the most complete example, but please notice how the first lines have accurate line/col info: `corelib/kernel.rb:521:23`._


UNRELEASED.md
```diff
@@ -49,3 +54,11 @@ Whitespace conventions:
 - Regexps assigned to constants are now frozen (#2007)
 - `Opal.$$` changed from being the constant cache of Object to being a shortcut
   for `Opal.const_get_relative` (#2038)
+  for `Opal.const_get_relative`
```
ahmadine: Duplicate line

elia: Thanks, fixed!

elia:
@Mogztter you see any problem for asciidoc related to these changes?

Mogztter:
The file `source-map-support.js` is only required in the CLI runner? In other words, there's no additional runtime dependencies (in the stdlib)?
We do not use the CLI runner directly in Asciidoctor.js so I think it should be fine.

.travis.yml
```diff
@@ -93,7 +93,7 @@ before_install:
   # Keep track of which version of node we're running the specs against
   - bin/git-submodule-fast-install
   - node -v
-  - npm install jshint uglify-js
+  - npm install
```
Mogztter: We should probably update the [HACKING.md](https://github.com/opal/opal/blob/master/HACKING.md) file:

```diff
- $ npm install -g jshint
+ $ npm install
```

elia: Thanks, moved everything to `bin/setup`, both for HACKING and in the README

elia:
> The file `source-map-support.js` is only required in the CLI runner? In other words, there's no additional runtime dependencies (in the stdlib)?

That's right 👍

PR #2050, opened 2019-12-31
SecureRandom#hex(n)

janbiedermann:
As the title says, with enabled specs.

### DIFF
--- a/stdlib/securerandom.rb
+++ b/stdlib/securerandom.rb
@@ -11,7 +11,8 @@ module SecureRandom
           string = '',
           temp;
       for (var i = 0; i < repeat; i++) {
-        temp = Math.floor(Math.random()*parseInt('ff'.repeat(6), 16)).toString(16);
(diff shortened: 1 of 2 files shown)


stdlib/securerandom.rb
```diff
@@ -1,4 +1,35 @@
...
+    raise ArgumentError, 'count of hex numbers must be positive' if count < 0
+    %x{
+      count = Math.floor(count);
+      var repeat = Math.floor(count / 6),
+          remain = count % 6,
+          remain_total = remain * 2,
+          string = '',
+          temp;
+      for (var i = 0; i < repeat; i++) {
+        temp = Math.floor(Math.random()*parseInt('ff'.repeat(6), 16)).toString(16);
```
elia: Can you replace `parseInt('ff'.repeat(6), 16)` with the static result?
The original code can be added as a comment just above with a brief explanation on how it's been obtained.

janbiedermann: Done.

PR #2052, opened 2019-12-31
Remove unused variables from runtime.js

janbiedermann:


elia:
No code is faster than _no code_! Thanks!

PR #2057, opened 2020-01-06
Continue or else ...!!!

janbiedermann:
Remove useless code.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -2299,10 +2299,6 @@
     for (i = 0, l = paths.length; i < l; i++) {
       path = Opal.normalize(paths[i]);
 
-      if (Opal.require_table[path]) {
-        continue;
-      }
-
       Opal.loaded_features.push(path);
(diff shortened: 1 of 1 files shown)


opal/corelib/runtime.js
```diff
@@ -2299,10 +2299,6 @@
     for (i = 0, l = paths.length; i < l; i++) {
       path = Opal.normalize(paths[i]);
 
-      if (Opal.require_table[path]) {
-        continue;
```
Mogztter: I don't think we can remove this code.
The above code:
```js
for (i = 0; i < 3; i++) { 
  if (i === 1) {
    continue;
  }
  console.log(i);
}
// print 0,2
```
is *not* equivalent to:

```js
for (i = 0; i < 3; i++) {
  console.log(i);
}
// print 0,1,2
```

In this case we dont want to push the same path in `Opal.loaded_features`.

janbiedermann: Oh crap, i was completely betriebsblind, sorry

PR #2058, opened 2020-01-06
Make Kernel#require overwritable.

janbiedermann:
Kernel#require was masked by Opal.require via Object#require.
This patch makes sure require is passed to Kernel#require, which can then be overwritten.
Overwriting Kernel#require is used by the (opal-)zeitwerk autoloader.

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -691,5 +691,9 @@ module Kernel
 end
 
 class Object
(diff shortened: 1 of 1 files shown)


janbiedermann:
It causes recursion, so obviously i fixed some other place too, but i dont remember which at the moment. It does not cause recursion in es6_modules_1_1. Will figure out this later on.

PR #2061, opened 2020-01-09
resolves #2060 set match on skip and scan_until

### LINKED ISSUE #2060: StringScanner returns nil on most recent match
Consider the following code:

```rb
require 'strscan'

stream = StringScanner.new('cat builder.rb')
if stream.skip(/cat/)
  puts stream[0].inspect
end
```
**MRI 2.6.3**

```
$ ruby test.rb  
"cat"
```

**Opal master (01f63335)**

```
$ bundle exec opal test.rb  
nil
```

Mogztter:
resolves #2060


stdlib/strscan.rb
```diff
@@ -80,6 +82,16 @@ def scan_until(pattern)
   end
 
   def [](idx)
+    if @match.empty?
```
Mogztter: I took inspiration from MRI: https://github.com/ruby/ruby/blob/4c5eac73238a203830d43bccf22eee27d242616f/ext/strscan/strscan.c#L1165

Mogztter:
@elia Do you want me to fix the codeclimate issues?

Mogztter:
To be honest, I don't really see the code duplication... is that a bug?

stdlib/strscan.rb
```diff
@@ -80,6 +82,16 @@ def scan_until(pattern)
...
+      return nil
+    end
+    if idx === Symbol
```
elia: Why not a case/when here?

```rb
case idx
when Symbol
…
```

Mogztter: Done :+1:

Mogztter:
> Also can you change the commit message subject to just describe the change?
You can keep the "Resolves" in the commit description if you want +1

Sorry I'm so used to the Asciidoctor Git commit convention :sweat_smile: 
I've edited the commit subject and I've added an item in the CHANGELOG.

PR #2063, opened 2020-01-17
Add a generic implementation of Kernel#caller

elia:
This implementation should be good enough for browsers and node.

Along with the recent changes to the node stacktrace will be making the development experience both of opal itself and of opal apps much better.


elia:
Replaced by #2064.

PR #2064, opened 2020-01-17
Add a generic implementation of Kernel#caller and #warn(uplevel:)

elia:
Along with the recent changes to the node stacktrace this PR should be making the development experience both of opal itself and of opal apps much better.


elia:
Replaced by #2065

PR #2065, opened 2020-01-17
Add a generic implementation of Kernel#caller and #warn(uplevel:)

### LINKED ISSUE #2068: Caller doesn't work
```
class Test
   def self.foo
      Fooz.bar
   end
end

class Fooz
  def self.bar
	puts caller[0][/`.*'/][1..-2]
  end
end

Test.foo
```

`caller_locations(1,1)[0].label` doesnt work too.
Is there an alternative or any way to fix this?

elia:
Along with the recent changes to the node stacktrace this PR should be making the development experience both of opal itself and of opal apps much better.

Fixes #2068


opal/corelib/kernel.rb
```diff
@@ -92,8 +92,13 @@ def at_exit(&block)
 
   # Opal does not support #caller, but we stub it as an empty array to not
```
iliabylich: ✂️

lib/opal/cli_runners/chrome.rb
```diff
@@ -53,6 +59,38 @@ def run
...
+          <head>
+            <meta charset='utf-8'>
+            <script src='./source-maps.js'></script>
+            <script>
+            sourceMapSupport.install({
+              retrieveSourceMap: function(path) {
+                if (path.endsWith('/index.js')) {
+                  return {
+                    url: './index.map',
+                    map: #{builder.source_map.to_json.to_json}
```
iliabylich: can the second `to_json` be replaced with `inspect`?

elia: I'm not 100% sure the escaping is the same between the two… I since moved the generation of the values to be written to the top of the method, so the two `to_json` don't appear one after the other anymore.

PR #2069, opened 2020-01-24
Fix Array keep_if, delete_if, reject!, select!

### LINKED ISSUE #2067: delete_if behaves differently from MRI
`[[1,2]].delete_if` should yield two values (1, 2) but yields an array `[1,2]`.

```
/Users/yhara/Dropbox/proj/dxopal % opal _1.0.2_  -e '[[1,2]].delete_if do |x, y| p c: [x, y] end'
{"c"=>[[1, 2], nil]}
/Users/yhara/Dropbox/proj/dxopal % opal _0.11.4_ -e '[[1,2]].delete_if do |x, y| p c: [x, y] end'
{"c"=>[[1, 2], nil]}
/Users/yhara/Dropbox/proj/dxopal % ruby -e '[[1,2]].delete_if do |x, y| p c: [x, y] end'
{:c=>[1, 2]}
```

There may be more array methods that have this problem (as far as I've tested, `Array#each` and `Array#find` works fine.)

elia:
Match CRuby behavior when exceptions are raised in the block and about
when the array is updated.

Fixes #2067

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -1,3 +1,5 @@
+# helpers: truthy, falsy
+
 require 'corelib/enumerable'
 require 'corelib/numeric'
 
@@ -14,6 +16,34 @@ class Array < `Array`
       } else {
         return klass.$allocate().$replace(#{`obj`.to_a});
       }
+    }
+
+    // A helper for keep_if and delete_if, filter is either Opal.truthy
+    // or Opal.falsy.
+    function filterIf(self, filter, block) {
+      var value, raised = null, updated = new Array(self.length);
+
+      for (var i = 0, i2 = 0, length = self.length; i < length; i++) {
+        if (!raised) {
+          try {
+            value = Opal.yield1(block, self[i])
+          } catch(error) {
+            raised = error;
+          }
+        }
+
+        if (raised || filter(value)) {
+          updated[i2] = self[i]
+          i2 += 1;
+        }
+      }
+
+      if (i2 !== i) {
+        self.splice.apply(self, [0, updated.length].concat(updated));
+        self.splice(i2, updated.length);
+      }
+
+      if (raised) throw raised;
     }
   }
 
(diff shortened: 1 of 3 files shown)


opal/corelib/array.rb
```diff
@@ -1357,16 +1368,27 @@ def keep_if(&block)
...
+          i2 += 1;
         }
       }
+
+      if (i2 !== i) {
+        self.splice.apply(self, [0, updated.length].concat(updated));
+        self.splice(i2, updated.length);
+      }
+
+      if (raised) throw raised;
```
iliabylich: MRI aborts iteration once the error has been thrown:

```ruby
[1,2,3].keep_if { |i| p i; raise 'a' }
1
# ...backtrace
RuntimeError (a)
```

So I guess this check-and-throw should be somewhere above (same for other methods)

opal/corelib/array.rb
```diff
@@ -1357,16 +1368,27 @@ def keep_if(&block)
...
-      for (var i = 0, length = self.length, value; i < length; i++) {
-        value = block(self[i]);
+      var value, raised = null, updated = new Array(self.length);
 
-        if (value === false || value === nil) {
-          self.splice(i, 1);
+      for (var i = 0, i2 = 0, length = self.length; i < length; i++) {
+        try {
+          value = Opal.yield1(block, self[i])
+        } catch(error) {
```
iliabylich: Also I guess it may catch and re-throw `Opal.breaker` and `Opal.returner` and thus, potentially exit the outer method

elia: From what I saw they're already managed by the compiler.

iliabylich: compiler manages only what is written in Ruby (but catch/re-throw here is in JS). maybe I'm wrong, but does it handle `break` and `return` inside `keep_if`? 👍 if so, but I'm afraid it swallows the initial `throw Opal.returner` and re-raises it once the iteration is done (MRI most probably does a hard `return` that aborts the loop)

Also, I'm not sure if it's actually that important. I've never seen such code 😄

elia: just out of curiosity, here's what I got:

```
opal:elia/fix-delete-if-keep-if ⤑ bin/opal -e'def a;4.times{|n| [1,2,3].delete_if { |i| p i; return }; p n => n};end; a'          ~/C/opal
1
opal:elia/fix-delete-if-keep-if ⤑ ruby -e'def a;4.times{|n| [1,2,3].delete_if { |i| p i; return }; p n => n};end; a'              ~/C/opal
1
opal:elia/fix-delete-if-keep-if ⤑                                                                                                 ~/C/opal
```

PR #2075, opened 2020-02-04
Enable strict-mode & frozen strings on corelib & stdlib

elia:


stdlib/matrix.rb
```diff
@@ -1,3 +1,6 @@
+# use_strict: true
+# frozen_string_literal: true
+
 # encoding: utf-8
 # frozen_string_literal: false
```
hmdne, line 5: force string literal is duplicated

stdlib/matrix/eigenvalue_decomposition.rb
```diff
@@ -1,3 +1,6 @@
+# use_strict: true
+# frozen_string_literal: true
+
 # frozen_string_literal: false
```
hmdne, line 4: frozen string literal is duplicated

stdlib/matrix.rb
hmdne, line 5: frozen string literal is duplicated

stdlib/matrix/lup_decomposition.rb
```diff
@@ -1,3 +1,6 @@
+# use_strict: true
+# frozen_string_literal: true
+
 # frozen_string_literal: false
```
hmdne, line 4: frozen string literal is duplicated

hmdne:
I just rebased it onto the current master.

PR #2079, opened 2020-02-17
Fix Regexp.new, replace \A to ^ and \z to $

ysakasin:
Opal replaces `\A` and `\z` in regexp literal but it doesn't replace in `Regexp.new`.
https://github.com/opal/opal/blob/178d9cd117444dfd4db6d4f6700b36838a8097d4/lib/opal/nodes/literal.rb#L191

This PR changes to replace `\A` to `^` and `\z` to `$` by `Regexp.new`

### DIFF
--- a/opal/corelib/regexp.rb
+++ b/opal/corelib/regexp.rb
@@ -83,6 +83,8 @@ class ::Regexp < `RegExp`
           #{::Kernel.raise ::RegexpError, "too short escape sequence: /#{regexp}/"}
         }
 
+        regexp = regexp.replace('\\A', '^').replace('\\z', '$')
+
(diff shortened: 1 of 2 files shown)


ahmadine:
I see one problem with that, but I'm unsure how to solve it.

```ruby
it 'works with escape characters' do
  /\\A/.should =~ '\A'
  /\\z/.should =~ '\z'
  /\\\z/.should =~ %s{hello\}
end
```

ysakasin:
I think that is another issue. That is behavior of regular expression literal from before.This pull request changes only `Regexp.new`.

(I know that problem, but I don't inform issue tracker because I have no idea how to solve it.

elia (approved):

PR #2083, opened 2020-02-26
Port prettyprint to Opal

### LINKED ISSUE #2080: pp (prettyprint)
Opal's `pp` seems like a stub. Is there any plan to suppot pretty printing?

Since CRuby's pp (pp.rb and prettyprint.rb) is written in Ruby, it shouldn't be so difficult to port it to Opal. Actually I could make it work with a few modifications. May I send a pull request?

yhara:
This PR replaces pp with the one of CRuby 2.7.0 (with minimal fixes for Opal).

fix #2080

### Demo

```
% bundle exec opal -r pp -e 'pp (1..9).map{|i| [i]*i }'
[[1],
 [2, 2],
 [3, 3, 3],
 [4, 4, 4, 4],
 [5, 5, 5, 5, 5],
 [6, 6, 6, 6, 6, 6],
 [7, 7, 7, 7, 7, 7, 7],
 [8, 8, 8, 8, 8, 8, 8, 8],
 [9, 9, 9, 9, 9, 9, 9, 9, 9]]
```


yhara:
> I see the value of keeping the diff from the original file minimal, that's why I would ask you to please generate a rubocop todo file out of the current failures and move its rules to the main .rubocop.yml so with a comment explaining that we want to stay close to the original.

There is "Files that were copied from MRI as is" section in the .rubocop.yml. How about just adding pp to this list(57c055b)?

> Other than that I would replace stdlib/pp.rb directly in the first commit, is there a reason for not doing that?

Not a big reason; I just wanted to emphasize I'm adding files, not modifying.

elia:
Thanks! 
_Sorry for letting you wait so long_

PR #2087, opened 2020-03-14
Kernel#sleep should return an integer

elia:
There's a spec for this [here](https://github.com/ruby/spec/blob/875a09e47a6efdd45c94ddea8227fd0df72f65b0/core/kernel/sleep_spec.rb#L10), but given the confusion between Integer and Float due to JavaScript numbers it didn't catch the error.


elia:
Closing in favor of https://github.com/opal/opal/pull/2077

PR #2090, opened 2020-03-16
Super node cleanup

elia:


hmdne:
Rebased this and corrected `helper :super` to `helper :find_super`

hmdne:
One question I have before I can do some review of this patchset... what's the difference between Super and Zuper/Zsuper? What's the logic between those names?

PR #2096, opened 2020-05-04
Fix compatibility for including plain JS files with opal-sprockets

hmdne:
opal-sprockets adds some OpalLoaded variable to keep track of plain JS files loaded. It may be a good idea to make use of this variable. Loading those files threw an error before this commit, because those were loaded but Opal expected an Opal module.

@janbiedermann This place may also be a good way to handle JS includes for Webpack. Seeing commit 0791b81 of your ES6 modules branch for 1.1, it looks like you came upon this problem as well.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -2396,6 +2396,10 @@
     if (module) {
       module(Opal);
     }
+    else if (typeof OpalLoaded !== "undefined" && OpalLoaded.indexOf(path) !== -1) {
+      // Compatibility for loading plain JS files via opal-sprockets
+      return false;
+    }
     else {
       var severity = Opal.config.missing_require_severity;
       var message  = 'cannot load such file -- ' + path;


hmdne:
https://github.com/opal/opal-sprockets/blob/master/lib/opal/sprockets.rb#L49

hmdne:
I don't know if it's the best solution anyway. There are two calls here:
https://github.com/opal/opal-sprockets/blob/master/lib/opal/sprockets.rb#L35
Line 36 overrides line 35 if stubs are defined.

elia:
> I don't know if it's the best solution anyway. There are two calls here:
> https://github.com/opal/opal-sprockets/blob/master/lib/opal/sprockets.rb#L35
> Line 36 overrides line 35 if stubs are defined.
> 
> This could have caused [opal/opal-rails#115](https://github.com/opal/opal-rails/issues/115).

I think this is the underlying issue, I'll fix it right away.

elia:
@hmdne 🤔 might not be that after all, line 36 includes `loaded` from line 35:

`loaded = "#{stubbed.to_json}.concat(#{loaded})" if stubbed.any?`

hmdne:
Right. And it will be the best to push this check in a different place.

PR #2098, opened 2020-05-16
Make sure that ignoreCase is either true or false (not undefined)

Mogztter:
Otherwise I get the following warning message:

```
warning: ignore case doesn't match for "\\/"
```

From line: https://github.com/opal/opal/blob/3c8d93ea8b3ed12b4c49c51cc96d092570c9f64c/stdlib/pathname.rb#L113

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -2319,7 +2319,7 @@
   // Combine multiple regexp parts together
   Opal.regexp = function(parts, flags) {
     var part;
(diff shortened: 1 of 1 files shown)


Mogztter:
@elia I've restarted the build in Travis (the build was green but somehow the notification was "lost" and the status on GitHub has not been updated)

Mogztter:
Thanks :tada: :tada: :tada:

PR #2100, opened 2020-05-17
Remove Number#step and implement Numeric#step

takaram:
This allows us to use the method with Numeric instances such as `Rational` and `BigDecimal`.


elia (approved):
Thanks for this! 🌟 
Sorry for the big delay! 🙏 

(just needs to be brought back to speed with the latest changes)

PR #2109, opened 2020-07-19
Restore compatibility with Chrome 38

hmdne:
This is an ancient browser, I know, but it's the basis of many smartphone default browsers.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1918,7 +1918,11 @@
 
     // Assign the 'length' value with defineProperty because
     // in strict mode the property is not writable.
-    Object.defineProperty(alias, 'length', { value: body.length });
+    // It doesn't work in older browsers (like Chrome 38), where
+    // an exception is thrown breaking Opal altogether.
+    try {
+      Object.defineProperty(alias, 'length', { value: body.length });
+    } catch (e) {}
 
     // Try to make the browser pick the right name
     alias.displayName       = name;


Mogztter:
`defineProperty` is supported since Chrome 5 according to MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Browser_compatibility

So I'm not sure what you are trying to solve... Do you have a stacktrace? But in any case, I don't think that we want to support a 6 years old browser.

hmdne:
This browser is used as WebView on some older (though still in use) smartphones. It's not about defineProperty, it's about just this particular use of defineProperty. I wouldn't call it a full support, just a prevention of a full failure. I can get a stacktrace, sure.

https://hastebin.com/emimopifar

Merging this little patch will reduce my local delta a little bit :)

hmdne:
By the way it was tested on https://github.com/opal/opal/commit/53593f7a82377deedd2d8082a92efd9b6848f1e0

hmdne:
And it all boils down to this:
`Object.defineProperty(function(){}, "length", {value: 10})`

hmdne:
By the way, to get a browser that fails this problem, I have the following podman/docker Dockerfile:

```
FROM debian:7

COPY sources.list /etc/apt/sources.list

RUN apt-get -o Acquire::Check-Valid-Until=false update && \
    apt-get -y install tightvncserver chromium openbox psutils iproute rxvt-unicode iceweasel xfonts-base; \
    apt-get clean

RUN echo 8736159e0a4d04d90f3de03f5d8b7d08 > /etc/machine-id; mkdir /root/.vnc; echo '' | vncpasswd -f > /root/.vnc/passwd; chmod 0700 /root/.vnc; chmod 0600 /root/.vnc/passwd

ENV USER=root

CMD vncserver :0; DISPLAY=:0 openbox
EXPOSE 5900/tcp
```

This requires the following sources.list file:

```
deb http://archive.debian.org/debian wheezy main contrib non-free
deb http://archive.debian.org/debian-security wheezy/updates main contrib non-free
```

Afterwards, one needs to build the image (replace podman with docker if you want to use docker):

```
podman build -t oldbrowsers/d7 .
```

And run it:

```
podman run -dt --rm --shm-size=512M -p 127.0.0.1:5900:5900 $DIR oldbrowsers/d7 $X
sleep 1
vncviewer :0
```

elia (approved):

elia:
@hmdne I rebased this PR but for some reason it's failing on minitest, although it works locally, you have any clue on why this happens?

hmdne:
@elia Just some entropy got between the cogwheels of the CI. I reran the tests and they work now.

PR #2115, opened 2020-09-26
Improve Array#shift performance on v8 > 7.1 (Node 12+)

Mogztter:
### Command

```
$ bundle exec rake bench:ips FILE=benchmark-ips/bm_array_shift.rb
```

#### Node 10

```
=== Files: benchmark-ips/bm_array_shift.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -rnodejs -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_array_shift.rb
""
        shift no arg    39.327k i/100ms
        shift no arg    979.605k (± 1.6%) i/s -      4.916M in   5.019468s
```

#### Node 12

```

=== Files: benchmark-ips/bm_array_shift.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -rnodejs -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_array_shift.rb
""
        shift no arg     2.364k i/100ms
        shift no arg     24.879k (± 0.5%) i/s -    125.292k in   5.036277s
```

#### Node 10

```
=== Files: benchmark-ips/bm_array_shift.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_array_shift.rb
        shift no arg    26.398k i/100ms
        shift no arg    449.226k (± 4.4%) i/s -      2.244M in   5.005441s
```

#### Node 12

```
=== Files: benchmark-ips/bm_array_shift.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_array_shift.rb
        shift no arg    37.305k i/100ms
        shift no arg    611.658k (± 1.0%) i/s -      3.059M in   5.001669s
```

### Summary

With this change, the code is now 2.18x times slower when using Node 10 but 24.58x times faster when using Node 12.

We cannot beat the highly optimized code produced by v8 pre 7.1 (Node 10) but the code is now "only" 1.6x times slower when using a recent version of v8/Node.

Please note that the performance on Node 14 is roughly equivalent to Node 12.

### Further improvement

The `Array#shift()` (JavaScript) function is also used in [`String#tr`](https://github.com/opal/opal/blob/53593f7a82377deedd2d8082a92efd9b6848f1e0/opal/corelib/string.rb#L1388) and [`String#tr_s`](https://github.com/opal/opal/blob/53593f7a82377deedd2d8082a92efd9b6848f1e0/opal/corelib/string.rb#L1533). It might be worth to replace this call by the optimized `shiftNoArg` function? In this case, where should I declare the `shiftNoArg` function? in `runtime.js`?

I also found an optimization for `splice` when the second argument is `1` (ie. `list.splice(position, 1)`). I found 8 occurrences in the code where the second argument is explicitly 1. When a variable is used as the second argument it might be worth to check if the value is actually 1 to use the optimized code (I think it's only worth if the array is large enough but that means that we need to add an additional check that will slow down the execution).


opal/corelib/array.rb
```diff
@@ -10,6 +10,16 @@ class Array < `Array`
   `Opal.defineProperty(self.$$prototype, '$$is_array', true)`
 
   %x{
+    function shiftNoArg(list) {
```
elia, line 13: Can you add a mention about the performance regression in a comment before the function definition?
I'm thinking about our future selves… and how we might want to remove the optimization if V8 goes back to optimizing the vanilla version.

elia (changes requested):

opal/corelib/array.rb
Mogztter: Sure, actually there are quite a few things to mention. I will also add a link to the original implementation (found in the Node.js codebase)

Mogztter: @elia Done

Mogztter:
@elia I found a more generic implementation:

```js
function spliceGen (list, start, deleteCount) {
  if (deleteCount <= 0) {
    return
  }
  const length = list.length
  for (var i = start, k = i + deleteCount, n = length; k < n; i += 1, k += 1)
    list[i] = list[k]

  list.length = i
  // or:
  //for (;i < length; i++)
  //  list.pop()
}
```

Depending on the list length, calling `list.pop()` can be faster than updating the list length. The goal is to remove "trailing" elements.

opal/corelib/array.rb
```diff
@@ -10,6 +10,28 @@ class Array < `Array`
...
+    // In short, Array operations are slow in recent versions of V8 when the Array.prototype has been tampered.
+    // So, when possible, we are using faster open-coded version to boost the performance.
+
+    // As of V8 8.4, depending on the size of the array, this is up to ~25x times faster than Array#shift()
+    // Implementation is heavily inspired by: https://github.com/nodejs/node/blob/ba684805b6c0eded76e5cd89ee00328ac7a59365/lib/internal/util.js#L341-L347
+    function shiftNoArg(list) {
+      var index = 0;
+      var r = list[index]
+      for (; index + 1 < list.length; index++) {
+        list[index] = list[index + 1];
```
elia, line 29: This is just a nit, but if you have time to fix it that'd be great!

Can you please use an explicit `0` for `var r`, like `var r = list[0]`.
Also I think it might be better to memoize the length and trade two `+ 1` for a `- 1`, like:

```js
var index = 1
var length
for (; index < list.length; index++) {
          list[index - 1] = list[index];
}
```

elia (approved):
Thanks for catching and taking care of this!
I left one nit, let me know if you don't have time to update the PR and I'll merge it as it is 👍

opal/corelib/array.rb
Mogztter: Sure :+1:

Mogztter:
@elia I will keep the focus of this pull request to just `Array#shift` and open a new pull request to generalize the usage of `spliceOne` / `spliceGen`.

opal/corelib/array.rb
Mogztter: Performance wise I don't see any change... I guess the bytecode produced by V8 is the same (or at least equivalent in term of performance)

Mogztter:
@elia I've added a changelog entry and rebased/squashed my work. Let me know if there's anything left to do.

elia (approved):

PR #2120, opened 2020-10-04
Add --rbrequire (-q) option to opal cmdline tool

hmdne:
The short rationale behind it is that for now it's impossible to
require opal libraries like opal-browser (close to impossible - it
needs Gem paths) or opal-erubi (impossible, because it needs to
modify the compiler context to support .erubi files).

The longer rationale is that opal-rspec depends on Opal::CLI. We
can't test opal-webassembly due to that.

### DIFF
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -7,7 +7,7 @@ require 'opal/cli_runners'
 module Opal
   class CLI
     attr_reader :options, :file, :compiler_options, :evals, :load_paths, :argv,
-      :output, :requires, :gems, :stubs, :verbose, :runner_options,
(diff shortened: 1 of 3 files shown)


hmdne:
A small request: can this be backported into the Opal 1.0 series? This will help me improve the documentation on some gems I manage.

hmdne:
I just found out it's not necessarily true about testing. Still I think it would be a good addition, so one could do something like:

`opal -qopal-browser -ropal-browser -c my_script.rb`

hmdne:
There is an `opal-build` command which actually has an `-r` switch which does what this proposed `-q` switch does. I don't know if having two commands doing actually mostly the same thing is really beneficial, but so be it :D.

elia (approved):

Issue #2114, opened 2020-09-26
Bridging syntax (class ... < ...) is confusing


davispuh:
```ruby
require 'native'

class Test < Native(`window`)
end

Test.new
```

This fails with
```js
Uncaught TypeError: can't convert undefined to object
```

It happens in Opal.bridge https://github.com/opal/opal/blob/master/opal/corelib/runtime.js#L1223
```js
$set_proto(native_klass.prototype, (klass.$$super || Opal.Object).$$prototype);
```

elia:
This is because bridging works with JavaScript classes (i.e. constructor functions), and even in that case the syntax would be:

```rb
class Test < `window.constructor`
  def foo; 123; end
end

win = `window`
win.foo # => 123
```

In Ruby the only thing that you can extend are classes, and that holds true in Opal.

_I'm closing this to keep the issues tidy, but feel free to reach out in the slack channel o on stack overflow (tag opalrb) for more help on how to handle native objects and inheritance._

elia:
PS. I didn't test the code above, consider it just demonstrative

davispuh:
There are several issues.
Firstly I want to create a child class of `Native` class (ie. extra methods added) of JavaScript object, not extend JavaScript object itself.
For example
```ruby
class MyWindow < Native(`window`)
  def self.url
    location.href
  end
end

MyWindow.url # => "https://github.com/opal/opal/issues/2114"
```

The main issue here is that `Native` is an instance of `Native::Object` (which is `BasicObject`) and not a Class.
Here Opal has a bug #2123 that it fails with JS error instead of saying that.
Anyway I'm working on implementing that my example works as provided, I've almost finished that.

As for 
```ruby
class Test < `window.constructor`
  def foo; 123; end
end
```

This does opposite, it changes original JS object, not creates a subclass and it works only for some objects. It doesn't work for `window`. It fails with `Uncaught TypeError: can't set prototype of this object`
Because browsers doesn't allow to change some prototypes
```js
> Object.setPrototypeOf(window.constructor.prototype, Object.create(null))
Uncaught TypeError: can't set prototype of this object
```

Also I find this syntax to be very counterintuitive. I don't see why Opal couldn't have used more straightforward way of
```ruby
class `Window`
  def foo; 123; end
end
```

which makes a lot more sense and is clearer/readable.

elia:
> ```ruby
> class MyWindow < Native(`window`)
>   def self.url
>     location.href
>   end
> end
> 
[...]

PR #2123, opened 2020-10-05
Fix `class Class < superclass` for invalid superclass

davispuh:
While trying to do #2114 I found this bug that
```ruby
class TestClass < BasicObject.new
end
```

This fails with `Uncaught TypeError: can't convert undefined to object` but it should raise error `TypeError: superclass must be a Class (BasicObject given)` like Ruby does.

Note that `Class.new(BasicObject.new)` works correctly.
In Ruby specs there is test for `Class.new` but not for `class Class < superclass` so they were passing, but I added such test aswell ruby/spec#792

### DIFF
--- a/spec/opal/core/runtime/bridged_classes_spec.rb
+++ b/spec/opal/core/runtime/bridged_classes_spec.rb
@@ -121,3 +121,19 @@ describe 'Bridged classes in different modules' do
     @bridged.new.some_bridged_method.should == [4, 5, 6]
   end
 end
+
+
+describe "Invalid bridged classes" do
+  it "raises a TypeError when trying to extend with non-Class" do
+    error_msg = /superclass must be a Class/
+    -> { class TestClass < `""`;                  end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `3`;                   end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `true`;                end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Math`;                end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Object.create({})`;   end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Object.create(null)`; end }.should raise_error(TypeError, error_msg)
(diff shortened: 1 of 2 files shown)


spec/opal/core/runtime/bridged_classes_spec.rb
```diff
@@ -121,3 +121,19 @@ def some_bridged_method
...
+  it "raises a TypeError when trying to extend with non-Class" do
+    error_msg = /superclass must be a Class/
+    -> { class TestClass < `""`;                  end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `3`;                   end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `true`;                end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Math`;                end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Object.create({})`;   end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `Object.create(null)`; end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `#{Module.new}`;       end }.should raise_error(TypeError, error_msg)
+    -> { class TestClass < `#{BasicObject.new}`;  end }.should raise_error(TypeError, error_msg)
```
elia, line 136: Maybe I'm missing something… isn't this the same as 

```rb
    -> { class TestClass < Module.new;       end }.should raise_error(TypeError, error_msg)
    -> { class TestClass < BasicObject.new;  end }.should raise_error(TypeError, error_msg)
```

elia (approved):
Just a question about the test case, otherwise looks good! 👍

elia:
@davispuh I amended the commit to get rid of the linting issue

spec/opal/core/runtime/bridged_classes_spec.rb
davispuh: There isn't any reason for it, did based on similarity from others and thought maybe it does make some difference. Anyway removed it. Updated PR with your changes aswell.

elia (approved):

PR #2127, opened 2020-10-12
Improve bridging (handle immutable prototype)

davispuh:
Currently if you'll try to bridge objects with immutable prototype (eg. window, Math) it will fail with
```
TypeError: can't set prototype of this object
```
This PR improves it so it will raise Ruby exception.

Also allow same object to be extended multiple times.


opal/corelib/runtime.js
```diff
@@ -1219,10 +1215,20 @@
...
-    $set_proto(native_klass.prototype, (klass.$$super || Opal.Object).$$prototype);
-    $defineProperty(klass, '$$prototype', native_klass.prototype);
+    if (!native_klass.hasOwnProperty('$$bridge')) {
+      var ourPrototype = (klass.$$super || Opal.Object).$$prototype;
+      try {
+        $set_proto(native_klass.prototype, ourPrototype);
+      } catch(error) {
+        if (error.name != "TypeError") {
+          throw error;
+        }
```
elia, line 1225: You think we can get to a more solid check here? e.g. by using `Opal.is_a(Opal.TypeError)` or something like that?

opal/corelib/runtime.js
```diff
@@ -1219,10 +1215,20 @@
...
+        throw Opal.ArgumentError.$new("Can't bridge immutable prototype");
+      }
+      $defineProperty(native_klass, '$$bridge', klass);
```
elia, line 1228: When a native class is bridged multiple times my understanding is that this one will hold a ref to the most recent one. I would instinctively go toward raising an error or at least a warning.

Do you have a use case for having multiple bridged classes? The current way is to have a single bridged class and, if needed, subclass it.

elia (changes requested):
Thanks for improving this! I have a question about multiple bridged classes from a single native one.

opal/corelib/runtime.js
davispuh: I don't think there's any difference, anyway updated.

opal/corelib/runtime.js
davispuh: It can't be warning or error because it's required for certain functionality. Basically I want to extend native class as I wrote in #2114. But because of current way bridging is implemented this is prerequisite. If you look at test case "can be extendable multiple times" you can see how I'm extending constructor 2 times. This invokes bridging codepath even if I don't really use it. And as you can see from test case it works fine. Technically `$$bridge` will point to 1st klass and not any later definitions but generally it's not really an issue.

Also what if I want to split bridging functionality across multiple files.
Ruby class/modules are open so we can update them in different files, but I actually haven't tested this case for bridging and tbh I don't really care about that...

But there's another use case, I'm doing hot-reloading and for that this is also needed. I unload by deleting `Opal` (and bunch of other properties). But of course this doesn't actually unload any bridged classes. So when I load Opal again it will try to bridge again which this PR will make work fine.

hmdne:
Further discussion will be needed for this PR, as maybe allowing for multiple bridges is not the best idea.

PR #2132, opened 2020-10-21
Implement Native to behave like a Class

### LINKED ISSUE #2113: Native doesn't inherit Object - very surprising
Currently `Native` doesn't inherit `Object` and thus methods like `send`, `method` are not available causing surprising behavior.

In Ruby almost everything inherits from `Object` so you can always use them to call a method. 
I was trying to do same in Opal with Native but it doesn't work.

```ruby
require 'native'

`window.$j = function(arg) { return arg }`
w = Native(`window`)

puts w.send(:$j, 1) # this returns nothing
puts w.public_send(:$j, 2) # same

#puts w.method(:$j).call(3) # raises undefined method `call' for nil

# can't call this way either
#puts w.$t(4)

# this is only way which works but isn't that pretty or Rubyish
puts w['$t'].call(5)
```

Output
```


5
```

davispuh:
This PR implements `Native` to behave both like a Module (as was before) and also like a Class which allows native objects to be extended as Ruby objects.

Consider this example
```ruby
class RubyWindow < Native(`window`)
    def self.url
        location.href
    end
end

puts RubyWindow.url
# https://github.com/opal/opal/
```

In this way this PR also resolves #2113 and makes that `#send` and `#method` works.

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -35,11 +35,9 @@ module Kernel
   def method(name)
     %x{
       var meth = self['$' + name];
-
       if (!meth || meth.$$stub) {
-        #{raise NameError.new("undefined method `#{name}' for class `#{self.class}'", name)};
+        #{::Kernel.raise NameError.new("undefined method `#{name}' for class `#{self.class}'", name)};
       }
-
       return #{Method.new(self, `meth.$$owner || #{self.class}`, `meth`, name)};
     }
   end
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -159,11 +159,11 @@
   // -----
 
   Opal.truthy = function(val) {
-    return (val !== nil && val != null && (!val.$$is_boolean || val == true));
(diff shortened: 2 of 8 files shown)


opal/corelib/kernel.rb
```diff
@@ -35,11 +35,9 @@ def <=>(other)
   def method(name)
     %x{
       var meth = self['$' + name];
-
       if (!meth || meth.$$stub) {
-        #{raise NameError.new("undefined method `#{name}' for class `#{self.class}'", name)};
+        #{::Kernel.raise NameError.new("undefined method `#{name}' for class `#{self.class}'", name)};
```
elia, line 39: This shouldn't be necessary as `#raise` is also implemented by `Kernel` so it should be already available in this context.

elia (changes requested):
The solution is interesting, but I'd like to preserve having Native inherit from BasicObject… is it possible to have a sort of `Native::Kernel` that will add the missing functionality when included in a `Native::Object`?

elia:
I'm closing this as we need to find a different approach to address this issue, feel free to ping me for reopening it 👍

PR #2133, opened 2020-10-27
Fix BuilderProcessors::Processor.handles for rbenv

webdev778:
Ensure to probably match multiple extensions so that it doesn't match `.rbenv`

This is a contribution from the [Interscript project](https://github.com/interscript) of [Ribose](https://github.com/riboseinc).


elia (approved):
Thanks! Nice catch!

webdev778:
Thanks for your review!

ronaldtse:
For the record, this issue was originally encountered here: https://github.com/interscript/interscript/issues/592#issuecomment-717728361

Thanks!

PR #2141, opened 2020-12-12
Implement String#getbyte

Mogztter:
Based on #2140

### DIFF
--- a/opal/corelib/string/encoding.rb
+++ b/opal/corelib/string/encoding.rb
@@ -56,10 +56,6 @@ class Encoding
     raise NotImplementedError
   end
 
-  def getbyte(*)
-    raise NotImplementedError
-  end
-
   def bytesize(*)
     raise NotImplementedError
   end
@@ -234,11 +230,13 @@ end
 class String
   attr_reader :encoding
   attr_reader :internal_encoding
+  `Opal.defineProperty(String.prototype, 'bytes', nil)`
   `Opal.defineProperty(String.prototype, 'encoding', #{Encoding::UTF_8})`
   `Opal.defineProperty(String.prototype, 'internal_encoding', #{Encoding::UTF_8})`
 
   def bytes
-    each_byte.to_a
+    @bytes ||= each_byte.to_a
+    @bytes.dup
   end
 
   def bytesize
@@ -289,7 +287,11 @@ class String
   end
 
   def getbyte(idx)
-    @internal_encoding.getbyte(self, idx)
+    string_bytes = bytes
+    idx = Opal.coerce_to!(idx, Integer, :to_int)
+    return if string_bytes.length < idx
+
+    string_bytes[idx]
   end
 
   # stub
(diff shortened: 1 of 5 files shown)


opal/corelib/string/encoding.rb
```diff
@@ -289,7 +289,11 @@ def force_encoding(encoding)
   end
 
   def getbyte(idx)
-    @internal_encoding.getbyte(self, idx)
+    string_bytes = bytes
```
Mogztter, line 292: Maybe we should memoize the value? Is there a Ruby idioms to do that? `String` is immutable right?

opal/corelib/string/encoding.rb
```diff
@@ -289,7 +289,11 @@ def force_encoding(encoding)
...
+    idx = Opal.coerce_to!(idx, Integer, :to_int)
+    return if string_bytes.length < idx
+
+    bytes[idx]
```
elia, line 296: Why we're not using string_bytes here?

opal/corelib/string/encoding.rb
elia: That can probably be done inside `def bytes` but the we must ensure we return a different array instance each time, anyway the most common idiom is:

```rb
def bytes
  @bytes ||= each_byte.to_a
  @bytes.dup
end
```

opal/corelib/string/encoding.rb
Mogztter: Ooops, my bad :sweat_smile:

opal/corelib/string/encoding.rb
```diff
@@ -232,13 +232,15 @@ def bytesize(string)
 end
 
 class String
+  attr_reader :bytes
```
Mogztter, line 235: I'm using this definition to workaround this issue because apparently, `bytes` is iterable/enumerable.

```
Iterable props defined by Opal on core JS objects is empty for strings FAILED
Expected ["0", "1", "2", "bytes"] to equal ["0", "1", "2"]
SpecExpectationNotMetError: Expected ["0", "1", "2", "bytes"] to equal ["0", "1", "2"]
```


But now the linter is unhappy :grimacing: 
Any idea on how to fix this issue (while making the linter happy!)

elia: I think for string we need to explicitly define any internal instance variable we use with `Opal.defineProperty` like we do for internal coding: `Opal.defineProperty(String.prototype, 'internal_encoding', #{Encoding::UTF_8})`

Mogztter: OK, I will give it a try :+1:

elia (changes requested):
Great! One last thing: can you please also remove the definition of `getbyte` on the base encoding class?

https://github.com/opal/opal/blob/bdbd1bf7f33f1d8baa8ba488310090975c6c9d69/opal/corelib/string/encoding.rb#L59-L61

Mogztter:
@elia Done! I've also added a changelog :wink:

elia (approved):
Thanks for fixing this @Mogztter!
(I squashed all the back and forth commits into a single one and fixed the placement of the new changelog entry)

Issue #2144, opened 2020-12-13
Define properties as Symbol instead of String to avoid conflicts


Mogztter:
The function `Opal.add_stubs` (and other functions as well) relies on `Object.defineProperty` to define new properties on an object.

Currently, we are using `$`-prefixed String as property name. For instance, if we have a (Ruby) class attribute named `id` we will define a property named `$id`. This property can then be accessed in JavaScript using the following code:

```js
console.log(obj.$id)
```

Using String as property name is potentially harmful because other librairies might use/declare the same name.

For instance, the official MongoDB client is the property `$id`. In the following example, the query will always return an empty array if Opal is loaded and `Opal.add_stubs(['$id'])` is called.

```js
const Opal = require('opal-runtime').Opal;
Opal.add_stubs(['$id']);
const { MongoClient } = require('mongodb');
const mongoUrl = "mongodb://localhost:27017";
const client = new MongoClient(mongoUrl);
​
async function main() {
  await client.connect();
  const th = await client.db('database').collection('collectionName').find().toArray();
  console.log(th);
}
main();
```

If the two first lines are removed, then MongoDB returns a list of objects (the MongoDB collection is not empty).

Using Symbol will avoid this kind of issue because a Symbol is unique:

```js
let Sym1 = Symbol("Sym")
let Sym2 = Symbol("Sym")
  
console.log(Sym1 === Sym2) // returns "false"
```

Reference: https://developer.mozilla.org/en-US/docs/Glossary/Symbol

Mogztter:
I can confirm that using Symbol resolves this specific issue with the MongoDB driver.

hmdne:
> It looks like double edged sword. Trading simplicity for safety.
> 
> Maybe it's worth it, having `$xxx` on native objects like `String` is not good... yet it's kinda short and convenient...
> 
> ```ruby
> class String
>   def some() "some"; end
> end
> 
> puts "".some
> puts `"".$some()`
> ```

The current idea is to generate a set of helper symbols, like:

```js
var itself$ = Opal.s("itself"); // ~Symbol.for("itself")
"abc"[itself$]();
```

We will still need to benchmark that for possible performance gains of losses. Though, I like the idea of also adding the legacy properties to non-bridged (ie. non-native) values.

You can also always use Opal.send, like follows:

```js
Opal.send("abc", "itself");
```

The signature of Opal.send is as follows:

```js
  // @param recv [Object] ruby object
  // @param method [Function, String] method body or name of the method
[...]

PR #2145, opened 2020-12-13
Define properties as Symbol

Mogztter:
Since this fix will require a lot of changes I want to make sure that we are on the same page.

For well-known attributes, we can declare symbols in `runtime.js`. For instance:

```js
var $$id_s = Symbol('$$id')
```

Then, we can use this symbol to access a property on an object.
I think we also need to 'export' the symbol:

```js
Opal.$$id_s = $$id_s
```

Note: the `_s` suffix is not necessarily needed. 

For user-defined attributes, we should probably use a Map to store the symbols:

```js
function getPropertySymbol(propertyName) {
  if (Opal.propertySymbols.has(propertyName)) {
    return Opal.propertySymbols.get(propertyName)
  }
  const symbol = Symbol(propertyName)
  Opal.propertySymbols.set(propertyName, symbol)
  return symbol
}
```

This is "mandatory" if we want to call (Ruby) methods in JavaScript:

```js
// proposed
const doc = Opal.Asciidoctor.Document[Opal.$$new_s]('foo')
console.log(doc[Opal.propertySymbols.get('$title')]())
```

For reference, here's how we can (Ruby) methods in JavaScript today:
```js
// current
const doc = Opal.Asciidoctor.Document.$new('foo')
console.log(doc.$title()) 
```

Again the naming is not definitive.

Alternatively, we could assign the symbols on the Opal object, and use `Object.getOwnPropertySymbols()` to retrieve them.

```js
Opal[Symbol('$name')] = '' // we need to set a value but we won't use it
console.log(Object.getOwnPropertySymbols(Opal)) // [Symbol('$name')]
```

ref #2144

### DIFF
--- a/opal/corelib/basic_object.rb
+++ b/opal/corelib/basic_object.rb
@@ -14,11 +14,11 @@ class BasicObject
 
   def __id__
     %x{
-      if (self.$$id != null) {
-        return self.$$id;
+      if (self[Opal.$$id_s] != null) {
+        return self[Opal.$$id_s];
       }
-      Opal.defineProperty(self, '$$id', Opal.uid());
-      return self.$$id;
+      Opal.defineProperty(self, Opal.$$id_s, Opal.uid());
+      return self[Opal.$$id_s];
     }
   end
 
--- a/opal/corelib/class.rb
+++ b/opal/corelib/class.rb
@@ -17,7 +17,7 @@ class Class
   def allocate
     %x{
       var obj = new self.$$constructor();
-      obj.$$id = Opal.uid();
+      obj[Opal.$$id_s] = Opal.uid();
       return obj;
     }
   end
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -68,6 +68,12 @@
     enable_stack_trace: true                  // true, false
   };
 
+  var $$id_s = Symbol('$$id')
+  Opal.$$id_s = $$id_s
+  Opal.propertySymbols = {
+    '$$id': $$id_s
+  }
+
   // Minify common function calls
   var $has_own   = Object.hasOwnProperty;
   var $bind      = Function.prototype.bind;
@@ -91,11 +97,11 @@
   // Retrieve or assign the id of an object
   Opal.id = function(obj) {
     if (obj.$$is_number) return (obj * 2)+1;
-    if (obj.$$id != null) {
-      return obj.$$id;
+    if (obj[$$id_s] != null) {
+      return obj[$$id_s];
     }
-    $defineProperty(obj, '$$id', Opal.uid());
-    return obj.$$id;
+    $defineProperty(obj, $$id_s, Opal.uid());
+    return obj[$$id_s];
   };
 
   // Globals table
@@ -2525,7 +2531,7 @@
   Opal.NilClass = Opal.allocate_class('NilClass', Opal.Object, $NilClass);
   Opal.const_set(_Object, 'NilClass', Opal.NilClass);
   nil = Opal.nil = new Opal.NilClass();
-  nil.$$id = nil_id;
+  nil[Opal.$$id_s] = nil_id;
   nil.call = nil.apply = function() { throw Opal.LocalJumpError.$new('no block given'); };
 
   // Errors


Mogztter:
@elia Could you please confirm (or not) that we are committed to this and give your opinion on the current implementation? I want to know if I should push forward this idea or not, thanks! :wink:

aepstein:
Is this on track to be merged?

elia:
@Mogztter sorry for the delay on this, it's a big thing and I think we can surely pursue it, but not hastily. Looks much like the symbol tables of ruby and has surely good properties, I'm just worried about performance and the impact of such a change. Not that I wouldn't do it, but needs to be properly assessed with benchmarks and all of that.

Additionally this surely will require a major release, so we're targeting v2 at least. 

Would be difficult to generate some micro benchmarks on the `$id` change in this PR?

Mogztter:
> sorry for the delay on this

No worries, thanks for your feedback.

> it's a big thing and I think we can surely pursue it, but not hastily. 

OK :+1: 

> Additionally this surely will require a major release, so we're targeting v2 at least.

I agree, this will definitely require a major release :smile: 

> Would be difficult to generate some micro benchmarks on the $id change in this PR?

I don't think so, I will give it a try.

Mogztter:
I cannot see any notable difference using the following benchmark code:

**bm_object_id.rb**
```rb
Benchmark.ips do |x|
  x.report('object_id') do
    "foo".object_id
    true.object_id
    1.object_id
    2.object_id
    5.object_id
  end

  x.compare!
end

```

##### Command

```
$ bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -ropal-parser -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_object_id.rb
```

##### master (Node v14.15.4)


```
           object_id   107.018k i/100ms
           object_id      1.046M (± 3.4%) i/s -      5.244M in   5.018278s
```

```
           object_id   104.345k i/100ms
           object_id      1.035M (± 1.9%) i/s -      5.217M in   5.041630s
```

##### branch (Node v14.15.4)

```
           object_id    94.928k i/100ms
           object_id      1.067M (± 5.2%) i/s -      5.316M in   4.999521s
```

```
           object_id   110.260k i/100ms
           object_id      1.101M (± 0.8%) i/s -      5.513M in   5.006067s
```

Mogztter:
Just notice that you did add a benchmark in late 2018:

`bm_js_symbols_vs_strings`
```
       global symbol   572.856k i/100ms
              symbol   662.695k i/100ms
               ident   542.531k i/100ms
              string   655.005k i/100ms
       global symbol      5.299M (± 7.6%) i/s -     26.351M in   5.003901s
              symbol      6.149M (± 7.5%) i/s -     31.147M in   5.097281s
               ident      6.228M (± 5.7%) i/s -     31.467M in   5.070651s
              string      5.902M (±11.0%) i/s -     29.475M in   5.060506s

Comparison:
               ident:  6227694.9 i/s
              symbol:  6148707.8 i/s - same-ish: difference falls within error              
              string:  5902396.3 i/s - same-ish: difference falls within error       
       global symbol:  5298877.6 i/s - 1.18x  (± 0.00) slower
````

Global symbol seems a bit slower but otherwise performance are same-ish.

**Comment by elia** [comment]
@Mogztter I ran [a quick experiment](https://github.com/opal/opal/compare/elia/symbols-experiment) storing symbols in `Opal.$`, calling method doesn't work but this is how compiled code looks like:

```rb
opal:elia/symbols-experiment ⤑ bin/opal -e 'def asdf; 123; end; def qwer?;456; end; p [asdf, qwer?]' -cEO                                                         ~/C/opal
/* Generated by Opal 1.1.0.rc1 */
(function(Opal) {
  var $asdf$1, $qwer$ques$2, self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.$$$, $$ = Opal.$$;

  Opal.add_stubs(['$p', '$asdf', '$qwer?']);
  
  
  Opal.def(self, Opal.$.asdf, $asdf$1 = function $$asdf() {
    var self = this;

    return 123
  }, $asdf$1.$$arity = 0);
  
  Opal.def(self, Opal.$['qwer?'], $qwer$ques$2 = function() {
    var self = this;

    return 456
  }, $qwer$ques$2.$$arity = 0);
  return self[Opal.$.p]([self[Opal.$.asdf](), self[Opal.$['qwer?']]()]);
})(Opal);
```

I also tried a variation on that benchmark to see how storing the symbol in a variable influenced performance, and basically there's no difference:

```
=== Files: benchmark-ips/bm_js_symbols_vs_strings.rb
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_js_symbols_vs_strings.rb
  live global symbol    89.663k i/100ms
stored global symbol    92.726k i/100ms
       stored symbol    95.003k i/100ms
               ident    96.683k i/100ms
              string    94.957k i/100ms
  live global symbol      6.806M (± 4.6%) i/s -     33.982M in   5.004508s
stored global symbol      7.258M (± 3.4%) i/s -     36.256M in   5.001710s
       stored symbol      7.158M (± 3.9%) i/s -     35.721M in   4.998628s
               ident      7.569M (± 4.8%) i/s -     37.803M in   5.006197s
              string      7.512M (± 4.6%) i/s -     37.508M in   5.004033s

Comparison:
               ident:  7568828.8 i/s
              string:  7511836.0 i/s - same-ish: difference falls within error
stored global symbol:  7258142.5 i/s - same-ish: difference falls within error
       stored symbol:  7158154.6 i/s - same-ish: difference falls within error
  live global symbol:  6805995.8 i/s - 1.11x  slower
```

Mogztter:
> I also tried a variation on that benchmark to see how storing the symbol in a variable influenced performance, and basically there's no difference:

That's a good news 👍

>  I ran a quick experiment storing symbols in Opal.$, calling method doesn't work but this is how compiled code looks like:

Looks good, I will extend your experiment 🤓

ggrossetie:
You should read the issue to understand why this change is needed/proposed.

https://github.com/opal/opal/issues/2144

The proposed syntax could be simplified/aliased to become much shorter/readable.

al6x:
Thanks. It kinda double edged sword. Trading simplicity for safety. Maybe it's worth it.

PR #2150, opened 2020-12-31
Parser 3.0: Support beginless and endless ranges

hmdne:
This commit adds support for beginless and endless ranges as in `(..1)` and `(1..)`.

I tried to test it to the best of my abilities, but Ruby specs are quite outdated as of now. So I mark it as a work in progress until those tests are updated.

### DIFF
--- a/opal/corelib/range.rb
+++ b/opal/corelib/range.rb
@@ -153,6 +153,7 @@ class Range
   def size
     infinity = Float::INFINITY
 
+    return 0 if (@begin == infinity && !@end.nil?) || (@end == -infinity && !@begin.nil?)
     return infinity if `infinite(self)`
     return nil unless Numeric === @begin && Numeric === @end
 
--- a/opal/corelib/range.rb
+++ b/opal/corelib/range.rb
@@ -21,7 +21,7 @@ class Range
   end
 
   %x{
-    function infinite(self) {
+    function is_infinite(self) {
       if (self.begin === nil || self.end === nil ||
           self.begin === -Infinity || self.end === Infinity ||
           self.begin === Infinity || self.end === -Infinity) return true;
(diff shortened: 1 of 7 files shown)


hmdne:
I added a commit 9b4e10c adding compiler tests btw

opal/corelib/range.rb
```diff
@@ -236,11 +266,11 @@ def bsearch(&block)
   end
 
   def to_s
-    "#{@begin}#{@excl ? '...' : '..'}#{@end}"
+    "#{@begin.nil? ? '' : @begin}#{@excl ? '...' : '..'}#{@end.nil? ? '' : @end}"
   end
 
   def inspect
-    "#{@begin.inspect}#{@excl ? '...' : '..'}#{@end.inspect}"
+    "#{@begin.nil? ? '' : @begin.inspect}#{@excl ? '...' : '..'}#{@end.nil? ? '' : @end.inspect}"
```
elia: These two probably work also with a simple `&&`, e.g.:

```rb
@end && @end.inspect
```

I don't think we should worry about having false here, as I'm quite sure it's not admissible as a value for a range edge.

opal/corelib/range.rb
```diff
@@ -20,15 +20,35 @@ def ===(value)
     include? value
   end
 
+  %x{
+    function infinite(self) {
```
elia, line 24: Would be an issue renaming this to `is_infinite`? it reads better

elia (approved):
Just a couple of nits, neither of them is blocking, I also 👍 on code climate.

(Please get rid of the commit I pushed, I think I wanted to fix something and merge but then got sidetracked and didn't cleanup 🙏)

opal/corelib/range.rb
hmdne: Well, weird as it may look but it is...

```ruby
[8] pry(main)> (true..true)
=> true..true
[9] pry(main)> (false..false)
=> false..
[10] pry(main)> (true..false) # Type mixing happens here and probably that's why it fails.
ArgumentError: bad value for range
from (pry):9:in `__pry__'
[11] pry(main)> 
```

PR #2151, opened 2020-12-31
Parser 3.0: methods

hmdne:
Itself not related to the parser, but implements a few methods that were added between Ruby 2.6 and 3.0.


opal/corelib/comparable.rb
```diff
@@ -80,7 +80,14 @@ def between?(min, max)
     true
   end
 
-  def clamp(min, max)
+  def clamp(min, max=nil)
+    if Range === min && max.nil?
+      max = min.end
+      max = Float::INFINITY if max.nil?
+      min = min.begin
+      min = -Float::INFINITY if min.nil?
```
elia: Can this be achieved with a simple `||`? I don't think we should consider `false` as a possible valid range edge

```rb
max = min.end || Float::INFINITY
min = min.being || -Float::INFINITY
```

opal/corelib/comparable.rb
```diff
@@ -80,7 +80,7 @@ def between?(min, max)
     true
   end
 
-  def clamp(min, max=nil)
```
elia: Can you please squash the rubocop fixes to the initial commit in which the change was introduced?

opal/corelib/comparable.rb
```diff
@@ -81,11 +81,15 @@ def between?(min, max)
...
-      min = -Float::INFINITY if min.nil?
+    if max.nil?
+      if Range === min
+        max = min.end
+        max = Float::INFINITY if max.nil?
+        min = min.begin
+        min = -Float::INFINITY if min.nil?
+      else
+        raise TypeError, "wrong argument type #{min.class} (expected Range)"
+      end
```
elia, line 92: Can you please add a quick explanation in the commit on why this was necessary?

elia (changes requested):
Great work! ✨ 
Just a question, and the request to squash the rubocop fixes to the original commits.

opal/corelib/comparable.rb
hmdne: false is a possible value and doesn't mean it's an infinite range

hmdne: i refactored this part anyway :)

hmdne:
I will need to do a little bit of more work on this. But it seems good enough, of 700 fails added by a version change, this changeset removes 100 of them.

opal/corelib/comparable.rb
elia: No, please if it's a possible range it's fine the way it is 👍 I just wasn't able to construct a range with a boolean, until now, it worked with `false..false`

elia:
> I will need to do a little bit of more work on this. But it seems good enough, of 700 fails added by a version change, this changeset removes 100 of them.

@hmdne I'm always ok with splitting if you want to have the part of this that's already ok merged right away

hmdne:
@elia If you see a problem to happen since some commit, we can split until that commit.

What do you think about the new approach to the #clamp? I basically ported this: https://github.com/ruby/ruby/blob/master/compar.c#L222-L252 (but to Javascript this time) and I found out that the entire module can be simplified.

hmdne:
By the way, this force-push is only changed from f1c0c4f onwards (not counting the squashes).

elia (approved):
Great work! 
(I refined a bit the commits so that each fixed spec is enabled along with the fix)

PR #2153, opened 2020-12-31
Parser 3.0: Forward arguments support: def a(...); puts(...); end

hmdne:
Support 2.7 + 3.0 argument forwarding.

Now instead of:

```
def method_missing(meth, *args, **kwargs, &block)
  send(:"do_#{ meth }", *args, **kwargs, &block)
end
```

We will be able to write:

```
def method_missing(meth, ...)
  send(:"do_#{ meth }", ...)
end
```

WIP again due to tests not present yet.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,6 +1,7 @@
 ### Added
 
 - Support for multiple arguments in Hash#{merge, merge!, update} (#2187)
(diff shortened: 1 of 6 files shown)


elia:
Can you please add a spec for this rewriter, similar to the ones we have for the other ones?

hmdne:
Done :)

PR #2159, opened 2021-01-04
Output less code for opal_engine_check

### LINKED ISSUE #1965: Remove blocks when the condition is unless RUBY_ENGINE == 'opal'
For example:
```rb
unless RUBY_ENGINE == 'opal'
  puts 'not opal'
end
```

produces:

```js
(function(Opal) {
  var self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.const_get_qualified, $$ = Opal.const_get_relative, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$==']);
  if ($$($nesting, 'RUBY_ENGINE')['$==']("opal")) {
    return nil
  } else {
    return nil
  }
})(Opal);
```

To reduce the generated file, it could be nice to remove these blocks:
```js
(function(Opal) {
  // noop
})(Opal);
```

Maybe we should add this feature behind a flag ?

hmdne:
The rationale for that is that if/else blocks already share the closure in JS (we don't output `let` variables at this time yet), so nothing is expected to break. Another rationale is that old code produced more code than needed, for example it created an empty
else block even if no else block was present, now the worst case scenario is adding an empty `nil`.

This commit fixes #1965.

### DIFF
--- a/lib/opal/rewriters/opal_engine_check.rb
+++ b/lib/opal/rewriters/opal_engine_check.rb
@@ -9,14 +9,12 @@ module Opal
         test, true_body, false_body = *node.children
 
         if skip_check_present?(test)
-          false_body = s(:nil)
+          process(true_body || s(:nil))
+        elsif skip_check_present_not?(test)
+          process(false_body || s(:nil))
+        else
+          super
         end
-
-        if skip_check_present_not?(test)
-          true_body = s(:nil)
-        end
-
-        node.updated(nil, process_all([test, true_body, false_body]))
       end
 
       def skip_check_present?(test)
--- a/spec/lib/rewriters/opal_engine_check_spec.rb
+++ b/spec/lib/rewriters/opal_engine_check_spec.rb
@@ -30,11 +30,11 @@ RSpec.describe Opal::Rewriters::OpalEngineCheck do
             s(:send, ruby_const_sexp, :==, opal_str_sexp)
           end
 
-          it 'replaces true branch with s(:nil)' do
+          it 'replaces the expression with the true branch' do
             expect_rewritten(
               s(:if, check, true_branch, false_branch)
             ).to eq(
-              s(:if, check, true_branch, s(:nil))
+              true_branch
             )
           end
         end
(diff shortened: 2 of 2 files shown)


elia:
I think the original reason for not dropping the unused branch was local variables:

```rb
if RUBY_ENGINE == 'opal'
  puts 'opal'
  a = 1
end

p a

unless RUBY_ENGINE == 'opal'
  puts 'not opal'
  b = 1
end

p b

```

The code above will fail if the unused branch is dropped:

```
opal
1
/private/var/folders/qk/p0zjl56n0klfq3pkm0d2j1j00000gn/T/opal-nodejs-runner-20210131-94791-kpp0tl:23350
  return self.$p(b);
                 ^

ReferenceError: b is not defined
```

To make this work we should at least initialize them with nil, but at that point I'm not sure it's still worth it.

In addition to that we might be better off relying on uglifyjs or similar which should be able to remove dead branches:

image

hmdne:
As far as I understand it currently, it's about (:begin) instructions being nested. I will investigate it further and maybe find some solution.

hmdne:
OK so let me explain this problem in a better way. The branches are _already_ dropped so this bug happens already. Currently a code of this form:

```ruby
something
if RUBY_ENGINE != 'opal'
  do_some_work
end
something_else
```

is converted to:

```ruby
something
if RUBY_ENGINE == 'opal'
  nil
else
  nil
end
something_else
```

any my patch changes the output to:

```ruby
something
nil
something_else
```

The idea of using Uglify at a later step may be fine, but neither Uglify/Terser nor Google Closure Compiler won't optimize out this code - it shouldn't, because it doesn't know if RUBY_ENGINE won't change at runtime. The fact that this rewriter already exists is already used in the "wild" isomorphic code mostly for shielding requires - and if only for that reason, it's a very important macro (actually, the only macro in Opal if I recall correctly).

elia:
@hmdne you're right, my main concern was with local variables rather than method calls but I just verified your example holds true even in that case:

```rb
a = 1
if RUBY_ENGINE != 'opal'
  b = 2
end
c = 3

p [a,b,c]
```

```js
/Users/elia/.asdf/shims/ruby -S opal -cEO -
/* Generated by Opal 1.1.1 */
(function(Opal) {
  var self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.$$$, $$ = Opal.$$, $truthy = Opal.truthy, a = nil, c = nil;

  Opal.add_stubs(['$!=', '$p']);
  
  a = 1;
  if ($truthy($$($nesting, 'RUBY_ENGINE')['$!=']("opal"))) {
    nil};
  c = 3;
  return self.$p([a, b, c]);
})(Opal);
```

So maybe we can even find a way to fix the semantics, but it's fine to merge this as long as we open an issue for fixing it later.

hmdne:
I would say let's merge it for now and improve on it later.

Issue #810, opened 2015-04-24
`call { true or next }` produces invalid js


elia:
```
> opal -e 'call { true or next }'
/private/var/folders/w0/yjfqr9n94ld7ft4j3hlz_fsm0000gn/T/opal-nodejs-runner-20150424-87011-ohchab:14182
  return ((($a = true) !== false && $a !== nil) ? $a : return nil;)}, TMP_1.$$
                                                       ^^^^^^
SyntaxError: Unexpected token return
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

Similar/duplicate issues that make for additional test cases are: https://github.com/opal/opal/issues/1492 https://github.com/opal/opal/issues/2091 https://github.com/opal/opal/issues/2103

elia:
the produced js code is:

`> opal -Oce 'call { true or next }'`
```js
/* Generated by Opal 0.8.0.dev */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  var $a, $b, TMP_1, self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$call']);
  return ($a = ($b = self).$call, $a.$$p = (TMP_1 = function(){var self = TMP_1.$$s || this, $a;

  return ((($a = true) !== false && $a !== nil) ? $a : return nil;)}, TMP_1.$$s = self, TMP_1), $a).call($b)
})(Opal);

/* Generated by Opal 0.8.0.dev */
(function(Opal) {
  Opal.dynamic_require_severity = "error";
  var self = Opal.top, $scope = Opal, nil = Opal.nil, $breaker = Opal.breaker, $slice = Opal.slice;

  Opal.add_stubs(['$exit']);
  return $scope.get('Kernel').$exit()
})(Opal);
```

elia:
another incarnation from #1492: `foo and next if bar`

PR #2160, opened 2021-01-06
Fix `call { true or next }` producing invalid js

elia:
Fixes https://github.com/opal/opal/issues/810

I'm not too happy with this solution in terms of produced code, on the other hand maybe some `if` can be simplified into ternary or `&&`/`||`.

I'll try to retain the simpler form when not needed before putting this into "ready for review".


hmdne:
I would suggest rewriting `and/or` as `if` only if the rhs is one of the `next/return/break` OR rhs.rhs is one of those, ad infinitum.

elia:
@hmdne I thought about it and did some performance testing, and after all I think it's better to converge both and/or and if to a single optimized implementation. If we're able to use a ternary operator whenever it makes sense for both of them is a double win, moreover we'll need to support in logical operators everything that is available for if, in terms of _returnability_ and other features.

@iliabylich you have any considerations on this or reasons not to do this?

lib/opal/rewriters/returnable_logic.rb
```diff
@@ -0,0 +1,33 @@
...
+module Opal
+  module Rewriters
+    class ReturnableLogic < Base
+      def next_tmp
+        @counter ||= 0
+        @counter += 1
+        "$ret_or_#{@counter}"
+      end
+
+      def reset_tmp_counter!
```
iliabylich, line 14: Does anyone use this method? It was a part of another rewriter to simplify testing (i.e. to have consistent `tmp_name_1` in tests)

lib/opal/rewriters/returnable_logic.rb
```diff
@@ -0,0 +1,33 @@
...
+        @counter = nil
+      end
+
+      def on_or(node)
+        lhs, rhs = *node.children
+        lhs_tmp = next_tmp
+
+        node.updated(:if, [s(:lvasgn, lhs_tmp, process(lhs)), s(:js_tmp, lhs_tmp), process(rhs)])
```
iliabylich, line 22: oh I have no idea why Opal got it now. Looks like are great idea to me, and I believe we could implement it much, much earlier.

IIRC sorbet has a very similar idea of "ast desugaring". Basically they rewrite AST to have as few nodes as possible by converting them into other "trivial" nodes that they support by default. For example, they rewrite `case` into a chain of `if`s

iliabylich (approved):
LGTM overall, sorry I should've included both comments into a review, but I missclicked 😄

lib/opal/rewriters/returnable_logic.rb
elia: Yeah, that's why I added it too initially, then I changed the tests in a way that doesn't require it, some cleanup is due there, maybe move the helper up the inheritance chain, I'll open an issue as a reminder 👍

lib/opal/rewriters/returnable_logic.rb
elia: This is great! https://github.com/sorbet/sorbet/blob/master/ast/desugar/Desugar.cc
I'll see if I can back port other stuff from it, and get less and more robust code 👍

PR #2161, opened 2021-01-07
Time#strftime %j should be padded to width 3

daniel-sousa-me:


daniel-sousa-me:
I found this bug and decided to directly submit a PR with it instead of opening an issue. I hope that's ok

hmdne:
```
# bundle exec rake mspec_ruby_nodejs RUBYSPECS=true INVERT_RUNNING_MODE=true
...
Expected 4 examples to fail:
  1) "Time#strftime should be able to print the julian day with leading zeroes"
  2) "Date#strftime should be able to print the julian day with leading zeroes"
```

You can remove those filters from `spec/filters` (so that we have 2 more passing MRI tests :) ). Otherwise, looks good to me.

daniel-sousa-me:
Those are a lot of "known bugs" O.O

PR updated.

hmdne:
Well, it's more like Opal will probably never be 100% bug-for-bug compatible with MRI and those are exceptions to the MRI's upstream test suite. I think we already pass like 3x more tests than we don't and I have ported some software from Ruby to Opal with almost no problem (except for the known issues with String, Regexp and C extensions), not to mention that most of those tests aren't actually checking for things that happen in the real code.

Anyway I'm approving the review of this commit and if you have any other small changes, feel free to submit them in any way you like :)

hmdne (approved):

elia (approved):

elia:
> I found this bug and decided to directly submit a PR with it instead of opening an issue. I hope that's ok

Do this as much as you want! We love this approach :shipit: 



> Those are a lot of "known bugs" O.O

As @hmdne this is out of that's out of the whole ruby spec, in addition some of them are caused by technical limitations (file manipulation, threads, etc.), anyway you can consider it a playground 🤸‍♀️ 😄

daniel-sousa-me:
> Those are a lot of "known bugs" O.O

This was not meant as any kind of criticism, it was just genuine surprise.

I've only come across opal because it's being used for 18xx.games and I've been developing there. Almost all the code is run both on native ruby (on the server) and in Opal (by the players). I think this is only the 2nd time I've come across any difference on their behaviour, so I didn't expect to see so many "known" ones (and that speaks for how big is the chunk that is already fully compatible)

daniel-sousa-me:
In summary: Thank you very much for the great project :)

PR #2184, opened 2021-02-27
Fix Array#difference takes multiple arguments

kachick:
Note 📝 
---

Initially I would like to check `arguments.length` in new `Array#-`, but the ruby/spec looks nothing  and given `arguments.length` always become `1`.
Checked the current compiled code of

```ruby
class Foo
  def bar(arg1, arg2)
    10 - arg1 - arg2 
  end
  
  def -(arg1, arg2)
    10 - arg1 - arg2 
  end
end

foo = Foo.new

p foo.bar(1, 2)
p foo.-(1, 2)
p foo.-(1)
p(foo - 1)
```

Just outputs `7`

スクリーンショット 2021-02-27 20 06 27

```javascript
/* Generated by Opal 1.0.3 */
(function(Opal) {
  function $rb_minus(lhs, rhs) {
    return (typeof(lhs) === 'number' && typeof(rhs) === 'number') ? lhs - rhs : lhs['$-'](rhs);
  }
  var self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.const_get_qualified, $$ = Opal.const_get_relative, $breaker = Opal.breaker, $slice = Opal.slice, $klass = Opal.klass, foo = nil;

  Opal.add_stubs(['$-', '$new', '$p', '$bar']);
  
  (function($base, $super, $parent_nesting) {
    var self = $klass($base, $super, 'Foo');

    var $nesting = [self].concat($parent_nesting), $Foo_bar$1, $Foo_$minus$2;

    
    
    Opal.def(self, '$bar', $Foo_bar$1 = function $$bar(arg1, arg2) {
      var self = this;

      return $rb_minus($rb_minus(10, arg1), arg2)
    }, $Foo_bar$1.$$arity = 2);
    return (Opal.def(self, '$-', $Foo_$minus$2 = function(arg1, arg2) {
      var self = this;

      return $rb_minus($rb_minus(10, arg1), arg2)
    }, $Foo_$minus$2.$$arity = 2), nil) && '-';
  })($nesting[0], null, $nesting);
  foo = $$($nesting, 'Foo').$new();
  self.$p(foo.$bar(1, 2));
  self.$p($rb_minus(foo, 1, 2));
  self.$p($rb_minus(foo, 1));
  return self.$p($rb_minus(foo, 1));
})(Opal);
```

Removing `:-` from https://github.com/opal/opal/blob/53593f7a82377deedd2d8082a92efd9b6848f1e0/lib/opal/nodes/call.rb#L18-L19 fixes the `Array#-` behavior.
But I don't have confident it is desirable way or not.
So just fixed current bugs of Array#difference in this PR 😅

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -210,32 +210,46 @@ class Array < `Array`
     `self.concat(other)`
   end
 
-  def -(other)
-    other = if Array === other
-              other.to_a
-            else
-              `$coerce_to(other, #{Array}, 'to_ary')`.to_a
-            end
-
+  def difference(*others)
     return [] if `self.length === 0`
-    return `self.slice()` if `other.length === 0`
+
+    others.map! do |other|
+      if Array === other
+        other.to_a
+      else
+        `$coerce_to(other, #{Array}, 'to_ary')`.to_a
+      end
+    end
+
+    result = []
+    hash = {}
 
     %x{
-      var result = [], hash = #{{}}, i, length, item;
+      var i, length, item;
+    }
 
-      for (i = 0, length = other.length; i < length; i++) {
-        $hash_put(hash, other[i], true);
+    others.each do |other|
+      %x{
+        for (i = 0, length = other.length; i < length; i++) {
+          $hash_put(hash, other[i], true);
+        }
       }
+    end
 
+    %x{
       for (i = 0, length = self.length; i < length; i++) {
         item = self[i];
         if ($hash_get(hash, item) === undefined) {
           result.push(item);
         }
       }
-
-      return result;
     }
+
+    result
+  end
+
+  def -(other)
+    difference(other)
   end
 
   def <<(object)
@@ -882,8 +896,6 @@ class Array < `Array`
     %x{filterIf(self, $falsy, block)}
     self
   end
-
-  alias difference -
 
   def dig(idx, *idxs)
     item = self[idx]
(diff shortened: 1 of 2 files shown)


opal/corelib/array.rb
```diff
@@ -210,32 +210,46 @@ def +(other)
...
     return [] if `self.length === 0`
-    return `self.slice()` if `other.length === 0`
+
+    others.map! do |other|
+      if Array === other
+        other.to_a
+      else
+        `$coerce_to(other, #{Array}, 'to_ary')`.to_a
+      end
+    end
```
elia, line 222: Can you please move this into the main cycle to save an iteration?

opal/corelib/array.rb
```diff
@@ -210,32 +210,46 @@ def +(other)
...
+    hash = {}
 
     %x{
-      var result = [], hash = #{{}}, i, length, item;
+      var i, length, item;
+    }
 
-      for (i = 0, length = other.length; i < length; i++) {
-        $hash_put(hash, other[i], true);
+    others.each do |other|
```
elia, line 231: As for the `Hash` methods it's better to stay native on such important objects such as Array, can you please replace the ruby cycle with a javascript `for` loop?

elia (changes requested):
Thanks for this, I added a couple of comments about moving stuff to a native for loop. 

But I was also wondering if it would make sense to put the single-item implementation inside `Array#-` and let `Array#difference` iteratively use `#-`. In this way `#-`, which is the most commonly used would be faster and `#differece` would still be correct. What are your thoughts on this?

opal/corelib/array.rb
kachick: Ah, it sounds I learned in https://github.com/opal/opal/pull/2187#discussion_r58901336 😓 
I'll try!

opal/corelib/array.rb
kachick: I'll try! 🙇

kachick:
>But I was also wondering if it would make sense to put the single-item implementation inside Array#- and let Array#difference iteratively use #-. In this way #-, which is the most commonly used would be faster and #differece would still be correct. What are your thoughts on this?

Hmm… it sounds reasonable way 🤔

hmdne:
We merged #2237 instead of this, because of a wider scope we were able to cover. Thank you for your work, @kachick!

kachick:
Sorry for late reply 🙇‍♂️ Thank you!

Issue #1960, opened 2019-03-21
Calling Hash#merge with no arguments or multiple arguments fails


mojavelinux:
Opal fails to merge a Hash when no arguments are given:

```rb
h = { a: 'a' }
h.merge
```

It also fails when multiple arguments are given:

```rb
h = { a: 'a' }
h.merge({ b: 'b' }, { c: 'c' })
```

It also fails if a single empty Hash is given and the argument list is not surrounded in brackets:

```rb
h = { a: 'a' }
h.merge {}
```

It works if a single empty Hash is given and the argument is surrounded in brackets (which indicates the previous example is a parser problem):

```rb
h = { a: 'a' }
h.merge({})
```

All of these succeed in CRuby 2.6 (as this is a 2.6 feature). The same examples fail in JRuby and CRuby < 2.6.

See http://ruby-doc.org/core-2.6/Hash.html#method-i-merge

mojavelinux:
Here's an implementation:

```rb
Hash.prepend(Module.new do
  def merge *args
    args.length > 1 ? args.inject(self) {|acc, arg| acc.merge arg } : (super args.fetch 0, {})
  end
end) if (Hash.instance_method :merge).arity == 1
```

elia:
Fixed by 4d93700d5 / #2187

PR #2187, opened 2021-02-27
Update Hash#{merge, merge!, update} as taking multiple arguments

kachick:
Fix https://github.com/opal/opal/issues/1960

### DIFF
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -767,16 +767,15 @@ class Hash
   end
 
   def merge!(*others, &block)
-    others.map! { |other| Opal.coerce_to!(other, Hash, :to_hash) }
-
-    `var i, other_keys, length, key, value, other_value;`
-    others.each do |other|
-      %x{
+    %x{
+      var i, j, other, other_keys, length, key, value, other_value;
+      for (i = 0; i < others.length; ++i) {
+        other = #{Opal.coerce_to!(`others[i]`, Hash, :to_hash)}
         other_keys = other.$$keys, length = other_keys.length;
 
         if (block === nil) {
-          for (i = 0; i < length; i++) {
-            key = other_keys[i];
+          for (j = 0; j < length; j++) {
+            key = other_keys[j];
 
             if (key.$$is_string) {
               other_value = other.$$smap[key];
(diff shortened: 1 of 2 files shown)


opal/corelib/hash.rb
```diff
@@ -762,57 +762,56 @@ def length
...
-      if (!other.$$is_hash) {
-        other = #{Opal.coerce_to!(other, Hash, :to_hash)};
-      }
+  def merge!(*others, &block)
+    others.map! { |other| Opal.coerce_to!(other, Hash, :to_hash) }
+
+    `var i, other_keys, length, key, value, other_value;`
+    others.each do |other|
+      %x{
+        other_keys = other.$$keys, length = other_keys.length;
```
elia, line 775: Unless there's a good reason I think it's better to keep the iteration in a JS `for` loop, to keep it as fast as possible.

`map!` and `merge` both introduce additional checks and overhead, but since the array is coming from `*others` we can expect it to always be a JS array (vs. Enumerator etc.).

A single for loop is also probably the implementation used by MRI:

```
>> (a= {}).merge!({a: 1}, 1, "a")
Traceback (most recent call last):
        6: from /Users/elia/.asdf/installs/ruby/3.0.0/bin/irb:23:in `<main>'
        5: from /Users/elia/.asdf/installs/ruby/3.0.0/bin/irb:23:in `load'
        4: from /Users/elia/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0/gems/irb-1.3.0/exe/irb:11:in `<top (required)>'
        3: from (irb):1:in `<main>'
        2: from (irb):2:in `rescue in <main>'
        1: from (irb):2:in `merge!'
TypeError (no implicit conversion of Integer into Hash)
>> a
=> {:a=>1}
>> 
```

As you can see the first item is merged even if the second raises an error.

elia (changes requested):
This is great 👍, just a concern on the iterations for coercion and merging

opal/corelib/hash.rb
kachick: Thanks for your pointing out!

>A single for loop is also probably the implementation used by MRI:

I didn't dig MRI implementation 🙇 Agreed!

Updated in https://github.com/opal/opal/pull/2187/commits/f429e1efca40eae18513c94272cb1b2f36e4d558 🙏

elia (approved):
👍👍👍

elia:
✅ Merged in 4d93700d5 with squashed commits and the changelog entry, thanks again!

kachick:
Thank you! 😄

PR #2191, opened 2021-03-07
Fix Number#to_i with huge number

ysakasin:
This pull request fixes the behavior of `Number#to_i` to that of `Math.trunc()` of JavaScript.

Before the fix, `Number#to_i` with a huge number in exponential notation returns a small number due to `parseInt()` specification of JavaScript. `parseInt()` converts numbers to strings and then parses it. `parseInt()` doesn't support exponential notation.

If `parseInt()` encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point.

## e.g.
```ruby
# In Opal (before the fix)
1504642339053716000000.to_s #=> "1.504642339053716e+21"
1504642339053716000000.to_i #=> 1
```

## Reasons for adding unsupported filter

- `Float#<=>`, `Integer#<=>`
  -  It compare `Infinity` and `Float::MAX.to_i*2`. `Float::MAX.to_i*2` will be Infinity in Opal
- `Time#yday`
  - `with_timezone` helper is not working in Opal. It uses `ENV['TZ']`

## Ref.
- parseInt()
  - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
- Math.trunc()
  - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,7 @@
 ### Fixed
 
 - Encoding lookup was working only with uppercase names, not giving any errors for wrong ones (#2181, #2183, #2190)
+- Fix `Number#to_i` with huge number (#2191)
 
 ### Changed
 ### Deprecated
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -865,7 +865,7 @@ class Number < Numeric
   end
 
   def to_i
-    `parseInt(self, 10)`
+    `self < 0 ? Math.ceil(self) : Math.floor(self)`
   end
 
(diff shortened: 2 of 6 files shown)


elia (approved):
Thanks for explaining the reason behind the failing specs, I guess we would need to integrate JS's BigInt support in order to fix them, also given now that's seamless in Ruby.

(I fixed a missing newline at the end of a file and the conflicts in the Changelog)

PR #2194, opened 2021-04-11
Fix String#bytes on strict mode

Mogztter:


opal/corelib/string/encoding.rb
```diff
@@ -232,7 +232,11 @@ class String
   `Opal.defineProperty(String.prototype, 'internal_encoding', #{Encoding::UTF_8})`
 
   def bytes
-    @bytes ||= each_byte.to_a
+    return @bytes.dup if @bytes
```
Mogztter, line 235: `@bytes ||= each_byte.to_a` does not work when running is strict mode because `self.bytes` is not allowed. The following exception is thrown:

```
Exception: Cannot create property 'bytes' on string 'foo'
```

tasks/testing.rake
```diff
@@ -354,12 +354,13 @@ task :minitest_node_nodejs do
   files = %w[
     nodejs
     opal-parser
-    nodejs/test_file.rb
     nodejs/test_dir.rb
+    nodejs/test_encoding.rb
```
Mogztter, line 358: Add the new test file and reorder alphabetically

tasks/testing.rake
```diff
@@ -370,7 +371,7 @@ task :minitest_node_nodejs do
   includes = "-Itest -Ilib -Ivendored-minitest"
 
   sh "ruby -rbundler/setup "\
-     "exe/opal #{includes} #{stubs} -R#{platform} -Dwarning -A --enable-source-location #{filename}"
+     "exe/opal #{includes} #{stubs} -R#{platform} -Dwarning -A --enable-source-location --use-strict #{filename}"
```
Mogztter, line 374: Compile and test with strict mode enabled

test/nodejs/test_opal_builder.rb
```diff
@@ -4,9 +4,13 @@
 
 class TestNodejsOpalBuilder < Test::Unit::TestCase
 
+=begin
+  # throws an error: "FrozenError: can't modify frozen String"
```
Mogztter, line 8: The following test is failing when strict mode is enabled.
Here's the complete stacktrace:

```
TestNodejsOpalBuilder#test_should_build_simple_ruby_file:
FrozenError: can't modify frozen String
    FrozenError: can't modify frozen String
      from <internal:corelib/runtime.js>:2446:7:in `Opal.set_encoding'
      from <internal:corelib/string/encoding.rb>:281:1:in `force_encoding'
      from parser/source/buffer.rb:92:48:in `force_encoding'
      from parser/source/buffer.rb:164:17:in `reencode_string'
      from <internal:corelib/runtime.js>:1729:5:in `Opal.send2'
      from <internal:corelib/runtime.js>:1719:5:in `Opal.send'
      from opal/compiler.rb:208:7:in `source='
      from opal/compiler.rb:198:7:in `parse'
      from opal/builder_processors.rb:84:11:in `compile'
      from opal/builder_processors.rb:0:1:in `compiled'
```

Yet another encoding issue.
Apparently, the source encoding is `ASCII-8BIT`. The parser is then calling `force_encoding` (parser/source/buffer.rb:92) to "reencode" the source to UTF-8 but "FrozenError: can't modify frozen String" is thrown!

https://github.com/opal/opal/blob/7d234115f99d9329d4f09ae9f58dd15f43ac26bf/opal/corelib/runtime.js#L2446

opal/corelib/string/encoding.rb
Mogztter: I'm confused... `bytes` should be available on any String since we are using:

```
Opal.defineProperty(String.prototype, 'bytes', nil)
```

Mogztter: OK I get it, the problem is that in strict mode we cannot assign a property on a `string`. 

```js
(function strict() { 'use strict'; 'foo'.bytes = '' })()
// Uncaught TypeError: Cannot create property 'bytes' on string 'foo'
```

```js
(function strict() { 'use strict'; new String('foo').bytes = '' })()
// OK
```

test/nodejs/test_opal_builder.rb
Mogztter: Found a solution 🎉 
My first analysis was wrong!

When the buffer source is assigned, `parser` will do:

```rb
input = input.dup if input.frozen?
input = self.class.reencode_string(input)
```

https://github.com/whitequark/parser/blob/21684f12eb73ac416fa88ecee7198739a1847596/lib/parser/source/buffer.rb#L163-L164

Unfortunately, `frozen?` will return false despite the fact that `input` is a `string` primitive.
I did implement `frozen?` on String to return `true` when "self" is primitive. Not 100% sure this is a good idea?

In addition, I've added a condition in `force_encoding` to "dup" the string when we are working with a primitive (otherwise `set_encoding` will throw a FrozenError: "can't modify frozen String").


@elia Alternatively, we could call `dup` on `Opal.set_encoding` instead of throwing a `FrozenError` but it might hide issues? We could also call "dup" only if we are running in strict mode. I found this solution to check if we are effectively running in strict mode: https://stackoverflow.com/questions/10480108/is-there-any-way-to-check-if-strict-mode-is-enforced. What do you think?

opal/corelib/string/encoding.rb
```diff
@@ -232,6 +232,13 @@ class String
   `Opal.defineProperty(String.prototype, 'internal_encoding', #{Encoding::UTF_8})`
 
   def bytes
+    # REMIND: required when running in strict mode, otherwise the following error will be thrown:
+    # Cannot create property 'bytes' on string 'abc'
+    %x{
+      if (typeof self === 'string') {
+        self = new String(self)
```
Mogztter, line 239: should we call `self.$dup()`?

opal/corelib/string.rb
```diff
@@ -1841,6 +1841,10 @@ def unpack(format)
   def unpack1(format)
     raise "To use String#unpack1, you must first require 'corelib/string/unpack'."
   end
+
+  def frozen?
+    `typeof self === 'string'`
+  end
```
Mogztter, line 1847: Could be removed... the more I think about it, the more I think we should remove it but waiting Elia's input 😉

opal/corelib/string/encoding.rb
```diff
@@ -277,6 +284,9 @@ def force_encoding(encoding)
 
       if (encoding === self.encoding) { return self; }
 
+      if (typeof self === 'string') {
+        self = self.$dup();
+      }
```
Mogztter, line 289: we call `dup` to create a String object otherwise `Opal.set_encoding` throws an FrozenError.

lib/opal/compiler.rb
```diff
@@ -151,7 +151,7 @@ def option_value(name, config)
 
     # @!method use_strict?
     #
-    # Adds source_location for every method definition
+    # Enables JavaScript's strict mode (i.e., adds 'use strict'; statement)
```
elia, line 154: 🤦‍♂️

test/nodejs/test_opal_builder.rb
elia: About the dup I'm a bit concerned with performance, let's treat strict mode like having `frozen_string_literal: true`, that should do the job for most things.

Since it's solved can you remove the comments?

opal/corelib/string/encoding.rb
elia: thoughts on just not memoizing the bytes array?

e.g.

```rb
    %x{
      if (typeof self === 'string') {
        return #{`new String(self)`.each_byte.to_a};
      }
    }
```

Also do we really need to memoize it?

tasks/testing.rake
elia: If we can control this through an env variable we can add an entry in the test matrix and test both with and without. That said we surely need to start moving in the direction of defaulting to strict mode over time.

opal/corelib/string.rb
elia: On face value it makes sense to me, what considerations are nudging you toward removing it?

opal/corelib/string/encoding.rb
elia: Replacing `self` is very risky and will break all other references to the string, let's keep the FrozenError and solve the issue in some other way. This would be consistent with Ruby using frozen string literals:

```rb
# frozen_string_literal: true
"asdf".force_encoding('ascii') # => FrozenError: can't modify frozen String: "asdf"
```

elia (changes requested):
Thanks for the patience! Finally got the time to review!

lib/opal/compiler.rb
Mogztter: No sweat, it happens 😉

tasks/testing.rake
Mogztter: It's a good idea 👍🏻

opal/corelib/string/encoding.rb
Mogztter: Probably not... sure it will be faster if we call `#bytes` on a loop but I doubt it's an actual/legit use case.

opal/corelib/string.rb
Mogztter: I felt like we were twisting the definition of `frozen?` but you made a good point about `frozen_string_literal`. Let's keep it 🤝

test/nodejs/test_opal_builder.rb
Mogztter: Done!

elia (changes requested):
Just one last nit, otherwise looks great!
Can you please also cleanup the commit history if you have time? I already took the liberty to rebase in order to fix the unrelated CI failures.

opal/corelib/string/encoding.rb
Mogztter: If I remove this line, `USE_STRICT=true bin/rake minitest_node_nodejs` is failing:

```
TestNodejsOpalBuilder#test_should_build_simple_ruby_file:
Opal::SyntaxError: can't modify frozen String
    test/nodejs/fixtures/hello.rb:in `puts "Hello world"'
    FrozenError: can't modify frozen String
      from <internal:corelib/runtime.js>:2446:7:in `Opal.set_encoding'
      from <internal:corelib/string/encoding.rb>:287:1:in `force_encoding'
      from <internal:corelib/string.rb>:243:5:in `force_encoding'
      from parser/lexer/literal.rb:245:7:in `b'
      from parser/lexer/literal.rb:47:22:in `coerce_encoding'
      from <internal:corelib/runtime.js>:1729:5:in `Opal.send2'
      from <internal:corelib/runtime.js>:1719:5:in `Opal.send'
      from <internal:corelib/class.rb>:39:1:in `new'
      from <internal:corelib/runtime.js>:1729:5:in `Opal.send2'
      from <internal:corelib/runtime.js>:1719:5:in `Opal.send'
```

elia (changes requested):
I think we're there, if we fix String#b this should be good to go!
Great work!

opal/corelib/string/encoding.rb
Mogztter: > If the problem is calling String#b in parser we can try to fix that, especially since the current implementation is wrong and tries to force the encoding instead of copying the string:

Sounds good, I will do that 👍🏻

Mogztter:
@elia Thanks for your review, it should be OK now 😉

PR #2218, opened 2021-05-22
A proper SecureRandom support + Random::Formatter module

### LINKED ISSUE #2170: SecureRandom: Math.random() is not secure
https://github.com/opal/opal/blob/master/stdlib/securerandom.rb uses Math.random() which is not cryptographically secure and the name SecureRandom implies that.

hmdne:
This fixes #2170, but also bring us closer to MRI otherwise.

### DIFF
--- a/lib/opal/repl.rb
+++ b/lib/opal/repl.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'opal'
+require 'securerandom'
 
 module Opal
   class REPL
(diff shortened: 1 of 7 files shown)


hmdne:
Can someone security conscious look at this code? Especially at this part https://github.com/opal/opal/pull/2218/files#diff-a9ce3be02847ca4980a871954401caf96f591528229776d33eb71c779ab0221aR39-R47 , because I'm unsure if that's the correct way of doing it

elia (approved):

Issue #746, opened 2015-03-12
mspec's bignum_value helper


vais:
Mspec defines `bignum_value` helper [as follows](https://github.com/rubyspec/mspec/blob/master/lib/mspec/helpers/numeric.rb#L10):
```ruby
  def bignum_value(plus=0)
    0x8000_0000_0000_0000 + plus
  end
```
This number is too big for javascript:
```
ruby -Ilib bin\opal -e "puts 0x8000_0000_0000_0000"
9223372036854776000

ruby -Ilib bin\opal -e "puts 0x8000_0000_0000_0000 + 1"
9223372036854776000
```
There are many specs using this helper. Any ideas on how to proceed? `not_supported_on :opal`?

Ajedi32:
Fyi, on MRI:

```ruby
(0x3FFF_FFFF).class #=> Fixnum
(0x3FFF_FFFF+1).class #=> Bignum
```

So Ruby solves this problem by using an entirely separate class for large numbers. Not sure whether that's feasible to do with Opal. In order to support huge numbers though, I think at some point a Bignum class will be required, even if it's not integrated with the language as well as it is with MRI.

vais:
From now on, any spec failure that involves bignums will be placed under `spec/filters/unsupported` if the entire example fails, or wrapped in `not_supported_on :opal` if only a sub-set of assertions within an example fails. @elia, @meh, @adambeynon, do you guys agree?

elia:
Fine by me

eregon:
One alternative could be to make `bignum_value` to be `INT_MAX` as many JS implems use `int` internally when it fits and goes to a "larger" (so double would be the "Bignum") type if not.

Or the maximum value which can be represented exactly as a `double` (around `2**52`) could also make sense as Fixnum are essentially immediate exact integral values, which `double` can provide in the approximate range `-2**52` to `2**52`.

vais:
@eregon you mean make this change in MSpec? In the end Opal still cannot pass specs that involve bignums though. I'm confused.

eregon:
@vais It's my mistake, I thought `bignum_value` was defined in terms of `fixnum_max` but it is arbitrarily `2**63 + x`. In any case, I think it would first be more useful to know whether Opal wants arbitrary-precision integers or not.

PR #2219, opened 2021-05-22
The Bignum experiment

hmdne:
I just found out that Javascript has Bignums. Well, so this is a very early experiment (no parser support yet)

```ruby
[user@localhost opal]# bin/opal-repl 
>> require 'corelib/bignum'
=> true
>> 2.to_bn + 2
=> 4
>> 2 + 2.to_bn
=> 4
>> 4.to_bn ** 4 ** 4
=> 13407807929942597099574024998205846127479365820592393377723561443721764030073546976801874298166903427690031858186486050853753882811946569946433649006084096
>> 24 > 23.to_bn
=> true
>> 23 > 24.to_bn
=> false
>> 2.to_bn << 100
=> 2535301200456458802993406410752
>> -2.to_bn.abs
=> 2
>> 2.to_bn == 2
=> true
>> 2 == 2.to_bn
=> true
>> 
```

Some Math operations work by coercing to Float.

```ruby
>> Math.sin(2.to_bn)
=> 0.9092974268256817
>> Math.sqrt(2.to_bn)
=> 1.4142135623730951
>> Math.asin(1.to_bn)
=> 1.5707963267948966
>> Math.sqrt(1000000000000000000.to_bn ** 2)
=> 1000000000000000000
>> Math.cbrt(1000000000000000000.to_bn ** 3)
=> 1000000000000000000
>> 
```

Float with Bignum...

```ruby
>> 2.to_bn + 2.2
=> 4.2
>> 2.2 + 2.to_bn
=> 4.2
>> (2.to_bn + 2).class
=> Bignum
>> (2.to_bn + 2.2).class
=> Number
>> 
```

Other methods

```ruby
>> 2.to_bn.times { puts "hi" }
hi
hi
=> 2
>> 
```


hmdne:
Previous issues/prs related to Bignum support:

#746, #862, #885

hmdne:
(Continuing discussion from https://github.com/opal/opal/issues/2231)

@martinvahi I have rebased and cleaned up a little this patch and ensured basic support of Matrix. I will likely need to work on it further., but in general it shouldn't be that complicated.

Note the notes I made:

```
    Preliminary Bignum support
    
    This adds limited Bignum support for Opal.
    - Unlike in recent Ruby, it provides a separate class for Bignum handling
    - Unlike in Ruby, Integers are not automatically promoted to Bignum, you
      need to manually call `#to_bn` on String (for big numbers) or Number
      (for small numbers).
    - To use it, you need to `require "corelib/bignum"`
```

Therefore, a number like `238923942893489389489234892894892983423894289489234` will be a Float. `238923942893489389489234892894892983423894289489234.to_bn` will be an approximate value. Only `"238923942893489389489234892894892983423894289489234".to_bn` will produce an exact value.

Also ensure you read the spec to note what's expected behavior: https://github.com/opal/opal/pull/2219/files#diff-b777712e47c483aa87f5f1d06974701a1287377c6992edd92f714659a59ec8f4

You can use this patch by adding the following to your Gemfile: `gem "opal", git: "https://github.com/opal/opal", ref: "hmdne/bignum"`. Or you can clone the repo: `git clone https://github.com/opal/opal; cd opal; git checkout hmdne/bignum` and use `gem "opal", path: "path/to/cloned/repo/of/opal"`. This patch will be merged in the future, so you will be able to switch to the `master` branch, or a release, once we also release it.

Keep in mind this patch is experimental and may not cover all operations. In particular, if you get an error: `Cannot convert a BigInt value to a number` (note `BigInt` which is JavaScript name for the type), this should be considered a bug and likely there's some method that is missing from the implementation. Feel free to make patches.

Also, keep in mind, that mathematical operations in the `Math` module (like Math.sin or Math.sqrt) first convert an argument to Float, thereby losing precision. This is consistent with what Ruby does.

hmdne:
(On a sidenote, while first working on this patch, I noted that the Numeric system of Ruby could allow to create some nice algebra library, with a syntax of let's say):

```ruby
Algsys { var(:a) == 4 + var(:a)**2 }.solve(:a)
```

PR #2220, opened 2021-05-22
Reimplement Promise to make it bridged with JS native Promise

### LINKED ISSUE #1942: problem with chaining a resolved promise
I am not sure if this is bug or a misunderstanding of how promises work.

```ruby
  1) reactive-record edge cases chaining a resolved promise
     Failure/Error:
       expect_evaluate_ruby do
         x = Promise.new
         x.resolve
         y = x.then { Promise.new }
         z = y.then { |r| r }
         [x.resolved?, y.resolved?, z.resolved?]
       end.to eq([true, false, false])
     
       expected: [true, false, false]
            got: [true, true, false]
```

### LINKED ISSUE #1967: no way to fail out of an Promise's `always` block
```ruby
p = Promise.new
p.always { |x| Promise.error('broken') }.fail { |z| puts "#{z}" }
```

does not work, you can do `raise 'broken'` but what if you want to pass a non exception object on to the failure block?

hmdne:
This commit is to reference discussions #2215 and #2213. This is a preliminary implementation of a bridged Promise. At this moment (and probably in the future) it doesn't implement all that the current Promise has. It's at the same time more and less powerful than the existing implementation.

Fixes #1942 
Fixes #1967


hmdne:
Talking about a difference between what we have currently and what JS has:

```js
> a = 5; Promise.resolve(5).then(function() { a = 6; }); a
5
> a
6
> 
```

We can't check the state in the same tick, we need to wait one. In our current implementation:

```ruby
[user@localhost opal]# opal <<< 'require "promise"; a = 5; Promise.value(5).then { p([:in_prom, a = 6]) }; p [:out_code, a]'
["in_prom", 6]
["out_code", 6]
[user@localhost opal]#
```

The implementation I propose:

```ruby
[user@localhost opal]# bin/opal <<< 'require "promise"; a = 5; Promise.value(5).then { p([:in_prom, a = 6]) }; p [:out_code, a]'
["out_code", 5]
["in_prom", 6]
[user@localhost opal]#
```

As far as I know, mspec doesn't follow Promises (unlike rspec-opal) so tests like this won't work

```ruby
  it 'works after a when' do
    x = 42

    Promise.value(1).then {
      Promise.when Promise.value(2), Promise.value(3)
    }.trace {|a, b|
      `console.log(#{[a,b]})`
      x = a + b[0] + b[1]
    }

    x.should == 6
  end
```

And I can't find any better option to make it work.

hmdne:
Those minitest tests... pass, but what's hidden behind Promises (that is - most) isn't running. For that we need to get #2221 (which is on top of this PR).

elia (commented):
This is great, I remember all the headaches I had when dealing with the single `#then` vs. the native JS behavior.

I think we should be careful with this change not to break any code until we switch to v2.0, so maybe we can let this be an alternative to the current promise implementation and add a warning about it being replaced in v2. About the specs you're right we should find a way to make them work async stuff if we start supporting it.

hmdne:
I actually ported minitest to support async in the async commit series.

elia (approved):
This is great! ✨

PR #2221, opened 2021-05-22
Async/await experiment

hmdne:
This pull request is on top of #2220 and references discussion #2213.

This is also my first attempt at modifying the nodes, so it should be seen mostly as an experiment. It implements a CoffeeScript idiom of handling async/await.

```ruby
require 'promise'
def sleep(s)
  Promise.new do |p|
    `setTimeout(#{p.method(:resolve).to_proc}, #{s}*1000)`
  end
end

def fun_with_await
  puts :hi
  sleep(1).await
  puts :ho
  sleep(1).await
  puts :hi
end

fun_with_await
```


hmdne:
As of now, await in global context is working (though not recommended - also you need to run it with --no-exit then). As for now, no support inside modules/classes (also wouldn't be recommended). There are a few edge cases where async/await isn't propagated (mostly related to nested `(function(){})()` calls) - if this happens, you will get an error suggesting that there's a missing identifier "await".

Also, it seems like uglify doesn't like async/await.

hmdne:
Oh and by the way, if you want to look what was needed to port minitest to async/await, see the last commit. It doesn't look like much, but if you miss something, the problems will propagate, functions won't get called, etc.

hmdne:
Well, now we have minitest fixed! ...Except for strict mode. But then a lot of other things broke. But I think we are on a right track in general :D

hmdne:
sleep will be a very useful method to debug await:

```bash
[user@localhost opal]# bin/opal -rawait <<< 'at_exit { puts 3 }; sleep(1).await; puts 1; sleep(2).await; puts 2; sleep(3).await'
Await functionality is a technology preview, which means it may change its behavior in the future unless this warning is removed. If you are interested in this part, please make sure you track the async/await/promises tag on Opal issues: https://github.com/opal/opal/issues?q=label%3Aasync%2Fawait%2Fpromises
PromiseV2 is a technology preview, which means it may change its behavior in the future until this warning is removed. If you are interested in this part, please make sure you track the async/await/promises tag on Opal issues: https://github.com/opal/opal/issues?q=label%3Aasync%2Fawait%2Fpromises
1
2
3
[user@localhost opal]#
```

hmdne:
Let's consider it mostly ready, working and shielded (ie. ES8 code won't happen unless you specifically use #await).

hmdne:
A message may be missed in commit logs, but: **await in top scope is now absolutely supported**

.eslintrc.await.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
hmdne, line 1: No, it can't be done as a magic comment:
https://github.com/eslint/eslint/discussions/14854?sort=top

hmdne:
I'm working on this btw. We decided that this feature should be disabled by default, but I'd like to extend this a little bit before merging. A proposal for a magic comment would be this:

```
# The default value, disables awaiting altogether
# await: false
# Enable #await
# await: true
# Enabled and also all methods ending with await/await!/await?/await= will await
# await: suffix
# Enabled and also some other method names will await:
# await: sleep, gets, irb
# Enabled, suffix and some other methods:
# await: suffix, sleep, gets, irb
```

janbiedermann:
I very much appreciate this effort, absolutely fantastic!

However i disagree with the syntax and the way of usage, because it seems i cant emulate that behaviour in matz ruby. I am thinking 'isomorphic environment' here or have to wrap code in `if RUBY_ENGINE ....`.

I am referring to the usage shown here:
https://github.com/opal/opal/blob/hmdne/async/test/nodejs/test_await.rb

Let me comment in detail in the following posts.

janbiedermann:
https://github.com/opal/opal/blob/4f57623519c3e52d93aea821b9ab15b8ca060e78/test/nodejs/test_await.rb#L8

is showing a mgic comment. That is ok, it doesnt hurt matz ruby and may enable opal compiler features.

L9 is requireing a module, thats fine too, however, this module could by opal be made available for matz ruby too, to emulate or enable async behavior in matz ruby.

janbiedermann:
Now for the syntax. Matz ruby allows to mark methods, eg:
```ruby
def aquire_world_domination
  # code to ask Pinky and Brain for help
end

private :aquire_world_domination
```
The same way async methods could me marked:
```
def aquire_world_domination
  # code to ask Pinky and Brain for help
end

async :aquire_world_domination
```
This could be emulated by matz ruby simply by defining methods on Module class.

janbiedermann:
for using those methods:
```ruby
def ask_pinky_and_brain_for_help(severity)
  # that must be async, of course
end
async :ask_pinky_and_brain_for_help

def aquire_world_domination
  await do
     ask_pinky_and_brain_for_help(:urgent)
  end
end
async :aquire_world_domination
```
Looks more rubyish and can be emulated by Matz ruby.
So looking into: 

https://github.com/opal/opal/blob/4f57623519c3e52d93aea821b9ab15b8ca060e78/test/nodejs/test_await.rb#L132

this would then look like:

```ruby
await { sleep(0.001) }
```
```ruby
sleep(0.001).__await__
```
What do you think? Would my proposal contradict implementation in any way?

hmdne:
Unfortunately, that API isn't isomorphic and we can hardly do anything about it. We had a lot of talks about the design and we aren't closing anything, marking it still as experimental even after a merge. And I'm not really happy about this solution, but I think it's a step in the right direction.

We can create a shim library for Ruby that defines __await__, each_await, etc.

`await { something }` is possible by just modifying the stdlib/async.rb, same with `await something` or `something.await` (the last one is supported) - provided that you `# await: await`

We can also abuse the magic comment to limit the number of calls that will need __await__.

>`async :aquire_world_domination`
I thought about it, but this nice syntax doesn't apply to blocks. Also, we have to know about what is async compile-time.

About Promises server-side, PromiseV2 is mostly API compatible with PromiseV1 (it implements a few new methods). Implementing await on top of a Promise will require us to run on some kind of a reactor and/or Fibers

janbiedermann:
I should add for clearity, with isomorphic i mean the extended version of isomorphic:
not just browser and node
but instead:
browser, node and Matz ruby

hmdne:
I got that point, but unfortunately isomorphism of this patchset is only happening if you are very careful. Since there's no concept of a Promise in the MRI, there's no simple mapping available for us. This patchset is mostly aimed at usecases like easy porting of software like RSpec.

Again, the API of this and PromiseV2 is marked as experimental which means we will only need to support it for the Opal 1.3 timeline, later (Opal 1.4 / 2.0) we will be able to replace everything with a different implementation or extend and stabilize it.

PR #2223, opened 2021-05-25
Use symbols instead of string for properties

s-leroux:
** draft pull request **


hmdne:
If I would to suggest something... maybe this could be a good idea:

```js
if (typeof Symbol !== "undefined") {
  Opal.s = function(name) {
    if (typeof Opal.s[name] !== 'undefined') return Opal.s[name];
    return Opal.s[name] = Symbol(name);
  };
}
else {
  Opal.s = function(name) {
    if (typeof Opal.s[name] !== 'undefined') return Opal.s[name];
    return Opal.s[name] = name;
  };
}
```

Then we would define a bunch of symbols with:
```js
Opal.s("$$id");
Opal.s("$$is_number");
```
and use them later on as
```js
Opal.s.$$id
Opal.s.$$is_number
```

This approach has a nice property of allowing old JS to work possibly (I know there aren't many). I don't care about it particularly, just that I don't like this Opal.$$id_s syntax. Also this approach will allow the libraries to use this infrastructure to also use their own symbolized names.

s-leroux:
> This approach has a nice property of allowing old JS to work possibly (I know there aren't many). I don't care about it particularly, just that I don't like this Opal.$$id_s syntax. Also this approach will allow the libraries to use this infrastructure to also use their own symbolized names.

I like that idea. I followed the syntax introduced by @Mogztter in  #2145, but I can make the change if he's ok with that.

Mogztter:
> If I would to suggest something... maybe this could be a good idea:

I like the `Opal.s` function.
`Symbol` is supported on all browsers (except IE but at this point IE is long dead and < 1%): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility


> Then we would define a bunch of symbols with [...] and use them later on as [...]

👍🏻 
Elia suggested something similar with `Opal.$`.

s-leroux:
> 👍🏻
> Elia suggested something similar with `Opal.$`.

I didn't fully understand Elia's suggestion. Is `Opal.$` something already existing? Or is this a new symbol like `Opal.s`?

s-leroux:
I've updated the PR with the `Opal.s.*` syntax suggested by @hmdne above.

s-leroux:
I've cleaned up the branch so only the relevant commits appear in the PR

s-leroux:
I have difficulties merging changes from Opal 1.2. hile running the `mspec_ruby_nodejs` test suite, the following tests fail here and I can't find why:

```
  1. Range#step with exclusive end and Float values returns Float values of 'step * n + begin < end'
                                                                                       
    Expected [1, 2.8, 4.6, 1, 19.2, 37.4, 55.599999999999994]                          
to have same value and type as [1, 2.8, 4.6, 1, 19.2, 37.4]                            
    SpecExpectationNotMetError: Expected [1, 2.8, 4.6, 1, 19.2, 37.4, 55.599999999999994]
to have same value and type as [1, 2.8, 4.6, 1, 19.2, 37.4]                            
  from <internal:corelib/runtime.js>:1899:5:in `Opal.send2'
  from <internal:corelib/runtime.js>:1889:5:in `Opal.send'
  from <internal:corelib/error.rb>:28:5:in `new'     
  from <internal:corelib/kernel.rb>:544:23:in `exception'         
  from mspec/expectations/expectations.rb:22:5:in `raise'  
  from mspec/expectations/should.rb:15:9:in `fail_with'   
  from ruby/core/range/step_spec.rb:213:9:in `should'
  from <internal:corelib/basic_object.rb>:119:1:in `instance_exec'             
  from <internal:corelib/runtime.js>:1899:5:in `Opal.send2'
  from <internal:corelib/runtime.js>:1889:5:in `Opal.send'
                          
  2. Hash#transform_keys! returns the processed keys if we broke from the block
                          
    Expected {"c"=>1, "d"=>4} == {"b"=>1, "c"=>2}          
to be truthy but was false                                
    SpecExpectationNotMetError: Expected {"c"=>1, "d"=>4} == {"b"=>1, "c"=>2}
to be truthy but was false                               
  from <internal:corelib/runtime.js>:1899:5:in `Opal.send2'
  from <internal:corelib/runtime.js>:1889:5:in `Opal.send'
  from <internal:corelib/error.rb>:28:5:in `new'      
  from <internal:corelib/kernel.rb>:544:23:in `exception'
  from mspec/expectations/expectations.rb:33:5:in `raise'         
  from mspec/matchers/base.rb:33:7:in `fail_predicate'     
  from mspec/matchers/base.rb:19:5:in `method_missing'
  from ruby/core/hash/transform_keys_spec.rb:93:7:in `=='
  from <internal:corelib/basic_object.rb>:119:1:in `instance_exec'
  from <internal:corelib/runtime.js>:1899:5:in `Opal.send2'
```

s-leroux:
I fixed most of the issues raised by the CI.
However, I have difficulties with that one:

```
rspec ./spec/lib/source_map/index_spec.rb:53 # Opal::SourceMap::Index points to the correct source line
```
The source code contains a comment that should help me updating the tests with the right line/column mapping -- but I don't understand what I'm supposed to do :/

```
  it 'points to the correct source line' do
    # To debug this stuff can be useful to print out all
    # the sources with #inspect_source:
    #
    #   p :compiled
    #   inspect_source compiled
    #
    #   p :generated_code
    #   source_map.source_maps.each do |sub_source_map|
    #     p generated_code: sub_source_map.file
    #     inspect_source sub_source_map.generated_code
    #   end
    #

    expect('jsline1(').to be_mapped_to_line_and_column(0, 4, file: 'js_file.js', map: source_map, source: compiled)
    expect('jsline2(').to be_mapped_to_line_and_column(1, 4, file: 'js_file.js', map: source_map, source: compiled)
    expect('jsline3(').to be_mapped_to_line_and_column(2, 4, file: 'js_file.js', map: source_map, source: compiled)
    expect('$rbline1(').to be_mapped_to_line_and_column(0, 2, file: 'rb_file.rb', map: source_map, source: compiled)
    expect('$rbline2(').to be_mapped_to_line_and_column(1, 2, file: 'rb_file.rb', map: source_map, source: compiled)
    expect('$rbline3(').to be_mapped_to_line_and_column(2, 2, file: 'rb_file.rb', map: source_map, source: compiled)
    expect('jsline4(').to be_mapped_to_line_and_column(3, 4, file: 'js2_file.js', map: source_map, source: compiled)
    expect('jsline5(').to be_mapped_to_line_and_column(4, 4, file: 'js2_file.js', map: source_map, source: compiled)
    expect('jsline6(').to be_mapped_to_line_and_column(5, 4, file: 'js2_file.js', map: source_map, source: compiled)
    expect('$rbline4(').to be_mapped_to_line_and_column(3, 2, file: 'rb2_file.rb', map: source_map, source: compiled)
    expect('$rbline5(').to be_mapped_to_line_and_column(4, 2, file: 'rb2_file.rb', map: source_map, source: compiled)
    expect('$rbline6(').to be_mapped_to_line_and_column(5, 2, file: 'rb2_file.rb', map: source_map, source: compiled)
  end
```

hmdne:
I bet that you would fix it with changing $rblineX( to [Opal.s("rblineX")]

Mogztter:
I'm willing to help get this merge if we all agree that this is the path forward?
Maybe we should take it step by step and split this merge request into smaller pull requests? What do you think?

Mogztter:
> And so, my proposal would be like this: first level is compiler support, so a compiler switch that would enable the new behavior. Second level is runtime switch, which would mean that Opal.s (or something similar) would return Symbols instead of Strings. Both disabled by default.

That's a good idea 👍🏻 

> It has been a long time since this PR was started and things got very different.

How so? Do you mean that it would be better to start from "scratch" (i.e., start a new branch from master)?

> This approach would allow us to slowly phase in the new system into a major release, just along a few other changes we have planned (like PromiseV2 becoming the default Promise).

👍🏻

hmdne:
> How so? Do you mean that it would be better to start from "scratch" (i.e., start a new branch from master)?
> 
Very possibly. master has diverged quite a lot since then, because 1.4 was quite big.

We talked with @elia about this today and @elia suggested this kind of code generation approach. After benchmarking we came to understand that it may even improve performance for some cases:

Currently code like `[1,2,3].join` generates this:

```js
Opal.queue(function(Opal) {/* Generated by Opal 1.4.1 */
  var nil = Opal.nil;

  Opal.add_stubs('join');
  return [1, 2, 3].$join()
});
```

What if it would generate this:

```js
Opal.queue(function(Opal) {/* Generated by Opal 1.5.x */
  var nil = Opal.nil, s$join = Opal.s('join');

  Opal.add_stubs('join');
  return [1, 2, 3][s$join]()
});
```

While in this particular case no gains were reported, rather it's calls like '$include?' that can't be called using the dot operator, but need allocation of a new string every time a call is made.

We lose quite a lot in debugging websites using DevTools then, but #2376 is planned.

Issue #2226, opened 2021-05-26
Proposal: Opal versioning regarding components


hmdne:
We are now at Opal v1.1.1, let's assume we won't broke compatibility until v2.0.0. The version dependencies of our components like opal-rspec, opal-jquery, etc. all have different dependencies baked in and they all need to be updated every release. Let's fix it, so we will have easier time doing v1.x releases.

elia:
Agree 👍 I merged everything, I'll add a note to the readme regarding the version policy.

PR #2227, opened 2021-05-28
Add a "Versioning" section to the README

elia:
Fixes https://github.com/opal/opal/issues/2226


hmdne:
Great idea! I would say this is the public interface we provide:

https://github.com/opal/opal/blob/master/docs/compiled_ruby.md#ruby-from-javascript

Additionally, we will start using private/public in corelib/stdlib.

hmdne (approved):

Issue #2099, opened 2020-05-16
Unable to compile code that contains unicode


Mogztter:
As mentioned on Slack, I get a `Invalid multibyte sequence` when compiling Asciidoctor to JavaScript.
It appears that the following line cannot be parsed: https://github.com/asciidoctor/asciidoctor/blob/8aadc1412267bf161f911e6c4800d08fa3d9263e/lib/asciidoctor/rx.rb#L19

I believe that the root cause is that we are now using `source.unpack('U*')` on the source file:

https://github.com/opal/opal/blob/9cc068f95e680549eb3b325d66b91b4e887d2fcb/lib/opal/parser/patch.rb#L12

Even if I remove the unicode letter: https://www.compart.com/fr/unicode/U+00EB the compilation is *very* slow. With the previous version it takes approximately 30 seconds to transpile Asciidoctor but now it takes more than 8 minutes (I've killed the process so it actually takes even longer).

Might be related: #2094

Issue #2094, opened 2020-05-03
accent crash opal


atomecorp:
any  accented char crash opal ex :
`puts "règle"`

elia:
I can get opal 1.0.3 running on node.js correctly printing the accented char, which version/platform are you getting the crashes on?

Please note the Opal assumes to run inside web-pages encoded in utf-8 when running in a browser.

atomecorp:
﻿Hello Elia,

I use Opal V 1.0.3. 
I have the problem my html file have the  utf-8 tag. 

I’ve try the following code : « puts  ’été’ » on  https://opalrb.com/try/ it show the problem 

Thanks for your help. 


Jean-Eric Godard

atome
CEO
+ 33 (0)6 52 57 55 17
jean-eric.godard@atome.io

74 bis avenue des Thermes - 63400 Chamalières

atomecorp:
it seems that it's the parser that have problem with encoding not Opal.

PR #2235, opened 2021-06-20
Correct the utf-8 unpack implementation

### LINKED ISSUE #2189: Opal 1.x does not eval (on the client) ascii chars with val >= 0x10
Chars with ascii code in the range 0x80 - 0xFF cannot be compiled on the client.

For example:

`bundle exec opal -e 'require "opal-parser"; eval("©")'`

will fail.

Looks like the error is here: 

https://github.com/opal/opal/blob/6f71df4882f385dd0369bd6d4e5fdf1e762897d2/opal/corelib/string/unpack.rb#L49

but that should read `if (charcode < 0x80) utf.push(charcode)`

instead of `if (charcode < 0x100) utf.push(charcode)`  (extra zero)

hmdne:
fixes #2189 #2094 #2099


hmdne:
So this broke unpack with: w, B, C, H, L, N, Q, S, U, V, b, c, h, l, s and v. It seems like the whole unpack implementation depends on this wrong assumption.

hmdne:
OK, the reason is actually very simple. The current implementation works correctly most of the time for both the BINARY encoding and Unicode alike. This is why it's <0x100. Most of the time, but it's not correct in fact, we need to take encoding into account. I'm supplying the correct implementation below.

hmdne:
I'll now rebase it on top of #2236 and remove some filters we were able to pass this time.

hmdne:
The next commit goes a little bit beyond the initial scope. We moved some functions from string.rb to string/encoding.rb, as more functions are now encoding-aware and I implemented a bit more of the encoding library.

We can go 2 ways now, either we make string/encoding.rb mandatory, which I see it already kind of is, or we write some "polyfill" functions. At first I implemented those functions, but decide to not include them in a patch.

```ruby
  # If corelib/string/encoding isn't loaded, below are the methods that
  # will be "correct most of the time".

  alias b itself
  def encode(encoding); self; end
  alias force_encoding encode
  def length; `self.length`; end
  alias size length
  alias bytelength length

  def each_char(&block)
    return enum_for(:each_char) { length } unless block_given?
    %x{
      for (var i = 0, length = self.length; i < length; i++) {
        #{yield `self.charAt(i)`};
      }
    }
    self
  end

  def each_byte(&block)
    return enum_for(:each_byte) { bytelength } unless block_given?
    %x{
      for (var i = 0, length = self.length; i < length; i++) {
        #{yield `self.charCodeAt(i)`};
      }
    }
    self
  end

  # Those don't need to be repeated:
  def bytes
    @bytes ||= each_byte.to_a
    @bytes.dup
  end

  def chars(&block)
    return each_char.to_a unless block

    each_char(&block)
  end
```

hmdne:
Oh and by the way, the main reason for those changes is that Ruby interprets "\x80" differently than "\u0080". Therefore pack and unpack may not be compatible with one another. After I finish this patch, the next one will try to allow strings in Opal like "\x80".

hmdne:
Well... from a seemingly simple problem now we are here. This is still not perfect, but hopefully most people won't notice the encoding problems.

The filters are a mess, but I can assure you there's only 1 new filter added which is related to unsupported EUC-JP encoding.

I'll do benchmarks to ensure there's no big performance regression (String#+ looks scary...). Afterwards, I will also try to optimize the dready `unpack("U*")` call.

hmdne:
I wrote a benchmark suite that measures Opal's performance on various fields, even since v0.8.0 (the interesting thing is that all those versions work on Ruby 3.0... I'll publish the code later on, but it's certainly not pretty). For this particular test, that in particular benchmarks ERB compilation on Opal compiled with Opal I noticed no statistically significant change on this patchset (so this tests both ERB compilation and ERB execution but just now I noticed it doesn't really use #+ too much, so let's try another, now less realistic, benchmark).

```ruby
`console.log("T:load:", +new Date)` # !@

require 'opal'

`console.log("T:iter1:", +new Date)` # @#

str = ''

200000.times do |i|
  str += i.to_s(2)
end

`console.log("T:iter2:", +new Date)` # #$

200000.times do |i|
  str += i.to_s(2)
end

`console.log("T:end:", +new Date)` # $%
```

This benchmark saw 5% slowdown (only between iter1 and end) after the #+ patch.

hmdne:
Now, for the ERB benchmark. The bigger value the worse:

984e9718 (first commit before 6fddca88 aka v1.0.0-237-g6fddca88): 46
in the middle (after 6fddca88 was introduced): 263
826e795 (the last commit of this PR): 61

So, it's still 33% slower than originally, but overall it's 4 times faster. And the benchmark takes into account only compiling being done by Opal compiled by Opal.

Mogztter:
> So, it's still 33% slower than originally, but overall it's 4 times faster. And the benchmark takes into account only compiling being done by Opal compiled by Opal.

Sounds good, I will compile Asciidoctor against your branch 👍🏻

hmdne:
@Mogztter just in case, I may rebase this branch, but it's one of the most complicated branches for the Opal 1.2 process (to end on 21st of July) and may not make it. Feel free to leave any feedback :D

Mogztter:
I can confirm that it's still slower but acceptable and that https://github.com/opal/opal/issues/2099 is fixed 🎉 
Transpiling Asciidoctor now takes 40 seconds (with the previous version it takes approximately 30 seconds).

hmdne:
The performance structure for Opal compiler compiled by Opal looks like this for my particular example of:
```ruby
  out = Opal.compile("def xxyx(a,c,d); if 5; 6; elsif 7; 8; end; end" * 1000)
```

image

It's 6.88s spent on parsing, 924ms on rewriting, 353ms on compiling (while 106ms on flattening the resulting structure).

This is AFTER the patch I just submitted that gained me ~10% compiler optimization on MRI, but may also bring some benefits to Opal-on-Opal.

hmdne:
(This one, I forgot to submit it before)

elia (approved):

Mogztter:
Woot woot 🎉

PR #2237, opened 2021-06-20
Parser 3.0: Array methods

hmdne:
This PR is on top of #2236. It also supersedes #2184 with a bigger context.

This PR includes two kinds of changes:
* union, difference, intersection to support multiple arguments
* to_a, slice/[], uniq, *, difference/-, intersection/&, union/|, flatten to return Array, not a subclass

The second may be a controversial change, as it breaks compatibility with what Ruby used to do. Ref: https://rubyreferences.github.io/rubychanges/3.0.html#array-always-returning-array

It removes 22 filters.

I can split the first part of the change for 1.2, second for 2.0, or we can take both for 1.2. What's your opinion on that?


opal/corelib/array.rb
```diff
@@ -2165,10 +2169,12 @@ def take_while(&block)
   end
 
   def to_a
-    self
+    self.class == Array ? self : Array.new(self)
```
elia, line 2172: Can you please reimplement this in JS to improve performance?

elia (approved):
Looks great, ready to merge once #to_a is fixed

PR #2242, opened 2021-06-27
Opal::Cache, an optional compiler cache

hmdne:
First, maybe how to use it:

```ruby
OPAL_CACHE=1 bundle exec rake mspec_opal_nodejs
```

I made it an ENV variable for now, not Opal::Config, mainly because there's a plethora of ways to call Opal and almost none of them support those options. This thing only works with Opal::Builder, so it doesn't work with opal-sprockets (but it would be trivial to add if it would be wanted, after all, Sprockets has its own cache - working better or worse).

As we all know, Opal compiles things slowly. So let's take probably the biggest Opal application out there, that is, `bundle exec rake mspec_ruby_nodejs`. So on my computer it takes 1 minute and 20 second (both with cache disabled and with cache full miss). But after cache is populated, it takes 23 seconds, almost a minute less.

Now, HOW does it work, well, this thing is a bit hacky I must say, but maybe someone will help me improve it. We cache Opal::Compiler instances (and Opal::SourceMap::File) with hand-crafted marshal_(dump,load) methods. We assume that some methods of Compiler/SourceMap won't be called, so we simply skip their instance variables if possible. Such a compiler would be useful for extracting what you usually want to extract from it, but no more.

But how and where it's stored? On disk, by 2 cache keys.

One cache key is an environment key, which is dependent on compiler options and... mtime of Opal compiler (ie. everything in lib/). The second is file key, which basically means a tuple of (filename, source, options). So you can work on Opal and have things run fast (unless you modify the compiler part, of course).

After running `mspec_opal_nodejs` the cache is 40MB. Deflating it is certainly a good idea.

The cache also isn't pruning itself automatically. It's up to user to do that at this point. At a later time we can think about some LRU, or other method so that it may be seamlessly included as a default option.

### DIFF
--- a/lib/opal/builder_processors.rb
+++ b/lib/opal/builder_processors.rb
@@ -80,11 +80,15 @@ module Opal
       end
 
       def compiled
-        @compiled ||= Cache.find_key_or_exec(self.class, @filename, @source, @options) do
+        @compiled ||= Opal.cache.fetch(cache_key) do
           compiler = compiler_for(@source, file: @filename)
           compiler.compile
           compiler
         end
+      end
+
+      def cache_key
+        [self.class, @filename, @source, @options]
       end
 
       def compiler_for(source, options = {})
--- a/lib/opal/builder_processors.rb
+++ b/lib/opal/builder_processors.rb
@@ -2,6 +2,7 @@
 
 require 'opal/compiler'
 require 'opal/erb'
+require 'opal/cache'
 
 module Opal
   module BuilderProcessors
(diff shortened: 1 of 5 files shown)


hmdne:
After gzipping, we are now down to something like 12MB.

elia:
@hmdne I pushed an extra commit, to allow different cache classes, including a null one for when the engine is opal itself, instead of using the flag.
Also now we have the global cache instance stored in `Opal.cache` meaning that it can be changed and customized, and it's easy to add alternative implementations.

hmdne (commented):
This wasn't submitted? Wow. I guess each day you learn something :D

hmdne:
Anyway, #2278 is a continuation.

PR #2243, opened 2021-06-28
Parser 3.0: Implement pattern matching.

hmdne:
Most of things work, except for the find pattern (`a => [*, b, *]`),
which is experimental and there are no tests for that in the current
suite. Some tests don't pass due to an eval bug (eval doesn't know
that local variables exist).

This implementation is split between compiler and corelib, which can
be thought as a little unorthodox. The reason is that I wanted this out
sooner and have it less buggy on release. In general it works in a way,
that compiler passes something similar to a sexp to a PatternMatching
class implemented in corelib and then corelib works on the actual
pattern matching logic.

Additionally, we added `Hash#except{,!}`. Also, corelib/pattern_matching
defines `#deconstruct{,_hash}` on Struct, Hash and Array, though those
are probably useless outside of pattern matching.

Additionally, we fixed a bug which prevented definition of instance
variables for Opal compiled with Opal.


hmdne:
I took the liberty of adding some new (pending) things on this branch, but I will tidy them up to separate PRs later.

hmdne:
I isolated the Binding and scope variables support to a separate PR, #2256.

elia (approved):

PR #2247, opened 2021-07-03
ObjectSpace & ObjectSpace::WeakMap

hmdne:


hmdne:
Now I have a moment to describe this pull request.

Well, all it does is add destructors and WeakMaps to Opal. In terms of a WeakMap we can have a few other things implemented, like: https://github.com/ruby/weakref/blob/master/lib/weakref.rb (or we can even implement them in terms of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef - but this is a much later addition).

We do "polyfill" the finalization registry (as in - old environments will simply ignore those calls), but not WeakMap, which is supported since IE 11: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

Most of those things work only for keys of type typeof var === "function" and "object", which is contrary to what Ruby does. JS WeakMaps are also non-iterable, contrary to how

hmdne:
We now support all primitive values on WeakMap by deferring to a regular Hash

elia (approved):

PR #2249, opened 2021-07-06
Fix `class ::String`.

### LINKED ISSUE #1475: redefining method on ::Object works oddly
Pretty obscure but still...

``` ruby
require 'spec_helper'

class ::Object

  define_method :some_bar do
    "bar1"
  end

  def some_bar
    "bar2"
  end

end

describe "TestMethod" do

  it 'you can redefine a method' do
    expect("string".some_bar).to eq("bar2")
  end

end
```

Fails ("string".some_bar) is equal to "bar1"

Works fine for some random class, just not ::Object

Was working in 0.8 as well.

hmdne:
The former behavior was to create a subclass of String (as
`"::".$$class` is a String), which was incorrect. The same is also
related to modules.

This should fix #1475.

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -533,7 +533,7 @@
   Opal.klass = function(scope, superclass, name) {
     var bridged;
 
-    if (scope == null) {
+    if (scope == null || scope == '::') {
       // Global scope
       scope = _Object;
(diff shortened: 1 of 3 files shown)


hmdne:
Sample code that was broken:

```ruby
class ::Object # Actually defined String::Object
end
```

elia (approved):

Issue #2136, opened 2020-11-08
Return inside attribution


daniel-sousa-me:
```
def foo bar
   a = case bar
   when 1
     return 3
   end
   return 2
end

puts foo(1)
```

This beautiful and very useful piece of code prints 3 in irb, but in [Opal](https://opalrb.com/try/) it prints 2

Issue #1600, opened 2016-12-05
variables created by case with explicit return differ from MRI


patrickkettner:
```rb
require 'opal'

def foo
  a = true
  b =
    case a
    when true
      puts 'should show'
      return a
    end
  puts 'shouldnt show'
end

foo
```

is Opal
```js
  Opal.defn(Opal.Object, '$test', TMP_1 = function $$test() {
    var self = this, a = nil, b = nil, $case = nil;

    a = true;
    b = (function() {$case = a;if (true['$===']($case)) {self.$puts("should show");
    return a;}else { return nil }})();
    return self.$puts("shouldnt show");
  }, TMP_1.$$arity = 0);
```

In MRI , the function breaks at the explicit return. In Opal, the case is created inside of an IFFE, as a result the return just ends that IFFE, rather than the parent function.

elia:
Thanks for reporting, can you specify the version in which you observed
this?

patrickkettner:
0.10.3

elia:
Ok, confirmed on master too

PR #2252, opened 2021-07-07
Eliminate case node with a rewriter.

hmdne:
This needs to be made more readable, tests to be written... otherwise, it should fix a lot of issues like #2136, #1600


elia:
This is great, the less nodes the better!

hmdne:
In this PR I also took liberty to experiment a little bit :D

With that experiment I created a new "SimpleIf" node to return an expression not a statement if it's possible so that we could get rid of the closures where they aren't needed. A similar thing could be done to Begin statements. Thing is, this doesn't improve performance as I thought it would, it only cleans up the resulting code. Maybe a minifier like Terser or Google Closure Compiler could do a better job here.

Later I found out a way to implement FlipFlops, which aren't as used in Ruby as much, but as a continuation of the If node changes, I decided to implement them.

Actually, I think that rewriters aren't the best with regards to compiler performance. Since all of them run a DFS algorithm on the whole tree, maybe it could be beneficial to connect some (especially those simple ones) rewriters into a single rewriter (running all in parallel). Then I wouldn't have such concerns about performance. With a careful thought we may be able to achieve this without sacrificing the code quality.

hmdne:
Right now I'm quite ambivalent about this. As in, reducing the number or Nodes is certainly with some benefits, but then, converting a when to a deeply nested if will cause an issue of a stack depth. Let's leave this patch for now, maybe revisit it in the future.

hmdne:
What if: we didn't recurse?

hmdne:
To note:
* Opal before allowed for about 5000 case branches for runtime on V8, virtually any number for compilation
* This branch reduces the number of case branches for compilation to 1700
* MRI breaks after 3000

Now I'm rebasing this atop of our compiler optimization branch; next I will do some optimizations for $truthy((x)['$==='](y))

hmdne:
Has been merged in a better form.

PR #2256, opened 2021-07-21
Binding and scope variables support for eval()

hmdne:
(This is implemented on top of #2243)

This is not a full implementation, but may somehow, maybe, possibly, allow us to implement `binding.irb` some day in the future when we find a way to implement correct async/await. Or it can be used with `window.prompt` :)

I made it so that we can pass a few more pattern matching tests. The only ones that don't pass, but are interesting are related to syntax like:

```ruby
a = 1
eval <<~RUBY
  [1,2,3] => [^a, *]
RUBY
```

Those are supported (sans eval), just that `a` isn't recognized as an instance variable in the parser.

For one, it doesn't support newly defined local variables. So things like:

```ruby
eval <<~RUBY
  a = 6
RUBY

a # => NameError
```

don't work and are unlikely to work.


hmdne:
So, I implemented a real binding in this pull request and replaced the previous shim implementation.

```bash
[user@localhost opal]# ruby <<< 'module A; refine Kernel do; def p(*); puts "no debugging!"; end; end; end; using A; p 5'
no debugging!
[user@localhost opal]# bin/opal <<< 'module A; refine Kernel do; def p(*); puts "no debugging!"; end; end; end; using A; p 5'
no debugging!
[user@localhost opal]# 
```

The code expanded:
```ruby
module A
  refine Kernel do
    def p(*)
      puts "no debugging!"
    end
  end
end

using A
p 5
```

The code compiled:

```js
  // using A
  ($refn$3 = [$$($nesting, 'A')], self);
  // p 5
  return $refined_send([$refn$3], self, 'p', [5]);
```

opal/corelib/main.rb
```diff
@@ -5,3 +5,8 @@ def self.to_s
 def self.include(mod)
   Object.include mod
 end
+
+# Compiler overrides this method
+def using(mod)
```
elia: I changed this to be `self.using` so it's defined only on main and not on Object

elia (approved):
I fixed up a couple of things, like missing EOF, one empty comment etc. and rebased on master

PR #2258, opened 2021-07-27
Fast Complex fix

hmdne:
With opal-optimizer, the undef step caused a regression.


hmdne:
Will be fixed in opal-optimizer.

PR #2263, opened 2021-08-07
Opal::Builder::Prefork for blazingly fast multicore compilation times

hmdne:
Well, yes, unfortunately this uses fork, but the architecture is provided so that we can also provide other kinds of multicore builders, like for instance this attempt to use Ractor, or a win32 specific builder (using some win32 equivalent of fork or process spawning - at this time unfortunately I don't have enough win32 experience to say something intelligent).

https://github.com/opal/opal/commit/611a2af66c4ac341e249028bda16021c99d950cc

(Both use a similar approach).

Unfortunately, Ractor is not really suitable for now. In short, I gave up when I found out I would need to patch Ragel:

https://github.com/adrian-thurston/ragel/blob/ragel-6/ragel/rubycodegen.cpp#L100-L104

Such a patch, while possible, (and while it should be done) would require a lot of logistics. Thing is, Ractors are really hard to do right. The illegal operation here is that when the main Ractor defines a class, other Ractors can't access its instance properties.

This patch is done on top of #2242 so that we may use the compiler Marshaling, also we touch similar targets (the thing below is applicable to both PRs). This pull request in general implements multithreading for the following pipelines (anything that uses Opal::Builder, not Opal::Compiler)
- Opal::SimpleServer
- opal command line utilities (opal, opal-repl)
- opal builder via rake
- opal via Tilt (eg for Roda)
- opal specs (rspec, mspec, minitest)

It doesn't touch the following:
- Opal compiled with Opal
- Opal::Sprockets, anything Sprockets (Sprockets unfortunately is single threaded)
- anything Webpack (but in my opinion it may be used for that purpose as well)

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -4,6 +4,7 @@ require 'opal/path_reader'
 require 'opal/paths'
 require 'opal/config'
 require 'opal/cache'
+require 'opal/builder_scheduler'
 require 'set'
 
 module Opal
@@ -55,8 +56,6 @@ module Opal
       processor_extensions.each { |ext| extensions << ext }
     end
 
-
-
     class MissingRequire < LoadError
     end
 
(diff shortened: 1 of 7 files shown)


hmdne:
I have revisited this and rebased it for the latest Opal. There are still issues with perhaps doing it in a similar way as we have pluggable mechanisms for cache.

I want to have this merged for 1.6 or 2.0 however we will call the next release so it doesn't diverge too much anymore.

hmdne:
This is finished now and ready for review.

elia (approved):
👍👍👍

PR #2264, opened 2021-08-09
Modernize Exceptions

hmdne:
This PR makes exceptions a little bit more compatible with contemporary Rubies


elia (approved):
Love this!!!

PR #2265, opened 2021-08-09
Support return from lambda (fixes opal#260)

hmdne:
The lambdas are now behaving exactly as they do in Ruby, they work just
like the def in terms of the return logic.

### DIFF
--- a/lib/opal/nodes/scope.rb
+++ b/lib/opal/nodes/scope.rb
@@ -106,6 +106,24 @@ module Opal
         @type == :def || @type == :defs
       end
 
+      def lambda?
+        iter? && @is_lambda
+      end
+
+      def is_lambda! # rubocop:disable Naming/PredicateName
+        @is_lambda = true
+      end
+
+      def defines_lambda
+        @lambda_definition = true
+        yield
+        @lambda_definition = false
+      end
+
+      def lambda_definition?
+        @lambda_definition
+      end
+
       # Is this a normal def method directly inside a class? This is
       # used for optimizing ivars as we can set them to nil in the
       # class body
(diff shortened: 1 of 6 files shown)


hmdne:
I'm surprised there are no MSpec tests for this behavior...

lib/opal/nodes/scope.rb
```diff
@@ -101,6 +101,24 @@ def def?
         @type == :def || @type == :defs
       end
 
+      def lambda?
+        iter? && @is_lambda
+      end
+
+      def is_lambda!
```
elia, line 108: The lint message can be fixed by append `# rubocop:disable Naming/PredicateName` to the `def` line

spec/filters/bugs/kernel.rb
```diff
@@ -411,4 +411,5 @@
   fails "Kernel.sprintf returns a String in the same encoding as the format String if compatible" # NameError: uninitialized constant Encoding::KOI8_U
   fails "Kernel.sprintf width specifies the minimum number of characters that will be written to the result" # Expected "         1.095200e+02" to equal "        1.095200e+02"
   fails "Kernel.srand returns the previous seed value on the first call" # NoMethodError: undefined method `insert' for "rubyexe.rb"
+  fails "Kernel.lambda treats the block as a Proc when lambda is re-defined" # Expected 2 == 1 to be truthy but was false
```
elia, line 414: Thoughts on supporting re-definition of `lambda` using a check on `$$pristine`?

elia (approved):
👍 

I left a suggestion on how to also support lambda redefinition, but this is more important and we can definitely merge and think about that later

spec/filters/bugs/kernel.rb
hmdne: Nope. This is a compiler-level feature, it doesn't know what runtime does.

hmdne: Unless... we change the returner, but then a returner would need to be compiled also for the parent scope.

PR #2266, opened 2021-08-10
Es6 modules 1 2

janbiedermann:


janbiedermann:
i should probably recreate this PR based on actual file differences so that huge load of commits gets reduced

janbiedermann:
probably obsolete

PR #2268, opened 2021-08-14
CI: Permissively check for introduced V8 deoptimizations

### LINKED ISSUE #1255: Add Auto Optimization/Benchmark Check
In order to keep track of which methods are optimize-able by JS runtimes, do something like this:

https://gist.github.com/wied03/76cb79e6076e48eee9a5

Probably need to focus on only methods in corelib as the "superset" to test against.

Key points from Gitter:
- Use text based output (not HTML reports)
- Fail the build (use Travis here) if new unoptimized methods appear
- Have a white list of class#method items that are "allowed" to be not optimized, and fail the build if a class#method that's not on that list is not optimized. This would catch regressions, and would force one to add a class#method to the list explicitly if there's no other way to implement the method. This would also let us track progress on optimization, as the list gets shorter over time.
- Might need each method check to run in its own process. The reason I mention it is that functions can get deoptimized for being too polymorphic. For example, if you call a function with more than some VM-specific number of different types of arguments (pretty sure in Safari it's 4, might be 3 in Chrome), it can decide to deoptimize the function because the cost (both performance and memory) of optimizing it for each different type of argument potentially outweighs the performance you gain from it.

hmdne:
Co-authored-by: Brady Wied <wied03@users.noreply.github.com>

This fixes #1255

What happens on deoptimization:
- https://github.com/opal/opal/runs/3328310772?check_suite_focus=true

With pull request we check against the branch we apply the pull request for (master most probably).

Otherwise, we check against the previous commit.

Deoptimization failures are not fatal. But they can serve us some idea what went wrong.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,11 @@ jobs:
             ruby: 3.0
           - name: timezone
             ruby: 3.0
+          - name: optstatus
+            ruby: 3.0
+            permissive: true
+            fetchdepth: '0'
+            command: bin/rake optstatus_compare
 
           # Currently failing:
           # - ruby: truffleruby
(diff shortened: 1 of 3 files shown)


hmdne:
Based on those tests - it all seems to work. Of course I don't understand how V8 really works but hopefully we will iterate based on this. Let's hope that the presence of this check will teach us some understanding. I will merge it now, so I can test it on some PRs. Just in case - if it doesn't prove to be a good thing, we can always revert it.

hmdne:
It doesn't work so well for pull request though... I will attempt to fix it later.

PR #2270, opened 2021-08-15
Alias for gvars, alias on main

hmdne:


elia (approved):
Great! 
I guess we can use the same technique can be used to define some dynamic global variables if we need to, like linking `$~` to `Regexp.last_match`.

hmdne:
Exactly. This could be a nice way to treat those gvars ($~, $_) as actually thread local vars

PR #2272, opened 2021-08-15
implement Regexp#names for compiler compatibility

janbiedermann:
There exists https://xregexp.com/ which has such goodies implemented in js

https://github.com/slevithan/xregexp/blob/d1302bd7b9b2eb94cc9fefd61d40ed244f5c04a4/src/xregexp.js#L513


hmdne:
This is a new version of #1822

Wow, I didn't know JS regexp supports named capture groups, but it does and it's supported quite universally:

https://caniuse.com/mdn-javascript_builtins_regexp_named_capture_groups

Anyway, I'm consulting an implementation of Regexp that will _in general_ try to be as compatible with Onigmo by compiling Regexps at runtime (or compile time if possible) with a small code size.

opal/corelib/regexp.rb
```diff
@@ -244,6 +244,15 @@ def match?(string, pos = undefined)
     }
   end
 
+  def names
+    %x{
+      var n = self.toString().match(/(?<=\(\?<)\w+(?=>.+\))/g);
+      if (!n) { return []; }
+      // make sure values are uniq
+      return Array.from(new Set(n));
```
hmdne, line 252: Can we here do something like:

```ruby
return #{`n`.uniq}
```

This isn't a hot path I think...

hmdne:
Superseded by #2272

elia (approved):

PR #2278, opened 2021-08-22
Improve a bit on the caching mechanism

hmdne:
* It is now implicitly loaded by default if any writable paths are
  found and a check for those is now more robust.
* The main path for cache is now ~/.cache/opal, otherwise we try
  /tmp/opal-cache-#{unix_username}.
* Also check for parser/ast modifications.
* Also check for Ruby version changes.
* Also check for file size changes as suggested by @janbiedermann
* Remove old cache entries, keep at most about 50MB of cache.

### DIFF
--- a/lib/opal/cache.rb
+++ b/lib/opal/cache.rb
@@ -13,7 +13,7 @@ module Opal
 
   def self.cache
     @cache ||=
-      if RUBY_ENGINE == 'opal' || %w[1 true TRUE].include?(ENV['OPAL_CACHE'])
+      if RUBY_ENGINE == 'opal' || ENV['OPAL_CACHE_DISABLE'] || !Cache::FileCache.find_dir
         Cache::NullCache.new
       else
         Cache::FileCache.new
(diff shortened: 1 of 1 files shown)


lib/opal/cache.rb
```diff
@@ -29,79 +29,147 @@ def fetch(*)
...
+          size_sum = entries.map { |i| File.size(i) }.sum
+          return unless size_sum > @max_size
+
+          # First, we try to get the oldest files first.
+          entries = entries.sort_by { |i| File.mtime(i) }
+          # Then, what's more important, is that we try to get the least
+          # recently used files first. Filesystems with relatime or noatime
+          # will get this wrong, but it doesn't matter that much, because
+          # the previous sort got things "maybe right".
+          entries = entries.sort_by { |i| File.atime(i) }
```
elia, line 86: Can these two be combined in a single cycle with `entries.sort_by! { |i| [File.mtime(i), File.atime(i)] }`?

lib/opal/cache.rb
```diff
@@ -29,79 +29,147 @@ def fetch(*)
       end
     end
 
-    class FileCache
-      def initialize(dir: nil)
-        @root = File.expand_path '..', Opal.gem_dir
-        @dir = dir || find_dir
-      end
+    unless RUBY_ENGINE == 'opal'
```
elia, line 32: Thoughts on moving FileCache to its own file and append the `unless` to the `require`?

PR #2279, opened 2021-08-23
Source map workshop

hmdne:
In this PR I will attempt to improve the source map support in Opal


hmdne:
This PR needs a little bit more work, but otherwise is mostly fine.

hmdne:
This is in general far from perfect, but it's a step ahead since before - backtraces are now at least complete.

elia (approved):
🤩

PR #2282, opened 2021-08-24
Refactor the performance_compare task

elia:
- Move shell commands from tasks to constants to avoid an extra shell-out
- Extract the stat collection to a lambda that can be used for both current and previous
- Use Float::INFINITY instead of an arbitrarily high number
- Show some output for when the V8 optimizations are the same


hmdne:
The problem was introduced with caching that compiler times weren't really meaningful. Other than that, I offered hosting a runner for this task specifically on a much more powerful CPU so we can run compiler like 7 times with even better granularity. Also, you can test this task on */ci-check - on pull requests it works so that it compares results with master, but on master, */ci-check, etc. it compares results with the previous push.

PR #2286, opened 2021-08-26
Add support for ECMAScript modules with an --esm option

janbiedermann:
Simply wraps the compiler result as self executing es6 module, mainly intended for Opal::Builder to create a importable opal code bundle
```javascript

// ... lots of opal code ...

// ... final wrap:

/* Generated by Opal 1.2.0 */
export default (function(Opal) {
  var self = Opal.top, $nesting = [], nil = Opal.nil, $$$ = Opal.$$$, $$ = Opal.$$;

  Opal.add_stubs(['$require']);
  
  self.$require("opal");
  return self.$require("opal-builder");
})(Opal);
```
for a example entry file:
```ruby
require 'opal'
require 'opal-builder'
```

Resulting file can be imported like this:
```javascript
import "resulting_file.js";
```
and after that Opal will be available

### DIFF
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -140,6 +140,7 @@ module Opal
         enable_source_location
         use_strict
         parse_comments
+        esm
       ]
     end
 
--- a/lib/opal/cli_options.rb
+++ b/lib/opal/cli_options.rb
@@ -180,6 +180,10 @@ module Opal
         options[:no_cache] = true
       end
 
+      on('--esm', 'Wraps compiled bundle as for ES6 module(requires -c or -L)') do
+        options[:esm] = true
+      end
+
       separator ''
     end
 
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -137,6 +137,11 @@ module Opal
     # Prepare the code for future requires
     compiler_option :requirable, default: false, as: :requirable?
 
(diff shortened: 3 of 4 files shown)


hmdne:
Well, anyway, only files that do `#await` in top will be affected, which probably means, for library code, none.

hmdne:
Other idea I have maybe... we have already a "use strict" flag. Maybe make it an option like

--es-level=[3,5,6]

I have an idea that we could also apply some es-level optimizations depending on the option chosen.

janbiedermann:
We could delegate that to the bundler: https://esbuild.github.io/content-types/#javascript

I think its 2 different things, JS level and packaging, es6_wrap is only about the packaging

hmdne:
So maybe something like: --es-package=[3,5,6] or [classic|strict|esm]

lib/opal/cli.rb
```diff
@@ -135,6 +135,7 @@ def compiler_option_names
         enable_source_location
         use_strict
         parse_comments
+        es6_wrap
```
elia, line 138: Thoughts on changing the option name to `--esm`?

janbiedermann: I think --esm is ok, when using Builder it seems clear, what to expect, when using the Compiler directly maybe some more options are needed in the future, it depends. For the moment probably --esm is best.

PR #2290, opened 2021-08-29
Bug: TrueClass === false; FalseClass === true #2230

### LINKED ISSUE #2230: Bug: TrueClass === false; FalseClass === true
**Describe the bug**

`TrueClass === false` returns true, `FalseClass === true` too.

Opal version: master

**To Reproduce**

```ruby
[user@localhost opal]# bin/opal-repl
>> TrueClass === false
=> true
>> 
[user@localhost opal]# pry
[1] pry(main)> TrueClass === false
=> false
[2] pry(main)> 
```

ppmasa8:
This fixes #2230 

Unfortunately, the implementation of the getter property setting has not been implemented successfully.
It is not a best practice and we would appreciate it if you could fix it.

### DIFF
--- a/opal/corelib/boolean.rb
+++ b/opal/corelib/boolean.rb
@@ -56,8 +56,33 @@ class Boolean < `Boolean`
   end
 end
 
-TrueClass  = Boolean
-FalseClass = Boolean
+class TrueClass < Boolean
+  `self.$$is_true_class = true`
+
+  class << self
+    def allocate
+      raise TypeError, "allocator undefined for #{name}"
+    end
+
+    def ===(object)
+      `Opal.is_obj_true(object, self)`
+    end
+  end
+end
+
+class FalseClass < Boolean
+  `self.$$is_false_class = true`
+
+  class << self
+    def allocate
+      raise TypeError, "allocator undefined for #{name}"
+    end
+
+    def ===(object)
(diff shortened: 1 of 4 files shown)


hmdne:
There's one problem remaining, this test doesn't pass:

```ruby
  it "<=> == === != =~ !~ have higher precedence than &&" do
    (false && 2 <=> 3).should == false
    (false && 3 == false).should == false
    (false && 3 === false).should == false
    (false && 3 != true).should == false

    class FalseClass; def =~(o); o == false; end; end
    (false && true =~ false).should == (false && (true =~ false))
    (false && true =~ false).should_not == ((false && true) =~ false)
    class FalseClass; undef_method :=~; end

    (false && true !~ true).should == false
  end
```

hmdne:
For those cases... we have method_missing, let's try to do something about that

hmdne:
Let's try this... unfortunately this test can't be passed, as `=~` is defined on Kernel. But I added this test and it works correctly:

```
  it 'allows defining methods on TrueClass/FalseClass' do
    class TrueClass
      def test_opal
        false
      end
    end

    class FalseClass
      def test_opal
        true
      end
    end

    expect(true.test_opal).to be_false
    expect(false.test_opal).to be_true
  end
```

elia (approved):
Thanks for this contribution!

PR #2293, opened 2021-09-01
Optimize truthiness/falsiness checks

hmdne:
I noticed we were first comparing to nil, then !boolean || true. Unless
I'm missing something, all we need to do is make sure it's not false or
nil.

This is a rebase of #1097, I created a new PR because I can't force push to the old one.

### DIFF
--- a/opal/corelib/basic_object.rb
+++ b/opal/corelib/basic_object.rb
@@ -1,3 +1,5 @@
+# use_strict: true
+
 class ::BasicObject
   def initialize(*)
   end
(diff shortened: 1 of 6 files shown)


hmdne:
Booleans... how do they work

hmdne:
This works now BUT: we can always get a boxed boolean via calls like Kernel#itself. And those booleans will be treated wrong. So I will try to continue this PR to ensure that boxing doesn't happen, for Booleans at least.

hmdne:
In general - this patchset would make `new Boolean(false)` truthy. And we ensure Opal does not create such a monstrosity.

This also would fix false.itself/false.tap boxing the values, cc @catmando

hmdne:
We are not getting this fully for v1.4. I will rework it slightly so we will at least get strict mode in a few modules to ensure we are not returning boxed values.

elia (approved):

PR #2299, opened 2021-09-15
Autoload n fixes

janbiedermann:
This is a extraction of #2266 , the minimum required set of patches for isomorfeus to function correctly
Ill separate them into single PRs, this PR just for refernce so i dont forget anything

PR #2301, opened 2021-09-15
ensure 0.digits == [0]

janbiedermann:
taken from #2299

### DIFF
--- a/opal/corelib/number.rb
+++ b/opal/corelib/number.rb
@@ -747,9 +747,15 @@ class Number < Numeric
     end
 
     %x{
+      if (self != parseInt(self)) #{raise NoMethodError, "undefined method `digits' for #{inspect}"}
+
       var value = self, result = [];
 
-      while (value !== 0) {
+      if (self == 0) {
+        return [0];
+      }
+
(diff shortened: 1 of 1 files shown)


hmdne:
There is problem with a === operator and self and I have a feeling it may cause a problem.

```js
> Object.prototype.itself = function() { return this; }
[Function (anonymous)]
> (5).itself()
[Number: 5]
> 5
5
> (5).itself() == 5
true
> (5).itself() === 5
false
> (5).itself() === (5).itself()
false
>

**Comment by hmdne** [comment]
It works though and I can't break it! Good!

**Comment by hmdne** [comment]
```js
Number.prototype.is_zero = function() {
  return this === 0;
}

Number.prototype.is_zero_strict = function() {
  "use strict";
  return this === 0;
}

console.log("is_zero: ", (0).is_zero())
console.log("is_zero_strict: ", (0).is_zero_strict())
```

Return:
```js
is_zero:  false
is_zero_strict:  true
```

elia:
Thanks @janbiedermann, I added a couple fixes and merged!

PR #2051, opened 2019-12-31
Module#autoload?

janbiedermann:

PR #2302, opened 2021-09-15
Refine autoloading capabilities

janbiedermann:
Taken from #2299, the largest part of it

This allows for:
```
require_tree 'components', autoload: true
```
and best used in cooperation with opal-zeitwerk autoloader.

Benefit is especially for large opal apps in a browser or otherwise interactive environment to reduce TTI (Time To Interactive) significantly on application load, or any other app, that frequently uses only part of its constants.

For reference see also #2056 and #2051 (which i will all close)


janbiedermann:
Some more autoloading specs work with this, but currently i cant run specs, specs maybe follow in another PR once i get to fix running spec on my machine, but dont wait for it

janbiedermann:
well, there seems to be problem, maybe i missed something

janbiedermann:
Alright, caught the missing part and added it again.

hmdne:
Rebased to master, let's see how the CI reacted to that.

PR #2307, opened 2021-09-15
Never return a debugger

hmdne:


hmdne:
Fixes opal/opal-browser#91

Even though not exactly, because this issue was caused by a local setup, this commit improves `debugger`.

elia (approved):

PR #2309, opened 2021-09-17
IO Refactor: Most importantly - add IO methods like IO#gets

### LINKED ISSUE #2092: gets: undefined method `gets' for main
I am trying to execute below the code and it's failing.  

````
puts "Enter a number between 1 and 10"
number = gets.chomp.to_i
````
Could you please help me to handle `gets` ruby call. 

Regards,
Hare

hmdne:
This is ongoing... fixes #2092


hmdne:
I'm changing the milestone to v1.3, but we may decide to postpone this merge. Especially, that it's not finished yet.

PR #2316, opened 2021-09-25
ESM for Opal::SimpleServer

hmdne:
* Add Opal::Config.esm
* bundle update the examples
* If Config.esm is enabled, SimpleServer does type="module"
* new rack-esm example


elia (approved):
🤩

PR #2321, opened 2021-10-06
Move a process patch to Rewriter::Base and optimize it further

hmdne:


hmdne:
This doesn't make sense, unfortunately.

PR #2323, opened 2021-10-06
Autoload refinement

janbiedermann:
Taken from #2299, the largest part of it

This allows for:

require_tree 'components', autoload: true

and best used in cooperation with opal-zeitwerk autoloader.

Benefit is especially for large opal apps in a browser or otherwise interactive environment to reduce TTI (Time To Interactive) significantly on application load, or any other app, that frequently uses only part of its constants.

For reference see also #2056 and #2051 (which i will all close)


janbiedermann:
Gosh, what did do? 😨 https://en.wikipedia.org/wiki/23_(film)

PR #2345, opened 2021-11-03
Fix defining multiple methods with the same block 

### LINKED ISSUE #497: define_method $block using all the same super
```rb
$block = proc {
  super()
}
class Base
   def foo
     puts "1"
   end
   def bar
    puts "2"
   end
end

class Derive < Base
  define_method :foo, $block
  define_method :bar, $block

end

d = Derive.new
d.foo #1
d.bar #2
# but opal inputs 2 in all two cases.
```

elia:
Fixes #497

### DIFF
--- a/opal/corelib/module.rb
+++ b/opal/corelib/module.rb
@@ -373,9 +373,10 @@ class Module
   end
 
   def define_method(name, method = undefined, &block)
-    if `method === undefined && block === nil`
-      raise ArgumentError, 'tried to create a Proc object without a block'
-    end
+    %x{
+      if (method === undefined && block === nil)
+        #{raise ArgumentError, 'tried to create a Proc object without a block'}
+    }
 
     block ||= case method
               when Proc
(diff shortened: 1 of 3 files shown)


opal/corelib/module.rb
```diff
@@ -395,14 +395,27 @@ def define_method(name, method = undefined, &block)
               end
 
     %x{
-      var id = '$' + name;
+      // Wrapping and forwarding is required to make super work, as the
+      // same block can be used to define multiple methods.
+      const wrapper = function() {
```
hmdne, line 400: Can we not break the compatibility yet and do `var`, not `const`?

opal/corelib/module.rb
```diff
@@ -395,14 +396,27 @@ def define_method(name, method = undefined, &block)
...
+        block.$$jsid        = name
+        block.$$p           = wrapper.$$p
+        block.$$s           = null;
+        block.$$def         = wrapper;
+        block.$$define_meth = true;
+
+        return block.apply(this, arguments)
+      }
+
+      var id = `$${name}`
```
hmdne, line 411: This is another thing I haven't noticed before. I would rather if it was '$'+name...

hmdne: Also, the linter has noticed it, but didn't fail the job:

```
* Parsing error: Unexpected character '`'
  - Compiled: tmp/lint/opal.js:3511:16
  - Original: corelib/module.rb:411:0
```

elia: sorry did that out of habit, will fix it along with the un-failed job 👍

hmdne (approved):

hmdne:
After a deeper look, I don't think it will be wrong to backport it to stable :D

PR #2347, opened 2021-11-04
Ruby 3.1 support branch

hmdne:
This branch is for a gradual Ruby 3.1 support.

I'm not updating the tests as for now, so that we may implement as much as possible. Also, some features like: `a = 5; {a: } # => {a: 5}` are not supported until parser gets a new version released.


hmdne:
Rebasing this on top of #2365

elia (approved):
Simply awesome!

(just waiting for the additional commit renaming `safe_inspect` to `inspect` in the runtime and we're good to go!)

PR #2353, opened 2021-11-07
Do not modify `$~` when calling `String#scan` from internal methods

### LINKED ISSUE #2352: Bug: If `$~` is evaluated/called then `$1`, `$2`... returns `nil`
**Describe the bug**

Opal version: v1.3.1

**To Reproduce**

```rb
'a:b'.gsub(/([a-z]):([a-z])/) do
  p $~
  target, content = $1, $2
  p target
  p content
end
```

```
$ ruby -v test.rb      
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
#<MatchData "a:b" 1:"a" 2:"b">
"a"
"b"
```

```
$ opal -v test.rb
Opal v1.3.1
#<MatchData "a:b" 1:"a" 2:"b">
nil
nil
```

If you remove (or comment) the line `p $~` then it's working as expected:

```
$ opal -v test.rb
Opal v1.3.1
"a"
"b"
```

hmdne:
Allow `String#scan` to be called with a private kwarg `no_matchdata`,
which ensures that `$~` isn't set. Then, ensure that matchdata methods
on searching for named captures use this.

This fixes #2352

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -992,7 +992,7 @@ class String < `String`
     `self.replace(/[\s\u0000]*$/, '')`
   end
 
-  def scan(pattern, &block)
+  def scan(pattern, no_matchdata: false, &block)
     %x{
       var result = [],
           match_data = nil,
@@ -1017,7 +1017,7 @@ class String < `String`
         }
       }
 
-      #{$~ = `match_data`}
+      if (!no_matchdata) #{$~ = `match_data`};
 
       return (block !== nil ? self : result);
     }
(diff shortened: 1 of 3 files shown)


opal/corelib/string.rb
```diff
@@ -1017,7 +1017,7 @@ def scan(pattern, &block)
         }
       }
 
-      #{$~ = `match_data`}
+      #{$~ = `match_data` unless no_matchdata}
```
elia, line 1020: Since it's a private API I think we can expect it to be either `true` or `false`, and move the if to JS:

```js
if (!no_matchdata) #{$~ = `match_data`}
```

hmdne:
Corrected and also added a test

elia (approved):

PR #2354, opened 2021-11-07
Stop interpreting falsey values as a missing constant in `Module#const_get`

### LINKED ISSUE #2351: Bug: `const_missing': uninitialized constant when using `const_get`
**Describe the bug**

Opal version: v1.3.1

**To Reproduce**

```rb
module SafeMode
  UNSAFE = 0
  SAFE = 1
  SERVER = 10
  SECURE = 20

  def self.value_for_name name
    const_get name.upcase, false
  end
end

puts SafeMode.value_for_name 'unsafe'
```

You can contrasting it to Ruby's behavior:

```
$ ruby -v test.rb
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
0
```

```
$ opal -v test.rb
Opal v1.3.1
<internal:corelib/module.rb>:357:20:in `const_missing': uninitialized constant SafeMode::UNSAFE (NameError)
	from <internal:corelib/runtime.js>:324:7:in `const_missing'
	from <internal:corelib/runtime.js>:341:5:in `const_get'
	from <internal:corelib/module.rb>:349:1:in `const_get'
	from test.rb:8:5:in `value_for_name'
	from test.rb:12:14:in `undefined'
	from <internal:corelib/runtime.js>:2693:7:in `<main>'
	from test.rb:1:1:in `null'
	from <js:internal/modules/cjs/loader.js>:1063:30:in `Module._compile'
	from <js:internal/modules/cjs/loader.js>:1092:10:in `Module._extensions..js'
```

For reference, it was working on Opal 1.2.0:

```
$ opal -v test.rb
Opal v1.2.0
0
```

hmdne:
This commit fixes a regression introduced by 9d3a43d. In runtime,
const_get_name had a check like: `if(cref.$$const[name])` which
broke if the const was `0`.

This fixes #2351

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -244,7 +244,7 @@
         throw e;
       }
       cref.$$autoload[name].required = true;
-      if (cref.$$const[name]) {
+      if (cref.$$const[name] != null) {
         cref.$$autoload[name].success = true;
         return cref.$$const[name];
       }
@@ -268,7 +268,7 @@
   // Get the constant in the scope of the current cref
   function const_get_name(cref, name) {
     if (cref) {
-      if (cref.$$const[name]) { return cref.$$const[name]; }
+      if (cref.$$const[name] != null) { return cref.$$const[name]; }
       if (cref.$$autoload && cref.$$autoload[name]) {
         return handle_autoload(cref, name);
       }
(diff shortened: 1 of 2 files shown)


hmdne:
Added a test

elia (approved):

PR #2358, opened 2021-11-12
[wip] Mutable strings

hmdne:
tl;dr: How to implement mutable Strings while preserving compatibility

For boxed strings we are allowed to declare toString and valueOf properties on them. Not length though. So what if we made those methods actually return a different string?

This is an early work in progress.


hmdne:
The issue with this is that all String methods and alike should return a mutable string value and none of them do - which, on the other hand, is good for performance and JS library compatibility - JS libraries often check for `typeof arg === 'string'` - and boxed strings would give us `object`, not `string`. Boxed values are in general a little known feature that some people think should be deprecated.

There are also a lot of issues with boxed values in JavaScript. Eg.

```js
> "hello" === new String("hello")
false
> new String("hello") == new String("hello")
false
> 
```

hmdne:
It is much performant to split a string than to push large arrays with strings. This is what Closure Compiler does. Let's help a little our Terser users.

```
                  Compile time: 6.088 -> 6.008 (change: -1.30%)
                      Run time: 0.285 -> 0.271 (change: -5.08%)
                   Bundle size: 5259054 -> 4812284 (change: -8.50%)
          Minified bundle size: 1264953 -> 1044964 (change: -17.39%)
            Mangled & minified: 812275 -> 721296 (change: -11.20%)

Issue #1382, opened 2016-03-04
Code Coverage


wied03:
Picking up where https://github.com/opal/opal/issues/613 left off:

I think this could be possible. The end game would be to implement http://ruby-doc.org/stdlib-2.3.0/libdoc/coverage/rdoc/Coverage.html

The compiler would probably need an opt in config setting that reports coverage data to a global data structure (and uses the compiler's source map info for line numbers, etc.).

The data structure would be appended to for each :stmt node (or maybe more) the compiler pushes.

This would enable tools like simplecov to be used.

Thoughts?

wied03:
This would also make https://github.com/opal/opal/pull/1411 possible. We might want to see how JRuby solved this.

hmdne:
https://github.com/gotwarlost/istanbul#the-instrument-command we could also base it on this...

PR #2362, opened 2021-11-18
Add a puppeteer-ruby runner

hmdne:
* Disable SauceLabs automation
* Support firefox and chromium runner & tests via puppeteer-ruby


hmdne:
Ref: YusukeIwaki/puppeteer-ruby#166

hmdne:
It could be a nice idea to generate a coverage report from this... apparently puppeteer-chrome has the tooling necessary (ref #1382)

PR #2369, opened 2021-12-08
Spec updates chore

hmdne:


hmdne:
The performance failure is expected as it has issues with submodules in general.

PR #2370, opened 2021-12-09
DCE for methods

elia:
This is a reprise of some older attempts at DCE, still WIP, but promising.

Any method definition for a method name that's not called anywhere is replaced with an alias to raise and a `/* DCE: foo */` comment.

```
opal:master * ⤑ bin/opal -ce "def foo; end; (1..100).maxmin"|wc -c
 1856900
opal:master * ⤑ env DCE=1 bin/opal -ce "def foo; end; (1..100).maxmin"|wc -c
 1058296
```

Right now it scrapes out 43% of the size.

### DIFF
--- a/lib/opal/builder.rb
+++ b/lib/opal/builder.rb
@@ -95,6 +95,21 @@ module Opal
 
     def build_str(source, rel_path, options = {})
       return if source.nil?
+
+      if options[:get_method_defs] != true && ENV['DCE']
+
+        options[:allowed_method_defs] = []
+
+        dup.build_str(source, rel_path, options.merge(get_method_defs: true)).processed.each do |asset|
+          # p [asset.class, asset.filename] => asset.method_calls.to_a.sort
+          options[:allowed_method_defs].push(*asset.method_calls)
+        end
+        options[:allowed_method_defs].uniq!
+        # p :allowed_method_defs
+        # puts options[:allowed_method_defs].to_a.sort
+        # exit 1
+      end
+
       abs_path = expand_path(rel_path)
       rel_path = expand_ext(rel_path)
       asset = processor_for(source, rel_path, abs_path, false, options)
--- a/lib/opal/builder_processors.rb
+++ b/lib/opal/builder_processors.rb
@@ -14,7 +14,7 @@ module Opal
         @required_trees = []
         @autoloads = []
       end
-      attr_reader :source, :filename, :options, :requires, :required_trees, :autoloads
+      attr_reader :source, :filename, :options, :requires, :required_trees, :autoloads, :method_calls
 
       def to_s
         source.to_s
@@ -106,6 +106,10 @@ module Opal
 
       def autoloads
         compiled.autoloads
+      end
+
+      def method_calls
+        compiled.method_calls
       end
 
       # Also catch a files with missing extensions and nil.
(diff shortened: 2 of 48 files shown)


hmdne:
A problem with this approach I see is that... moving of aliases may be a good idea where we have control, but not where we don't. Opal-Optimizer does this with eliminating the alias/undef calls as well.

hmdne:
There's another thing I see. Your algorithm, as far as I can see, works this way:

- Compile the code first
- Collect all calls
- Compile again, but without the methods that aren't called.

Opal-Optimizer works this way:

- Compile the code (outside of scope of Opal-Optimizer)
- Loop (while X > 0)
  - Collect all calls
  - Eliminate the methods that aren't called. Save a number of eliminated methods into a variable X.

elia:
oh, I see, so it's able to eliminate nested calls, that's a good point. 
Sounds like we can do the same here but maybe we can cap the number of iterations (think of `-O3` for C compilers)…

The earlier attempt I had, from some years ago, included tracking from which method every call was sent and work the minimal number of calls from that graph, this is admittedly more of a shotgun approach. I'm ok with deploying something like this and refine it over time to be more precise and modular.

Thoughts?

hmdne:
Now that I thought about it, how about doing it in reverse? We start from top and don't traverse into defs. After we have collected a number of calls from top, we traverse into untraversed named defs and repeat this a number of times until there are no new defs found. During this process for each file we build an array of calls. And then we remove the rest. This way we will only need one iteration and also we don't need full compilation, just the AST, after rewriting, let's say, and we can reuse this AST for the compilation process.

hmdne:
Just for the record, in Opal-Optimizer (which... works well enough for my production codebase to never break :D):

A definition is one of the following:
- def call
- defs call
- defn call
- udef call (yep, it's a definition :D)
- alias call (where first arg is the definition name)

A call is one of the following:
- a regular Ruby call
- alias call (where second arg is the call name)
- a call to send with a first parameter being a String

janbiedermann:
What about:
```ruby
o = Object.new
m = 'to_s'

o.send(m)
```
?

hmdne:
@janbiedermann this is a thing that we obviously can't support without doing a full evaluation :( for this case we may add a magic comment to note that our code calls #to_s. most of the time though, send is used by `method_missing` in a delegation pattern, which is perfectly fine!

hmdne:
(Please ensure you took a look at my edits :D)

hmdne:
Also... let's take a look at the visitor pattern that is used by parser and friends. We would need then to support wildcards, like eg.:

```ruby
# dont_treeshake: on_*
```

hmdne:
Also @janbiedermann while this is related, we talked with @elia yesterday about things like, what if we extended the `parser` gem to find a magic comment like `# js_xstrings: true` and then let's say a code like:

```ruby
hello(:world, `window.prompt(#{message})`)
```

Would produce a sexp like:

```ruby
s(:send, nil, :hello,
  s(:args,
    s(:sym, :world),
    s(:js_call,
      s(:js_dotaccessor,
        s(:js_var, :window)
        s(:js_method, :prompt)
      ),
      s(:js_args,
        s(:lvar, :message)
      )
    )
  )
)
```

ie. we dispatch to the JS parser for x-strings
This would have potential for both tree shaking and also maybe considering something like connecting RBS to TypeScript?

PR #2372, opened 2021-12-16
Fix io puts

elia:
An empty `puts` was not writing a newline, now the behavior is 1:1 with CRuby except for recursion management

### DIFF
--- a/bin/opal-mspec
+++ b/bin/opal-mspec
@@ -7,7 +7,5 @@ specs = ARGV.map do |s|
 end
 
 env = {}
-env['MSPEC_PATTERN'] = "{#{specs.join(',')}}" if specs.any?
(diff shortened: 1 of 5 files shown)


stdlib/nodejs/io.rb
```diff
@@ -43,7 +43,6 @@ def self.binread(path)
 end
 
 STDOUT.write_proc = ->(string) { `process.stdout.write(string)` }
-STDERR.write_proc = ->(string) { `process.stderr.write(string)` }
```
hmdne, line 46: Why this removal?

elia: I'm dumb, I looked at it multiple times and couldn't see what was different between the two lines! 🤦‍♂️

hmdne (approved):

PR #2374, opened 2021-12-19
Docs & opal/platform fixes

elia:
I was trying to read a file with nodes after requiring opal-platform, but that resulted in an infinite loop because we were only loading nodejs/io. Now every known platform  loads a file with the same name.


hmdne:
We don't require nodejs by opal-platform for a reason. The default require is to make it slightly compatible with a browser, eg. for the test suites.

PR #2383, opened 2021-12-27
Runtime optimization

hmdne:
This patchset will focus on runtime optimization.


hmdne:
This PR is basically a bunch of experiments that may give off something. The last commit is particularly interesting! The rest are negligible.

For Opal-Parser:

Before the commit:
real	0m13.804s

After the commit:
real	0m8.901s

PR #2385, opened 2021-12-30
Fix zsuper with args named with js reserved word

### LINKED ISSUE #2361: Bug: New implicit super doesn't rename variables
**Describe the bug**
The new implicit super doesn't rename variables.

Opal version: >=v1.3.0

**To Reproduce**

```ruby
[user@localhost opal]# bin/opal
class A
  def a(arg)
    p arg
  end
end

class B < A
  def a(native)
    super
  end
end

B.new.a(1)
-:9:5:in `a': native is not defined (Exception)
	from -:13:6:in `undefined'
	from <internal:corelib/runtime.js>:2693:7:in `<main>'
	from -:1:1:in `null'
	from node:internal/modules/cjs/loader:1101:14:in `Module._compile'
	from node:internal/modules/cjs/loader:1153:10:in `Module._extensions..js'
	from node:internal/modules/run_main:79:12:in `executeUserEntryPoint'
	from node:internal/main/run_main_module:17:47:in `undefined'
[user@localhost opal]# 
```

Comparing this to Opal v1.2.0 (MRI exhibits the same behavior):

```ruby
[user@localhost opal]# opal _1.2.0_
class A
  def a(arg)
    p arg
  end
end

class B < A
  def a(native)
    super
  end
end

B.new.a(1)
# => 1
[user@localhost opal]# 
```

Reason:

```javascript
    return (Opal.def(self, '$a', $B_a$2 = function $$a(native$) {
      var $iter = $B_a$2.$$p, $yield = $iter || nil, self = this;

      if ($iter) $B_a$2.$$p = null;
      return $send2(self, $find_super(self, 'a', $B_a$2, false, true), 'a', [native], $iter)
    }, $B_a$2.$$arity = 1), nil) && 'a'
```

Do note how `native` is renamed to `native$`, but not in argument list to `$send2`

takaram:
Fixes #2361

### DIFF
--- a/lib/opal/nodes/super.rb
+++ b/lib/opal/nodes/super.rb
@@ -179,28 +179,28 @@ module Opal
         same_arg_counter = Hash.new(0)
 
         def_scope.original_args.children.each do |sexp|
-          arg_name = sexp.meta[:arg_name]
+          lvar_name = sexp.children[0]
 
           case sexp.type
           when :arg, :optarg
-            arg_node = s(:lvar, arg_name)
+            arg_node = s(:lvar, lvar_name)
             #   def m(_, _)
             # is compiled to
             #   function $$m(_, __$2)
             # See Opal::Node::ArgsNode
-            if arg_name[0] == '_'
-              same_arg_counter[arg_name] += 1
-              arg_node = s(:js_tmp, "#{arg_name}_$#{same_arg_counter[arg_name]}") if same_arg_counter[arg_name] > 1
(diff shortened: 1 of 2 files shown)


hmdne:
Thank you for this patch!

We are going to review and merge it on Wednesday, as usually

Happy new year! :D

elia (approved):

sarhm:
> Fixes #2361

tnx

PR #2387, opened 2022-01-23
added URI.decode_www_form

HoneyryderChuck:
Adding stdlib module URI, with the implementation of `decode_www_form`.

 I didn't manage out to find out how to test (though was an rspec test, but it was running CRuby `uri` instead). Also, implementation omits 2 arguments, because the features may not be implemented in JS (are string encodings a thing in JS?).

Help would be appreciated :)

### DIFF
--- a/stdlib/uri.rb
+++ b/stdlib/uri.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module URI
+	def self.decode_www_form(str, separator: '&', isindex: false)
+		%x{
+			var str = #{str};
+			var ary = [];
+			if (str.length == 0) {
+				return ary;
+			}
+			var parts = str.split(#{separator});
+			for (var i = 0; i < parts.length; i++) {
+				var string = parts[i];
+				var comps = string.split('=');
+				var key = comps[0];
+				var val = comps[1];
+				if (#{isindex}) {
+					if (comps.length < 2) {
+						val = key;
+						key = '';
+					}
+				}
+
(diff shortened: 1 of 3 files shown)


hmdne:
Hm... yeah, unfortunately the specs aren't present yet. See:

https://github.com/ruby/spec/blob/master/library/uri/decode_www_form_spec.rb

But in general, to get URI specs enabled we will have to do this:

```diff
diff --git a/spec/ruby_specs b/spec/ruby_specs
index cfd0d74bf..915458afa 100644
--- a/spec/ruby_specs
+++ b/spec/ruby_specs
@@ -204,3 +204,4 @@ ruby/library/stringio/each_line_spec
 ruby/library/stringio/each_spec
 ruby/library/stringscanner
 ruby/library/optionparser
+ruby/library/uri
```

Your patch is cool :D I corrected some slight styling issues.

Btw. we mostly do patch reviews on Wednesdays every week.

hmdne:
@elia found out there are some tests in CRuby's test suite (we use CRuby tests in addition to RubySpec aka MSpec), but they will not all will pass. For the reference, the test suite is here, not all will pass, but we will factor this in into Opal:

https://github.com/ruby/ruby/blob/master/test/uri/test_common.rb#L167-L193

hmdne:
Let's merge it! Thanks!

In any case, if you have any urgent questions about Opal, we are available on Slack for real-time communication :D

HoneyryderChuck:
thx everyone!

PR #2392, opened 2022-02-14
Support binding.irb

### LINKED ISSUE #2376: `binding.irb` for a web browser
The idea:

Server-side:
- listen on some websocket / provide ajax interface with logic from opal-repl
- do compilation there

Client side:
- connect to such websocket / do ajax request which would interface with the server-side part (using replutils - just like opal-repl does)
- use binding.js_eval to evaluate the Opal-compiled JS

Challenges:
- IO.read_proc will need to become async-capable

This should create a seamless solution where you can use binding.irb either on server or on client

hmdne:
This fixes #2376

But does it a little differently from what we assumed there. In browser, we attempt to load `opal-parser` from the CDN using SJAX whenever user attempts to use `Opal.irb` (or `binding.irb`) or `Opal.eval` in a browser. We leverage `prompt` for synchronous communication.

If a debugger creates a breakpoint, we can use `Opal.irb(s => eval(s))` to launch the irb prompt.

I wish for `corelib/binding` and `corelib/debug` to be included by default.

This needs a little bit more work.


hmdne:
Ways to test it:

In browser:
```
bin/opal -Rserver -e 'a = 5; b = "a"; binding.irb'
```

In node:
```
bin/opal -ropal/platform -e 'require "opal-replutils"; require "opal-parser"; a = 5; b = "a"; binding.irb'
```

hmdne:
The plan is to rename it to corelib/irb. And also to enable it by default!

elia (approved):

PR #2398, opened 2022-02-20
DateTime and Date refactor

### LINKED ISSUE #1743: Feature Request - Implement DateTime from date_core.c --> ISO8601 DateTime is essential for Databases.
Dear Developers:

I want to suggest that you port the DateTime kernel object from date_core.c ( found here: https://github.com/ruby/ruby/blob/b366c4d4104d9273d011f7daec343bdced8bde17/ext/date/date_core.c ) over to Opal core and ensure the Marshal::dump/load functions are exactly like regular Ruby with respect to this class. This way a much needed basic element can be serialized, passed, and reconstituted without additional 'shim' code.

I tried to do a Opal.eval("require 'date_core'"); but DateTime remains undefined. It is an essential part of all my Database code as it can span -999,999 to +999,999 years (**Ideal** for scientific and archival records) whereas if you attempt a date prior to 1970, Jan 1 - midnight with a Time object, you get jenky, or inaccurate, results - not reliable I have found. Time class simply won't suffice for the need. Date class is insufficient as it lacks time and time zone information. Only a real DateTime will do the job.

So ... purdy-please will you implement DateTime in Opal.core as a Kernel class, and ensure it serializes/reconstitutes properly?

Thank you for your attention to this issue.

### LINKED ISSUE #2391: Feature: DateTime
**Is your feature request related to a problem? Please describe.**

In Ruby, when I `require 'date'`, I get a `DateTime` class. I noticed that I don't get it in Opal. 

I just upgraded to Opal 1.4.1, and I was surprised to find `DateTime` still missing from its API considering the announcement that it is fully compatible with Ruby 3.1.0. I thought it would be feature-complete, and that `DateTime` is finally implemented.

**Describe the solution you'd like**

A `DateTime` class that behaves exactly like the one we get in Ruby:

https://rubyapi.org/3.1/o/datetime

**Describe alternatives you've considered**

A while ago while using Opal 1.1.1, I implemented `DateTime` on my own in [glimmer-dsl-opal](https://github.com/AndyObtiva/glimmer-dsl-opal):

Here is the rough implementation, which might not be complete, but addressed my needs in the past:

```ruby
require 'date'
require 'time'

class DateTime < Date
  def initialize(*args, &block)
    @time = Time.new(*args, &block)
    methods_to_exclude = [:to_date, :to_time, :==, :eql?, :class]
    methods_to_define = @time.methods - methods_to_exclude
    methods_to_define.each do |method|
      singleton_class.define_method(method) do |*args, &block|
        @time.send(method, *args, &block)
      end
    end
  end
  
  def to_date
    @time.to_date
  end
  
  def to_time
    @time
  end
  
  def ==(other)
    return false if other.class != self.class
    year == other.year and
      month == other.month and
      day == other.day and
      hour == other.hour and
      min == other.min and
      sec == other.sec
  end
  alias eql? ==
end

class Date
  def to_datetime
    # TODO support timezone
    DateTime.new(year, month, day, hour, min, sec)
  end
end

class Time
  class << self
    alias new_original new
    def new(*args)
      if args.size >= 7
        Glimmer::Config.logger.debug "Dropped timezone #{args[6]} from Time.new(#{args.map(&:to_s)}) constructor arguments since Opal does not support it!"
        args = args[0...6]
      end
      new_original(*args)
    end
  end
  
  def to_datetime
    # TODO support timezone
    DateTime.new(year, month, day, hour, min, sec)
  end
end
```

At least, reusing `Time` as `DateTime` prevented Ruby API calls from breaking in Ruby gems! If you could at least provide an implementation like that at minimum, that would definitely help!

**Additional context**

Here is the code I expect to work in Opal, which works in Ruby (inside IRB);

```ruby
require 'date'
puts DateTime.new(2029, 10, 22, 11, 3, 0).inspect
# => #<DateTime: 2029-10-22T11:03:00+00:00 ((2462432j,39780s,0n),+0s,2299161j)>
```

hmdne:
This fixes #2391 and this fixes #1743


elia (approved):
It was about time! 🥁 😅

PR #2401, opened 2022-02-27
Improve performance with optimized common method/iter implementation shortcuts

hmdne:


opal/corelib/runtime.js
```diff
@@ -2809,7 +2809,30 @@
     return $truthy(arg['$!']());
   }
 
+  // Optimized function generator for simple functions
+  function $value_fun(arg) {
```
hmdne, line 2813: Think about the names...

hmdne:
I added optimizations for code like:

```
def some_delegator
  @ivar.method
end
```

I am afraid this may break backtraces. Do we gain something from that? Perhaps we can introduce something like optimization levels, eg.:

```
-O0: return output maximally compatible with MRI
-O1: (default) return output maximally compatible with ISO Ruby (MRuby)
-O2: sacrifice potential stacktraces
```

I know `-O` is used, so it's just a food for thought for now.

hmdne:
For a code like:

```ruby
def x
  y
end

def y
  z
end

def z
  error
end

x
```

We get this stacktrace:

image

I wonder if this is acceptable, though the gains are quite clear.

lib/opal/nodes/iter.rb
```diff
@@ -1,52 +1,19 @@
 # frozen_string_literal: true
 
-require 'opal/nodes/node_with_args'
+require 'opal/nodes/def'
```
elia, line 3: Not sure why this change, looks like we're still inheriting from NodeWithArgs, is there another reason?

elia (approved):
Just one nit but 👍👍👍 overall!

lib/opal/nodes/iter.rb
hmdne: My mistake. I wanted to do one thing inheriting another.

elia (commented):
After discussing this together we agreed to put this behind an optimisation flag, e.g. `--optimize-defs` due to the loss of precision in the stacktrace

hmdne:
We agreed to get a limited version for v1.5, the part that doesn't mess up backtraces. For v1.6 we will ship a massively improved changes.

PR #2403, opened 2022-03-08
Fix exception during Hash#each and Hash#each_key if keys get deleted during loop

janbiedermann:
Given the following ruby:
```ruby
o = { a: 1, b: 2, c: 3 }
h = o.dup
h.each { |k,v| h.delete(k) }
i = o.dup
i.each_key { |k| i.delete(k) }
puts "o: #{o}\nh: #{h}\ni: #{i}"
```
and executed with matz ruby results in the following expected output:
```
o: {:a=>1, :b=>2, :c=>3}
h: {}
i: {}
```
However, using opal master the following exceptions are thrown (need to comment out the .each ain above code to get the second exception):
```
<internal:corelib/hash.rb>:429:1:in `each': Cannot read properties of undefined (reading '$$is_string') (Exception)
        from <internal:corelib/runtime.js>:1900:5:in `Opal.send2'
        from <internal:corelib/runtime.js>:1888:5:in `each'
        from test.rb:3:2:in `undefined'
        from <internal:corelib/runtime.js>:2760:7:in `<main>'
        from test.rb:1:1:in `null'
        from node:internal/modules/cjs/loader:1103:14:in `Module._compile'
        from node:internal/modules/cjs/loader:1155:10:in `Module._extensions..js'
        
<internal:corelib/hash.rb>:450:1:in `each_key': Cannot read properties of undefined (reading '$$is_string') (Exception)
        from <internal:corelib/runtime.js>:1900:5:in `Opal.send2'
        from <internal:corelib/runtime.js>:1888:5:in `each_key'
        from test.rb:5:2:in `undefined'
        from <internal:corelib/runtime.js>:2760:7:in `<main>'
        from test.rb:1:1:in `null'
        from node:internal/modules/cjs/loader:1103:14:in `Module._compile'
        from node:internal/modules/cjs/loader:1155:10:in `Module._extensions..js'
```

This happens because the original $$keys aray is used for the loop. If that gets modified during the loop, the indexes fail to match during the loop, resulting in `null` keys and failing.

The patch copies the $$keys using .slice() (the supposedly fastest way to copy arrays) before entering the loop:

```
o: {"a"=>1, "b"=>2, "c"=>3}
h: {}
i: {}
```

### DIFF
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -423,7 +423,7 @@ class ::Hash
     return enum_for(:each) { size } unless block
 
     %x{
-      for (var i = 0, keys = self.$$keys, length = keys.length, key, value; i < length; i++) {
+      for (var i = 0, keys = self.$$keys.slice(), length = keys.length, key, value; i < length; i++) {
         key = keys[i];
 
         if (key.$$is_string) {
@@ -444,7 +444,7 @@ class ::Hash
     return enum_for(:each_key) { size } unless block
 
     %x{
-      for (var i = 0, keys = self.$$keys, length = keys.length, key; i < length; i++) {
+      for (var i = 0, keys = self.$$keys.slice(), length = keys.length, key; i < length; i++) {
         key = keys[i];
 
         block(key.$$is_string ? key : key.key);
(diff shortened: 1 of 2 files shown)


hmdne:
Thanks! Let's merge!

PR #2406, opened 2022-03-23
Add an exception for Tokyo fractional timezone. Fixes #2405

### LINKED ISSUE #2405: Bug: `Time.new` for old date returns wrong results when `utc_offset` given
**Describe the bug**
`Time.new` returns a weird time when:

- 7th argument `utc_offset` is not nil
- System timezone is not UTC
- 1st argument `year` is around 19th century or before (this seems to depend on the timezone)

Because of this bug, DateTime tests fail on my environment.

Opal version: master (v1.4.1)

**To Reproduce**
Below is tested on a system whose timezone is `Asia/Tokyo`.
```
$ ruby -e 'p Time.new(1800, 1, 1, 0, 0, 0, 0)'
1800-01-01 00:00:00 +0000
$ opal -e 'p Time.new(1800, 1, 1, 0, 0, 0, 0)'
1799-12-31 23:59:01 UTC
```

This seems to be caused by the difference of timezone offset before and after standard time started to be used.
```
$ node
Welcome to Node.js v16.7.0.
Type ".help" for more information.
> new Date(2022, 0, 1).getTimezoneOffset()
-540
> new Date(1800, 0, 1).getTimezoneOffset()
-558
```

hmdne:
This fixes #2405.

I would recommend some consultation before merging this.

### DIFF
--- a/opal/corelib/time.rb
+++ b/opal/corelib/time.rb
@@ -135,7 +135,7 @@ class ::Time < `Date`
 
   def self.new(year = undefined, month = nil, day = nil, hour = nil, min = nil, sec = nil, utc_offset = nil)
     %x{
-      var args, result, timezone;
+      var args, result, timezone, tz_offset;
 
       if (year === undefined) {
         return new Date();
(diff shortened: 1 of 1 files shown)


elia:
Maybe we should consider using `Intl`, although I'm not sure exactly how it can be used it seems like it knows better WRT timezones:

```
> new Intl.DateTimeFormat('en-US', {timeZone: 'Asia/Tokyo',hour: 'numeric', minute: 'numeric', second: 'numeric',}).format(Date.UTC(1887, 11, 20, 3, 0, 0, 200))
< "12:18:59 PM"
> new Intl.DateTimeFormat('en-US', {timeZone: 'Asia/Tokyo',hour: 'numeric', minute: 'numeric', second: 'numeric',}).format(Date.UTC(1888, 11, 20, 3, 0, 0, 200))
< "12:00:00 PM"
```

elia:
cc @takaram

elia:
Delayed to when we'll have a more general solution, meanwhile the patch in this PR can be used to make tests pass locally.

hmdne:
Outdated, closing

PR #2409, opened 2022-03-24
Compress output of nested scopes

hmdne:
This is the next part of the minification/optimization series. Perhaps it will provide some nice benefits...


elia (commented):
Do we have examples of this happening in real world code?
I expect most of the time we fall into one of the cases that will abort compression, but could be wrong of course.

hmdne:
If you use the Zeitwerk pattern it happens _all the time_ (as can be seen from the `performance` task). The only situation which we can't optimize fully for is the following:

```ruby
module Opal; module Rewriters; class PatternMatching < Base
end; end; end
```

This is (kind of - because we still add `Opal` to the `$nesting`) optimized to:

```ruby
module (module Opal; self; end)::Rewriters; class PatternMatching < Base
end; end
```

PR #2413, opened 2022-03-27
String methods always return Strings even when overloaded

hmdne:
This is how modern Ruby works... maybe not a big change but perhaps we can gain some performance.


elia (approved):
This is great! Very glad to be cast-free here 👍👍👍

PR #2415, opened 2022-03-30
Logic optimization of runtime.js

hmdne:
This pull request includes non-controversial logic changes to runtime.js


hmdne:
This actually happens to decrease the run time. I anticipated easy gains. Will need some rework.

elia (approved):
⚡

PR #2419, opened 2022-04-20
Uninline non-typical argument handling

hmdne:
Some experiments on uninlining things. I'm curious about the performance benefits from that.


elia (approved):
🙌

PR #2426, opened 2022-06-18
Make `Time.new` not depend on `Date.prototype.getTimezoneOffset()`

### LINKED ISSUE #2405: Bug: `Time.new` for old date returns wrong results when `utc_offset` given

takaram:
Fixes #2405

This is more general way than #2406.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,7 @@ jobs:
             command: bin/rake lint
             ruby: '3.0'
           - name: timezone
+            command: bin/rake mspec_nodejs TZ="Pacific/Fiji"
             ruby: '3.0'
           - name: performance
             ruby: '3.0'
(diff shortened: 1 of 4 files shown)


takaram:
Thank you!

PR #2433, opened 2022-09-14
Fix an edge case of if in the most complex form not returning

### LINKED ISSUE #2424: Bug: edge case in which the final expression is not returned
**Describe the bug**

Opal version: 1.5 / master

**To Reproduce**

```
⤑ bin/opal -v -e '                                                                                                            ~/C/opal
                              def m
                                @m = begin
                                  bar while false

                                  case 1
                                  when Integer
                                    2
                                  end
                                end
                              end
                              p m
                              '
Opal v1.5.0
nil
```

```
⤑ ruby -v -e '                                                                                                                ~/C/opal
                              def m
                                @m = begin
                                  bar while false

                                  case 1
                                  when Integer
                                    2
                                  end
                                end
                              end
                              p m
                              '
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]
2
```

in the compiled code the `case` is wrapped in a function that is not returned:

```
⤑ bin/opal -v -e '                                                                                                            ~/C/opal
                              def m
                                @m = begin
                                  bar while false

                                  case 1
                                  when Integer
                                    2
                                  end
                                end
                              end
                              p m
                              ' -cOE --no-source-map
Opal v1.5.0
Opal.queue(function(Opal) {/* Generated by Opal 1.5.0 */
  var nil = Opal.nil, $gvars = Opal.gvars;

  return ($gvars.VERBOSE = true)
});

Opal.queue(function(Opal) {/* Generated by Opal 1.5.0 */
  var self = Opal.top, $nesting = [], $$ = Opal.$r($nesting), nil = Opal.nil, $truthy = Opal.truthy, $eqeqeq = Opal.eqeqeq, $def = Opal.def;

  Opal.add_stubs('bar,===,p,m');
  
  
  $def(self, '$m', function $$m() {
    var $a, self = this, $ret_or_1 = nil;

    return (self.m = (function() {
    (function() {while ($truthy(false)) {
      self.$bar()
    }; return nil; })();
    (function() {if ($eqeqeq($$('Integer'), ($ret_or_1 = 1))) {
      return 2
    } else {
      return nil
    }})();})())
  }, 0);
  return self.$p(self.$m());
});
```

hmdne:
This fixes #2424

### DIFF
--- a/lib/opal/compiler.rb
+++ b/lib/opal/compiler.rb
@@ -558,7 +558,7 @@ module Opal
             returns(true_body),
             returns(false_body)
           ]
-        )
+        ).tap { |s| s.meta[:returning] = true }
       else
         if sexp.type == :send && sexp.children[1] == :debugger
           # debugger is a statement, so it doesn't return a value
--- a/lib/opal/nodes/if.rb
+++ b/lib/opal/nodes/if.rb
@@ -61,12 +61,18 @@ module Opal
         end
 
         if expects_expression?
+          return_kw = 'return ' if returning_if?
+
           if scope.await_encountered
-            wrap '(await (async function() {', '})())'
+            wrap "#{return_kw}(await (async function() {", '})())'
           else
-            wrap '(function() {', '})()'
+            wrap "#{return_kw}(function() {", '})()'
           end
         end
+      end
+
+      def returning_if?
+        @sexp.meta[:returning]
       end
 
       def truthy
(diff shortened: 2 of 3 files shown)


elia:
I think this is the bug I was trying to resolve with that random commit in the DCE branch, thanks for getting rid of it 👍

PR #2436, opened 2022-09-17
Server CLIRunner refactor

hmdne:
Refactor the Server CLI Runner to use Opal::SimpleServer
    
* We extend the API here - if runner.dynamic_builder? then it doesn't
  provide the runners with a Builder, but a proc that returns a Builder
  which will always correspond to an up-to-date code.
* As a result - now the server runner will supply an up-to-date code
  each refresh.


elia (changes requested):
Awesome, thoughts on going a step further and always using a dynamic-builder for all runners? 

My guess is that the real difference is that some runners will only fetch the builder once, while others will do it multiple times. That way we can avoid some complexity.

elia (approved):
Live reviewed together and paired on the previously requested chages

PR #2437, opened 2022-09-18
ESM import & export

hmdne:
This is an early experimental branch aimed at supporting proper ESM import and export. At this point I see many issues with static import, namely:

- imports can be only done at top level. NodeJS support needs `fs` at least, then if we load `opal/platform`, then it simply won't work if ran on other platforms.
- they are incompatible with how require() works, for instance path resolution doesn't work for me, I need more time to understand the issue

But in short, it works in the following way:

```ruby
fs = ::JS.import("fs") # => toplevel: import * as _tmpvar_ from "fs";, Opal.esm_imports["_something_"] = _tmpvar;
                       # => fs = Opal.esm_imports["_something_"]
::JS.export("name", 10) # => toplevel: export name Opal.esm_exports["_somethingelse_"];
                        # => Opal.esm_exports["_somethingelse_"] = 10
```


hmdne:
This PR is superseded by #2502.

PR #2438, opened 2022-09-28
String#length should be available when using opal/mini.rb

dan42:
So String#length and String#size definition should be moved from string/encoding.rb (which is not loaded by opal/mini.rb) to string.rb


elia (approved):
Nice one! Thanks for catching this 👍👍👍

PR #2439, opened 2022-09-29
Chrome CLI Runner to facilitate an async exit.

hmdne:
* We now load a headless chrome platform ONLY if Chrome CLIRunner is being used
* We inhibit a default exit from CDP interface if Kernel#exit hasn't been called (and opal-platform has been loaded)
* We modify Kernel#exit to run a special alert message to communicate with Chrome CDP interface (we would have sent an exception, but it doesn't bubble in the async mess - perhaps it will once we migrate to PromiseV2)

Thanks to Brandon Gastelo for spotting this bug.

Ref: opal-rspec#89

### DIFF
--- a/lib/opal/cli_runners/chrome.rb
+++ b/lib/opal/cli_runners/chrome.rb
@@ -80,6 +80,7 @@ module Opal
             <meta charset='utf-8'>
             <script src='./source-map-support.js'></script>
             <script>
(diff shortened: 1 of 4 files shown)


elia (approved):
Thanks! Exiting from chrome is a great addition 👍

PR #2440, opened 2022-10-04
Opal fixes for RSpec update

hmdne:
Those fixes will be needed for the related RSpec update


lib/opal/cli_runners/chrome_cdp_interface.rb
```diff
@@ -106,17 +106,35 @@
...
             `Page.handleJavaScriptDialog({accept: false})`
           end
+        elsif `dialog.type` == 'alert' && `dialog.message` == 'opalheadlesschromeexit'
+          # A special case of an alert with a magic string "opalheadlesschromeexit".
+          # This denotes that `Kernel#exit` has been called. We would have rather used
+          # an exception here, but they don't bubble sometimes.
+          %x{
+            Page.handleJavaScriptDialog({accept: true});
+            Runtime.evaluate({ expression: "window.OPAL_EXIT_CODE" }).then(function(output) {
+              client.close();
```
elia, line 116: Leftover?

lib/opal/nodes/call.rb
```diff
@@ -99,7 +99,7 @@ def csend?
 
       def default_compile
         if auto_await?
-          push 'await '
+          push '(await '
```
elia, line 102: Thoughts on adding a spec/test that would be failing without this patch?

stdlib/await.rb
```diff
@@ -81,3 +83,35 @@ def async?
     @method.async?
   end
 end
+
+class BasicObject
+  def instance_exec_await(*args, &block)
+    ::Kernel.raise ::ArgumentError, 'no block given' unless block
+
+    nil.await
```
elia, line 91: Can we add a quick comment explaining this workaround?

stdlib/await.rb
```diff
@@ -81,3 +83,35 @@ def async?
     @method.async?
   end
 end
+
+class BasicObject
+  def instance_exec_await(*args, &block)
```
elia, line 88: Thoughts on adding a test for this one?

elia (changes requested):
🙌 great stuff, left a couple of minor comments around explaining and testing pieces here and there

PR #2442, opened 2022-10-08
Hash#freeze

janbiedermann:
Mostly working, well pragmatically working in a acceptable way

I still need to check performance impact of using Object.freeze, maybe it has to go and using the $$frozen flag has to be sufficient, or maybe Object.freeze can be optional.

Object.freeze on the Hash instance causes lots of problems with other specs.

If in matz ruby one does:
```
a = Hash.new.freeze
a.define_singleton_method(:b) { true } # >> FrozenError
```

But that doesnt happen yet with this implementation.
Either $$frozen can be accepted globally, so that other methods can check, or Object.freeze on the instance can be made to work. Not sure yet, need to investigate the problems with the other specs further.

### DIFF
--- a/opal/corelib/error.rb
+++ b/opal/corelib/error.rb
@@ -224,7 +224,6 @@ class ::ZeroDivisionError   < ::StandardError; end
 class ::NameError           < ::StandardError; end
 class ::NoMethodError         < ::NameError; end
 class ::RuntimeError        < ::StandardError; end
-class ::FrozenError           < ::RuntimeError; end
 class ::LocalJumpError      < ::StandardError; end
 class ::TypeError           < ::StandardError; end
 class ::ArgumentError       < ::StandardError; end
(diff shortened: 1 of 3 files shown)


janbiedermann:
The specs, that are not enabled yet, depend on Object#freeze or String#freeze

janbiedermann:
Performance penalty by Object.freeze is not too bad for the use cases i tried, adds ~1.4% for example when freezing state and props during render cycle.

opal/corelib/error.rb
```diff
@@ -224,7 +224,6 @@ class ::ZeroDivisionError   < ::StandardError; end
 class ::NameError           < ::StandardError; end
 class ::NoMethodError         < ::NameError; end
 class ::RuntimeError        < ::StandardError; end
-class ::FrozenError           < ::RuntimeError; end
```
hmdne, line 227: This shouldn't be done here, the error tree is of an importance for maintaining.

opal/corelib/error.rb
```diff
@@ -224,7 +224,6 @@ class ::ZeroDivisionError   < ::StandardError; end
...
 class ::LocalJumpError      < ::StandardError; end
 class ::TypeError           < ::StandardError; end
 class ::ArgumentError       < ::StandardError; end
```
hmdne, line 229: eg. below is UncaughtThrowError, despite it being extended later

janbiedermann: done

janbiedermann:
Ill open another PR with Hash#freeze, Array#freeze and Module#freeze soon, lets see how that goes

janbiedermann:
Superseeded by #2448

PR #2444, opened 2022-10-09
#freeze and #frozen?

janbiedermann:
Generic #freeze and #frozen? based on Object.seal and property getters/setters and a $$frozen flag.


opal/corelib/kernel.rb
```diff
@@ -233,6 +248,72 @@ def extend(*mods)
...
+        #{
+          if self.is_a?(String) || self.is_a?(Array)
+            return self
+          end
+        }
+        var type = typeof(self);
+        if (type === "object") {
+          // make setters of instance variables throw FrozenError
+          var prop, prop_type, current_value;
+          for(prop in self) {
```
hmdne, line 265: `for in` is not an ES5 compatible keyword

opal/corelib/kernel.rb
```diff
@@ -233,6 +248,72 @@ def extend(*mods)
...
+              continue;
+            }
+            current_value = self[prop];
+            var ct = typeof(current_value);
+
+            // create closure for s and v
+            // for Opal 2.0 let for s and v should do the trick, instead of a function
+            (function() {
+              var s = false, p = prop, v;
+              Object.defineProperty(self, prop, {
```
hmdne, line 281: this is nice!

opal/corelib/kernel.rb
```diff
@@ -233,6 +248,72 @@ def extend(*mods)
...
+          if (!self.hasOwnProperty('$$id')) { $prop(self, '$$id', null); }
+          if (!self.hasOwnProperty('$$meta')) { $prop(self, '$$meta', null); }
+          // These are for Regexp patterns
+          if (!self.hasOwnProperty('$$gm')) { $prop(self, '$$gm', null); }
+          if (!self.hasOwnProperty('$$g')) { $prop(self, '$$g', null); }
+
+          // seal the Object
+          Object.seal(self);
+        }
+      } catch(e) { console.err('freeze', e); }
```
hmdne, line 307: for debugging this is fine, but for final code this will need to be removed...

opal/corelib/kernel.rb
hmdne: sorry, I'm wrong... it's fine

opal/corelib/kernel.rb
janbiedermann: Sure

janbiedermann:
Problems that showed today:
- Spec expect frozen strings in various places, howver in non strict mode of Javascript a automatic conversion of string primitives to String objects happens (string primitives are always frozen, string objects not by default)
- Using String object at those places instead of primitives causes internal problems because the === operator is used, so that comparision of string primitive with a String object will always fails
- adding instance vars to sealed objects in non strict mode silently fails, should throw ideally

opal/corelib/kernel.rb
```diff
@@ -233,6 +248,88 @@ def extend(*mods)
...
+                });
+              })();
+              // must set value once
+              self[prop] = current_value;
+            }
+          }
+        }
+
+        // set $$id
+        if (!self.hasOwnProperty('$$id')) { $prop(self, '$$id', $uid); }
```
hmdne, line 300: Shouldn't this be a call to the uid function?

janbiedermann: Good catch, yes it should be, ill fix

opal/corelib/module.rb
```diff
@@ -407,8 +427,18 @@ def define_method(name, method = undefined, &block)
 
   def remove_method(*names)
     %x{
-      for (var i = 0, length = names.length; i < length; i++) {
-        Opal.rdef(self, "$" + names[i]);
+      var length = names.length;
+
+      if (length > 0) {
```
hmdne, line 432: Seems like an extraneous check...

opal/corelib/module.rb
```diff
@@ -645,8 +683,18 @@ def to_s
 
   def undef_method(*names)
     %x{
-      for (var i = 0, length = names.length; i < length; i++) {
-        Opal.udef(self, "$" + names[i]);
+      var length = names.length;
+
+      if (length > 0) {
```
hmdne, line 688: Seems like an extraneous check here too... why isn't undef_method an alias to remove_method? Does it differ somehow?

opal/corelib/regexp.rb
```diff
@@ -119,6 +119,36 @@ def =~(string)
...
+    # Specialized version of freeze, because the $$gm and $$g properties need to be set
+    # especially for RegExp. At the same time instance variables are ignored here.
+
+    return self if frozen?
+
+    %x{
+      self.$$frozen = true;
+
+      // set $$id
+      if (!self.hasOwnProperty('$$id')) { $prop(self, '$$id', $uid); }
```
hmdne, line 132: $uid here too should be a call

opal/corelib/regexp.rb
```diff
@@ -119,6 +119,36 @@ def =~(string)
...
+
+      // ensure $$meta can be set lazily, $$meta is frozen when set in runtime.js
+      if (!self.hasOwnProperty('$$meta')) { $prop(self, '$$meta', null); }
+
+      // $$comparable is used internally and set multiple times
+      // defining it before sealing ensures it can be modified later on
+      if (!self.hasOwnProperty('$$comparable')) { $prop(self, '$$comparable', null); }
+
```
hmdne, line 140: Perhaps consider using super in this function to make it smaller?

opal/corelib/module.rb
janbiedermann: One calls Opal._r_def, one calls Opal._u_def

opal/corelib/regexp.rb
janbiedermann: Done

opal/corelib/regexp.rb
janbiedermann: Obsolete, using super

opal/corelib/module.rb
janbiedermann: Done

opal/corelib/module.rb
janbiedermann: Check removed

hmdne:
We have reviewed this with @elia today, so my comments are ours together. Overall, the patch is all fine and we want it for v1.6. Unfortunately, we merged another patch which started some conflict with `spec/filters/unsupported/freeze.rb` (as always, before the final merge, please squash and rebase; we prefer to have in history logical parts of the changeset, not history of changes while in development). Also I see there's a failing test from mspec_opal suite (run `bundle exec rake mspec_opal_nodejs` to review it).

Other thing we noted is that there is quite a significant decrease in performance due to this patch. This is understandable, probably stems from the usage of the runtime macro basically everywhere. I thought about adding a compiler option to disable the freeze functionality altogether where it isn't needed, but let's not worry about it for now.

So, in short, it's great to get closer to MRI when we though in the past it won't be possible to implement freeze at all. Thanks for all the great work :D

opal/corelib/kernel.rb
```diff
@@ -233,6 +248,88 @@ def extend(*mods)
...
+
+            desc = Object.getOwnPropertyDescriptor(self, prop);
+            if (desc && desc.enumerable && desc.writable) {
+              current_value = self[prop];
+
+              // create closure for s and v
+              // for Opal 2.0 let for s and v should do the trick, instead of a function
+              (function() {
+                // set v to undefined, as if the property is not set
+                var s = false, p = prop, v = undefined;
```
hmdne, line 282: This logic with single letter variables is as we understand for debugging... `s` is always false, right?

janbiedermann: Oh, i see, i think i can simplify this a bit more ... i will try ...

janbiedermann: p was left over, s is now not required at all

janbiedermann:
Superseeded by #2448

janbiedermann:
Reopening for research purposes

janbiedermann:
Took over some improvements from #2448 and squashed it

janbiedermann:
For the record:
Compile time: 4.396 (+26.56%) -> 4.404 (+20.05%) (change: +0.18%)
    Run time: 1.756 (+3.72%) -> 1.753 (+14.00%) (change: -0.15%)

spec/filters/unsupported/freeze.rb
```diff
@@ -1,182 +1,22 @@
 # NOTE: run bin/format-filters after changing this file
 opal_unsupported_filter "freezing" do
-  fails "A method definition inside a metaclass scope raises FrozenError if frozen" # Expected FrozenError but no exception was raised ("foo" was returned)
```
hmdne, line 3: Since freeze is now supported, this filter file should be moved to bugs I think

hmdne:
This looks great to me. I will aim to do a final review on Wednesday and finally merge it!

janbiedermann:
da53da8 shows a clear benefit in benchmarks, that specifically test the touched areas, like Array#<<, however the benefit is barely measurable in performance:compare.
The intention was, to generally compensate for $deny_frozen_access, but that seems to work.
Maybe better not merge da53da8?

janbiedermann:
Ill remove da53da8, needs more research.

elia (approved):
This is just awesome! 🙌 
Great job!

PR #2449, opened 2022-10-18
Improve method block performance 

janbiedermann:
delete is expensive, seems like there is no need to delete $$p, setting it to null seems enough


janbiedermann:
                  Compile time: 7.117 (±0.54%) -> 7.091 (±0.40%) (change: -0.37%)
                      Run time: 1.723 (±1.07%) -> 1.693 (±7.59%) (change: -1.74%)
                      
Looks good :-)

elia (approved):
Good call!

PR #2450, opened 2022-10-21
Windows performance:compare

janbiedermann:


janbiedermann:
Now ready

tasks/testing.rake
```diff
@@ -1,4 +1,8 @@
+require_relative "#{__dir__}/../lib/opal/os"
```
elia, line 1: nit: also here using __dir__ shouldn't be necessary

elia (approved):
@janbiedermann thanks for this PR! Looks great, just left a couple of comments for minor changes, can you please try them on windows?

Meanwhile I'm pre-approving 👍

tasks/performance.rake
```diff
@@ -1,6 +1,9 @@
 # A set of tasks to track performance regressions by the CI
 
-require 'opal/util'
+require_relative "#{__dir__}/../lib/opal/util"
+require_relative "#{__dir__}/../lib/opal/os"
```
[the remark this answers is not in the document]
janbiedermann: I had to use __dir__ so that it would use the patched version

tasks/testing.rake
janbiedermann: I had to use __dir__ so that it would use the patched version

janbiedermann:
@hmdne OS abstraction is working great

PR #2452, opened 2022-10-27
Eliminate var constructor in allocate_class, seems redundant

janbiedermann:


janbiedermann:
Also helps performance and bundle size a tiny bit

elia (approved):

PR #2455, opened 2022-10-27
[experimental] performance:compare : Take best time if WITHIN_VARIATION results are within MAX_VARIATION

janbiedermann:
Keeps the performance task running until WITHIN_VARIATION results are within best_time + MAX_VARIATION %

MAX_VARIATION here is set to 3%
WITHIN_VARIATION here is set to 10
=>  If 10 results are within best_time + 3%, take the best time to compare

Verifying a performance improvement

Lets say a improvement of 1.5% is expected.
- Set MAX_VARIATION to below 1.5%, maybe 1%
- run the task
- if change is larger than variation and around the expected 1.5%, the improvement is verified within margin
- run multiple times to verify

Limits:
Improvements in Opal currently are usually below 2%, most often below 0.5%
Best achievable variation for 10 best times on my machine out of a total of 128 tries is around 0.65%
On my machine setting MAX_VARIATION to below 1% is useless,
thus accurately verifying improvements below 1% is not possible (unless setting tries to a very high number)

However, github results should be more reliable, as variation is kept below 3%.
So if the -5% change is hit, its probably accurate and not due to high variation.

Please play a bit with it, set MAX_VARIATION to 0.1 and max_tries to 128 and see whats the best variation that can be achieved on your machine

### DIFF
--- a/tasks/performance.rake
+++ b/tasks/performance.rake
@@ -22,8 +22,8 @@ class Timing
   # expecting 10 results within MAX_VARIATION margin
   def margin_achieved?
     @times.sort!
-    return true if @times.size >= WITHIN_VARIATION && error < max_error
-    return true if @times.size >= @max_tries
+    return true if tries >= WITHIN_VARIATION && error < max_error
+    return true if tries >= @max_tries
     false
   end
 
--- a/tasks/performance.rake
+++ b/tasks/performance.rake
@@ -7,7 +7,7 @@ OS = Opal::OS
 
 class Timing
   MAX_VARIATION = 3 # percent
-  WITHIN_VARIATION = 10 # results within MAX_VARIATION
+   = 10 # results within MAX_VARIATION
 
   def initialize(max_tries: 64, &block)
     @max_tries = max_tries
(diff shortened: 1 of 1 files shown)


janbiedermann:
Limitations of github actions become visible:

                  Compile time: 3.573 (+5.31%) -> 3.547 (+2.56%) (change: -0.72%)
                      Run time: 1.514 (+1.73%) -> 1.510 (+1.75%) (change: -0.29%)
                      
It seems like short benchmarks can be more accurate than longer ones.
The target of 3% for the shorter run time benchmarks has been hit very quickly.
But even 64 tries have not been enough to keep the 10 best results within the 3% variation for the compile time benchmark.

janbiedermann:
github actions second result:

                  Compile time: 2.854 (+6.60%) -> 2.890 (+6.30%) (change: +1.27%)
                      Run time: 1.204 (+2.00%) -> 1.196 (+2.49%) (change: -0.67%)

janbiedermann:
github actions third result, similar:

                  Compile time: 4.071 (+4.34%) -> 3.958 (+5.02%) (change: -2.79%)
                      Run time: 1.692 (+1.21%) -> 1.691 (+1.78%) (change: -0.03%)

janbiedermann:
@hmdne

In all 3 github action runs above, the actual change should have been 0 or at least close to 0.

Conclusion:
- the compile the compile time benchmark provides no or false information due to high variation when running on github action machines with a max change of -2.79%
- averaging over a multitude of smaller benchmarks with shorter runtime may provide us with a much better performance trend

PR #2456, opened 2022-10-27
[experimental] failed attempt to reduce performance variation

janbiedermann:
This PR, based on #2455 tries to minimize obstructions like:
- node startup
- file io
- garbage collection

It measures within node using performance.*

Its focused on asciidoctor run time.

But still variation is too high, basically no advantage over #2455, to accurately measure small performance improvements.


janbiedermann:
This PR needs ruby 3.1 becasue of speednode

janbiedermann:
Run time: 1.793 (+1.52%) -> 1.805 (+0.97%) (change: +0.72%)
                Speed run time: 1.848 (+0.75%) -> 1.837 (+0.78%) (change: -0.63%)

PR #2458, opened 2022-10-29
Behave like Ruby, treat \r\n as \n when reading a file, not binary

janbiedermann:
In MRI, when reading a file with \r\n as line separator, File#read and File.open.read return \r\n as \n.
File#binread or File.open.read with mode 'b' will still return \r\n as \r\n

### DIFF
--- a/stdlib/nodejs/file.rb
+++ b/stdlib/nodejs/file.rb
@@ -127,7 +127,7 @@ class File < IO
 
 
   def self.read(path)
-    `return executeIOAction(function(){return __fs__.readFileSync(#{path}).toString()})`
+    `return executeIOAction(function(){return __fs__.readFileSync(#{path}).toString().replaceAll("\r\n", "\n")})`
   end
 
   def self.write(path, data)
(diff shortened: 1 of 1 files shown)


hmdne:
Do we get some new tests passed probably?

hmdne:
Ah, nevermind, we don't test Node stdlib with mspec...

janbiedermann:
I checked, did not find any filtered ones, that would apply

hmdne:
After some research, I think that commit is kind of wrong. While you are right, that read and binread (or open without and with "b" flag) have different semantics, they have only on Windows:

```
[3] pry(main)> File.write("/tmp/test", "abc\r\n")
=> 5
[4] pry(main)> File.read("/tmp/test")
=> "abc\r\n"
[5] pry(main)> File.open("/tmp/test", &:read)
=> "abc\r\n"
[6] pry(main)>
```

The further question is, should Opal differentiate Windows systems or always follow Unix semantics for portability?

janbiedermann:
Actually your example works likewise on Windows ruby, so this PR is kinda bogus. I was trying to fix something else and came accross this inconsitency. The cause may be somewhere else, i will continue investigating.

PR #2457, opened 2022-10-28
Fix performance:compare asset size calculation

janbiedermann:

PR #2451, opened 2022-10-26
Ensure UTF8 encoding of sourcesContent of SourceMaps

janbiedermann:
Avoids exceoptions: JSON::GeneratorError and Encoding::UndefinedConversionError

sourcesContent will be valid UTF8. invalid UTF8 will be replaced by �

The previously used #force_encoding will keep invalid characters which may cause failures during subsequent handling of sourcesContent within opal or external consumers

PR #2460, opened 2022-10-29
[wip] Fix specs on Windows

janbiedermann:
Trying to fix remaining issues when running specs on Windows.

This PR depends on and currently merges #2457 #2454 #2451, will cleanup later.

Problem for specs ralted to Exceptions and stack traces:
Position information is completely wrong, related to sourcemaps

In chrome positions are offset by the length of runtime.js lines.
The difference to Linux seems to be that git checked out runtime.js with \r\n line breaks instead of \n, which may or may not cause this issues.

Improvements so far:
e470117 provides a tiny performance improvement
255d8e8 considers \r\n and \n line breaks when splitting fragments

So far problems persist.


janbiedermann:
Yessssss! Fixed it, finally. When reading sources as binary, must also write them as binary! Otherwise, they get mangled. This may affect other places too, maybe Sprockets, etc. need to check.

janbiedermann:
Anything that uses Opal::Builder.build.to_s and writes the result to a filesystem, must write the result in binary mode.

janbiedermann:
Depends on #2454 #2451 to properly work on Windows

janbiedermann:
Now all specs pass on Windows too (with #2454 #2451)

:-)

lib/opal/eof_content.rb
```diff
@@ -15,9 +15,12 @@ def eof
       eof_content = @source[last_token_position..-1]
       return nil unless eof_content
 
-      eof_content = eof_content.lines.drop_while { |line| line == "\n" }
+      # On Windows token position is off a bit, because Parser does not seem to compensate for \r\n
+      # The first eof_content line on Windows may be for example "end\r\n"
+      # Must match for it and \r\n and \n
+      eof_content = eof_content.lines.drop_while { |line| /\A.*\r?\n\z/.match?(line) && !line.start_with?('__END__') }
```
hmdne, line 21: `\r?\n\z` can be shortened to `\Z` afaik

.github/workflows/build.yml
```diff
@@ -46,8 +46,7 @@ jobs:
             ruby: '3.0'
             command: bin/rake smoke_test
           - name: windows
-            # These two fail because of broken stacktraces on windows: minitest_node_nodejs mspec_nodejs
-            command: bundle exec rake rspec minitest_nodejs
+            command: bundle exec rake
```
hmdne, line 49: I would suggest adding another pipeline perhaps, because I think this particular one may run for too long

hmdne:
So this is probably why it worked well on Windows for me, because I have no conversion enabled in my Git MSYS setup.

lib/opal/eof_content.rb
janbiedermann: done

.github/workflows/build.yml
janbiedermann: done

janbiedermann:
```
  1) Failure:
TestNodejsFile#test_windows_file_expand_path [nodejs/test_file.rb:189:5]:
should add a trailing slash when the path is d:.
Expected: "d:/"
  Actual: "d:/a/opal/opal"
```
I can't reproduce and i have no idea, how that happens.

janbiedermann:
Well, it looks to me like a classic c programming failure issue.

janbiedermann:
Really, d: is a symlink

janbiedermann:
All green

janbiedermann:
All green > again <

lib/opal/eof_content.rb
```diff
@@ -15,9 +15,12 @@ def eof
       eof_content = @source[last_token_position..-1]
       return nil unless eof_content
 
-      eof_content = eof_content.lines.drop_while { |line| line == "\n" }
+      # On Windows token position is off a bit, because Parser does not seem to compensate for \r\n
+      # The first eof_content line on Windows may be for example "end\r\n"
+      # Must match for it and \r\n and \n
+      eof_content = eof_content.lines.drop_while { |line| /\A.*\Z/.match?(line) && !line.start_with?('__END__') }
```
hmdne, line 21: unfortunately we can't use \Z here (and \Z is slightly equivalent to \n?z), because Opal doesn't support it... :(

lib/opal/eof_content.rb
```diff
@@ -15,9 +15,12 @@ def eof
...
 
-      if eof_content[0] == "__END__\n"
+      if /\A__END__\r?\Z/.match?(eof_content[0])
```
hmdne, line 23: take a look here, above you have \Z, here \r?\Z, both should be \r?\n?\z

janbiedermann: ... i trusted you ...

elia (approved):
I still don't completely understand how this fix works for the EOF contents regarding this line:

```rb
eof_content = eof_content.lines.drop_while { |line| /\A.*\r?\n?\z/.match?(line) && !line.start_with?('__END__') }
```

but approving anyway as I wasn't able to brake it 😊

PR #2461, opened 2022-10-31
Update rake:bench and add more runners

### LINKED ISSUE #2157: Benchmark support broke after submodules update
1. Benchmarks are now named without bm_ in their file names
2. Some benchmarks are now .yml files

janbiedermann:
Makes benchmarks work again

- removes bench:opal

- adds bench:ruby_vs_opal - benchmarks ruby vs opal node
- adds bench:all - benchmarks most important engines - ruby, node, chrome, firefox

- uses rubys standard benchmark_driver to allow to run ruby benchmarks

- adds a Deno runner, basic port to Deno
- adds a Firefox CDP runner
- CDP runners (Chrome and Firefox) make use of a tiny http server
- CDP runners can write files using File#write (required by benchmark_driver)
- CDP runners open a new "Target" for each execution, preventing problems with faulty or crashed targets
- new OPAL_PLATFORMs: headless-firefox, deno

- fixes a Problem with Node 18 and existing chrome runner:
```
node:net:1300:16:in `afterConnect': connect ECONNREFUSED ::1:9222 (Exception)
rake aborted!
```
- fixes a Problem where environment variables have been ignored by runners

- updates CliRunners to register only runners available on current platform
- benchmark.rake and testing.rake create tasks for runners registered by CliRunners

Will fix #2157
Fixing remaining issues of Opal running in Firefox or Deno is out of scope of this PR, will do later on, once merged


janbiedermann:
Also provides nice output:
`
                     C:/Ruby32-x64/bin/ruby.exe  bin/opal-bench.bat
          app_answer                     36.897              73.580 i/s -       1.000 times in 0.027103s 0.013591s

Comparison:
                       app_answer
  bin/opal-bench.bat:        73.6 i/s
C:/Ruby32-x64/bin/ruby.exe:        36.9 i/s - 1.99x  slower
`

hmdne:
Perhaps related to #2157

janbiedermann:
```
C:\Users\jan\workspace\opal>bundle exec benchmark-driver -e C:/ruby32-x64/bin/ruby.exe benchmark/bm_array_add.rb 
        bm_array_add                    16.051k i/s -       1.000 times in 0.000062s (62.30μs/i)
        bm_array_add                    14.577k i/s -       1.000 times in 0.000069s (68.60μs/i)
        bm_array_add                    15.015k i/s -       1.000 times in 0.000067s (66.60μs/i)
        bm_array_add                    15.601k i/s -       1.000 times in 0.000064s (64.10μs/i)
        bm_array_add                    12.920k i/s -       1.000 times in 0.000077s (77.40μs/i)
        bm_array_add                    14.599k i/s -       1.000 times in 0.000068s (68.50μs/i)
```

This does not help at all with variation of results. For good measure multiple, many runs remain necessary, and then taking the best.

janbiedermann:
Ready.
rake bench:opal_chrome may or may not work. Thats fine, keep the task, i will add puppeteer based runners next and replace the current chrome runner for this task with a puppeteer based one and add another task for firefox too.

bin/opal-bench-chrome
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+bundle exec ruby exe/opal -Ilib -Rchrome %*
```
hmdne, line 2: In bash, `%*` is spelled `"$@"` (with `"`s, very important)

janbiedermann:
The firefox runner half way works. It creates a temp profile, starts firefox, executes code, CDP connects, but it does not seem to receive any events, also it doesnt exit automatically. Sure it has the CORS problems when executing specs, but most specs work.

CDP is updated to match firefox, firefox_cdp_interface.rb

chrome_cdp_interface needs a update, as it uses Console.messageAdded, which is deprecated it sems, and should use Runtime.consoleAPICalled instead.

janbiedermann:
For the record: attempts to use puppeteer-ruby with firefox failed so far.

janbiedermann:
Messages get send by firefox, just chrome-remote-interface cannot handle them, needs some fixing

janbiedermann:
Updating chrome-remote-interface and experimenting with ii, i found, that c-r-i is not reliable, messages get lost (assuming its not a firefox issue, sending sometimes only half the messages). Trying a different approach.

janbiedermann:
chrome and firefox publish their implemented CDP protocol at the remote-debugging-port via http://localhost:9222/json/protocol

bin/opal-bench-chrome
janbiedermann: done

janbiedermann:
Firefox on Linux in --headless still needs X, otherwise it fails opening the --remote-debugging-port, can be run with `xvfb-run bundle exec ...`

After overcoming some Windows/Git/Node/Linux bamboozles mspec-firefox runs on Linux and Windows in github actions.

janbiedermann:
Example, rake bench:all:
```

Benchmarking benchmark/bm_hash_replace_string.rb started at 2022-11-05 09:59:13 +0100:
                       bin/opal-bench-node.bat  bin/opal-bench-chrome.bat  bin/opal-bench-firefox.bat  C:/Ruby32-x64/bin/ruby.exe
bm_hash_replace_string                   1.090                      1.137                       1.422                      69.664 i/s -       1.000 times in 0.917683s 0.879300s 0.703000s 0.014355s

Comparison:
             bm_hash_replace_string
C:/Ruby32-x64/bin/ruby.exe:        69.7 i/s
bin/opal-bench-firefox.bat:         1.4 i/s - 48.97x  slower
bin/opal-bench-chrome.bat:         1.1 i/s - 61.26x  slower
bin/opal-bench-node.bat:         1.1 i/s - 63.93x  slower
```

janbiedermann:
Benchmarks currently fail on Linux for some reason

janbiedermann:
Specs and benchmarks work with all runners. Only thing left, is to somehow make repl work with browsers.

janbiedermann:
Reverted 6ef5abea33259822ef5ee10d1121025c58b40687, everything working!

janbiedermann:
This branch, the updated CDP runners, also fixes a problem with Node 18, when running specs for example:
```
node:net:1300:16:in `afterConnect': connect ECONNREFUSED ::1:9222 (Exception)
rake aborted!
```

lib/opal/cli_runners/chrome_cdp_interface.rb
```diff
@@ -3,143 +3,185 @@
...
   host: #{ENV['CHROME_HOST'] || 'localhost'},
   port: #{ENV['CHROME_PORT'] || 9222}
 };
 
+function not_found(res) {
+  res.writeHead(400, { "Content-Type": "text/plain" });
+  res.end("NOT FOUND");
+}
+
+var server = http.createServer(function(req, res) {
```
hmdne, line 28: since chromium runner can work with a server, perhaps we can collapse the two cdp interfaces?

lib/opal/cli_runners/node_modules/.bin/chrome-remote-interface.cmd
```diff
@@ -0,0 +1,17 @@
+@ECHO off
```
hmdne, line 1: can we have node_modules updates in a separate commit? would make a review much simpler

lib/opal/nodes/call.rb
```diff
@@ -8,7 +8,7 @@
 module Opal
   module Nodes
     class CallNode < Base
-      handle :send, :csend
```
hmdne, line 11: here I see a revert of the csend patch, should definitely go into a separate commit imo

hmdne: unless we could find an edge case that causes the issue...

benchmark/run.rb
```diff
@@ -1,15 +1,40 @@
...
+# ruby exes
+opal_chrome_exe = OS.windows? ? 'bin/opal-bench-chrome.bat' : 'bin/opal-bench-chrome'
+opal_firefox_exe = OS.windows? ? 'bin/opal-bench-firefox.bat' : 'bin/opal-bench-firefox'
+opal_node_exe = OS.windows? ? 'bin/opal-bench-node.bat' : 'bin/opal-bench-node'
+ruby_exe = Gem.ruby
+
+# rubies
+rubies = ''
+
+if ARGV.delete('-node')
```
hmdne, line 16: opal now supports optparse, perhaps it would be a good idea to rewrite it in terms of it?

lib/opal/cli_runners/chrome_cdp_interface.rb
janbiedermann: Probably we can support more features in the future (coverage, performance, profiling, debugging), however, thats where FF and Chrome start to differ even more, thats why i like to keep them separate. They already differ slightly, they just appear similar (eg. stack representation).

lib/opal/cli_runners/node_modules/.bin/chrome-remote-interface.cmd
janbiedermann: I made 3 commits

lib/opal/nodes/call.rb
janbiedermann: As we now know, that its not the new CDP runner code that causes the issues, but instead #2467 , i simply removed this

benchmark/run.rb
janbiedermann: done

lib/opal/cli_runners/firefox.rb
```diff
@@ -0,0 +1,356 @@
...
+              'C:/Program Files/Mozilla Firefox/firefox.exe'
+            ].each do |path|
+              next unless File.exist? path
+              return path
+            end
+          elsif OS.macos?
+            '/Applications/Firefox.app/Contents/MacOS/Firefox'
+          else
+            %w[
+              firefox
```
hmdne, line 180: Debian users have firefox-esr by default, may be a good idea to add it as well

lib/opal/cli_runners/chrome_cdp_interface.rb
```diff
@@ -3,143 +3,185 @@
...
+      } else if (path.endsWith(".map")) {
+        content_type = "application/json"
+      } else {
+        content_type = "application/javascript"
+      }
+      res.writeHead(200, { "Content-Type": content_type });
+      res.end(fs.readFileSync(path));
+    }
+  } else if (req.method === "POST") {
+    if (req.url === "/File.write") {
```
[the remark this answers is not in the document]
janbiedermann: We need POST so that the browser can write bnechmark results to a file.
benchmark_driver creates a ruby file with the benchmark, thats executed and writes the times to a file, using File#write, which is then read by benchmark_driver to get the benchmark result.

elia:
@janbiedermann we moved this to v1.7 so it's not blocking the imminent release of v1.6 and also because we mean to release v1.7 on Christmas.

lib/opal/cli_runners/chrome_cdp_interface.rb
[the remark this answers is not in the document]
janbiedermann: Please check [286dc35](https://github.com/opal/opal/pull/2461/commits/286dc350b7e0c5065c54203d76fbdbdc689e4690)
But well, CDP is locally open anyway, so you may use that, to get the secret ...

Anyway, that scheme doesn't seem to work with Firefox ...

Still investigating ...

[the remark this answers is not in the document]
janbiedermann: You could use CDP to overwrite any file using the screenshot, pdf or save page functions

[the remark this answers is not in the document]
janbiedermann: No, i am wrong, they changed that, screenshot, pdf return a io stream, but https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-setDownloadBehavior alloes for any path

janbiedermann:
Test Profiling on Chrome runner with:
```
bundle exec ruby exe/opal -Ilib -Istdlib -rheadless_browser -Rchrome --runner-options '{"profiling":"prof.json"}' -e 'puts "test"'
```
Test Coverage on Chrome runner with:
```
bundle exec ruby exe/opal -Ilib -Istdlib -rheadless_browser -Rchrome --runner-options '{"coverage":"cov.json"}' -e 'puts "test"'
```
The json is not compatible with the Profiler tab in Chrome devtools, too much information is missing. There seems to be no way to get that additional information via CDP. It needs some other tool to visualize or maybe generate HTML. Suggestions welcome.

janbiedermann:
Yeah, chrome waiting forever for some unfulfilled promise ... anyway, i think, for coverage its possible to implement the interface of rubys coverage.rb to some extend and then use simplecov to some extend ... lets see ...

janbiedermann:
Try the synchronous Coverage support in chrome runner with:
```
require 'coverage'

Coverage.start
puts "test"
puts Coverage.result
```

Result is a bit raw from CDP, not ruby ompatible yet, WIP

lib/opal/cli_runners/firefox.rb
janbiedermann: done

hmdne:
Ref: #1382

janbiedermann:
Example Coverage result:
```
:lines format:
corelib/string.rb: ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,1,1,1,1
:oneshot_lines format:
corelib/string.rb: 250,251,252,253,254
```

janbiedermann:
Supporting :methods and :branches format seems possible to some extend, if we can transform the cryptical javascript back to ruby method names/blocks, maybe its an idea for the compiler to provide hints within those names or something like that.
I leave it at :lines and :oneshot_lines support for the moment, cleanup, squash and will try to port Coverage support to the firefox runner and implement a node CDP runner, s othat we can have Coverage with node too, maybe, neeed to research whats possible

janbiedermann:
No Coverage support for Firefox, because:
```
request: {
    method: 'Debugger.enable',
    params: undefined,
    sessionId: undefined
  },
  response: {
    message: 'Debugger.enable',
    data: 'RemoteAgentError@chrome://remote/content/cdp/Error.sys.mjs:20:5\n' +
      'UnknownMethodError@chrome://remote/content/cdp/Error.sys.mjs:103:7\n' +
      'execute@chrome://remote/content/cdp/domains/DomainCache.sys.mjs:92:13\n' +
      'receiveMessage@chrome://remote/content/cdp/sessions/ContentProcessSession.sys.mjs:79:45\n'
  }
```

I tried different ways, same result. Moving on to node ...

janbiedermann:
There is a problem with node and CDP, well, sure there is, when doing Debug.enable() the script is not executed ... so at the moment it runs scripts, buts its useless for Coverage. I tried to do the same as VSCode https://github.com/microsoft/vscode-js-debug/blob/47c60558ec31902f42c255abb9b460078df02f9d/src/targets/node/nodeTarget.ts#L154
Needs more research, whats going on.

janbiedermann:
Removing Coverage via CDP and nodecdp runner. In theory CDP could provide us with some goodies for performance and coverage and debugging, but in practice implementations don't work as expected. Maybe its worth to try again in a couple of years, if/when CDP implementations in practice catch up to theory.

janbiedermann:
Benchmarking Ruby WASM or other Rubies

To benchmark Ruby WASM there are 2 ways:
1. make Ruby WASM the current ruby, so calling `ruby` would execute Ruby WASM
2. set the `OPAL_BENCH_EXTRA_RUBIES` environment variable to point to Ruby WASM

Optcarrot

Optcarrot provides a benchmark.yaml for benchmark_driver, this can be used as arg to rake bench:
rake bench:opal_nodejs[path/to/optcarrot/benchmark.yaml]

provided that paths in benchmark.yaml are set up correctly

janbiedermann:
Ready

.github/workflows/build.yml
```diff
@@ -24,6 +24,12 @@ jobs:
           - name: mspec-chrome
             ruby: '3.0'
             command: bin/rake mspec_chrome
+          - name: mspec-firefox
```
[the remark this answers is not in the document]
janbiedermann: Ok, using minitest_firefox for now

docs/headless_browsers.md
```diff
@@ -27,7 +32,7 @@ The runner also listens for any exceptions and prints formatted stracktraces bac
 
 ## Using exit codes
 
-By default headless chrome runner explicitly sets exit code to 1 when there was any error in the code.
+By default headless breowser runner explicitly sets exit code to 1 when there was any error in the code.
```
hmdne, line 35: typo

hmdne:
We have reviewed all the patchset today - it's really great :D, but we want to have a better understanding of this patchset, so this week we will split it into parts and merge them.

hmdne:
This has been fully merged, but split into those 6 pull requests. Thanks again!

PR #2463, opened 2022-11-01
Retry if file cache write operation exits with Zlib::BufError

hmdne:
This error shouldn't happen and is likely a bug in Zlib. This is a temporary measure, also for older Rubies, so that tests won't break randomly.

This started happening since we merged Prefork

### DIFF
--- a/lib/opal/cache/file_cache.rb
+++ b/lib/opal/cache/file_cache.rb
@@ -23,6 +23,11 @@ module Opal
         out = Marshal.dump(data)
         out = Zlib.gzip(out, level: 9)
         File.binwrite(file, out)
+      rescue Zlib::BufError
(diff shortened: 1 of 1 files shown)


lib/opal/cache/file_cache.rb
```diff
@@ -23,6 +23,11 @@ def set(key, data)
         out = Marshal.dump(data)
         out = Zlib.gzip(out, level: 9)
         File.binwrite(file, out)
+      rescue Zlib::BufError
+        # This sometimes happens, unsure why, makes no sense, possibly
+        # some race condition
+        warn '[Opal]: Zlib::BufError; retrying'
+        retry
```
elia, line 30: Let's re-evaluate this code for loops before releasing 1.6

elia (approved):

PR #2470, opened 2022-11-15
Add Opal::BuilderProcessors::RubyERBProcessor

hmdne:


elia (approved):
Thoughts on compiling the generated ruby source in the RubyERBProcessor?

PR #2472, opened 2022-11-17
Fix ruby-head runner

hmdne:
This is being done by forcing the versions of rack and sinatra, so that we don't pull rack 3.0.0 (which moves rackup to a separate gem) mistakenly.
We would like to upgrade this later on, but for now Sinatra does not support rack 3.0


hmdne:
Also, this was probably caused by a change to the dependency resolver in Ruby 3.0

PR #2482, opened 2022-12-01
Targeted patches for Opal-Parser

hmdne:
1. Reduce code that assigns a result to a variable then returns this variable. This is mostly for racc-generated code, but other applications may be reduced as well
2. Optimize large arrays produced by lexer, but mainly we are interested in improving compile times, by reducing the tree for the further compilation efforts (also reducing the bundle size a bit)


elia (approved):
Can you move the JS helper directly to the runtime? It's short enough to live there I think

PR #2485, opened 2022-12-03
Compiler runner rework (mainly: --watch)

hmdne:
1. NodeJS: `File.symlink`
2. Rearrange the CLI options (so `opal --help` makes more sense)
3. Add `--output` and `--watch` support

The patchset is mostly ready for review, but in the end it will also include a couple more features :)


hmdne:
The --directory option (in the next commit once I clean it up):

```
[user@localhost opal]# bin/opal --directory -ce 'p 123' -o _tx
[user@localhost opal]# node _tx/
123
[user@localhost opal]# rm -rf _tx
[user@localhost opal]# bin/opal --esm --directory -ce 'p 123' -o _tx
[user@localhost opal]# node _tx/index.mjs 
123
[user@localhost opal]# chromium-freeworld --no-sandbox --allow-file-access-from-files _tx/index.html
(outputs 123 in web console)
```

hmdne:
@janbiedermann this may be useful for you as well - this format may be better suited for input to `esbuild` - I remember it had problems with index (combined) maps - those are file maps, each file has an individual map for itself

lib/opal/nodes/call/js.rb
```diff
@@ -0,0 +1,122 @@
...
+      # Warning - this may (or may not!) return a Promise, you need to manually await it!
+      #
+      # The `what` argument denotes what do we want to import. For ESM:
+      # - `:default`       => import("x").then(_mod => _mod.default) (default value)
+      # - `:*`             => import("x")
+      # - `:none`          => import("x").then(() => nil)
+      # - `:anything_else` => import("x").then(_mod => _mod.anything_else)
+      # For CJS:
+      # (mode of operation the same as for regular import)
+      add_special :dynimport, const: :JS do |compile_default|
```
janbiedermann, line 82: i would prefer :dynamic_import here, for clarity and its not used often

lib/opal/nodes/call/js.rb
```diff
@@ -0,0 +1,122 @@
...
+        end
+      end
+
+      # TODO
+      add_special :export, const: :JS do |compile_default|
+      end
+
+      # root directory denotes the root directory of our compiled program
+      # ::JS.npm_dependency("glob", "7.1.3", map: {"glob" => "node_modules/glob/glob.js"})
+      add_special :npm_dependency, const: :JS do |compile_default|
```
janbiedermann, line 114: Please have a look at https://deno.land/manual@v1.28.3/node/npm_specifiers

Maybe its a good idea, to use the deno way above for :import, and let the compiler on demand produce the node way

Maybe its a good idea to refactor platform support, to have a --platform option for the compiler, so it can produce platfrom specific output

hmdne:
The directory part is split to another PR

elia (approved):

PR #2499, opened 2022-12-15
Optimize argument passing

hmdne:
Some various optimizations regarding argument passing - performance-wise and bundle-size-wise


elia (approved):
🚀

PR #2500, opened 2022-12-20
Ruby 3.2 support branch

hmdne:


elia (approved):
We'll merge even with that single failure on windows+firefox, as it's just an issue with the CI configuration

PR #2507, opened 2023-01-01
Run Firefox specs by default

janbiedermann:
depending on #2506 for a clean run

PR #2508, opened 2023-01-03
Platform specific spec filters

janbiedermann:
May add `spec/filters/platform/#{platform}/*.rb` files

Will be required by safari runner and #2507

### DIFF
--- a/tasks/testing.rake
+++ b/tasks/testing.rake
@@ -80,7 +80,7 @@ module Testing
       specs
     end
 
-    def filters(suite)
+    def filters(suite, platform)
       opalspec_filters = Dir['spec/filters/**/*_opal.rb']
 
       if ENV['INVERT_RUNNING_MODE']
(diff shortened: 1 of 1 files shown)


tasks/testing.rake
```diff
@@ -80,15 +80,17 @@ module Testing
...
 
       if ENV['INVERT_RUNNING_MODE']
         # When we run an inverted test suite we should run only 'bugs'.
         # Unsupported features are not supported anyway
         rubyspec_filters = Dir['spec/filters/bugs/*.rb'] - opalspec_filters
       else
-        rubyspec_filters = Dir['spec/filters/**/*.rb'] - opalspec_filters
+        platform_filters = []
+        platform_filters = Dir["spec/filters/platform/#{platform}/*.rb"] if Dir.exist?("spec/filters/platform/#{platform}")
+        rubyspec_filters = (Dir['spec/filters/bugs/*.rb'] + Dir['spec/filters/unsupported/*.rb'] + platform_filters) - opalspec_filters
```
elia, line 93: I think this can be simplified to:

```rb
rubyspec_filters = Dir["spec/filters/{unsupported,bugs,platform/#{platform}}/*.rb'] - opalspec_filters
```

as `Dir[]` doesn't require a check for existence

elia (commented):
One small nit, but otherwise legit 🙌

elia (approved):

PR #2510, opened 2023-01-03
Fix CLI file reading for macOS

elia:
- Avoid relying on the `OriginalARGV` global var for communication between parts of the project
- Use `File.read` when possible to allow macOS read fresh data from the FS

### DIFF
--- a/lib/opal/cli_runners/compiler.rb
+++ b/lib/opal/cli_runners/compiler.rb
@@ -91,7 +91,7 @@ class Opal::CliRunners::Compiler
   end
 
   def reexec
-    system($0, *OriginalARGV) && exit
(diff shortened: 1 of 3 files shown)


lib/opal/cli_runners/compiler.rb
```diff
@@ -91,7 +91,7 @@ def watch_compile
   end
 
   def reexec
-    system($0, *OriginalARGV) && exit
+    Process.kill('USR2', Process.pid)
```
hmdne, line 94: From what I distinctly remember, Windows may have issues with signals, eg. it doesn't support SIGTERM, I wonder if it does support USR2

PR #2511, opened 2023-01-03
Trick Safari/WebKit parser to execute code with literal lookbehind RegExp

janbiedermann:


lib/opal/nodes/literal.rb
```diff
@@ -158,11 +158,30 @@ def compile_static_regexp
         case value
         when ''
           push('/(?:)/')
+        when /\(\?<=/
```
hmdne, line 161: why not:

```ruby
when /\(\?<=/, /\(\?<!/
```

or

```ruby
when /\(\?<=|\(\?<!/
```

janbiedermann: Fixed, thanks

elia (approved):

PR #2519, opened 2023-01-07
Fix links in docs/compiler.md

vzvu3k6k:
This PR adds a missing link reference definition and removed an unused one in [docs/compiler](https://opalrb.com/docs/guides/v1.7.1/compiler).


elia (approved):
Thanks!

PR #2521, opened 2023-01-28
Disallow to define a singleton class for Number

takaram:
Makes `1.singleton_class` or `class << 1; end` raise TypeError instead of Exception.


hmdne (approved):
Looks great!

PR #2529, opened 2023-03-13
Fix for Opal::Cache::FileCache.dir_writable?

dan42:
which previously returned truthy regardless of the dir being writeable or not, resulting in `dir_writable?(Dir.home, '.cache', 'opal')` being always true, even for non-login user (with $HOME set to /nonexistent)

### DIFF
--- a/lib/opal/cache/file_cache.rb
+++ b/lib/opal/cache/file_cache.rb
@@ -78,19 +78,16 @@ module Opal
         nil
       end
 
-      # This complex piece of code tries to check if we can robustly mkdir_p a directory.
+      # Check if we can robustly mkdir_p a directory.
       def self.dir_writable?(*paths)
-        dir = nil
-        paths = paths.reduce([]) do |a, b|
-          [*a, dir = a.last ? File.expand_path(b, a.last) : b]
+        return false unless File.exist?(paths.first)
+
+        until paths.empty?
+          dir = File.expand_path(paths.shift, dir)
+          ok = File.directory?(dir) && File.writable?(dir) if File.exist?(dir)
         end
 
-        File.exist?(paths.first) &&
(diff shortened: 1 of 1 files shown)


lib/opal/cache/file_cache.rb
```diff
@@ -78,19 +78,14 @@ def get(key)
...
+          ok = File.directory?(dir) && File.writable?(dir) if File.exist?(dir)
         end
-
-        File.exist?(paths.first) &&
-          paths.reverse.all? do |i|
-            !File.exist?(i) || (File.directory?(i) && File.writable?(i))
-          end
-
-        dir
+        dir if ok
```
elia, line 88: Nice improvement!
Since `expand_path` accepts `nil` as the second argument it can be skipped, which makes it compatible with windows, also added a check on the first path segment to ensure it exist, it being the "root".

```rb
       def self.dir_writable?(*paths)
         return false unless File.exist?(paths.first)
         
         until paths.empty?
           dir = File.expand_path(paths.shift, dir)
           ok = File.directory?(dir) && File.writable?(dir) if File.exist?(dir)
         end

         dir if ok
       end
```

elia (changes requested):
Thanks for the contribution, left a comment about windows compatibility and checking for the root path existence.

lib/opal/cache/file_cache.rb
dan42: Ah, Windows compatibility, didn't think of that one.

The PR is marked as "changes requested" but you already made the changes, so I'm not sure if you're waiting on something from me?

hmdne: We haven't committed that change to the branch. We did now :) Merging in a moment

PR #2531, opened 2023-03-25
Fix Kernel#.Integer with `exception:` option

ysakasin:
Support keyword argument `exception` of `Kernel#.Integer`.

RDoc: https://ruby-doc.org/3.2.1/Kernel.html#method-i-Integer


elia (approved):

ysakasin:
Thank you merging!

PR #2532, opened 2023-03-25
Fix Kernel#.Float with `exception:` option

ysakasin:
Support keyword argument `exception` of `Kernel#.Float`.

RDoc: https://ruby-doc.org/3.2.1/Kernel.html#method-i-Float

### DIFF
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -477,7 +477,7 @@ module ::Kernel
     }
   end
 
-  def Float(value)
+  def Float(value, exception: true)
     %x{
       var str;
 
@@ -504,6 +504,12 @@ module ::Kernel
 
       return #{::Opal.coerce_to!(value, ::Float, :to_f)};
     }
+  rescue
+    unless exception
+      return nil
+    end
+
+    ::Kernel.raise
   end
 
   def Hash(arg)
--- a/opal/corelib/kernel.rb
+++ b/opal/corelib/kernel.rb
@@ -482,7 +482,11 @@ module ::Kernel
       var str;
 
       if (value === nil) {
-        #{::Kernel.raise ::TypeError, "can't convert nil into Float"}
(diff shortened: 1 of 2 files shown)


opal/corelib/kernel.rb
```diff
@@ -504,6 +504,12 @@ def Float(value)
 
       return #{::Opal.coerce_to!(value, ::Float, :to_f)};
     }
+  rescue
+    unless exception
+      return nil
+    end
+
+    ::Kernel.raise
```
elia, line 512: Thoughts on using the `$coerce_to` helper instead of rescuing whenever `exception:` is false?

Something like:

```rb
%x{
  // …
  return (#{eception} ? #{::Opal.coerce_to!(value, ::Float, :to_f)} : $coerce_to(value, #{::Float}, 'to_f')
}
```

elia (commented):
Thanks, left a comment with an alternative that could save us from using `rescue`

opal/corelib/kernel.rb
elia: 🤔 There's also the `Kernel.raise` in the inner `if`… that would mean nesting an `if (!exception) return nil` there.

Please let me know your thoughts, the `rescue` complicates the compiled code and worsen performance significantly, but I also see a point in using a simpler implementation or ruby's side.

ysakasin: `$coerce_to` brings no too complicated code for us, I think.

I used `rescue` because  I didn't now `$coerce_to`. I'll replace `rescue` with `$coerce_to`.

ysakasin:
@elia Could you rerun failed actions? They seems to fail `bundle install`. 😢

hmdne:
This is just a macos failure and not related to Opal or your patches. Unfortunately it persists when rerun.

ysakasin:
Thank you merging!

PR #2533, opened 2023-04-25
Improve performance of Array#intersect? and intersection

takaram:
This PR improves performance of `Array#intersect?` and `Array#intersection`.

### Array#intersect?
It returns true immediately when first common element is found instead of calculating the whole of intersection.
This is what CRuby (and probably most other implementations) do.

### Array#intersection
These two intersection method calls give identical results, but former is significantly slow on current master.
```ruby
large_array = (1..10000).to_a
large_array.intersection(large_array, [1]) # slow
large_array.intersection([1], large_array) # fast
```
It is now optimized with this PR.

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -1417,7 +1417,7 @@ class ::Array < `Array`
 
     # First, calculate intersection of argument arrays.
     # Array#& is faster when the argument size is small.
-    # So `largest & shotest & second_shortest & ...` would be the fastest.
+    # So `largest & shortest & second_shortest & ...` would be the fastest.
     largest = `arrays.pop()`
     intersection_of_args = arrays.reduce(largest, &:&)
 
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -142,11 +142,13 @@ class ::Array < `Array`
   end
 
   def &(other)
-    return [] if `self.length === 0`
-
-    other = `convertToArray(other)`
-
     %x{
+      other = convertToArray(other)
+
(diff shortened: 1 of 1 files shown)


opal/corelib/array.rb
```diff
@@ -142,6 +142,8 @@ def self.try_convert(obj)
   end
 
   def &(other)
+    return [] if `self.length === 0`
+
     other = `convertToArray(other)`
```
elia: Thoughts on also adding `if (other.length === 0) return []`?

nit, optional: Also probably the two initial lines can be moved inside the `%x{…}` block.

opal/corelib/array.rb
```diff
@@ -1397,7 +1399,30 @@ def inspect
...
+
+    arrays = arrays.sort_by(&:length)
+    # When self is the smallest among the arrays
+    if `self.length < arrays[0].length`
+      return arrays.reduce(self, &:&)
+    end
+
+    # First, calculate intersection of argument arrays.
+    # Array#& is faster when the argument size is small.
+    # So `largest & shotest & second_shortest & ...` would be the fastest.
```
elia, line 1420: typo: `shotest` → `shortest` (missing r)

elia (approved):
Looks great!

Left a couple of small potential improvements 🙌

opal/corelib/array.rb
takaram: Fixed. Thank you!

takaram:
@elia 
Thank you for your review!
I'm sorry for the delay. I didn't notice the reply.

PR #2534, opened 2023-05-08
SourceMap support for Kernel#eval

janbiedermann:
image

Kinda working in Chrome Browsers

### DIFF
--- a/opal/corelib/string/encoding.rb
+++ b/opal/corelib/string/encoding.rb
@@ -103,6 +103,7 @@ class ::Encoding
 
   class ::EncodingError < ::StandardError; end
   class ::CompatibilityError < ::EncodingError; end
+  class UndefinedConversionError < ::EncodingError; end
 end
 
 ::Encoding.register 'UTF-8', aliases: ['CP65001'], ascii: true do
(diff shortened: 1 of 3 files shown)


janbiedermann:
image
Also works in Firefox 112

opal/corelib/string/encoding.rb
```diff
@@ -103,6 +103,7 @@ def bytesize(*)
 
   class ::EncodingError < ::StandardError; end
   class ::CompatibilityError < ::EncodingError; end
+  class UndefinedConversionError < ::EncodingError; end
```
elia, line 106: Can you please add the double column here? It makes the lookup faster.

elia (approved):
This is great, I just left a comment for a quick fix and we good to go 👍👍👍

opal/corelib/string/encoding.rb
janbiedermann: That wouldn't be correct, it must be Encoding::UndefinedConversionError, see https://github.com/opal/opal/blob/89661b5226079ebf20230412c7c3360108d8a3ae/lib/opal/source_map/map.rb#L14

elia: Oh, right!

PR #2541, opened 2023-06-02
ChatGPT4 experiment for Opal coding

hmdne:
I will expand the description later on.


elia (approved):
👍👍👍 

_Lint should be ok when merging_

PR #2543, opened 2023-06-03
[compat] Add an option to enable x-string compilation to JS

hmdne:
For now, it's enabled by default. But if it's not specified and there are backticks in the code, it will issue a warning.

In Opal 2.0 we will switch the default to disabled.

The rationale is simple: we aim to be as compatible with Ruby as possible. This will give us a big headstart into being able to compile various existing Ruby libraries for NodeJS. Now, if there are backticks, the code must be ported, because backticks with usual Ruby semantics compile to invalid JavaScript. Yet still, for NodeJS we do implement ```Kernel#` ```, so this code may work out of the box.

On the other hand, there is a lot of Opal-dependent code, but compared to above - people who wrote that code know that Opal exists. So, with proper guidance, they should be able to migrate their code by adding a magic comment.


hmdne:
The lint failure is due to Rubocop upgrade I suppose.

elia (approved):
🙌 awesome!!!

PR #2544, opened 2023-06-05
Fix String#split with limit and capturing regexp

### LINKED ISSUE #1599: splitting a string with a regex group and limit returns non compatible result
In every ruby engine I have been able to test in (except mruby, which doesn't allow for regex splits)
```rb
'aBa'.split(/(B)/, 2)
=> ["a", "B", "a"]
```

in Opal
```js
'aBa'.$split(/(B)/, 2)
=> ["a", "B"]
```

Which is consistent with javascript, but not ruby. Considering that the `2` is supposed to represent `"at most [the] number of fields [to] be returned"`, I am not sure why ruby is giving three. There does not seem to be any ruby spec covering this behavior, so I am trying to read through the C to figure out what the logic is.

takaram:
Closes #1599

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -1060,6 +1060,8 @@ class ::String < `String`
           string = self.toString(),
           index = 0,
           match,
+          match_count = 0,
+          valid_result_length = 0,
           i, ii;
 
       if (pattern.$$is_regexp) {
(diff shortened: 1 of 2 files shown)


hmdne:
Wow, this is great :D 

(Don't worry about a failing lint, it is due to Rubocop upgrade)

hmdne (approved):

PR #2546, opened 2023-06-05
Bridging: call original constructor for subclasses of level larger than 1

### LINKED ISSUE #2540: Bug: No backtrace on exceptions under Firefox
Problem:

> In Firefox there is no backtrace when an exception is thrown. Chrome does have it however.

File was built with:

```
bundle exec opal -c --include ./src src/backtrace.rb --watch -o js-gen/backtrace.js
```

System:
* Opal v1.7.3
* CRuby 3.2.2

Gemfile:
```
# frozen_string_literal: true

source "https://rubygems.org"

# gem "rails"

gem 'opal', git: 'https://github.com/opal/opal.git'
gem "rack", "~> 3.0"

gem "opal-browser", "~> 0.3.3"

gem "listen", "~> 3.8"
```

(using latest from Opal master branch, commit #83030fc)

Browsers:

* Chrome: Version 113.0.5672.92 (Official Build) (x86_64)
* Firefox 113.0.1 (64-bit)

Chrome console:

Chrome console output

Firefox console:

Firefox console output

HTML code used to reproduce bug:

```html
<html>
<head>
  <title>Ruby Opal | Backtrace Test</title>
  <link rel="icon" href="data:;base64,iVBORw0KGgo=">
</head>
<body>

  <!-- Include Opal compiled code -->
  <script type="text/javascript" src="./js-gen/backtrace.js"></script>

  <script type="text/javascript">

    window.addEventListener("error", function(event) {

      console.log(event);

      var suppressErrorAlert = true;
      // If you return true, then error alerts (like in older versions of 
      // Internet Explorer) will be suppressed.
      return suppressErrorAlert;
    });

  </script>

</body>
</html>
```

Ruby code used to reproduce bug:
```ruby
puts "Hello"

raise "Boom!"
```

hmdne:
This commit introduces a function `descends_from_bridged_class(klass)` to check if a class or any of its ancestors is a bridged class. During the allocation of a new class, the `descends_from_bridged_class()` function is now used to determine if the new class is a descendant of a bridged class, and if so, it ensures the original JavaScript constructor is correctly called. This enables more accurate and robust handling of class hierarchies involving bridged classes.

This problem has manifested in Firefox which didn't capture stacktraces for RuntimeError, but did so for StandardError.

This fixes #2540

### DIFF
--- a/opal/corelib/error.rb
+++ b/opal/corelib/error.rb
@@ -36,6 +36,16 @@ class ::Exception < `Error`
     `self.message = (args.length > 0) ? args[0] : nil`
   end
 
+  # Those instance variables are not enumerable.
+  def copy_instance_variables(other)
+    super
+    %x{
+      self.message = other.message;
+      self.cause = other.cause;
+      self.stack = other.stack;
+    }
+  end
+
   %x{
     // Convert backtrace from any format to Ruby format
     function correct_backtrace(backtrace) {
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -538,6 +538,13 @@
   Opal.$$$ = Opal.const_get_qualified;
   Opal.$r = Opal.const_get_relative_factory;
 
+  function descends_from_bridged_class(klass) {
+    if (klass == null) return false;
+    if (klass.$$bridge) return klass;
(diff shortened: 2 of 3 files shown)


elia (approved):
Left one optional comment, feel free to ignore

PR #2547, opened 2023-06-05
yield1/yieldX: Use $$arity if available.

### LINKED ISSUE #2525: don't use javascript block.length to determine number of arguments in block definition
Currently `Opal.yield1` and `Opal.yieldX` use `block.length` to determine the blocks.  This causes sadness if used in combination with the splat operator and javascript minimification.

For example consider

```bash
irb(main):010:0> [[1, 2], [3, 4]].collect { |x, *| x }
=> [1, 3]
```

Opal translates the block to something like this:

```javascript
($$1 = function(x, $a){var self = $$1.$$s || this, $post_args;
    if (x == null) {
      x = nil;
    };
    $post_args = Opal.slice.call(arguments, 1, arguments.length);
    ;
    return nil;}, $$1.$$s = self, $$1.$$arity = -2, $$1))
})
```

BUT because `$a` is never used js minification may drop it.  Resulting in `$$1.length` being 1 rather than 2.

This will then cause the elements of the array (i.e. [1, 2] and [3, 4]) not be splatted because yield1 and yieldX rely on the blocks length rather than the arity.

Fix which I believe will always work is to change `block.length` to `Math.abs(block.$$arity || block.length)` in both yield and yieldX

This can also be patched as follows in the meantime:

```javascript
%x{
  Opal.yield1 = function(block, arg) {
    if (typeof(block) !== "function") {
      throw Opal.LocalJumpError.$new("no block given");
    }

    var has_mlhs = block.$$has_top_level_mlhs_arg,
        has_trailing_comma = block.$$has_trailing_comma_in_args,
        length = Math.abs(block.$$arity || block.length);

    if (length > 1 || ((has_mlhs || has_trailing_comma) && length === 1)) {
      arg = Opal.to_ary(arg);
    }

    if ((length > 1 || (has_trailing_comma && length === 1)) && arg.$$is_array) {
      return block.apply(null, arg);
    }
    else {
      return block(arg);
    }
  };

  // handles yield for > 1 yielded arg
  Opal.yieldX = function(block, args) {
    if (typeof(block) !== "function") {
      $raise(Opal.LocalJumpError, "no block given");
    }

    if (Math.abs(block.$$arity || block.length) > 1 && args.length === 1) {
      if (args[0].$$is_array) {
        return block.apply(null, args[0]);
      }
    }

    if (!args.$$is_array) {
      var args_ary = new Array(args.length);
      for(var i = 0, l = args_ary.length; i < l; i++) { args_ary[i] = args[i]; }

      return block.apply(null, args_ary);
    }

    return block.apply(null, args);
  };
}
```

hmdne:
The rationale behind this is that JavaScript compressors can sometimes reduce a number of function arguments therefore changing `block.length`.

This is important for yield1/yieldX, because it changes block call semantics.

This fixes #2525

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1667,13 +1667,16 @@
     }
 
     var has_mlhs = block.$$has_top_level_mlhs_arg,
-        has_trailing_comma = block.$$has_trailing_comma_in_args;
+        has_trailing_comma = block.$$has_trailing_comma_in_args,
+        length = block.$$arity || block.length;
 
-    if (block.length > 1 || ((has_mlhs || has_trailing_comma) && block.length === 1)) {
+    if (length < 0) length = -length;
+
+    if (length > 1 || ((has_mlhs || has_trailing_comma) && length === 1)) {
       arg = Opal.to_ary(arg);
     }
 
-    if ((block.length > 1 || (has_trailing_comma && block.length === 1)) && arg.$$is_array) {
+    if ((length > 1 || (has_trailing_comma && length === 1)) && arg.$$is_array) {
       return block.apply(null, arg);
     }
     else {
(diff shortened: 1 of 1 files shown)


hmdne:
This is unfortunately incorrect.

PR #2556, opened 2023-06-22
Closure: Generate a JS object, not an Error

hmdne:
This is so that we don't need to capture the stacktrace each time we want to use Ruby semantics of closures.

This should make Asciidoctor work ~15% faster.

Authored by: @merceroncode


elia (approved):
👏

PR #2557, opened 2023-06-22
[WIP] Do not create Objects when using null as arg, instead inline {}

janbiedermann:
no point using Object.create with null as arg, as it only returns a empty object, which can be inlined, no function call needed, faster

### DIFF
--- a/benchmark-ips/bm_hash_create.rb
+++ b/benchmark-ips/bm_hash_create.rb
@@ -1,0 +1,32 @@
+Benchmark.ips do |x|
+  %x{
+    function old_way(keys, smap) {
+      var hash = new Opal.Hash();
+  
+      hash.$$smap = smap;
+      hash.$$map  = Object.create(null);
+      hash.$$keys = keys;
+  
+      return hash;
+    };
+
+    function new_way(keys, smap) {
+      var hash = new Opal.Hash();
+  
+      hash.$$smap = smap;
+      hash.$$map  = {};
+      hash.$$keys = keys;
+  
+      return hash;
+    };
+   }
+
+  values = [123,243,35,"sd",false,nil,123413234,120412,0,1234.1234,0.34,false,false,true,"sadfasf","","0",13,123,nil,Object.new,[]]
+
+  x.time = 32
+
+  x.report('old_way') { `old_way([], {})` }
+  x.report('new_way') { `new_way([], {})` }
+
+  x.compare!
+end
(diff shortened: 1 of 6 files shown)


janbiedermann:
according to the benchmark report, that failed, unless something went wrong
according to my synthetic benchmarks run locally, it should be faster

need to investigate

janbiedermann:
ips for $hash2 shows a benefit:
             old_way   166.073k i/100ms
             new_way   171.379k i/100ms
             old_way      4.942M (± 3.5%) i/s -    157.935M in  32.003285s
             new_way      5.263M (± 5.8%) i/s -    167.780M in  32.021167s

janbiedermann:
Ok, benefit became visible, however, i understand now, that Object.create(null) is used to get clear Objects {}, because Opal modifies the Object prototype. This may safe some memory, but using {} is faster anyway.
Ill experiment a bit more, maybe can have both benefits.

janbiedermann:
For the record
```
                  Compile time: 4.797 (+23.29%) -> 4.776 (+32.43%) (change: -0.44%)
                      Run time: 1.379 (+12.84%) -> 1.365 (+36.10%) (change: -1.00%)
                   Bundle size: 4640.89 kB -> 4640.45 kB (change: -0.01%)
          Minified bundle size: 998.81 kB -> 998.62 kB (change: -0.02%)
            Mangled & minified: 703.92 kB -> 703.73 kB (change: -0.03%)
```

janbiedermann:
No, i was wrong, Opal does not modify the prototype of {}.

janbiedermann:
Talking @hmdne , the trade off is between object creation vs property access:
```
             old_way   184.811k i/100ms
             new_way   183.579k i/100ms
             old_way      6.085M (± 3.6%) i/s -    194.421M in  31.994209s
             new_way      5.708M (± 3.8%) i/s -    182.478M in  32.021663s
```
property access for Objects created with .create is faster, because there is no prototype to check

janbiedermann:
Another way to create Objects inline without prototype would be to use `{ __proto__: null }`
However benchmarks show:
```
.create(null)     55 Mio. ops/s ± 3.62%
{}             505 Mio. ops/s ± 4.65%
{ __proto__: null }    31 Mio. ops/s ± 2.87%
```

janbiedermann:
So with this PR i confirmed, that using Object.create(null) at places overall is currently the best solution

PR #2558, opened 2023-06-23
[WIP] [Experimental] Use typeof instead of $has_own for Hash and speed

janbiedermann:
As $$smap is mostly a Object.create(null) and it is completely in control of Opal it can be assumed, that all its props are "own", so checking for existence with typeof $$smap[prop] === "undefined" should be sufficient and multiple times faster than using $has_own.

There is one place, where regular Objects are used for $$smap, that is, when the compiler creates a $hash2 creating a regular {} inline object from context.

Lets see the effect first ...

### DIFF
--- a/benchmark-ips/bm_hash_get.rb
+++ b/benchmark-ips/bm_hash_get.rb
@@ -1,0 +1,30 @@
+Benchmark.ips do |x|
+  %x{
+    var $has_own   = Object.hasOwn || $call.bind(Object.prototype.hasOwnProperty);
+
+    function hash_get_new(hash, key) {
+      if (key.$$is_string) {
+        if (typeof hash.$$smap[key] !== "undefined") {
+          return hash.$$smap[key];
+        }
+        return;
+      }
+    };
+
+    function hash_get_old(hash, key) {
+      if (key.$$is_string) {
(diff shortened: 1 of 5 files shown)


janbiedermann:
i previously messed up $$smap and $$map

janbiedermann:
```
bundle exec opal --dynamic-require ignore --missing-require ignore -ropal/platform -gbenchmark-ips -rbenchmark/ips -A benchmark-ips/bm_hash_get.rb
         old_version   172.296k i/100ms
         new_version   177.152k i/100ms
         old_version      5.488M (± 3.3%) i/s -     27.395M in   4.998200s
         new_version      6.040M (± 4.5%) i/s -     30.116M in   4.998255s

Comparison:
         new_version:  6040384.6 i/s
         old_version:  5487952.3 i/s - 1.10x  slower
```
~10% win for $hash_get, simple case

janbiedermann:
Summary:
- Hash access can be 10% faster
- consistent use of null prototype objects nullifies overall performance gain for asciidoctor

janbiedermann:
Now showing a tiny overall gain for asciidoctor

janbiedermann:
Closing this in favour of #2561

PR #2559, opened 2023-06-24
[WIP] [Experimental] Bridge Hash to Map

janbiedermann:
use Javascript Map class as Opal Hash

currently partially working

### DIFF
--- a/lib/opal/nodes/args/extract_kwoptarg.rb
+++ b/lib/opal/nodes/args/extract_kwoptarg.rb
@@ -21,7 +21,7 @@ module Opal
 
           add_temp lvar_name
 
-          line "#{lvar_name} = $kwargs.$$smap[#{key_name.to_s.inspect}];"
+          line "#{lvar_name} = $kwargs.get(#{key_name.to_s.inspect});"
 
           return if default_value.children[1] == :undefined
 
--- a/lib/opal/nodes/call.rb
+++ b/lib/opal/nodes/call.rb
@@ -366,7 +366,7 @@ module Opal
       end
 
       add_special :__OPAL_COMPILER_CONFIG__ do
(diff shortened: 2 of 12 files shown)


janbiedermann:
Why use Map?
- keeps order of assignment like Hash
- can use any js type/object as key, like Hash
- is in some cases even faster than JS Objects
- for sure faster than maintaining 2 Objects and a Array, what Hash currently does internally
- saves some code, as all the code for maintaining 2 Objects and a Array can go

hmdne:
I was working on a similar patchset. One thing to note is that...

Ruby hash has 2 modes:
- compare_by_identity - `#__id__` semantics (this I think may be close to what JS does)
- default - `#hash` and `#eql?` semantics for Hash (internally Ruby creates a number of buckets keyed by `#hash` and additionally checks `#eql?`)

This patchset would break both of those semantics. Question is if that's OK for us.

janbiedermann:
And another problem is the "key by reference"-issue:
```
m.set([1], 2)
m.get([1]) => undefined

a = [1]
m.set(a, 2)
m.get(a) => [1]
```

still needs a workaround for that.

janbiedermann:
So non literal Hash key access may be slow, as it walks through keys comparing them. That unfortunatley includes JS Symbol and others, so these may need conversion to a literal, as future optimization if JS Symbol is used as Ruby Symbol or something else.

janbiedermann:
Ruy Hash specs test for certain implementation details, that, using Map as bridged Class, are not necessary to be implemented and executed. Implementing them would even hurt performance. Anyway, Hash < Map is possible, can be made to work, would probably work well enough, saves some code, but because of missing spec compliance i will pause this attempt and look for another way.

janbiedermann:
Closing in favour of #2563

PR #2561, opened 2023-06-25
Create Hash $$map on demand for speed and reduced memory usage

janbiedermann:
This makes $hash2 ~23% faster, likewise $hash_init.
Also benefits general Hash performance for the common case using string keys.
I measured no penalty for object keys.

This is a experiment, meant as companion for #2558 , compensating the penalty of using Object.create(null) consistently everywhere for Hash $$map and $$smap.

### DIFF
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -6,7 +6,7 @@ require 'corelib/enumerable'
 # ---
 # Internal properties:
 #
-# - $$map         [JS::Object<String => hash-bucket>] the hash table for ordinary keys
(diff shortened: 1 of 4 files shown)


janbiedermann:
There is a quirk though https://github.com/opal/opal/blob/0976f9e33ae1961f211564ec86a43f3ff4f4b103/opal/corelib/runtime.js#L2312

keeping it for now, will resolve in #2558 when ready

janbiedermann:
No longer experimental instead ready. Applications with heavy usage of $hash2 or hash_init with $$is_string keys should benefit.

janbiedermann:
closing in favour of #2567

PR #2562, opened 2023-06-26
[WIP] [Experimental] Hash without $$smap

janbiedermann:
After the positive experience of #2561 this PR explores if the $$smap optimization of Hash is still justified.

It removes the $$smap completely, having one less Object.create(null), saving time and memory.

Also the code having less branches and smaller functions, may allow modern JS engines to optimize better and overall produce better results.

In addition this PR serves as guidance how to maybe implement Hash < Map in a better way than #2559 does, basically then using self as $$map while keeping $$keys.

### DIFF
--- a/lib/opal/nodes/args/extract_kwoptarg.rb
+++ b/lib/opal/nodes/args/extract_kwoptarg.rb
@@ -21,7 +21,7 @@ module Opal
 
           add_temp lvar_name
 
-          line "#{lvar_name} = $kwargs.$$smap[#{key_name.to_s.inspect}];"
+          line "#{lvar_name} = ($kwargs.$$map[#{key_name.to_s.inspect}]) ? $kwargs.$$map[#{key_name.to_s.inspect}].value : undefined;"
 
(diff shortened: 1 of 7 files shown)


janbiedermann:
Now look at that:
```
                  Compile time: 4.472 (+20.06%) -> 4.290 (+24.87%) (change: -4.06%)
                      Run time: 1.305 (+7.84%) -> 1.299 (+21.72%) (change: -0.49%)
                   Bundle size: 4640.89 kB -> 4657.63 kB (change: +0.36%)
          Minified bundle size: 998.81 kB -> 1016.60 kB (change: +1.78%)
            Mangled & minified: 703.92 kB -> 722.13 kB (change: +2.59%)
```
Its even faster!

Well, on github that is, it may be flawed, please somebody run on your own machine, to verify and report. Thanks.
(my machines shows equal results, but also it has problems with throttling due to summer heat, producing unreliable results)

janbiedermann:
closing in favour of #2567

PR #2563, opened 2023-06-26
[WIP] [Experimental] Hash < `Map` V2

janbiedermann:
based on #2562

Lets see, if there is performance to gain from using Map ...

### DIFF
--- a/lib/opal/nodes/args/extract_kwoptarg.rb
+++ b/lib/opal/nodes/args/extract_kwoptarg.rb
@@ -21,7 +21,7 @@ module Opal
 
           add_temp lvar_name
 
-          line "#{lvar_name} = $kwargs.$$smap[#{key_name.to_s.inspect}];"
+          line "#{lvar_name} = ($kwargs.$$map[#{key_name.to_s.inspect}]) ? $kwargs.$$map[#{key_name.to_s.inspect}].value : undefined;"
 
           return if default_value.children[1] == :undefined
 
--- a/lib/opal/nodes/args/extract_kwoptarg.rb
+++ b/lib/opal/nodes/args/extract_kwoptarg.rb
@@ -16,12 +16,13 @@ module Opal
         children :lvar_name, :default_value
 
         def compile
+          helper :hash_get
           key_name = @sexp.meta[:arg_name]
           scope.used_kwargs << key_name
 
           add_temp lvar_name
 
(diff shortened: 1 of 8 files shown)


janbiedermann:
```
                  Compile time: 5.769 (+23.35%) -> 5.692 (+18.59%) (change: -1.32%)
                      Run time: 2.013 (+7.74%) -> 1.967 (+28.49%) (change: -2.30%)
                   Bundle size: 4640.89 kB -> 4621.20 kB (change: -0.42%)
          Minified bundle size: 998.81 kB -> 989.06 kB (change: -0.98%)
            Mangled & minified: 703.92 kB -> 694.75 kB (change: -1.30%)
```
A bit faster, there may be some room for optimizations here and there.

janbiedermann:
Now there is a literal Hash mode, for tyeof "string", "number" and global "symbol" keys, which uses Hash < Map itself, the JS Map, literally, for key and values. This makes `$hash2()` obsolete, replaced by `(new Map())`

If a object is passed as key, Hash is tronsformed to object mode, which uses a self.$$buckets array for buckets and uses the map  as `key_hash => bucket list` store, well just the $$map before.

The good news is, in literal Mode a JS Map is interchangable with a Ruby Hash, although #freeze is not honored by JS.

The bad news, in object mode, interchanging a JS Map with a Hash is mostly useless, as before, and 30 specs are still failing for no reason, apparently ;).

janbiedermann:
```
                  Compile time: 4.649 (+18.10%) -> 4.553 (+15.33%) (change: -2.05%)
                      Run time: 1.393 (+5.69%) -> 1.358 (+7.82%) (change: -2.54%)
                   Bundle size: 4640.89 kB -> 4628.17 kB (change: -0.27%)
          Minified bundle size: 998.81 kB -> 990.23 kB (change: -0.86%)
            Mangled & minified: 703.92 kB -> 698.44 kB (change: -0.78%)
```
Performance and asset size look good.

janbiedermann:
There is a performance trade off here. Performance is gained because of using literal Map for kwargs ans Hashes with simple keys, but at the same time performance is lost when iterating through entries, because arrays for entries or copies of $$bucket are created each time, more often than before. See the introduced $hash_each() helper.

janbiedermann:
closing in favour of #2567

PR #2565, opened 2023-06-27
Improve `Array.push()` performance when pushing many items

janbiedermann:
in a few places

Also initialize doesnt need to splice if length is 0, faster

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -76,6 +76,22 @@ class ::Array < `Array`
       }
       return #{`array`.to_a};
     }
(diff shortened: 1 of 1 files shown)


janbiedermann:
Looks like it has a positive on the compiler too:
```
Compile time: 4.550 (+26.76%) -> 4.311 (+21.89%) (change: -5.25%)
       Run time: 1.329 (+11.89%) -> 1.327 (+33.74%) (change: -0.14%)
```

elia (approved):
👍

PR #2566, opened 2023-06-28
Add CGI::Util#(un)escapeURIComponent

janbiedermann:
Could not find specs for it in spec/ruby

### DIFF
--- a/stdlib/cgi.rb
+++ b/stdlib/cgi.rb
@@ -1,0 +1,14 @@
+class CGI
+  module Util
+    def escapeURIComponent(c)
+      `encodeURI(c)`
+    end
+
+    def unescapeURIComponent(c)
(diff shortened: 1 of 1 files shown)


stdlib/cgi.rb
```diff
@@ -0,0 +1,13 @@
+class CGI
+  module Escape
```
hmdne, line 2: why not CGI::Util like in upstream?

https://github.com/ruby/cgi/blob/master/lib/cgi/util.rb

janbiedermann: Well, i looked here https://docs.ruby-lang.org/en/master/CGI/Escape.html, otherwise no reason, can change it, no problem

elia (approved):

PR #2567, opened 2023-07-08
Hash < Map V3

janbiedermann:
Building on V2 #2563 , but there where 2 issues in general when using `Hash < Map`, one issue is, that keys of a Map are references and the second issue are hash collisions, where previously buckets where used as linked list of items with the same hash.

However, if two different objects have the same hash, they still have different references, so we can use that together with the "keys are references" feature of Map and use the objects as keys directly.

Now of course they would become unreachable, so we also use a $$keys Map internally, on demand, to map hash_keys of objects to the actual objects that are used as keys, basically mapping hash_keys to references, this way keeping objects reachable.

So we can directly map keys to values, no more buckets are needed.

This further simplifies Hash code overall.

Primitives "string", "number" and "symbol" (must be global) are directly used as keys.

Now with this a ruby Hash can be exchanged with a Javascript Map freely, those primitive types above are directly accessible, like in V2, but now also objects as keys are directly accessible from JS ... by reference.

Not sure how JS is supposed to have a reference to Opal Ruby objects or the other way around, how practical that is, if it ever gets used that way, but it works.

If a Hash is passed to JS, modified by JS and received back from JS, and it has Object keys added or modified by JS, calling Hash#rehash may be advisable to keep Objects accessible, but also may fail, if a JS object used as key doesn't have a .$hash() function.

In other words, if a Opal Ruby Hash is passed to JS, within JS as Map it behaves like Hash#compare_by_identity

### DIFF
--- a/docs/bridging.md
+++ b/docs/bridging.md
@@ -91,9 +91,9 @@ With bridging creating equivalent Ruby and JavaScript objects, no explicit type 
 
 ## Bridged Classes in Corelib
 
-Opal's core library bridges several commonly used JavaScript classes, including `Array`, `Boolean`, `Number` (think: `Float` which also may act as an `Integer`), `Proc`, `Time`, `RegExp`, and `String`, allowing interaction with these JavaScript instances as their Ruby equivalents.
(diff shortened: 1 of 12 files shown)


janbiedermann:
```
Compile time: 4.723 (+23.87%) -> 4.487 (+34.14%) (change: -5.00%)
                      Run time: 1.409 (+6.54%) -> 1.380 (+35.46%) (change: -2.06%)
                   Bundle size: 4640.89 kB -> 4618.10 kB (change: -0.49%)
          Minified bundle size: 998.81 kB -> 988.38 kB (change: -1.04%)
            Mangled & minified: 703.92 kB -> 697.18 kB (change: -0.96%)
```
Looking good

Issue #2514, opened 2023-01-05
Bug: kwrestarg (**kwargs) doesn't support non-String (Symbol) keys


hmdne:
**Describe the bug**
The non-String keys are dropped, because kwargs extraction logic is simplified.

Opal version: all/1.7

```
# opal -v -e 'def a **kwargs; kwargs; end; p a true => false'
Opal v1.7.0
{}
```

```
# ruby -v -e 'def a **kwargs; kwargs; end; p a true => false'
ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
{true=>false}
```

janbiedermann:
with #2567 this can directly be supported in Hash by adding "boolean" to places like this https://github.com/opal/opal/blob/ec1f37963e27bb2f634fe0a4b37b5cd4ad0b401e/opal/corelib/runtime.js#L2243

janbiedermann:
actually any object is possible, but if objects are used as keys and the compiler inlines new Map(), Hash#rehash must be called for object keys to be accessible (and a condition in there be changed, testing for $$keys)
Without object key kwarg:
`(new Map([['kwarg', '23']])`
with object key kwarg:
`(new Map([[some_object, '23']]).$rehash()`

hmdne:
This is fixed now.

PR #2568, opened 2023-07-09
Hash < `Map` V3 and support for Object kwargs

janbiedermann:
Building on #2567 and Fixing #2514

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,5 +33,6 @@ module.exports = {
     "Reflect": "readonly",
     "Uint8Array": "readonly",
     "WeakRef": "readonly",
+    "Map": "readonly",
   }
 };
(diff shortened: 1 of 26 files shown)


janbiedermann:
This unifies Hash compilation by always using `(new Map())` in variations, depending on entries.
This also in way allows Hashes with String objects as keys, if somehow inserted by JS, Hash#rehash must be called.
If object keys are generated by the compiler, it `$hash_rehash(new Map([', ']))` rehashes.

janbiedermann:
```
Compile time: 4.796 (+20.75%) -> 4.391 (+22.14%) (change: -8.44%)
                      Run time: 1.326 (+7.66%) -> 1.280 (+31.78%) (change: -3.44%)
                   Bundle size: 4640.89 kB -> 4616.75 kB (change: -0.52%)
          Minified bundle size: 998.81 kB -> 988.40 kB (change: -1.04%)
            Mangled & minified: 703.92 kB -> 697.24 kB (change: -0.95%)
```
Still looking good

PR #2569, opened 2023-07-22
Date DateTime Time ::parse improvements, implement Date::_parse

janbiedermann:
- use common constants
- implement Date::_parse
- use that for Date and DateTime::parse
- add a few more formats
- makes 27 more specs work

Still to do:
- add time formats
- move _parse code to Time and attach to Opal.date_time_parse (because Time is loaded first and also can use it for Time::parse)
- use Opal.date_time_parse from Date::_parse
- also attach constants to Opal

Bonus, optional maybe todo, would be nice to have:
- implement #cwyear


hmdne:
I have rebased it on top of `master`, then ran `bundle exec rubocop -a` to fix the remaining lint errors. Ensure you mark it as ready for review once it is :D

janbiedermann:
Thanks, it still needs some time until i have time to continue working on it

PR #2571, opened 2023-07-27
Hash#rehash - optimize for the common case, no hash collision, avoid calling $slice

janbiedermann:


janbiedermann:
So synthecially measuring this line, this patch improves execution time for this line by ~60% for the common case, but here on github, this patch shows a slight negative overall effect in our benchmarks. What can i say ...

PR #2572, opened 2023-08-02
Fix Module#clone/dup

takaram:
`Module#clone` and `Module#dup` had been defined, but it virtually didn't work since methods defined on the module were not copied.
`Class#clone` and `Class#dup` had been returning `Module` instead of `Class`.

This PR implements `Module#clone`, `Module#dup`, `Class#clone` and `Clone#dup`.

Note that this PR still doesn't support cloning bridged classes. e.g. `String.clone.new` raises an exception.


elia (commented):
Awesome work!

Just left a comment about method cloning

elia (approved):
Thanks, this is great!

PR #2573, opened 2023-08-07
The small symbolization

hmdne:
Ref: #2144

This PR will explore ways to gradually move to new API that sets method keys using JavaScript symbols and make that API available before we do the full migration.

### DIFF
--- a/lib/opal/nodes/iter.rb
+++ b/lib/opal/nodes/iter.rb
@@ -69,7 +69,8 @@ module Opal
           in_closure(Closure::JS_FUNCTION | Closure::ITER | (@is_lambda ? Closure::LAMBDA : 0)) do
             body_code = stmt(returned_body)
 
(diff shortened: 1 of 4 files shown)


hmdne:
```
                  Compile time: 4.536 (+25.22%) -> 4.490 (+24.91%) (change: -1.01%)
                      Run time: 1.224 (+8.87%) -> 1.223 (+36.85%) (change: -0.02%)
                   Bundle size: 4625.64 kB -> 4718.50 kB (change: +2.01%)
          Minified bundle size: 990.33 kB -> 1074.65 kB (change: +8.51%)
            Mangled & minified: 698.46 kB -> 687.68 kB (change: -1.54%)
```

The performance is almost the same (even though we haven't introduced real Symbols yet). Bundle size increase for minified is expected.

PR #2575, opened 2023-08-16
corelib/runtime/truthy: break early if we passed a literal true

merceroncode:
Also I'm removing the whole instanceof and valueOf check, since apparently it's not required anymore

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -183,7 +183,9 @@
   // -----
 
   var $truthy = Opal.truthy = function(val) {
-    return false !== val && nil !== val && undefined !== val && null !== val && (!(val instanceof Boolean) || true === val.valueOf());
+    if (true === val)
+      return true;
+    return false !== val && nil !== val && undefined !== val && null !== val;
   };
 
   Opal.falsy = function(val) {


opal/corelib/runtime.js
```diff
@@ -183,7 +183,9 @@
   // -----
 
   var $truthy = Opal.truthy = function(val) {
-    return false !== val && nil !== val && undefined !== val && null !== val && (!(val instanceof Boolean) || true === val.valueOf());
```
elia, line 186: There are two specs that are failing in which they explicitly use a boxed boolean, not sure if anything changed so that it's a super contrived example and basically impossible to recreate that situation, but we need to be sure that's the case before dropping support for this.

hmdne: I have worked together with @merceroncode and we noted that just simply adding `return true` improved performance slightly.

I have worked with you @elia together in the past on this line of code. I am in support of dropping support for boxed booleans for a couple of reasons:

* They don't occur naturally in Opal code
* They aren't even supported by JavaScript

elia: 🙌 Happy to drop support given it's all expected to be strict code now:

```rb
`typeof #{true.itself}` # => 'boolean'
```

PR #2576, opened 2023-09-03
Fix `String#object_id`, `String#__id__`, `String#hash`

### LINKED ISSUE #2257: Bug: Object Id is not being returned
**Describe the bug**

When I try to run the code "hello".object_id the output I am getting is hello but it should have actually returned the object_id.

**To Reproduce**
Screenshot 2021-07-21 at 3 37 20 PM

You can try it out here in this [link](https://academy.bigbinary.com/learn-ruby/pages/what-is-symbol)

takaram:
Closes #2257

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -13,17 +13,27 @@ class ::String < `String`
     var string_id_map = new Map();
   }
 
-  def __id__
-    %x{
-      var id, value = self.valueOf();
-      if (string_id_map.has(value)) {
-        return string_id_map.get(value);
+  # Force strict mode to suppress autoboxing of `this`
+  %x{
+    (function() {
+      'use strict';
+      #{
+        def __id__
+          %x{
+            if (typeof self === 'object') {
+              return #{super}
+            }
+            if (string_id_map.has(self)) {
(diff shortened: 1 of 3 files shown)


hmdne:
That's a very cool idea, but there are also wrapped Strings to be concerned about. The `#__id__` method should correspond to unique objects and wrapped Strings are unique (they may contain info about encoding, let's say). So for wrapped strings this could be just delegated to the default implementation, since they are mutable objects, unlike string literals (except for the non-strict mode, but we don't really want to care about it anymore).

Also, for the non-wrapped strings, we could use WeakMap, as this will facilitate garbage collection.

takaram:
> So for wrapped strings this could be just delegated to the default implementation

Thank you, I changed the code.

> for the non-wrapped strings, we could use WeakMap

Unfortunately, WeakMap doesn't accept primitive strings as keys.
I agree that we should allow strings be garbege collected, but I need to investigate how I can realize that.

hmdne:
Thanks! Well, if we can't use WeakMap, that likely means that string literals are never GC'ed. But let me take it from there, as I have an idea of changing `#hash` implementation throughout the codebase to be more akin to how JS handles things.

PR #2579, opened 2023-09-10
Pass value in PromiseV2#always just like PromiseV1#always does it

janbiedermann:
Makes PromiseV2 more compatible with PromiseV1

### DIFF
--- a/stdlib/promise/v2.rb
+++ b/stdlib/promise/v2.rb
@@ -144,6 +144,7 @@ class PromiseV2 < `Promise`
       `Promise.resolve(#{value})`.tap do |prom|
         prom.instance_variable_set(:@type, :resolve)
         prom.instance_variable_set(:@realized, :resolve)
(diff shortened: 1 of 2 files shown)


elia (commented):
Great catch!

I added a commit with a couple of tests that pass using PromiseV1 and are partially failing with this code, can you please look into it? I couldn't find any obvious solution 🙏

janbiedermann:
Things i would still like to do:
- mark the other private methods private
- make V2 work in c-Ruby
- make it isomorphically available in lib, not necessarily required, but would be nice

hmdne:
>make V2 work in c-Ruby
Well... I'm afraid we would need to use V1 not V2, but both are mostly compatible (we are finding incompatibilities just now).
>make it isomorphically available in lib
Creating an isomorphic core is a great idea. But - I'd say, we would want to rename it to Opal::Promise perhaps.

PR #2582, opened 2023-09-13
`#hash` now returns integers

hmdne:
This is a followup to opal/opal#2576. The idea is to ensure that `#hash` return integers, as Ruby does. We use a very lightweight hashing algorithm we have created for this purpose, called `Opal32` which should most of the time return unique values.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
     "Proxy": "readonly",
     "Reflect": "readonly",
     "Uint8Array": "readonly",
+    "Int32Array": "readonly",
     "WeakRef": "readonly",
     "Map": "readonly",
   }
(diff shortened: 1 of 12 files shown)


hmdne:
The updated changeset keeps the previous version of a Map of Strings/Floats, but only for calls on `#__id__`. Hash now correctly computes a hash for each type. `#__id__` needs uniqueness though.

Perhaps a solution could be to use WeakMap with wrapped values. It would be great to benchmark that somehow, but unfortunately WeakMap does not seem to be inspectable.

cc: @elia @takaram

elia:
@hmdne even if weakmap is not inspectable we can assess its performance from outside, especially if we can generate lots of objects and trigger garbage collection programmatically. That would ensure we can understand both the performance profile and the memory footprint.

PR #2584, opened 2023-09-13
 Use a Map instead of a POJO for the jsid_cache

### LINKED ISSUE #2583: Bug: jsid_cache breaks for `"toString"`, and I suspect other methods of JS Object
**Describe the bug**

Opal version: v1.7.0 - v1.7.3

The cache added in #2481 breaks for the method `toString`, and I suspect will break for other Object methods, like `valueOf`.

`jsid_cache["toString"]` evaluates to the `toString()` method on the cache object itself, instead of resulting in a cache miss and building the string `"$toString"`.

https://github.com/opal/opal/blob/f02fa40875c51b388de9c1b534459d363bcf951e/opal/corelib/runtime.js#L1868

**To Reproduce**

Call `Opal.add_stubs('toString')`, or write some Ruby code that will compile with such a call.

This was discovered while trying to upgrade [18xx.games](https://18xx.games/) to use Opal 1.7. We have a class which wraps the JS [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class, and it needs to call `toString` on the wrapped object, leading to a runtime error:

https://github.com/tobymao/18xx/blob/49a570c9237ac9739023db780f69a58c336fcd1b/assets/app/lib/params.rb#L36-L38

michaeljb:
Fixes #2583, introduced in #2481

Would it be possible for this fix to be back-ported, for a 1.7.4 release?

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1863,9 +1863,14 @@
   }
 
   // Optimization for a costly operation of prepending '$' to method names
-  var jsid_cache = {}
+  var jsid_cache = new Map();
   function $jsid(name) {
-    return jsid_cache[name] || (jsid_cache[name] = '$' + name);
+    var jsid = jsid_cache.get(name);
+    if (!jsid) {
+      jsid = '$' + name;
+      jsid_cache.set(name, jsid);
+    }
+    return jsid;
   }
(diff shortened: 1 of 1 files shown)


elia (approved):
Thanks!

PR #2586, opened 2023-09-18
[WIP] Bridge improvements

janbiedermann:


opal/corelib/class.rb
```diff
@@ -59,7 +68,15 @@ def inherited(cls)
 
   def new(*args, &block)
     %x{
-      var object = #{allocate};
+      var object;
+
+      if (self.$$bridge && self.$allocate == Opal.Class.prototype.$allocate) {
```
hmdne, line 73: What if... we subclass a bridged class?

```ruby
class Array < `Array`; end
class MyArray < Array; end
class YourArray < MyArray; end
```

(The lack of awareness of that edge case caused in the past a hard-to-debug issue where we missed backtraces in Firefox - due to not launching a constructor - if Exception was subclassed more than once - see #2540 ).

hmdne:
Also, take a look at the prior work on this matter:

https://github.com/opal/opal/pull/2344

opal/corelib/class.rb
hmdne: For comparison, we have something that's called `pristine` methods. This is an option for us to define a method to be unmodified from the default implementation. Perhaps this is a better idea.

janbiedermann: Alright, i will pay attention to that case, i still need to understand better whats going on

janbiedermann:
So the last patch makes bridge for subclasses work and no longer cuts of the prototype chain
```
class A { a() {return 'a'}}
class B extends A { b(){return 'b'}}
class C extends B { c(){return 'c'}}
Opal.klass(Opal.Kernel, C, 'C')

d = Opal.Kernel.C.$new()
d.c() -> 'c' - worked before
d.b() -> 'b'  - before b() was gone, now works
d.a() -> 'a' - before a() was gone, now works
```
There are still issues though

janbiedermann:
At this stage the following works:
```ruby
require 'engine'
# Engine now allow access to globalThis, so n the browser for example
Engine.history.length # -> 2
```
manually in the browser console in js:
```
Engine.$send('history').$send('length') // -> 2
```
#send is required, because method_missing is used

janbiedermann:
further works:
```
require 'engine'
Engine.history.class.name # -> Engine::History

a = Engine::ArrayBuffer.new()
a.class.name # -> Engine::ArrayBuffer
```
the last one in js:
```
Opal.Engine.$const_get('ArrayBuffer').$new().$class().$name() // -> 'Engine::ArrayBuffer'
```

stdlib/engine.rb
```diff
@@ -0,0 +1,137 @@
...
+    def [](property)
+      method_missing(property)
+    end
+
+    def method_missing message, *args, &block
+      if message.end_with? '='
+        message = message.chop
+        property_name = ::Engine.property_for_message(message, self)
+        arg = args[0]
+        arg = arg.to_n if `arg && typeof arg.$to_n === 'function'`
```
hmdne, line 14: what if $to_n is a stub? It's still a function then.

hmdne:
I kind of fail to understand what Opal::Engine does. Is it yet another wrapping solution? Wouldn't it be a better idea to split that to another PR?

janbiedermann:
Possibly, its handy for testing things, as its the goal to have automatic, easy and compact access to JS APIs and values. Once everything is working i may make that another PR or not, we will see

janbiedermann:
For correctness when using Engine for testing the bridges proeprties must be accessed by [], eg:
```
Engine[:history]
Engine[:history][:length]
Engine[:history].back
```
May work more or less, anymway
Continuing from here ...

PR #2589, opened 2023-09-19
Use literal instead of Object.create(null) for speed

janbiedermann:
tiny improvement


elia (approved):
image
tiny indeed, but looks like it's a standardized way of setting the proto

_pending the conflicts_

janbiedermann:
grafik

Try the same with assignment, how we use it

janbiedermann:
grafik

Not so tiny at all

PR #2591, opened 2023-09-24
Unexpected return/break should raise LocalJumpError

takaram:
Orphaned `Proc` objects will throw `LocalJumpError` on return or break.
See https://ruby-doc.org/3.2.2/Proc.html#class-Proc-label-Orphaned+Proc

Some related bugs are fixed too.

### DIFF
--- a/lib/opal/nodes/closure.rb
+++ b/lib/opal/nodes/closure.rb
@@ -102,7 +102,7 @@ module Opal
         def generate_thrower(type, closure, value)
           id = closure.register_catcher(type)
           closure.register_thrower(type, id)
-          push id, '.$throw(', expr_or_empty(value), ')'
+          push id, '.$throw(', expr_or_nil(value), ', ', scope.identify!, '.$$is_lambda)'
           id
         end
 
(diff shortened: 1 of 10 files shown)


lib/opal/nodes/closure.rb
```diff
@@ -223,11 +225,13 @@ def compile_catcher
             end
             line "throw $e;"
           end
-          line "}"
+          line "} finally {", *catchers_without_eval_return.map { |type| "$t_#{type}.is_orphan = true;" }, "}"
```
hmdne, line 228: Can we skip the finally clause if `catchers_without_eval_return` is empty?

takaram: Yes, I changed the code.

elia (approved):
This is great! Not the easiest part to tackle 🙌 🙌 🙌

hmdne:
This PR caused a significant failure at some codebase I'm working on (namely - https://github.com/plurimath-js/plurimath-js ). I will give more info when I get to debug it.

PR #2592, opened 2023-09-28
Compute `$@` dynamically

hmdne:
This should improve performance a little, so that we don't have to reparse the backtrace for exceptions when we don't need it.


elia (approved):
@hmdne I pushed an additional commit removing the added runtime helper, I think if we do it, we should apply it everywhere

PR #2594, opened 2023-10-02
Ensure that the identifiers are correctly represented in Web Consoles.

hmdne:


opal/corelib/runtime.js
```diff
@@ -601,13 +608,15 @@
...
     } else {
       klass = function(){};
     }
-
-    if (name && name !== nil) {
-      $prop(klass, 'displayName', '::'+name);
-    }
+  
+    $prop_gs(klass, "name", false, $mod_name_getter);
+    $prop_gs(klass, "displayName", false, $mod_name_getter);
```
elia, line 619: Would be possible to define these on `Class` itself (or, even better, `Module`!) and rely on `$$name` dynamically?

`$set_proto(klass, Opal.Class.prototype);`

opal/corelib/runtime.js
```diff
@@ -773,14 +782,18 @@
   //
   // @return [Module]
   function $allocate_module(name) {
-    var constructor = function(){};
-    var module = constructor;
+    var module;
+    if ($truthy(name)) {
+      module = Function("return function "+name+" (){}")();
```
elia, line 787: Can you add a comment explaining why we need this?

elia (approved):
Left a couple of comments, but it already looks good 👍

opal/corelib/runtime.js
hmdne: No. `name` is set during function definition, which overrides setting on prototype.

PR #2595, opened 2023-10-02
Optimize Kernel#loop

hmdne:


takaram:
This will cause a problem:

```console
$ bin/opal -e 'loop(&-> { return })'

/tmp/opal-system-runner20231004-47362-rqbq0f.js:5473
        } else { throw $err; }
                 ^
{
  '$thrower_type': 'return',
  '$throw': [Function: $throw],
  is_orphan: false,
  '$v': klass {
    '$$id': 4,
    apply: [Function (anonymous)],
    call: [Function (anonymous)],
    '$$frozen': true,
    '$$comparable': false
  }
}
```

Actually this problem was there (for all `yield` calls) until v1.8.0.alpha1 and fixed in #2591.

hmdne:
Thanks for the heads up! In fact, when doing the closure tracking part, I ensured that such an expression would simply compile to `return nil`, I guess I will have to revisit that part.

takaram:
This can be fixed by `Proc#call`.
However, considering there are many `block()` calls in corelib codes, maybe it's better to make lambda function handle the thrower  by itself rather than fixing this patch.

PR #2598, opened 2023-10-12
Improve `Range#include?`/`member?`/`cover?`/`===`

takaram:
These four methods are very similar, but bit different.This PR makes these methods more compatible with CRuby.


elia (approved):
✨ Nice!

PR #2607, opened 2023-11-08
Move YAML from `nodejs/yaml` into `yaml` so that it can be used outside Node

hmdne:


stdlib/nodejs/yaml.rb
```diff
@@ -1,19 +1 @@
...
-    load(`#{File}.__fs__.readFileSync(#{path}, 'utf8')`)
-  end
-
-  def self.load(data)
-    loaded = `__yaml__.safeLoad(data)`
-    loaded = Hash.new(loaded) if native?(loaded)
-    loaded
-  end
-end
+warn "DEPRECATED: requiring nodejs/yaml is deprecated: use `require 'yaml'` instead"
```
elia, line 1: can you add the `require 'yaml'` just after the warning, so that `require 'nodejs/yaml'` will still work and we'll drop the whole file at the next major

PR #2619, opened 2023-11-26
Improve string interpolation

### LINKED ISSUE #1068: if to_s does not return a string, things blow up
``` ruby

class Foo < String
end

class Bar 
  def to_s
    Foo.new
  end
end

puts "abc#{Bar.new}def"

```

dies with

`TypeError: Cannot convert object to primitive value`

### LINKED ISSUE #2618: Bug: String interpolation with bridged classes does not call `to_s`
**Describe the bug**
When some of bridged classes, such as `Array`, `Proc`, `Time`, are interpolated into a string, Opal does not call `to_s`.

Opal's string interpolation depends on JavaScript's `Object.prototype.toString` which is overwritten by Opal. However, since some of the native JavaScript classes have their own `toString`, they take priority over Object's one.

Opal version: v1.8.2 / v2.0.0dev

**To Reproduce**
```console
$ ruby -ve 'puts "#{[1, 2, 3]}\n#{Time.new}\n#{-> {}}"'
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
[1, 2, 3]
2023-11-26 19:40:53 +0900
#<Proc:0x00007fd0a26fd180 -e:1 (lambda)>
```

```console
$ bin/opal -ve 'puts "#{[1, 2, 3]}\n#{Time.new}\n#{-> {}}"'
Opal v2.0.0dev
1,2,3
Sun Nov 26 2023 19:40:47 GMT+0900 (Japan Standard Time)
function() {
      return arg;
    }
```

takaram:
No longer use `+` for string concatenation and introduce `Opal.interpolate` to handle object stringification.

- Call `Object#to_s` directly, instead of depending on `Object.prototype.toString`.
  - Closes #2618
- Fall back to the default `Object#to_s` implementation if `Object#to_s` does not return a string.
  - Closes #1068

### DIFF
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -259,27 +259,14 @@ module Opal
       handle :dstr
 
       def compile
-        if children.length > 1 && children.first.type == :str
-          skip_empty = true
-        else
-          push '""'
+        helper :interpolate
+
+        push '$interpolate('
+        children.each_with_index do |part, index|
+          push ', ' unless index == 0
+          push expr(part)
         end
-
-        children.each do |part|
-          if skip_empty
-            skip_empty = false
-          else
-            push ' + '
-          end
-
-          if part.type == :str
-            push expr(part)
-          else
-            push '(', expr(part), ')'
-          end
-
-          wrap '(', ')' if recv?
-        end
+        push ')'
       end
     end
 
--- a/lib/opal/nodes/literal.rb
+++ b/lib/opal/nodes/literal.rb
@@ -259,14 +259,19 @@ module Opal
       handle :dstr
 
       def compile
-        helper :interpolate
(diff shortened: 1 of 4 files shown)


hmdne:
There's a consistent slight performance regression (~3% on AsciiDoctor). This involves use of `Array.map` I believe. Let me try to correct that.

hmdne:
Also, from my other experience, kinda counterintuitively `"abc"+"abc"` is more performant than `["abc", "abc"].join("")`

hmdne:
I have attempted various ideas, but from my tests, this one is the best for performance. We'll likely merge it the next review window (next Thursday).

Congratulations on your consistent work of getting us closer to MRI, one step at a time, @takaram :D

PR #2621, opened 2023-12-03
Make Regexp with n flag compile

janbiedermann:
Required to compile Rack


lib/opal/nodes/literal.rb
```diff
@@ -173,6 +173,10 @@ def compile_static_regexp
           # errors) - at least since Node 17.
           static_as_dynamic(value)
         else
+          if flags.include?('n')
+            value.encode!('ASCII-8BIT')
```
hmdne, line 177: `encode!` may not work under Opal

hmdne:
Other than the comment, can you add a simple test? I know encodings are a mess, but just a test to prove `n` is now accepted by Opal.

PR #2633, opened 2024-01-30
Speed up {Array,Hash}#hash and a couple of Hash operations

janbiedermann:
I have a use case where Hash#key? computing hashes for a not too large Hash/Array combo takes ~ 3 seconds, for the combination if things going on 10 seconds are wasted.
Using a Map instead of Object.create(null) for $hash_ids seems to reduce that to a 11ms for Hash#key? and the combination ~100ms or so.

Have not tried other use cases or anything else, let see result of the actions ...

### DIFF
--- a/opal/corelib/array.rb
+++ b/opal/corelib/array.rb
@@ -1296,10 +1296,12 @@ class ::Array < `Array`
       }
 
       try {
-        values = $hash_ids.values();
-        for (item of values) {
-          if (#{eql?(`item`)}) {
-            return $opal32_add(result, 0x01010101);
+        if (!top) {
+          values = $hash_ids.values();
+          for (item of values) {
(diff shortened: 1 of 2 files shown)


janbiedermann:
Something broke, needs some tuning ...

janbiedermann:
The huge speedup is due to error on my side, bit a more realistic ~10% speedup remains with things fixed, for my use case, using large Hash/Array combo as key with Hash#key?

janbiedermann:
Another 10% speedup by avoiding function calls

janbiedermann:
Another 10%, from 10s down to 7s.
Using the native Map iterator may benefit other places in Hash too

janbiedermann:
Now down to below 7 seconds from 10 for the complete use case and down to 1.5s from 3s for Hash#key?, nice.

PR #2634, opened 2024-02-01
[WIP] Update Firefox runner to fix Ubuntu/Snap problems

janbiedermann:
When running `bundle exec rake mspec_firefox` on ubuntu 23.10 with Firefox 122 installed from Snap (there is no alternative way provided out of the box) Firefox does not open the remote debugging port.
Firefox fails to find the generated profile.

The firefox executable in the path, `/snap/bin/firefox` in fact is a link to the snap executable, which, when executed as firefox, mangles firefox options to provide paths from the snap directories. If a option with arg is separated by space, like "--profile /dir", the option is somehow mistreated and not recognized. If a option with arg is connected by '=', like `--profile=/dir,` the option is recognized and correctly passed on to the real firefox-bin executable.

Changing the way opts and args are passed an the patch at least makes firefox open the remote debugging port, but execution currently still fails with:
`chrome://remote/content/cdp/domains/content/Runtime.sys.mjs:285:5:in 'Error: context is null (evaluate': context is null (evaluate@chrome://remote/content/cdp/domains/content/Runtime.sys.mjs:285:5 ...`

This patch also updates options and prefs to the latest ones as used by latest puppeteer.


janbiedermann:
Btw, mspec-firefox from ci is not affected, because it installs firefox in a different way that avoids using snap.

hmdne:
Unfortunately it is :(. It is now running for multiple hours after this patch

janbiedermann:
Good things need time  ... ;)

janbiedermann:
failing specs are unrelated

lib/opal/cli_runners/firefox.rb
```diff
@@ -134,6 +134,15 @@ def run_firefox_server
       rescue Timeout::Error
         puts 'Failed to start firefox server'
         puts 'Make sure that you have it installed and that its version is > 100'
+        if !OS.windows? && !OS.macos?
+          # The firefox executable within snap is in fact the snap executable which sets up paths and then calls the
+          # real firefox executable. It also mistreats passed options and args and always tries to reuse a existing
+          # instance. Thus firefox from snap, when called by this runner, will start with the wrong options, the
+          # wrong profile and will fail to setup the remote debugging port
```
hmdne, line 141: Wild idea: What if we resolve a correct path to Firefox from snap, ie. /snap/asdad/bin/firefox and call it directly?

janbiedermann: I had that wild idea. The question is how and how to make that reliable?

PR #2635, opened 2024-02-02
Initial `fork()` support for Node.js

janbiedermann:
Working, it may happen though, that the parent process exits before the child was able to read its javascript file. 
Happens with SystemRunner and its tempfile, which gets deleted when the parent exits. It will fork and then print:
`Error: Cannot find module '/tmp/opal-system-runner20240202-905467-v7fj7d.js'`

In ruby, such situation will still work, because fork relies on the image in memory. In node, the original javascript file is still required when forking.

That of course is not a issue at all, if the parent process keeps running anyway.

This updates the node requires in nodejs/kernel.rb to use the modern form.

### DIFF
--- a/stdlib/nodejs.rb
+++ b/stdlib/nodejs.rb
@@ -1,5 +1,6 @@
 require 'nodejs/base'
 require 'nodejs/kernel'
+require 'nodejs/process'
 require 'nodejs/file'
 require 'nodejs/dir'
 require 'nodejs/io'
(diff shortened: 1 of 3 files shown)


janbiedermann:
```
if RUBY_ENGINE == 'opal'
require 'nodejs'
end

def cpu_intensive_process
  puts "Pid: #{Process.pid}"
  x = 0
  1000000000.times do |i|
    x = i + x
  end
  puts x
end

# either this way, may fail with the error mentioned above, if there wouldn't be the other forks below
# keeping this process busy
fork do
  cpu_intensive_process
end

# or that way
fork
cpu_intensive_process
```

Example for testing

PR #2638, opened 2024-02-04
Profiling convenience with a nice flame graph

janbiedermann:


janbiedermann:
But i just had the idea, to make that OPAL_FLAME and also find a similar solution for the bun and deno runners, need to investigate further

janbiedermann:
Deno team is working on that: https://github.com/denoland/deno/issues/22102
Bun currently has nothing to offer. Lets go with NODE_FLAME for the moment.

hmdne:
Here it's how it looks like:

image

PR #2639, opened 2024-02-04
Improve instance variable performance

janbiedermann:
This patch improves $each_ivar, with positive effects on Up! benchmark results for the env.to_s benchmark:
from 8328.87 req/s to 13105.05 req/s, thats ~50% up. Probably has positive effects on other apps too.


elia (approved):
🎈 
🏡

Issue #1703, opened 2017-09-30
Incompatibility in multiple assignment


yhara:
I found this code behaves differently between Opal (git master) vs. Ruby (2.4.2).

```rb
((x, y)) = [[1,2]]
```

Here is the results:

```
$ ./bin/opal -e '((x, y)) = [[1,2]]; p(["x", x, "y", y])'
["x", [1, 2], "y", nil]
$ ruby -e '((x, y)) = [[1,2]]; p(["x", x, "y", y])'
["x", 1, "y", 2]
```

elia:
Yes, from what I found it ignores the double parens on the lhs of the assignment:

```
opal:master ⤑ ruby -ve '(x, y) = [[1,2]]; p(["x", x, "y", y])'                                                                         ~/C/opal
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]
["x", [1, 2], "y", nil]
```


That's reflected in the syntax tree too:

```
opal:master ⤑ be bin/opal -e '((x, y)) = [[1,2]]' --sexp                                                                               ~/C/opal
s(:masgn,
  s(:mlhs,
    s(:lvasgn, :x),
    s(:lvasgn, :y)),
  s(:array,
    s(:array,
      s(:int, 1),
      s(:int, 2))))
opal:master ⤑ be bin/opal -e '(x, y) = [[1,2]]' --sexp                                                                                 ~/C/opal
s(:masgn,
  s(:mlhs,
    s(:lvasgn, :x),
    s(:lvasgn, :y)),
  s(:array,
    s(:array,
      s(:int, 1),
      s(:int, 2))))
```

so it seems to be parser gem issue too:

```
opal:master ⤑ ruby-parse -e '((x, y)) = [[1,2]];'                                                                                                                     ~/C/opal
warning: parser/current is loading parser/ruby24, which recognizes
warning: HEAD-compliant syntax, but you are running 2.4.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
(masgn
  (mlhs
    (lvasgn :x)
    (lvasgn :y))
  (array
    (array
      (int 1)
      (int 2))))
opal:master ⤑ ruby-parse -e '(x, y) = [[1,2]];'                                                                                                                       ~/C/opal
warning: parser/current is loading parser/ruby24, which recognizes
warning: HEAD-compliant syntax, but you are running 2.4.0.
[...]

PR #2642, opened 2024-02-15
Add an option to use Prism parser

takaram:
Ruby 3.3 introduced Prism parser, and it is also released as a gem.
The latest version of Prism includes `Prism::Tarnslaion::Parser`, a wrapper class to make Prism compatible with parser gem (`whitequark/parser`).

This PR tries to replace the current parser (`Parser::Ruby32`) with `Prism::Tranlation::Parser`.
Since this is just an experiment, the default parser remains to be that of parser gem. Instead I added `--prism` CLI option to switch the parser.

Here are some advantages with Prism:
- Compile becomes faster
- Parser gem-related issue (#1703) is resolved

Todo:
- [ ] Support using Prism in opalopal
  - Prism can be run in JavaScript through wasm
- [ ] Run tests with Prism

cc: @kddnewton

### DIFF
--- a/lib/opal/cli.rb
+++ b/lib/opal/cli.rb
@@ -75,6 +75,8 @@ module Opal
         @file = @stdin
       end
 
+      Opal::Parser.use_prism if options.delete(:prism)
+
       raise ArgumentError, "unknown options: #{options.inspect}" unless @options.empty?
     end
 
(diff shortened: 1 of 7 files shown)


kddnewton (commented):
Amazing!

hmdne:
There's a small issue for opalopal as of now which regards loading of WASM modules - they can be only loaded async (at least in Chromium). I have a plan to support such workflow, but it will take some time.

Other than that, before we flip the default, we will also have to check regressions in source maps.

Issue #2545, opened 2023-06-05
Feature: Opalfile


hmdne:
**Is your feature request related to a problem? Please describe.**
A problem we have currently is that we lack a good packaging system. The current approach is that we do something like Opal-Browser does - an Opal package is basically a Ruby package, where eg. `lib/opal-browser.rb` adds `opal` directory to Opal's LOAD_PATH and also it adds `paggio` Gem's `lib` directory.

**Describe the solution you'd like**
When compiling a file /a/b/c/x.rb, upper directories are searched for an `Opalfile` file, first `/a/b/c`, then `/a/b`, then `/a`.

An Opalfile file would contain a Ruby code that uses `Opal::Project::Opalfile` DSL.

Said DSL would be used to add Opal's LOAD_PATHS and Gem dependencies (like `Opal::Builder#use_gem`)

If either `Opal::Builder#use_gem` or equivalent directive of `Opalfile` would be used, it would:
- either evaluate said gem's dependencies as we do currently and add its `lib` (or other directory specified in gemspec) to Opal's LOAD_PATH
- or if that gem contains `.opalrc`, it would use it

In the future, `Opalfile` could also be used to contain NPM dependencies. (Consider how to ensure forward compatibility: method_missing that warns? Recommend version checks?)

**Describe alternatives you've considered**
Opal as a platform for rubygems. I have a view, that rubygems in general is a mess and also we still need to include some features for rubygems to be compiled into Opal.

**Additional context**
This provides us also with a way forward to make Opal a full alternative Ruby implementation.

PR #2645, opened 2024-03-11
Opal::Project

hmdne:
This commit creates a concept of a project, which is a concept similar to a Gem, but rather for JavaScript parts.

A single Gem can contain multiple projects. A project is most often defined by adding an Opalfile file that contains an Opalfile DSL. A project is added to compilation by adding an appropriate require path or by compiling a file,

As of now, this DSL contains just two directives:
- add_gem_dependency - declares that files from a given gem should be available for use by Opal's require directive
- add_load_path - adds a load path. This is useful for Opal gems, so that you could specify that eg. "opal" directory contains Opal files and "lib" contains Ruby files.


hmdne:
This fixes #2545

PR #2647, opened 2024-03-13
Split CallNode into a couple files (and a couple of other cosmetic changes)

hmdne:
Those are mostly cosmetic changes, no need to include in changelog.


hmdne:
This is partially taken from #2502 to make the rebased patch smaller.

PR #2649, opened 2024-04-10
`Object#method` supports methods invoked through `method_missing`

takaram:
```ruby
class MyClass
  def method_missing(name, *args)
    if name == :foo
      puts "foo called"
    else
      super
    end
  end

  def respond_to_missing?(name, *args)
    name == :foo
  end
end

MyClass.new.method(:foo).call
```

This code used to raise NameError. Now it works correctly as CRuby.


elia (approved):
🙌

PR #2650, opened 2024-04-16
Implement --directory CLI option

hmdne:
As of now, it only works for the compiler runner and system runner (ie. Node) but nothing prevents us to make it work for other runners in the future as well.

This aims to produce an output that is more easily digestible by modern JS bundlers and that may work better for import maps. This is also a starting point for full ESM support.

This is partially extracted from #2502 (and I intend to split #2502 into some more separate branches, to make it more digestible).


elia (approved):
👍👍👍 📁

PR #2651, opened 2024-04-30
Opal::Watcher and Opal::RackHandler

janbiedermann:
Depends on #2650

  - Implement a Opal::Watcher for watching directories/files in a way that works with opalopal and ruby
  - Update opal -c --watch to use new Opal::Watcher
  - Implement Opal::RackHandler for delivering Opal assets on demand and hot reloading
  - Update Opal::SimpleServer and examples to use Opal::RackHandler
    
Still needs some docs


lib/opal/simple_server.rb
```diff
@@ -20,66 +21,69 @@ class Opal::SimpleServer
...
-    end
-  rescue NotFound => error
-    [404, {}, [error.to_s]]
+  attr_accessor :main, :index_path, :hot_updates, :hot_ruby, :hot_javascript
+
+  attr_reader :app_dir
+
+  def app_dir=(d)
+    @app_dir = d
+    Opal.append_path @app_dir
```
hmdne, line 54: Global state? I know we used to rely on it...

lib/opal/simple_server.rb
```diff
@@ -20,66 +21,69 @@ class Opal::SimpleServer
...
-    Opal.append_path path
+    app_call = proc do |env|
+      call_index
+    rescue NotFound => error
+      [404, {}, [error.to_s]]
+    end
+
+    yield self if block_given?
+    @handler = Opal::RackHandler.new(app_call, { prefix: @prefix, main: @main, builder: builder,
+                                                 hot_updates: hot_updates == false ? : false : true,
```
hmdne, line 43: Well, I guess it's not working yet...

lib/opal/simple_server.rb
janbiedermann: Well, hot updates use the compiler directly, and it has no path otherwise.
Ill switch to using a Builder processor

lib/opal/simple_server.rb
janbiedermann: Works for me

hmdne:
Just noting - this PR got closed by mistake. GitHub changed behavior. If you can change the target branch to `master` and reopen it... but possibly you can't (because I can't too), so we will need to open a new PR I guess.

PR #2653, opened 2024-05-03
Builder namespace cleanup

janbiedermann:


hmdne:
I wanted to do it myself actually - and probably a similar thing with Compiler (nodes etc.). I would consider adding compatibility constants, but since we aim for a compatibility-breaking 2.0 release, we can do that.

PR #2655, opened 2024-05-04
[WIP] Opal::RackHandler and Opal::Watcher

janbiedermann:
- Implement Opal::Watcher for watching directories/files in a way that works with opalopal and ruby
    - Update opal -c --watch to use new Opal::Watcher
    - Implement Opal::RackHandler for delivering Opal assets on demand and hot reloading
    - Update Opal::SimpleServer and examples to use Opal::RackHandler
    
    Feature wise its finished, but after discussion with @hmdne, using the Builder, which is already used to build the assets in the first place, also for hot updates seems preferable to using the Compiler. This would resolve the necessity to append build paths globally. Also this would require a few changes to Builder, possibly making Watcher a property of Builder, like { watch: true } (for example to avoid additional filesystem scans, as builer reads all required files anyway). Maybe adding a few methods to builder, like #watch, #diff, #poll_updates.


janbiedermann:
Maybe like this:
Builder#watch - watch and rebuild dir/file
Builder#updates? - check for updates -> true/false
Builder#diff - Array of compiled files since last call to diff or build -> [{path: 'file.rb', js: 'function(){ ...' }, ...]

janbiedermann:
Now depends on #2656 which makes Opal::Watcher obsolete.
Still needs testing and fixing.

PR #2656, opened 2024-05-05
Add watch and update capability for hot reloading to Builder

janbiedermann:
Adds 2 methods to Builder:

Builder#watch(&block)
  loops and watches for changes, calls block with block.call(builder, changes)
  where changes is a Hash like: { added: [], modified: [], removed: [] }
  if a error occured, changes has another key :error which contains the exception

Builder#update
  checks for changes and updates itself, returns the changes like watch above

Builde delegates the actual update to the processed asset, which is a Processor instance, via Processor#update

Although CI is green, i have not tested it yet. I will update #2655 to use this PR and test using that.


lib/opal/builder.rb
```diff
@@ -92,6 +92,7 @@ def self.build(*args, &block)
 
     def build(path, options = {})
       build_str(source_for(path), path, options)
+      self
```
hmdne, line 95: This breaks the API :(

lib/opal/builder.rb
```diff
@@ -109,7 +110,7 @@ def build_str(source, rel_path, options = {})
       # Don't automatically load modules required by the module
       process_requires(rel_path, requires, asset.autoloads, options.merge(load: false))
       processed << asset
-      self
+      asset
```
hmdne, line 113: Why should this method return an asset and other not?

lib/opal/builder.rb
```diff
@@ -223,8 +220,77 @@ def compiled_source(with_source_map: true)
       compiled_source
     end
 
+    # Builds and watches all paths for changes and then rebuilds, does not return
+    def watch
```
hmdne, line 224: Can this, and the rest of the methods related to watching, be moved to `Builder::Watcher` which would be included here?

janbiedermann:
Also added specs, which fail, because i came to a point, where more changes are required to the assets. For example in case of require_tree and a added file, its not enough to just add the new asset to the set, it must also be added to the outer asset as required asset. Thus delegating the detection of a new file to the outer asset may be appropriate. Work in progress.

PR #2657, opened 2024-05-06
Builder namespace cleanup 2

hmdne:
This commit improves upon #2653 by adding a couple of other Builder-adjacent classes (ie. ones that are realistically only used with Builder) - and moves them to a Builder namespace as well.

In addition, there's a small deprecation system added, so that the old names will still work, but show a warning.


janbiedermann (commented):
Very nice!

PR #2664, opened 2024-06-16
Import files required by require_tree in Builder directory mode

janbiedermann:
Opal::Builder directory mode did not import files required via require_tree. This PR fixes that, however it probably moves SourceMap info a couple of lines too, which currently is not my concern, ill open a separate bug report for that.

The correct import is tested by specs of #2656


janbiedermann:
Contrary to above comment I added a spec here and line offset compensation for source maps.

Issue #2670, opened 2024-08-01
Bug: Hash from a native object that contains an Hash returns unexpected values


lionelchauvin:
**Describe the bug**

In ruby the first argument of Hash.new is the default value returned by the hash when a key doesn't exist.
With opal the first argument can be used to initialize the hash from a native javascript object by using backticks:
```
Hash.new(`{a: "A"}`)
```

Opal allows to insert an opal hash in javascript by using #{}:

```
`var h = #{ {a: "A"} }`
```

These two features allow to build an Hash from a native javascipt object with a property that contains an opal hash:
```
h = {a: "A"}
g = Hash.new(`{h: #{h}}`)
```

I am expecting the nested hash to return nil when a key doesn't exist:
```
expect(g['h']['b']).to eq(nil) # because h['b'] == nil
```

but with opal 1.8.2 it returns the nested hash `{"a" => "A"}`.

There is a bug somewhere in
https://github.com/opal/opal/blob/9d48ab627dc253ba26b1507caa768f772c361cbf/stdlib/native.rb#L542

Opal version: 1.8.2

**To Reproduce**

```
# backtick_javascript: true
h = {a: "A"}
g = Hash.new(`{h: #{h}}`)

p g['h']['a'] # "A"
p g['h']['b'] # {"a" => "A"} instead of nil
```

nothing special in generated code:

```
Opal.queue(function(Opal) {/* Generated by Opal 1.8.2 */
  var self = Opal.top, $nesting = [], $$ = Opal.$r($nesting), nil = Opal.nil, h = nil, g = nil;

  Opal.add_stubs('new,p,[]');
  
  h = (new Map([["a", "A"]]));
  g = $$('Hash').$new({h: h});
  self.$p(g['$[]']("h")['$[]']("a"));
  return self.$p(g['$[]']("h")['$[]']("b"));
});

```

lionelchauvin:
I am not sure because it can be called recursively by initialize, but this change seems to fix the problem:

```
diff --git a/stdlib/native.rb b/stdlib/native.rb
index e3f4675c2..e974b8cd0 100644
--- a/stdlib/native.rb
+++ b/stdlib/native.rb
@@ -544,7 +544,11 @@ unless Hash.method_defined? :_initialize
           (value.constructor === undefined ||
             value.constructor === Object ||
             value instanceof Map)) {
-         $hash_put(hash, key, #{Hash.new(`value`)});
+          if (value.$$is_hash) {
[...]

PR #2671, opened 2024-08-01
Native/Hash - Use correct var

janbiedermann:
as identified in #2670
(but not a fix for #2670)


janbiedermann:
Obsolete, fixed by #2672

PR #2672, opened 2024-08-01
Native/Hash - Fix Hash from native object contains unexpected values

janbiedermann:
Fix for #2670


janbiedermann:
now using instanceof Array too

janbiedermann:
Now also recursive

elia (approved):

PR #2673, opened 2024-08-01
Remove $$is_hash, use instanceof Map instead

janbiedermann:
A tiny bit cleaner


janbiedermann:
some other gems may depend on $$is_hash though

PR #2674, opened 2024-08-01
Remove $$is_array, use instanceof Array instead

janbiedermann:


janbiedermann:
$$is_array may be used by other gems

hmdne (approved):

PR #2675, opened 2024-08-01
Remove `String#untaint` from exe/opal-build

delonnewman:
`String#untaint` was deprecated in Ruby 2.x and removed in 3.2


hmdne (approved):

hmdne:
Thanks!

PR #2678, opened 2024-08-09
$hash_each: use iterator directly

janbiedermann:


janbiedermann:
No, wont work, i forgot

PR #2679, opened 2024-08-09
Hash optimize internal each usage and improve js interoperability

janbiedermann:
For case, where a new Hash is created from self without modifying self there is no need to use the $hash_each helper, self.forEach is sufficient. This avoids creating a extra Array, which is not needed, and thus should benefit memory usage and performance.

Also for #compact and #compact!:
If #compact(!) is called on a Map passed from JS with `null` values, i would expect #compact(!) to remove those too.
So this PR adds a `null` check.

### DIFF
--- a/opal/corelib/hash.rb
+++ b/opal/corelib/hash.rb
@@ -222,12 +222,12 @@ class ::Hash < `Map`
     %x{
       var hash = new Map();
 
-      return $hash_each(self, hash, function(key, value) {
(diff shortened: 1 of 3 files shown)


janbiedermann:
This still has a issue with `undefined`, working on it

janbiedermann:
Fixed, added spec, everything ok

elia (approved):

PR #2683, opened 2024-08-24
Remove unused Builder attributes

janbiedermann:


elia (approved):
Those are useful options for some edge cases, but I guess they're not that popular, and not referenced anywhere, let's simplify! ✂️

PR #2684, opened 2024-08-31
Bridging `BigInt` to `Integer`

elia:


hmdne:
You can also refer to #2219 which I have recently rebased

elia:
> You can also refer to #2219 which I have recently rebased

Thanks for the link to prior art 🙌 , I wanted to attempt an all or nothing change and see what actually needs to change 🤔

hmdne:
One idea I had, was to convert things like, let's say:

```ruby
%x{
  for (i = 0; i < 100; i++) {
...
```

to:

```ruby
%x{
  for (i = 0; i < #{100}; i++) {
...
```

so that perhaps we could think of some way to make this an optional change. But that all depends on if the performance impact is noticeable.

PR #2688, opened 2024-09-15
Implement Threaded Scheduler for JRuby and TruffleRuby

janbiedermann:
One thing to note: default thread stack size for TruffleRuby leads to stack overflow exception when compiling parser, must be raised to around 2MB 
This PR depends on #2687 
lets see what CI reports ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,8 +48,10 @@ jobs:
             ruby: '3.0'
           - name: jruby
             ruby: 'jruby'
-          - name: truffleruby
-            ruby: 'truffleruby'
+          # truffleruby: disabled, does not finish on github actions
+          # probably not enough memory
+          # - name: truffleruby
+          #   ruby: 'truffleruby'
           - name: smoke-test
             command: bin/rake smoke_test
           - name: windows-mspec-nodejs
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,6 @@ inherit_from:
 
 require: rubocop-performance
 
-
 AllCops:
   TargetRubyVersion: 3.0
   NewCops: disable
@@ -214,7 +213,7 @@ Naming/BinaryOperatorParameterName:
 
 Naming/PredicateName:
   # Ruby has "has_key?" method
-  ForbiddenPrefixes: 
+  ForbiddenPrefixes:
     - is_
     - have_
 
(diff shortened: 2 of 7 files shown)


janbiedermann:
all good - take them all #2686 #2687 #2688

janbiedermann:
Some notes:

Performance:
JRuby and TruffleRuby performance for Opal is very bad, 5 to 10 times slower than Matz Ruby, probably because the parser methods, and possibly others too, are too large to be optimized by their vms. We have a similar issue with V8. However for J-/TruffleRuby the situation is worse, as they have the problem twice, first with the Ruby -> bytecode compiler and second within the vm, so they mostly interpret code, which for Opal is critical, it seems.

Stability:
JRuby may fail from time to time for no apparent reason.
TruffleRuby is much more consistent but has a few compatibility issues concerning core classes/methods.
When porting code to use real threads, TruffleRuby is the better choice, because of its consistent behavior.

hmdne (approved):
Great! Let's merge it in the next merge window

hmdne:
@janbiedermann We have merged your JRuby and TruffleRuby patches today, aiming to also merge this one, but unfortunately it wasn't passing the tests, so we just did #2690 to finalize this series (there was some lint issue due to requiring `etc`, because you broke the macro and tried to overcome this with a special case to `etc` which we have also removed).

@elia has rebased this PR to the latest `master`. The CI on JRuby breaks, because apparently sometimes you try to join a thread from the same thread. We weren't able to find the culprit of the failure, so no merge happened today. Keep in mind, that `process_requires` may happen at any point of time, inside a main thread, or in any other thread. Can you take a look and correct the issue?

janbiedermann:
@hmdne , @elia , no, i am trying to join new threads only. In case threads have been shut down, for a a later call to process_requires, it will create new threads. For the moment I blame JRuby, as this code works in TruffleRuby fine. Ill investigate a bit more.

headius:
I am trying to investigate the JRuby issue, but it is very hard to reproduce. I got it to go twice yesterday, but a multi-hour loop of the specs has failed to reproduce it again today.

janbiedermann:
@headius it seems to work now, i blame @janbiedermann for not seeing, that with a higher amount of cores there is a likely interference of @threads #<< and @threads #empty? and i blame JRuby too for messing with Thread ids or something.

@hmdne @elia it seems to work now, i squeeze the commits and will run JRuby action a couple of times more during the next days. We do not run the TruffleRuby action for the moment, cause since 24.1 it has memory problems it seems, but running the JRuby action should work now.

headius:
@janbiedermann Thanks for continuing to dig into this... I had a test loop running all weekend and was unable to reproduce the issue anymore. I'm not exactly sure what sequence of events can cause the calling thread to be in the `@threads` list, but we'll cross our fingers that synchronizing that list has resolve the issue.

Let me know if you need anything else to support this PR.

janbiedermann:
@headius alright, thanks for your efforts too

PR #2697, opened 2024-10-08
Let opal build standalone applications via the --compile-to-exe option

janbiedermann:
Ill expand on the example and docs after my platform work.
Also apps of type osa currently crash most probably, because osascript requires some platform work. Well, depends on what you do. The simple example should work.

### DIFF
--- a/examples/exe/create_apps.sh
+++ b/examples/exe/create_apps.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+# Comment those below, which are not installed or available for your platform,
+# then execute this script to create some very simple apps.
+
+# with bun installed:
+bundle exec opal --compile-to-exe bun -o simple_bun ./simple.rb
+
+# with deno installed:
+bundle exec opal --compile-to-exe deno -o simple_deno ./simple.rb
+
+# with node installed:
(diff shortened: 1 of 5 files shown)


janbiedermann:
Adapted to --build-app and some small fixes for macOS an Windows

janbiedermann:
Oh, exe, i was to quick with app. What about --compile-to-exe? Let me quickly change that.

janbiedermann:
We have:
 -c, --compile                    Compile to JavaScript
 Now we have:
  --compile-to-exe TYPE           Compiles and builds a standalone application with runtime of TYPE. The following TYPEs are supported: bun, deno, node, quickjs
  
I think thats most consistent

lib/opal/app_creator.rb
```diff
@@ -0,0 +1,117 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'tmpdir'
+require 'opal/os'
+
+module Opal
+  class AppCreator
```
hmdne, line 8: Here it's still called AppCreator and not something like ExeBuilder

janbiedermann: Oh, sure, i forgot to change the class name. Next time then.

PR #2685, opened 2024-09-09
Refreshing some npms (introduces ESLint 9)

janbiedermann:

PR #2698, opened 2024-10-10
Unify CDP

### LINKED ISSUE #2498: Code duplication between chrome and firefox CDP runners
This should be worked out by for example merging some identical codepaths

Ref #2491

janbiedermann:
Depends on #2694
Fixes #2498


janbiedermann:
Ok, also works with old Firefox, so the Firefox communication problems got fixed in chrome-remote-interface, #2685. Very nice.

elia (approved):

PR #2699, opened 2024-10-10
Enable async Opal startup and REPL, use correct REPL runner

janbiedermann:
Taken from my platform work.
Use case: Conditional dynamic import() of platform specific modules directly after runtime.js and before corelib ruby parts.
Lets see what specs say ...

### DIFF
--- a/lib/opal/cli_options.rb
+++ b/lib/opal/cli_options.rb
@@ -196,6 +196,10 @@ module Opal
         options[:irb] = true
       end
 
+      on('--await', 'Enable async/await support') do
+        options[:await] = true
+      end
+
       on('-M', '--no-method-missing', 'Disable method missing') do
         options[:method_missing] = false
       end
--- a/lib/opal/nodes/top.rb
+++ b/lib/opal/nodes/top.rb
@@ -88,7 +88,7 @@ module Opal
 
         if compiler.requirable?
           unshift "#{async_prefix}function(Opal) {"
-        elsif compiler.eval?
+        elsif compiler.eval? || compiler.irb?
           unshift "(#{async_prefix}function(Opal, self) {"
         else
           unshift "Opal.queue(#{async_prefix}function(Opal) {"
@@ -104,9 +104,9 @@ module Opal
             # require absolute paths from CLI. For other cases
             # we can expect the module names to be normalized
             # already.
-            line "Opal.load_normalized(#{module_name.inspect});"
+            line "Opal.queue(()=>{ Opal.load_normalized(#{module_name.inspect}) });"
           end
-        elsif compiler.eval?
+        elsif compiler.eval? || compiler.irb?
           line "})(Opal, self);"
         else
           line "});\n"
--- a/lib/opal/repl.rb
+++ b/lib/opal/repl.rb
@@ -12,9 +12,10 @@ module Opal
 
     attr_accessor :colorize
 
-    def initialize
+    def initialize(runner_type = nil)
       @argv = []
       @colorize = true
+      @runner_type = runner_type || :nodejs
 
       begin
         require 'readline'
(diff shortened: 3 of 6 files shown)


janbiedermann:
All good, nice.

janbiedermann:
No, not all good, it causes issues when using `eval()`

janbiedermann:
Actually not, the issues i hit are caused by a pending promise, a error somewhere else, so still all good.

janbiedermann:
No, not all good, because the unresolved promise is caused by this change.

janbiedermann:
For the moment, not queuing code for IRB solves the problem. But this needs further work to enable async code for REPL/IRB.

janbiedermann:
Alright, all good

lib/opal/repl.rb
```diff
@@ -44,6 +45,7 @@ def run(argv = [])
 
     def load_opal
       runner = @argv.reject { |i| i == '--repl' }
+      runner += ['-R', @runner_type.to_s, '--await']
```
hmdne, line 48: Makes repl work again with other runners? :O

lib/opal/nodes/top.rb
```diff
@@ -104,9 +104,9 @@ def closing
             # require absolute paths from CLI. For other cases
             # we can expect the module names to be normalized
             # already.
-            line "Opal.load_normalized(#{module_name.inspect});"
+            line "Opal.queue(()=>{ Opal.load_normalized(#{module_name.inspect}) });"
```
hmdne, line 107: I don't really remember this code, will take a closer look at a proper review. But - async top requires need to be `require("something").__await__`.

Ie. won't this break such code?

```ruby
# file1.rb
$a = 1
require "file2"
$a == 2 or raise "Wrong result! Expected 2, got 1!"

# file2.rb
$a = 2
```

If it won't, what about:

```ruby
# file1.rb
$a = 1
require("file2").__await__
$a == 2 or raise "Wrong result! Expected 2, got 1!"

# file2.rb
# await: true
$a = 2.__await__
```

lib/opal/cli_options.rb
```diff
@@ -196,6 +196,10 @@ def initialize
         options[:irb] = true
       end
 
+      on('--await', 'Enable async/await support (experimental)') do
```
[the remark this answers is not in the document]
hmdne: Anyway, this should be on by default in some future release.

lib/opal/repl.rb
janbiedermann: Sure, as long as they can do IO

lib/opal/cli_options.rb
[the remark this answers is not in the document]
janbiedermann: Alright, i will adjust

lib/opal/nodes/top.rb
janbiedermann: I will test with this case

janbiedermann: I created file1.rb as above and added a `puts $a` after the `require "file2"` line
I created file2.rb as above
I compiled with `bundle exec opal -c -I. -o file1.js file1.rb`
I executed with `node file1.js`
It printed: `2`

So, seems to work

janbiedermann: First case, without __await__

hmdne: Looks like the second case also works

hmdne:
My test case:

```
>> puts 123; p [1,2,3,4,5].each_await { |i| p i }.__await__; puts 456
123
1
2
=> #<PromiseV2:0x31e>
>> 
3
=> nil
>> 
4
=> nil
>> 
5
=> nil
>> 
=> nil
>> 
[1, 2, 3, 4, 5]
456
=> nil
```

Somehow it returned too early...

janbiedermann:
@hmdne sure, i removed that await again, because if you do:
`>> a = Promise.new`
it will wait forever

PR #2703, opened 2024-10-18
Implement IO::Buffer

janbiedermann:
including tests taken from Ruby and slightly adapted. A few features cannot be supported, specifically mmap, modifying strings, number precision issues for numbers > 2^53. Some tests are marked unsupported due to Tempfile issues. Otherwise pretty complete.
Its not used by anything yet, just available.
I expect to make IO depend on it.

### DIFF
--- a/opal/corelib/io/buffer.rb
+++ b/opal/corelib/io/buffer.rb
@@ -1,0 +1,644 @@
+# backtick_javascript: true
+
+require 'corelib/string/encoding'
+
+class ::IO
+  class Buffer
+    # Types that can be requested from the buffer:
+    #
+    # :U8: unsigned integer, 1 byte
+    # :S8: signed integer, 1 byte
+    # :u16: unsigned integer, 2 bytes, little-endian
+    # :U16: unsigned integer, 2 bytes, big-endian
+    # :s16: signed integer, 2 bytes, little-endian
+    # :S16: signed integer, 2 bytes, big-endian
+    # :u32: unsigned integer, 4 bytes, little-endian
+    # :U32: unsigned integer, 4 bytes, big-endian
+    # :s32: signed integer, 4 bytes, little-endian
+    # :S32: signed integer, 4 bytes, big-endian
+    # :u64: unsigned integer, 8 bytes, little-endian
+    # :U64: unsigned integer, 8 bytes, big-endian
+    # :s64: signed integer, 8 bytes, little-endian
+    # :S64: signed integer, 8 bytes, big-endian
+    # :f32: float, 4 bytes, little-endian
+    # :F32: float, 4 bytes, big-endian
+    # :f64: double, 8 bytes, little-endian
+    # :F64: double, 8 bytes, big-endian
+
+    include Comparable
+
+    class AccessError < ::StandardError; end
+    class LockedError < ::StandardError; end
+
+    DEFAULT_SIZE = 4096
+
+    EXTERNAL = 1
+    INTERNAL = 2
+    MAPPED = 4
+    SHARED = 8
+    LOCKED = 32
+    PRIVATE = 64
+    READONLY = 128
+
+    %x{
+      let size_map = new Map()
+      .set('U8', 1).set('S8', 1)
+      .set('u16', 2).set('U16', 2).set('s16', 2).set('S16', 2)
+      .set('u32', 4).set('U32', 4).set('s32', 4).set('S32', 4)
+      .set('u64', 8).set('U64', 8).set('s64', 8).set('S64', 8)
+      .set('f16', 2).set('F16', 2)
+      .set('f32', 4).set('F32', 4)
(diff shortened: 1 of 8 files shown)


janbiedermann:
Bun 1.1.30 works, Bun 1.1.31 fails ... :-(

janbiedermann:
Actually i was wrong. The problem is much bigger. Bun 1.1.31 has severe problems when accessing `process.stdout` and friends, causing it eventually to crash. It tries to access Opal code from within its internal tty code, where it tries to copy props for some reason. No idea why that would be a good idea or what for, it fails. So not matter what, Opal won't work with Bun 1.1.31 unless all references to `process.stdout` and friends are removed.

janbiedermann:
All good.

janbiedermann:
Just for reference i include the Bun stack trace:
```
Error
    at $deny_frozen_access (/Users/jan/workspace/opal/tmp/mspec_bun.js:2578:25)
    at $rehash (/Users/jan/workspace/opal/tmp/mspec_bun.js:17999:7)
    at <anonymous> (internal:primordials:35:62)
    at forEach (native:1:11) 
    at <anonymous> (internal:primordials:31:15)
    at <anonymous> (internal:primordials:1:11)
    at anonymous (native)
    at <anonymous> (internal:validators:1:11)
    at anonymous (native)
    at <anonymous> (node:tty:1:11)
    at anonymous (native)
    at <anonymous> (native:1:11)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:22747:22) typeof(proccess.stdout)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2813:17)
    at $require (/Users/jan/workspace/opal/tmp/mspec_bun.js:6002:19)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:22783:18)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2813:17)
    at $require (/Users/jan/workspace/opal/tmp/mspec_bun.js:6002:19)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:28193:11)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:28214:6)
    at <anonymous> (native:11:43)
```

There are multiple question marks here. The problem is triggered by loading `corelib/io`, which towards its end accesses `typeof(proccess.stdout)` which causes the problem.

1. Why tries accessing `process.stdout` try to copy props of Opal Ruby Objects?
https://github.com/oven-sh/bun/blob/f3b658d9f7b0d3f45dcfcdf845be44ea80e05055/src/js/internal/primordials.js#L37
2. Why does it call Hash#rehash at this stage, Opal still booting up, loading `corelib/io`, when nowhere in corelib Hash#rehash is called?
3. Is this stack trace maybe bogus, even though it is consistent, due to a previous severe error (stack overflow)?

opal/opal/mini.rb
```diff
@@ -14,5 +14,6 @@
 ::Object.require 'corelib/method'
 ::Object.require 'corelib/regexp'
 ::Object.require 'corelib/variables'
-::Object.require 'corelib/io'
 ::Object.require 'opal/regexp_anchors'
+::Object.require 'corelib/io/buffer'
```
hmdne, line 18: Should we really need this in opal/mini? Perhaps a better place for it would be opal/full, since not many people use this interface.

janbiedermann: I put it in mini because io is in mini and i expect io to depend on io/buffer.
But i am still working on io, platform support. I thought using io/buffer in io is a good idea, but i am not 100% sure about this anymore. Depending on further progress with my platform work. Cannot decide yet. Feel free to change, if you prefer, or otherwise wait a bit.

opal/corelib/io/buffer.rb
```diff
@@ -0,0 +1,644 @@
...
+      end
+      self
+    end
+
+    def each_byte(offset = 0, count = nil)
+      # Iterates over the buffer, yielding each byte starting from offset.
+      raise(AccessError, 'Buffer has been freed!') if null?
+      return enum_for(:each_byte, offset, count) unless block_given?
+      count ||= size
+      while count > 0
```
hmdne, line 361: Off by one?

opal/opal/mini.rb
hmdne: Same goes with let's say pattern matching, but pattern matching is rarely used for now, so it takes up valuable asset size. I'll adjust it to `opal/full` and merge.

PR #2707, opened 2024-10-26
Limit parallelism of CI matrix to have a better reliability and more accurate results

janbiedermann:
Lets see how it goes ...


janbiedermann:
A matrix parallelism of 8 seems like a good compromise in terms of speed and reliability.

janbiedermann:
Ha, and just as i wrote that, there is one failure

janbiedermann:
Ok, lets use a parallelism of 6 then

PR #2705, opened 2024-10-18
Actions: Specify runner engine versions, least supported and latest, run on Edge

janbiedermann:
JavaScript engines move fast, deprecating, removing or adding features. To ensure Opal works, we need to run tests on older engine versions and latest ones. This PR allows to specify exact versions for the engines and enables actions for:

Older engines:
- nodejs 18 (lts/hydrogen), supported until 2025
- Firefox 115 esr, supported until 2025
- Chrome 115, not supported

Newer engines:
- nodejs latest, currently 23.0
- Firefox latest, currently 131
- Chrome stable, currently 131
- Edge stable, currently 129
- Safari, depending on macos-latest provided by github actions
- OSAScript, depending on macos-latest provided by github actions (actually this engine is only partially newer)
- bun, currently fixed to 1.1.30, due to problems with latest 1.1.31
- deno latest stable, currently 2.0.2

PR #2708, opened 2024-10-27
Require ::Math for mspecs, its expected to be available by Ruby constants_spec.rb

janbiedermann:


janbiedermann:
already included in #2707 and #2705

elia (approved):

Issue #2709, opened 2024-11-12
Bug: String methods returning wrong results for Strings containing UTF-16 surrogates, further String issues to JavaScript UCS2/UTF-16 issues


janbiedermann:
JavaScript/Browsers String implementation is "funky" in so far, as it can handle to display UTF-16 correctly, also do transformations correctly in UTF-16, mostly, yet some methods treat strings as UCS2. As Opal relies on String with Opal String being a bridged class, these issues leak to Ruby space. For example:

Opal:
```
>> 'a𝌆'[2]
=> "�"
>> 'a𝌆'[1]
=> "�"
>> 'a𝌆'.size
=> 3
```
Ruby:
```
:002 > 'a𝌆'[2]
 => nil 
 :003 > 'a𝌆'[1]
 => "𝌆"
 :004 > 'a𝌆'.size
 => 2
```

For reference: https://mathiasbynens.be/notes/javascript-encoding)

This should be considered for #2231

janbiedermann:
Fixed by #2710

PR #2710, opened 2024-11-13
Let String correctly handle Unicode astral plane surrogates

janbiedermann:
fix for #2709, at least for methods and method parts that directly work with strings.
However problem applies to Regexp too, thats why this is WIP so far.

Lets first see what specs say and if/how performance is affected.


takaram (commented):
Since spread syntax (`[...str]`) is expensive when `str` is large, we should try to avoid it if possible.
I wrote some idea of optimization in the comment, but I believe we can optimize more.

janbiedermann:
Oh my, performance is indeed very bad. Thanks for your suggestions @takaram , i will try to make it fast ...

opal/corelib/string.rb
```diff
@@ -369,7 +385,7 @@ def chop
   end
 
   def chr
-    `self.charAt(0)`
+    `[...self][0]`
```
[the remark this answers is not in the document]
janbiedermann: i use that, thx

janbiedermann:
So far I reduced code and applied a few optimizations, but sadly with little effect on asciidoctor performance.

hmdne:
IIRC, in Ruby `String#[]` increased complexity from O(1) to O(n) when introducing Unicode support (1.9):

```ruby
[3] pry(main)> r = "ü"+("a"*1000000); s = "ü"+("a"*100000); t = "üa"; def bm(a); t0 = Time.now; a[a.length-1]; Time.now-t0; end; [bm(r), bm(s), bm(t)] 
=> [0.00028107, 2.4601e-05, 3.5e-07]
[4] pry(main)> 
```

janbiedermann:
There are a few ways to use the String iterator:
- creating a Array and using Array methods`[...string].findIndex()` - which is fast, uses lots of memory and gets slower with larger strings
- creating a Array and using a number based for loop - is fast too, uses lots of memory, but seems a bit faster for larger strings
- using a `for (char of string)` loop, which uses the iterator, saves memory as the Array creation can be avoided, yet seems slower then the Array versions
- using the String Iterator directly with `iterator.next()`, sames memory too, yet also seems slower than all the others

In additions for the latter two, as often a index is needed, the loops require housekeeping of a index variable.

So its possible to avoid the creation of a Array with the spread syntax, but except for saving memory there is no benefit.
Also results seem to depend on engines very much. Saving memory is a good idea, benefits performance too, i will try `for (char of string)` where applicable.

janbiedermann:
Also it could help here, if Ruby Strings would be JavaScript String objects, as we could cache a few results.

hmdne:
@janbiedermann You can also take a look at this patch, which touches the UTF-16 but in different places:

#2610

janbiedermann:
@hmdne excellent, thx

janbiedermann:
Further experimentation shows, that avoiding the String iterator and using number based `for` loops is fastest.

opal/corelib/string.rb
```diff
@@ -175,7 +191,7 @@ def =~(other)
 
   def [](index, length = undefined)
     %x{
-      var size = self.length, exclude, range;
+      let str = [...self], size = str.length, exclude, range;
```
hmdne, line 194: My idea for some optimization:

A UTF-16 character of index N won't happen later than in N*2th 16bit word, so we could just expand a slice. And calls like `a[0]`, `a[1]` will be the most common, then `a[a.length-1]` will be the second most common, for which we could think of doing a similar but in reverse.

hmdne: Hm, for the reverse, length will be a harder to procure though...

janbiedermann: We should not need length with a negative index, we can just walk from the back.

janbiedermann:
Depending on use case, i retract my previous statement and suggest the opposite.

janbiedermann:
I ran some benchmarks for `String#[]`
orig: the O(1) original implementation with astral bugs
spread: the `[...self]` implementation as in this PR currently, with correct astral results
forof: using `for(c of self)` with correct astral results
fori: using `for(let i=0; ... ) { self.codePointAt(i)` with correct astral results

Benchmark results:
```
for a 2 byte string, index 0
  orig short      2.414M (± 6.4%) i/s -     12.014M in   5.015036s
 forof short      2.314M (± 1.1%) i/s -     11.673M in   5.044107s
  fori short      2.338M (± 3.6%) i/s -     11.760M in   5.040488s
spread short      1.912M (± 3.5%) i/s -      9.618M in   5.039323s
 
50 byte string, index size - 2
  orig medi+      2.444M (± 3.7%) i/s -     12.247M in   5.022135s
 forof medi+      1.175M (± 5.6%) i/s -      5.896M in   5.036771s
spread medi+    620.410k (± 8.2%) i/s -      3.080M in   5.023374s
  fori medi+    167.376k (± 0.6%) i/s -    837.816k in   5.005780s
 
64k byte string, index size - 2
  orig long+      2.458M (± 0.8%) i/s -     12.389M in   5.039579s
 forof long+      1.909k (± 0.8%) i/s -      9.690k in   5.077547s
spread long+    430.329  (± 3.7%) i/s -      2.184k in   5.082824s
  fori long+    135.007  (± 3.0%) i/s -    676.000  in   5.012608s
```

forof looks best so far.
But for reverse look up with negative index, without length, using fori would still be very slow.

hmdne:
I have ran our asciidoctor benchmark by adding method-by-method. So, ie. I start with just with changing [], then by adding chr (so both [] and chr), etc.

The results are as follows, I stopped after adding #length, but then continued with other methods that I think may cause a performance issue:

```
+ String#[]              
              Run time on node: 0.755 (+16.04%) -> 0.784 (+9.73%) (change: +3.91%)
+ String#chr
              Run time on node: 0.747 (+13.85%) -> 0.784 (+17.32%) (change: +4.95%)
+ String#delete_{pre,suf}fix
              Run time on node: 0.748 (+12.19%) -> 0.785 (+10.89%) (change: +4.85%)
+ String#end_with?
              Run time on node: 0.746 (+13.04%) -> 0.835 (+15.38%) (change: +11.88%)
+ String#include?
              Run time on node: 0.748 (+15.09%) -> 1.321 (+12.29%) (change: +76.73%)
+ String#index
              Run time on node: 0.750 (+10.85%) -> 1.321 (+13.40%) (change: +76.01%)
+ String#length
              Run time on node: 0.751 (+5.95%) -> 1.310 (+9.57%) (change: +74.49%)
+ String#start_with?
              Run time on node: 0.753 (+9.21%) -> 1.408 (+6.88%) (change: +86.98%)
+ String#tr
              Run time on node: 0.743 (+10.51%) -> 1.428 (+8.47%) (change: +92.37%)
+ full patch
              Run time on node: 0.749 (+13.61%) -> 1.424 (+12.10%) (change: +90.05%)
```

So, from what I see here, it's `end_with?`, `start_with?` and `include?` that degraded performance a lot. And I strongly think that those three methods don't need UTF-16 awareness at all (except for edge cases which I think we can skip). Others may not be used by AsciiDoctor a lot, but also I don't think they need UTF-16 awareness, in particular: `delete_{pre,suf}fix`.

I suggest we start with methods that obviously need UTF-16 awareness. Then we can optimize for the most obvious cases.

janbiedermann:
This iteration avoids [...self] for #[] , #include? and #size and uses custom `for of` loops instead.
It also includes a new benchmark in benchmark-ips, just for playing around, i may remove that in the final version.
Lets see ...

janbiedermann:
ah, looking much better, i will continue on the `for of` path

janbiedermann:
This iteration further improves #[] and this time also #ljust, #rjust, #center.
#center is now ~ 2 times faster than before:
```
      current center:   729390.4 i/s
         orig center:   378744.2 i/s - 1.93x  slower
```
included benchmark-ips updated

janbiedermann:
Actually i messed up the benchmark for #center, truth is, #center is ~ 4 times faster now

opal/corelib/string.rb
```diff
@@ -427,9 +443,12 @@ def delete_prefix(prefix)
       if (!prefix.$$is_string) {
         prefix = $coerce_to(prefix, #{::String}, 'to_str');
       }
+      let str = [...self],
+          prefix_s = [...prefix],
+          prefix_l = prefix_s.length;
 
-      if (self.slice(0, prefix.length) === prefix) {
-        return self.slice(prefix.length);
+      if (str.slice(0, prefix_l).join('') === prefix) {
```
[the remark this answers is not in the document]
janbiedermann: I also take this idea in principle, with a correction, it seem to be required to test specifically for the leading surrogate range 0xD800 ... for the last character of the prefix

janbiedermann:
... much better, only ~ +17%

janbiedermann:
This iteration improves #index

janbiedermann:
This iteration improves #next, #partition and #reverse, with #reverse being multiple times faster than before

janbiedermann:
This iteration improves #rindex, #rpartition, #tr, #tr_s and #upto. #tr and #tr_s are now multiple times faster than the original implementation. (thats also what i meant for #reverse and #center)

janbiedermann:
Current affairs on my machine with node 20.18.0:
```
                  Compile time: 4.359 (+30.55%) -> 4.803 (+23.21%) (change: +10.18%)
              Run time on node: 1.441 (+2.83%) -> 1.603 (+18.47%) (change: +11.23%)
               Run time on bun: 1.622 (+4.03%) -> 2.006 (+11.05%) (change: +23.63%)
                   Bundle size: 4663.95 kB -> 4673.62 kB (change: +0.21%)
          Minified bundle size: 1009.62 kB -> 1011.43 kB (change: +0.18%)
            Mangled & minified: 706.68 kB -> 707.71 kB (change: +0.15%)
```
YMMV

ill try to unicodify the Regexp parts of String now

janbiedermann:
This iteration adds the u flag to Regexp, almost always. Adding the u flag changes behavior of JS RegExp a little, so still to do now: literal brackets in RegExpes must be escaped now.
For Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Quantifier

janbiedermann:
Just a rebase

janbiedermann:
I had to update Firefox for specs to latest-esr, well, for Grapheme support at least Firefox 125 is required.
Also Deno is making problems  ... "FrozenError" ... maybe too cold outside?

janbiedermann:
Further improvements.
mspec_ruby master runs 13964 examples
this PR 14200 examples

Deno has problems, expected.

janbiedermann:
14231 examples passing

Ready to merge

PR #2713, opened 2024-12-15
[WIP] Implement mutable ::String

janbiedermann:
This demonstrates the concept. All is documented at the end of corelib/string.rb.

It implements at the moment String#<< and String#concat .

This has further implications though, to make it work for real. For example:
JS Strings are always considered frozen, even when boxed, so they are immutable.
::String.new strings are mutable  (until frozen).

Some compiler changes will be required to choose the correct one. Compiling to a literal/primitive will result in a frozen String, but that's not always wanted. In other cases it should compile to Opal.String.$new() or so.
Ruby 3.4 is also making some changes concerning frozen Strings.

@hmdne + @elia please check, see if this concept is ok for you. If you are ok with this concept i would then continue on that path.

### DIFF
--- a/opal/corelib/string.rb
+++ b/opal/corelib/string.rb
@@ -1,4 +1,4 @@
-# helpers: coerce_to, respond_to, global_regexp, prop, opal32_init, opal32_add, transform_regexp
+# helpers: coerce_to, respond_to, global_regexp, prop, opal32_init, opal32_add, transform_regexp, deny_frozen_access
 # backtick_javascript: true
 
 # depends on:
@@ -14,6 +14,7 @@ class ::String < `String`
     const MAX_STR_LEN = Number.MAX_SAFE_INTEGER;
 
     Opal.prop(#{self}.$$prototype, '$$is_string', true);
+    self.$$is_string_class = true;
 
     var string_id_map = new Map();
 
@@ -426,6 +427,12 @@ class ::String < `String`
       }
       return ascii;
     }
+
+    function str_deny_frozen_access(str) {
+      if (str["$frozen?"]()) {
+        #{raise(FrozenError, "can't modify frozen String: " + `str.$inspect()`)};
+      }
+    }
   }
 
   # Force strict mode to suppress autoboxing of `this`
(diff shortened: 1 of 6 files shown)


janbiedermann:
Specs are expected to fail at this time. Only 23 failures, some due to different #frozen? semantics. And at the same time the first String#<< and String#concat specs passing!

opal/corelib/string.rb
```diff
@@ -2413,12 +2447,167 @@ def freeze
...
+  proto.$$bridge_do_not_touch = true;
+
+  // Here now all the functions that mutate this.value.
+  // Though these functions here are simple and do not validate args or anything, this
+  // is done by the Ruby methods above instead, as they need to raise Ruby exceptions
+  // anyway. So usual JavaScript rules apply. 
+  proto.append = function(s) { return this.value += s.toString(); }
+
+  // Make MutableString globally available.
+  Opal.global.MutableString = MutableString;
```
hmdne, line 2508: Do we need it on global? Or is it just a convenience?

opal/corelib/string.rb
```diff
@@ -2413,12 +2447,167 @@ def freeze
...
+
+# Essentially we have two String implementations floating around, with the
+# immutable JavaScript String being hidden, mostly, unless its directly accessed
+# as `String`. A "implementation detail".
+# We must take care to keep them both synchronized. Also its important to note,
+# that the Opal.String <-> MutableString bridge is not a "clean" bridge, but one
+# used with some underlying "magic". So even internally within Opal its preferrable
+# to use Opal.String.$new() over new MutableString(). Best practice is not to use
+# MutableString directly at all, except for debugging.
+class String < `MutableString`
```
hmdne, line 2523: This seems pluggable. Can we perhaps do `corelib/string/mutable.rb` ?

test.rb
```diff
@@ -0,0 +1,12 @@
+require 'opal'
+puts 'ready'
+
+t = String.new('bla')
```
hmdne, line 4: So, this is the only way we can get a mutable String, right? Perhaps we can also do `String#+`, maybe `String#dup` and maybe also by default for `dstr` (ie. expressions like `"test#{other}"`). If it can be pluggable, then we would just do `String.new` on it.

hmdne:
Overall, nice implementation, but it feels kinda hacky to me. Perhaps we could get rid of some hacks? Also I left a couple of comments.

opal/corelib/string.rb
janbiedermann: if the bridge would work perfectly i think it would be nice to have a shortcut for internal use, like new Map() for ::Hash. But currently its a convenience for debugging.

opal/corelib/string.rb
janbiedermann: Probably, i will experiment further.

test.rb
janbiedermann: Yes, currently its the only way. So everywhere we return a new_string it should be ::String.new(new_string)

janbiedermann:
@hmdne Thanks for looking over it! I agree, its a bit hacky. So this iteration improves Opal.bridge to work with arbitrary JS Classes,  so also with MutableString. It also improves MutableString to be a real JS class. Its a bit less hacky now. Its not yet pluggable, i still want to experiment a bit "within the same realm".

janbiedermann:
This is maybe is so far the least hacky approach, but still hacky. It removes the shadowed Ruby Class, uses a nice JS Class, and a Proxy ... although it solves some problems of the previous attempts, it creates new ones. Main Problem is now, need to take care to return the Proxy to keep things working. However, the impact on asciidoctor performance is less than expected, so using a Proxy to cover the length issue for String may be a better way with your (@hmdne) previous approach in #2358. Basically a mutable String would then be just a Proxy of String.

PR #2715, opened 2024-12-18
[WIP] Implement mutable ::String via direct String Proxy

janbiedermann:
This implements mutable Strings via a direct Proxy of String building on the previous approach in #2358. This is far the cleanest approach and only a few of the existing specs fail, mostly because of encoding issues. Yet, accessing the length property of String even via Proxy  fails with: 
`TypeError: 'get' on proxy: property 'length' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '3' but got '7')`

Conclusion:
To get a proper length property the only way is to explicitly attach the String.prototype to another constructor.
This way however index access via [] is lost, which can only be recreated via a Proxy.

This leaves the latest attempt of #2713 as the most promising, if correct .length property access and index access via [] are considered a requirement.

But lets see what performance says about this String Proxy approach ...


janbiedermann:
Nice:
```
               Run time on node: 1.570 (+4.07%) -> 1.577 (+9.71%) (change: +0.50%)
               Run time on bun: 1.858 (+22.80%) -> 2.149 (+14.91%) (change: +15.63%)
```

Issue #2706, opened 2024-10-21
Bug: Broken on Bun > 1.1.30


janbiedermann:
On Bun version 1.1.31 the following happens:

Error
    at $deny_frozen_access (/Users/jan/workspace/opal/tmp/mspec_bun.js:2578:25)
    at $rehash (/Users/jan/workspace/opal/tmp/mspec_bun.js:17999:7)
    at <anonymous> (internal:primordials:35:62)
    at forEach (native:1:11) 
    at <anonymous> (internal:primordials:31:15)
    at <anonymous> (internal:primordials:1:11)
    at anonymous (native)
    at <anonymous> (internal:validators:1:11)
    at anonymous (native)
    at <anonymous> (node:tty:1:11)
    at anonymous (native)
    at <anonymous> (native:1:11)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:22747:22) typeof(proccess.stdout)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2813:17)
    at $require (/Users/jan/workspace/opal/tmp/mspec_bun.js:6002:19)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:22783:18)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2813:17)
    at $require (/Users/jan/workspace/opal/tmp/mspec_bun.js:6002:19)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:28193:11)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:2778:20)
    at <anonymous> (/Users/jan/workspace/opal/tmp/mspec_bun.js:28214:6)
    at <anonymous> (native:11:43)

There are multiple question marks here. The problem is triggered by loading corelib/io, which towards its end accesses typeof(proccess.stdout) which causes the problem.

    Why tries accessing process.stdout try to copy props of Opal Ruby Objects?
    https://github.com/oven-sh/bun/blob/f3b658d9f7b0d3f45dcfcdf845be44ea80e05055/src/js/internal/primordials.js#L37
[...]

PR #2718, opened 2024-12-20
Towards Mutable Strings, Step 1 of X: Introduce global $str helper 

janbiedermann:
... for creating Strings
Consistently use $str withing String whenever encoding is set
Enable Deno specs again


janbiedermann:
Although this PR enables Deno again, it does not help all with the Bun problems of #2706, which i just verified.

PR #2720, opened 2024-12-29
Implement EUCJP, JIS and SJIS encodings, add Encoding#decode(io_buffer)

janbiedermann:
Encoding#decode(io_buffer) will be required by IO


hmdne:
my nitpicks:
 * perhaps not 'corelib/string/encoding_sjis` but `corelib/string/encoding/sjis`?
 * ensure they are required by `opal/full` (so that they are picked up by tests)
 * how are the tables generated? can we perhaps add some script for that? (for future maintenance)

janbiedermann:
how are the tables generated?

I included the generator and table sources

janbiedermann:
Caught a load order issue, changed tables to .rb to improve loading behavior.

janbiedermann:
Encoding#decode and #each_byte_buffer will be used by IO, allowing to read and write strings correctly, even strings in Japanese encodings.

elia (approved):

PR #2722, opened 2025-01-04
Fix `Module#include` after `#prepend`

### LINKED ISSUE #2721: Bug: Ancestor chain not correctly constructed
**Describe the bug**

Opal version: 1.8.2

**To Reproduce**

Hey there 👋 Thanks for Opal, it's amazing!!

### The problem

I ran across a curious bug today and thought I'd dig in to see if I could figure it out. Here's what I believe to be a minimal reproduction case. On the latest Opal (v1.8.2), this code results in infinite recursion (i.e. I get a `Maximum call stack size exceeded (Exception)` from JavaScript):

```ruby
module Api
  def self.included(base)
    base.prepend(MethodOverrides)
  end

  module MethodOverrides
    def initialize
      super
    end
  end
end

module Utilities
end

module Behavior
  def self.included(base)
    base.include(Api)
    base.include(Utilities)
  end
end

class Test
  include Behavior
end

Test.new
```

The same code does not error under MRI v3.3 and just exits normally.

### What's going wrong

The `initialize` method in the `MethodOverrides` module calls `super`, which should call `BasicObject#initialize` but instead calls itself over and over.

If we take a look at `Test`'s ancestors, we see that `Api::MethodOverrides` is listed twice. It's only listed once under MRI 3.3.

<table>
<thead>
<tr>
<td>Opal 1.8.2</td>
<td>MRI 3.3.0</td>
</thead>
<tbody>
<tr>
<td>

```ruby
[
  Api::MethodOverrides
  Test
  Utilities
  Api::MethodOverrides
  Object
  Kernel
  BasicObject

]
```

</td>
<td>

```ruby
[
  Api::MethodOverrides
  Test
  Utilities
  Api
  Behavior
  Object
  Kernel
  BasicObject
]
```

</td>
</tr>
</tbody>
</table>

Moreover, we also see that the `Api` and `Behavior` modules are _not_ listed as ancestors under Opal.

### The solution

Initially I thought the right thing to do was to filter out duplicates from the ancestor chain. I implemented this and it worked, but I later realized it wouldn't account for the missing `Api` and `Behavior` ancestors. So unfortunately I don't have a solution at the moment. Looking for ideas from the community and maintainers 😄

takaram:
Closes #2721

### DIFF
--- a/opal/corelib/runtime.js
+++ b/opal/corelib/runtime.js
@@ -1090,7 +1090,13 @@
   }
 
   function own_included_modules(module) {
-    var result = [], mod, proto = Object.getPrototypeOf(module.$$prototype);
+    var result = [], mod, proto;
+
+    if ($has_own(module.$$prototype, '$$dummy')) {
+      proto = Object.getPrototypeOf(module.$$prototype.$$define_methods_on);
+    } else {
+      proto = Object.getPrototypeOf(module.$$prototype);
+    }
 
     while (proto) {
       if (proto.hasOwnProperty('$$class')) {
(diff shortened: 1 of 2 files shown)


hmdne:
Thank you @takaram ! I tried fixing that, but I don't really understand how include/prepend work.

PR #2728, opened 2025-01-27
corelib/array: Create a fastpath for #any?

hmdne:
This optimizes opalopal by a consistent 1.2% on a large dataset.


opal/corelib/array.rb
```diff
@@ -556,7 +556,25 @@ def []=(index, value, extra = undefined)
...
+
+      // A special optimized fastpath for Array#any? if no argument
+      // or block is given. Since Array#any? in this form is often
+      // used instead of !Array#empty?, and Enumerable#any? uses
+      // closure exceptions, this shortpath may provide some
+      // optimization gains.
+      if (pattern == null && !#{block_given?}) {
+        for (var i = 0; i < self.length; i++) {
+          var elem = self[i];
+          if (elem != null && elem !== false && elem !== nil)
```
takaram, line 570: You need to care about boxed boolean (i.e. `new Boolean(false)`).
Maybe we should use `$truthy()`?

hmdne: @takaram I am all for dropping support for boxed booleans. Supporting them was crucial before we ensured `use_strict` (you are probably aware of it, but for the reference, in non-strict mode, when adding functions to a prototype - not when calling - `this` is automatically boxed; in strict mode it's kept as-is), now they don't happen anymore (unless from user provided structures, but even then, they are a very obscure feature). Though @elia thinked differently last time we talked about this, basically he said that the cost of supporting them is very low. So we kept them.

`$truthy` is a good idea, I forgot about this. Thanks.

PR #2729, opened 2025-02-06
::Array fix Exception (Maximum call stack size exceeded)

janbiedermann:


elia (approved):
I added some erro messages

PR #2730, opened 2025-02-15
Towards Mutable Strings, Step 2 of X: Introduce global $dstr helper and use it for :dstr Nodes 

janbiedermann:


janbiedermann:
lint complaining is totally unrelated

janbiedermann:
The truth of the last commit message is just slightly projected into the near future.

PR #2733, opened 2025-02-16
Dead Code Elimination

hmdne:
Earlier approaches: #2370, #1357

For now, we had a pipeline:

```
Parser -> Rewriters -> Compiler -> Builder
```

This replaces the pipeline to:

```
Parser -> Rewriters -> Compiler -> PostProcessor -> Builder
```

In particular, we come into more and more situations where we need to somehow interface Builder from Compiler. That's a bad approach, since Compiler should be concerned with building a single file, and not a whole bundle (or directory, whatever).

After this patch, Compiler may generate hints to Postprocessor, which has access to Builder.

The first user of this PostProcessor system is a DCE implementation. It works, even with a parser. And it can even trim runtime. I have implemented the following algorithm:

1. Get all files, built normally (but with hints)
2. Compute the method names that are called by the TOP or by other methods
3. Take method names called by the TOP, put them into a Set
4. Take all methods from that Set and also add to the Set methods that are called by those methods
5. Repeat 4
6. Go through all the files and remove all methods that are not in that Set
7. Then pass that to Builder and we continue as usually

For the most part, we just do a dirty thing, we just regexp all JS files and Ruby fragments for anything that looks like a call. Some can't be caught that easily - we have a compiler hint `dce_use` that can be added to ensure that we capture a method call.

For if a method call is in a method or not, we do have two helpers: `dce_def_begin` and `dce_def_end`. This also marks parts that we can cut from resulting file if a method is not needed.

I have confirmed this approach also works with `opal-parser` (with some special handling), though the first thing you should strip for code size optimization is `opal-parser` itself.

### DIFF
--- a/spec/lib/builder/post_processor/dce_spec.rb
+++ b/spec/lib/builder/post_processor/dce_spec.rb
@@ -1,0 +1,92 @@
+require 'lib/spec_helper'
+require 'opal/builder'
+
+RSpec.describe Opal::Builder::PostProcessor::DCE do
+  let(:builder) { Opal::Builder.new(compiler_options: {cache_fragments: true}, dce: dce_types) }
+
+  def build(code)
+    builder.build_str(code, 'input.rb').to_s
+  end
+
+  context "when dce removes unused instance methods" do
+    let(:dce_types) { [:method] }
+
+    it "strips methods that are never referenced" do
+      output = build(<<~'RUBY')
+        class Foo
+          def used; 1; end
+          def unused; 2; end
+        end
+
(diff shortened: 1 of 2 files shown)


hmdne:
Other earlier work for reference:

- https://github.com/hmdne/opal-multishake-demo
- https://github.com/hmdne/opal-optimizer
  - This won't work with Opal 2.0, but I used this in production pre-2.0.

hmdne:
The postprocessor subsystem has been merged earlier. We are merging only the DCE part now.

PR #2740, opened 2025-03-08
nodes/literal: Don't annotate regexps when adding "u" flag.

hmdne:
This should improve performance, as annotate_regexp is quite a heavy call.


hmdne:
Not an improvement I expected. Still, worth it IMO, even if only for a marginal bundle size improvement.

PR #2746, opened 2025-04-02
Remaster platform support with a common corelib ruby implementation and small platform drivers and more ...

janbiedermann:
### Goals of this PR

- Have a perfectly working corelib on all platforms*
- Improve maintainability of platforms and across platforms
- Improve situation for self hosted Opal

### Goodies

- Many bugs fixed along the way
- Functionality improved
- **Overall 18084 specs working on node/linux** with `bundle exec rake mspec_ruby_node`
- Perfect #format*
- Perfect #fnmatch*
- Perfect #glob*
- Perfect #expand_path and #absolute_path*
- Perfect #dirname*
- fileutils
- tmpdir
- complete Pathname
- FileTest
- Kernel#system, Kernel#backtick

### Implementation

Before Opal had custom implementations of everything platform dependent for each platform, which required a lot of code and left most platforms badly and partially supported. This PR improves the situation by providing a common corelib implementation of methods, that depend on platform dependent code, that works across all platforms and requires therefore only small platform drivers to implement the final system calls or provide a emulation or raise a error, depending on platform capabilities.

### Notes for Review

There are still a few todos and see platform footnotes below.

### Platform Status

~~strike~~ means unable to run specs reliably or at all

#### Servers

| driver | support | platform | version | os  | mspec_ruby | mspec_opal | minitest | notes |
| ------ | --- | ----------- | ------- | --- | ------------ | ------------ | ----- | --- |
| x | x | node | 22.16 | lin | **18084**  | 673 | 2627 |   |
| x | x | node | 22.16 | win | 17416 | 673 | 1414 | *1 |
| x | x | node | 22.16 | mac | 17696 | 673 | 3949 | *1 |
| - | - | deno | 2.3.3 | lin/win/mac | - | - | - | #2748  |
| - | - | bun | 1.2.13 | lin/win/mac | - | - | - | #2706 | 
| x | - | graalnodejs | 24.2.0 | lin | - | - | - | *4|

#### Browsers

| driver | support | platform | version | os | mspec_ruby | mspec_opal | minitest | notes |
| ------ | --- | ----------- | ------- | --- | ------------ | ------------ | ----- |--|
| x | x | chrome | 136 | lin | 14537 | 547 | 772, 1e |  |
| x | x | chrome | 136 | win | 14536 | 547 | 772, 1e |  |
| x | x | chrome | 136 | mac | 14536 | 547 | 772, 1e |  |
| x | x | edge | 136 | win | 14536 | 547 | 772, 1e |  |
| x | x | firefox | 138 | lin | 14534 | 547 | 769, 1e |  |
| x | x | firefox | 138 | win | 14533 | 547 | 769, 1e |  |
| x | x | firefox | 138 | mac | 14533 | 547 | 769, 1e |  |
| x | x | safari | 18.5 | mac | 14531 | 547 | 769, 1e |  |

#### Others
| driver | support | platform | version | os | mspec_ruby | mspec_opal | minitest | notes |
| ------ | --- | ----------- | ------- | --- | ------------ | ------------ | ----- |--|
| - | - | gjs | 1.82.1 | lin | ~~0~~ | ~~0~~ | ~~0~~ | *5 |
| - | - | cjs | 128.0 | lin | ~~0~~ | ~~0~~ | ~~0~~ | *5 |
| - | x | quickjs | 2025-04-26 | lin |14514 | ~~545, 2f~~ | ~~0~~ | *8 |
| - | - | quickjs-ng | 0.9.0 | lin | ~~0~~ | ~~0~~ | ~~0~~ |  |
| - | - | miniracer | 0.18.1 | lin | ~~0~~ | ~~0~~ | ~~0~~ | *6 |
| - | - | osascript | 15.4 | mac | ~~0~~ | ~~0~~ | ~~0~~ | *7 |

### Still to do

(no specific order)
- [x] Document String/Encoding
- [x] Update github actions
- [x] `__dir__` and `__FILE__`
- [x] opal_engine_check rewriter and RUBY_PLATFORM mismatch
- [x] rake lint
- [x] Some platforms are missing some JS features, mainly TextDecoder : this PR provides a small generic TextDecoder shim 
- [x] document that on Windows opal must be cloned "as is" for specs to pass ( no \n to \r\n conversion )
- [x] take care of #spawn, #system, etc.
- [x] make sure benchmarks work
- [x] corelib/variables, remove line 16 ::ARGV.shift if ::ARGV.first == '--'
- [x] XMLHttpRequest solution for servers
- [x] improve exit
- [x] IO wait/nonblock/console

### Desired Extractions

- [ ] String/Encoding
- [ ] Regexp transpiler

*as much as possible
*1 numbers not updated
*4 graalnodejs is 10 times too slow, not running tests, but in terms of correctness its very, very close to nodejs results
*5 gjs and cjs suffer from a totally unstable API in terms of IO, not practical for any real use, specific runner and driver removed
*6 mini_racer just exits, best i can get is: in 'MiniRacer::Context#eval': unsupported type b (MiniRacer::InternalError)
*7 no specific driver for osascript anymore, works mostly but specs don't finish, mspec_opal and minitest work but statstics not visible, runner removed
*8 quickjs, as a platform with a runner but no driver, serves as example for a 'unknown' platform


spec/filters/platform/safari/literal_regexp.rb
```diff
@@ -1,6 +1,6 @@
 # NOTE: run bin/format-filters after changing this file
 opal_unsupported_filter "Literal Regexp" do
-  fails "Literal Regexps handles a lookbehind with ss characters"
-  fails "Literal Regexps supports (?<! ) (negative lookbehind)"
-  fails "Literal Regexps supports (?<= ) (positive lookbehind)"
+  fails "Literal Regexps handles a lookbehind with ss characters"
+  fails "Literal Regexps supports (?<! ) (negative lookbehind)"
+  fails "Literal Regexps supports (?<= ) (positive lookbehind)"
```
hmdne, line 5: This should not be the case in modern Safari

opal/corelib/constants.rb
```diff
@@ -1,4 +1,5 @@
-::RUBY_PLATFORM       = 'opal'
+# backtick_javascript: true
+::RUBY_PLATFORM       = `Opal.platform.ruby_platform`
```
hmdne, line 2: From the gist of it, it looks like a breaking change?

janbiedermann: This is required so that ruby specs correctly can identify if they are running on Windows or Linux and select platform specific specs. So in case of the node driver, this returns 'opal win' on Windows or 'opal linux' on Linux or 'opal' on others. Other drivers may return 'opal macos' or something.

janbiedermann: Platform OS identification usually happens by regexp matching RUBY_PLATFORM against /win/, /linux/ etc.

opal/corelib/kernel.rb
```diff
@@ -99,6 +100,10 @@ def Array(object)
     }
   end
 
+  def __dir__
+    File.realpath(File.dirname(__FILE__))
```
hmdne, line 104: __FILE__ here is opal/corelib/kernel.rb, so __dir__ is always opal/corelib

opal/corelib/kernel.rb
```diff
@@ -270,6 +275,27 @@ def extend(*mods)
     self
   end
 
+  def fork
```
hmdne, line 278: Nice :D

opal/corelib/kernel.rb
janbiedermann: This __dir__ ___FILE___ thing probably needs some further discussion/work. I add it to the TODO list above.

opal/corelib/variables.rb
```diff
@@ -12,9 +12,9 @@
 $/ = "\n"
 $, = nil
 
-::ARGV = []
+::ARGV = `Opal.platform.argv`
+::ARGV.shift if ::ARGV.first == '--'
```
hmdne, line 16: Maybe this should be handled by a runner?

janbiedermann: Good catch, line 16 should be handled by platform driver.

opal/corelib/constants.rb
hmdne: What I mean is opal_platform_check rewriter which strips the code based on where it's used. And currently Opal code may use `==` to check if something is running on Opal, along with RUBY_ENGINE. This is what we will need to investigate more thoroughly.

janbiedermann: I add it to TODO

opal/corelib/variables.rb
janbiedermann: Well, when you compile a script and later run it, lets say with node or deno, there is no runner. So the driver should correctly identify script args.

janbiedermann:
This adds a tiny Virtual File System for browsers, which is required so that specs can run without exiting in between. The spec suite as run by node, deno, bu, etc. executes specs that during setup require IO. When these specs run without IO the whole suite exits with "example has not been reset". As we don't have a conditional file exclusion and as the drivers make it easily possible, i added a VFS, used by browsers, available via 'opal/full'. Its very simple, create, write, read, delete File or Dir works, even File.stat. Most IO/File/Dir specs still fail at this point, but specs overall run.

janbiedermann:
@hmdne @elia Thanks for reviewing part 1, now part 2 is ready for review, which inludes all the drivers. I updated the topmost comment above accordingly.

janbiedermann:
rebase

janbiedermann:
@hmdne @elia Part 3 is ready, fixes and adapts mspec_opal_*

janbiedermann:
fileutils, tmpdir, pathname may seem out of scope of this PR, as it says "fully functional corelib", but corelib specs are not complete and implementing these helped fix some bugs and tests for some ambiguities in corelib. Also fileutils, tmpdir and pathname are very, very useful :-)

janbiedermann:
@hmdne @elia Part 4 is ready

opal/corelib/dir.rb
```diff
@@ -1,23 +1,798 @@
...
+      # Forms an array entry_names of the entry names selected by the arguments.
+      # Argument patterns is a string pattern or an array of string patterns;
+      # note that these are not regexps.
+
+      # This a more or less straightforward port of dir_s_glob and related functions from ruby/dir.c.
+      base = if `base === UNDEFINED`
+              nil
+            elsif base.nil?
+              getwd
+            else base
```
takaram, line 126:
```suggestion
            else
```

opal/corelib/env.rb
```diff
@@ -1,40 +1,41 @@
...
-    `Object.keys(process.env)`
+    `$platform.env_keys()`
   end
 
   def delete(name)
     %x{
-      var value = process.env[#{name}] || nil;
-      delete process.env[#{name}];
+      let value = $platform.env_get(name) || nil;
+      delete $platform.env_del(name);
```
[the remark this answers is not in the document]
janbiedermann: Thanks, good catch!

janbiedermann:
https://github.com/opal/opal/pull/2746/commits/9f3ff8f6cc735bed39ab17b9e8ccc06ae06a800c changes the driver interface a bit, which only works with node and breaks all other platforms. Working on it ...

janbiedermann:
Ooops, the process.rb part is still missing, working on it too ..., so specs may not run at all at this stage

opal/corelib/dir.rb
janbiedermann: Thanks, fixed

janbiedermann:
@hmdne @elia Part 5 is ready, it runs specs for node only, as i changed driver interface a bit and need to update other drivers, and specs for node do not even run clean, 4 issues in spec setups. Part 5 contains important changes:
- new exit code
- backtick default to false with a working Kernel#backtick
- removal of MSpec platform rewriters
and fixes from @takaram 

This contains kinda working process spawning methods, but they are still in progress, specifically the args checking needs to be common, coming next.

This should have been the "minispecs working part", but well, thats maybe for part 6 or 7.

stdlib/stringio.rb
```diff
@@ -1,116 +1,979 @@
...
+    s = @buffer.size
+    return nil if @eof || @pos >= s
+    c = @buffer.get_raw_string(@pos, `Math.min(16, s - self.pos)`, @ext_enc) # account for multi byte chars
+    c = c[0]
+    @pos += c.bytesize
+    @eof = true if @pos >= s
+    c
+  end
+
+  def gets(sep = $/, limit = nil, chomp: false)
```
hmdne, line 374: Ruby StringIO has IO::generic_readable and IO::generic_writable as ancestors - which seem like to be an internal thing.

Still we may want to simulate it somewhat, so that certain functions are not duplicated.

opal/corelib/string.rb
```diff
@@ -8,7 +8,18 @@
 class ::String < `String`
```
hmdne, line 8: Can the changes on String/Encoding be separated to another PR?

opal/corelib/process.rb
```diff
@@ -1,59 +1,415 @@
...
-      monotonic = function() {
-        var hrtime = process.hrtime(hrtime_base);
-        var us = (hrtime[1] / 1000) | 0; // cut below microsecs;
-        return ((hrtime[0] * 1000) + (us / 1000));
-      };
-    }
-  }
+    def argv0
+      # Returns the name of the script being executed.
+      ARGV[0]
```
hmdne, line 25: This implementation is wrong:

```
[user@localhost ~]$ pry -- asd asd asd
[1] pry(main)> ARGV
=> ["asd", "asd", "asd"]
[2] pry(main)> ARGV[0]
=> "asd"
[3] pry(main)> Process.argv0
=> "/home/user/bin/pry"
[4] pry(main)> 
```

opal/corelib/process/status.rb
```diff
@@ -1,21 +1,75 @@
...
-      @status
+    def ==(other)
+      # Returns whether the value of to_i == other
+      to_i == other
     end
 
-    attr_reader :pid
+    # Returns true if the process generated a coredump when it terminated, false if not.
+    def coredump?
+      __not_implemented__
```
hmdne, line 20: Perhaps we should do some helper for not implemented functions, like:

```ruby
Opal::Raw.not_implemented :coredump?
```

What this helper could do as well, is to make `respond_to?` return `false` on those methods as it does in CRuby.

opal/io/wait.rb
```diff
@@ -0,0 +1,34 @@
+class ::IO
```
hmdne, line 1: Maybe it would make more sense to move those into the `stdlib`?

hmdne:
We have identified a lot of parts that can be extracted as separate PRs that we can merge much faster (like let's say changes on regexp transpiler). The primary issue is that this PR is huge and certain design decisions should be understood better. By extracting certain parts we may be able to reduce the delta.

opal/corelib/kernel.rb
```diff
@@ -299,436 +419,609 @@ def frozen?
     }
   end
 
+  # gem - not supported yet
+
   def gets(*args)
-    $stdin.gets(*args)
+    # Returns (and assigns to $_) the next line from the list of files in ARGV (or $*), or from standard input.
+    ::ARGF.gets(*args)
```
takaram, line 426: Calling `Kernel.gets` results in an infinite loop.

```console
$ bin/opal -e 'gets'
Arguments.slice (<anonymous>): Maximum call stack size exceeded (Exception)
        from <internal:corelib/kernel.rb>:1:1:in `send2'
        from <internal:runtime/send.rb>:59:1:in `send'
        from <internal:runtime/send.rb>:45:1:in `gets'
        from <internal:corelib/kernel.rb>:426:11:in `send2'
        from <internal:runtime/send.rb>:59:1:in `send'
        from <internal:runtime/send.rb>:45:1:in `gets'
        from <internal:corelib/kernel.rb>:426:11:in `send2'
        from <internal:runtime/send.rb>:59:1:in `send'
        from <internal:runtime/send.rb>:45:1:in `gets'
        from <internal:corelib/kernel.rb>:426:11:in `send2'
        from <internal:runtime/send.rb>:59:1:in `send'
        from <internal:runtime/send.rb>:45:1:in `gets'
        from <internal:corelib/kernel.rb>:426:11:in `send2'
...
```

`Kernel.gets` calls `ARGF.gets`, but `ARGF.class#gets` is undefined because `corelib/argf.rb` is not loaded.

spec/filters/platform/safari/literal_regexp.rb
janbiedermann: Thanks, done.

opal/corelib/string.rb
janbiedermann: I added another section to the top comment here "Desired Extractions", please add everything you want in a separate PR there or mention in a comment and i will add and i will then attend to them once everything is working.

opal/corelib/process.rb
janbiedermann: Thanks, fixed in the latest commit.

opal/io/wait.rb
janbiedermann: These are tested in spec/ruby/core so i would prefer to have them in corelib. They don't have a real implementation yet, i add it to todo

opal/corelib/kernel.rb
janbiedermann: Thanks, fixed in the latest ARGF commit.

janbiedermann: But there is a new, node specific error invalid seek Errno::ESPIPE.

janbiedermann:
@hmdne , @elia , @takaram part 6 is ready,  it does not yet fix all found issues, but makes mspec_ruby, mspec_opal and minitest work with node, along with ARGF, ARGV and ENV

janbiedermann:
@hmdne , @elia , @takaram part 7 is ready, this makes mspec_ruby, mspec_opal, minitest work on Bun, Deno, Node; Bun of course, with errors. Out of frustration i totally removed support for quickjs, gjs, cjs, mini_racer and osascript. We don't test them, they don't work, they can't run specs. osascript, cjs, gjs are broken, they have a autogenerated API where nobody verified if functions are callable at all, crashing all the time. mini_racer is terrible to debug, so is quickjs. I love quickjs, but its os an std API is not helpful, so whatever we do, running on quickjs is considered broken, especially IO. So i removed support for quickjs too.

opal/corelib/process/status.rb
janbiedermann: I added a simple check for a alias to __not_implemented__ to #respond_to? in the latest commit [ae6127a](https://github.com/opal/opal/pull/2746/commits/ae6127ab8d22d21be5df9d6b5a2b0bd8c4de1895). This allows for overwriting the alias and #respond_to? will automatically be doing the right thing.

janbiedermann: However, what is __not_implemented__ clearly depends on drivers too, need to think about that a bit, how i can bring those 2 together

opal/corelib/kernel.rb
takaram: Errno::ESPIPE is raised when `$stdin` is tty.

When `$stdin` is not a tty, no error is raised but `gets` always returns nil.
This seems to be caused by `IO#eof` which is broken with stdin.

https://github.com/janbiedermann/opal/blob/ae6127ab8d22d21be5df9d6b5a2b0bd8c4de1895/opal/corelib/io.rb#L797

```console
$ echo foo | bin/opal -e 'p gets'
nil
$ echo foo | bin/opal -e 'p $stdin.eof'
true
```

janbiedermann: Fixed in the last commits.

janbiedermann:
@hmdne , @elia , @takaram part 8 is ready, fixes bugs found by @takaram and makes linter happy. Most important change is probably the new platform specific spec selection. mspec ruby/opal, minitest run on all platforms with only a few errors remaining.

janbiedermann:
See #2748

janbiedermann:
Github Actions almost done, asciidoctor needs some # backtick_javascript: true comment to function properly for the performance task.

janbiedermann:
Benchmarks fixed

janbiedermann:
Some rspec actions where apperaring green although they should have been red, so i missed them. As they now appear red i removed the "ready to merge" label and handle the failing spec.

PR #2747, opened 2025-05-07
Add Compiler WhileNode retry look ahead, fix #2743

### LINKED ISSUE #2743: Bug: while loop with begin ... rescue ... retry breaks 'break' and 'return' from loop
Opal version: master

For `return`:
```
def name
  while cond
    return if another_cond # <---- here
    begin
      case
      when something
      end
    rescue
      retry
    end
  end
end
```
The return above does not return the method but executes another loop.
Compiled code:
```
while (condition) {
                    (function() {
                        try {
                            if (another_cond) {
                                return
                            }
```

For `break`:
```
def name
  while cond
    break if another_cond # <---- here
    begin
      case
      when something
      end
    rescue
      retry
    end
  end
end
```
Compiling results in an error:
```
/tmp/opal-system-runner20250323-109580-ax051o.js:30379 
         break 
         ^^^^^ 
 
SyntaxError: Illegal break statement 
```

Workaround:
```
def name
  while cond
    if another_cond
        cond = false  # make condition false so that while does not execute another loop
        return        # and use return
    end
    ...
```

janbiedermann:
This adds some slowness though
Fixes #2743

### DIFF
--- a/lib/opal/nodes/base.rb
+++ b/lib/opal/nodes/base.rb
@@ -228,6 +228,19 @@ module Opal
         line = @sexp.loc.line
         "['#{file}', #{line}]"
       end
+
+      def node_has?(node, type)
+        # look ahead if a child with specified type is in the tree
+        if node
+          node.children.each do |child|
+            if child.is_a?(::AST::Node) || child.is_a?(::Opal::Nodes::Base)
+              return true if child.type == type
+              return true if node_has?(child, type)
+            end
+          end
+        end
+        false
+      end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


janbiedermann:
No, seemingly no performance disadvantage for compiling asciidoctor

janbiedermann:
The spec i tried to implement did not fail with old master, so i removed them. However this bug has been triggered by Kernel#format in #2746, so other specs will fail, should this problem surface again.

hmdne:
JRuby consistently fails here (probably because of JRuby update, not on our side). But otherwise, I can certify this patch is just a little deoptimization on a rarely used part of Ruby.

PR #2781, opened 2026-03-18
spec: reduce rspec runtime overhead

hmdne:
Reduce the RSpec suite runtime by removing expensive setup and execution paths that several examples did not actually need for the behaviors they assert. This brings the suite from 1 minute 7.36 seconds to 9.67 seconds by shifting slow specs toward the cheapest code paths that still preserve their intent.

The REPL spec now reuses a duplicated memoized builder template and a memoized silencer instead of rebuilding that setup for every evaluation. The `rake dist` smoke coverage now builds only the JavaScript artifacts that the spec actually loads, writes them into a temporary directory, and cleans that directory up afterward instead of paying for the full dist pipeline. Several CLI specs now assert compiler output rather than spawning a full runtime when the behavior under test is argument assembly, require resolution, or emitted JavaScript. Pipe coverage is still exercised, but it now uses `Open3.capture2` with stdin input rather than a heavier shell pipeline. Builder and simple server specs likewise avoid compiling the full Opal runtime when a smaller fixture or a lighter builder setup is enough to verify the same behavior.

These changes keep the suite coverage focused on the semantics each example is meant to protect while removing avoidable work that dominated wall clock time. The result is a materially faster feedback loop for compiler and runtime work without dropping the relevant assertions from the optimized specs.

Before:
- Finished in 1 minute 7.36 seconds (files took 0.2438 seconds to load)

After:
- Finished in 9.67 seconds (files took 0.26181 seconds to load)

### DIFF
--- a/lib/opal/repl.rb
+++ b/lib/opal/repl.rb
@@ -76,8 +76,8 @@ module Opal
     end
 
     def eval_ruby(code)
-      builder = Opal::Builder.new
-      silencer = Silencer.new
+      builder = build_repl_builder
+      silencer = (@silencer ||= Silencer.new)
 
       code = "#{@incomplete}#{code}"
       if code.start_with? 'ls '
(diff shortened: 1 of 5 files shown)


elia (approved):
🏎️

PR #2787, opened 2026-05-01
builder: fix JRuby 10.1 rspec

hmdne:
This is really weird... Somehow during an upgrade of JRuby from 10.0 to 10.1 the CI stopped working, with JRuby actually messing up our compiler cache (replacing the output with a source map).

As of now, I know it's related to 3 components: `{ nil => Set[] )` being added to our compiler options (or any other key, but with a Set value) hash and marshaling.

The fix has been found by my agent (removing this useless nil key from compiler options hash), but it is unable yet to produce a minimal reproduction.

For now, this PR fixes JRuby CI.

### DIFF
--- a/lib/opal/builder/scheduler/threaded.rb
+++ b/lib/opal/builder/scheduler/threaded.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'etc'
+
 module Opal
   class Builder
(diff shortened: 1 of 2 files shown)


hmdne:
Let's close it for now. I have pushed a fix to JRuby, as it actually is a JRuby bug.

PR #2788, opened 2026-05-05
stdlib: add Opal::JS interface

hmdne:
Introduce Opal::JS as the opt-in Ruby-facing interface for working with JavaScript values. The new stdlib entrypoint provides the $js global wrapper, live Object/Array/Function wrappers, explicit Ruby-to-JS and JS-to-Ruby boundary conversion, callback handling, stable wrapper/function identity, and conversion errors for unsupported or cyclic projections.
    
Add the Opal::JS::Wrapper DSL for typed JavaScript wrappers, including js_object, js_array, js_constructor, wrapper registration, js_method/js_accessor helpers, and argument projection for APIs that expect callbacks, rest args, and keyword option objects in a different order.
    
Expose Ruby values back to JavaScript through Opal.$ using a Proxy facade. The export side supports constant lookup, construction, method/setter dispatch, JS-facing reflection names, exact escapes through __send__/__const__, raw argument shaping through __send_raw__, callable member conversion back to Ruby, and Promise-friendly projection without exposing Ruby internals.
    
Document the beginner-facing Opal::JS workflow, wrapper patterns, and Opal.$ usage, and cover the new behavior with focused Opal stdlib specs plus stress cases for conversion cycles, reflection, wrapper registration, and argument shaping.

### DIFF
--- a/docs/opal_js.md
+++ b/docs/opal_js.md
@@ -1,0 +1,252 @@
+# Opal::JS
+
+`Opal::JS` is the usual way for Ruby code compiled by Opal to work with the
+JavaScript world. Use it when you need to read browser or Node globals, call JS
+methods, pass callbacks, or wrap a JS class with a Ruby API.
+
+```ruby
+require 'opal/js'
+```
+
+Requiring `opal/js` defines `$js`, a wrapped JavaScript global object.
+
+## Ruby To JavaScript
+
+Use `$js` to read properties, assign properties, and call JavaScript methods.
+
+```ruby
+$js.document.title = 'Hello from Opal'
+
+element = $js.document.query_selector('#app')
+element[:textContent] = 'Loaded'
+```
+
+Dynamic Ruby method names try the exact JavaScript property first, then a
+camelCase translation. This means both forms can work:
+
+```ruby
+$js.document.querySelector('#app')
+$js.document.query_selector('#app')
+```
+
+Use bracket access for exact JavaScript property names. Bracket access never
+translates names.
+
+```ruby
+$js[:document][:body]
+$js.document[:querySelector].call('#app')
+```
+
+If bracket access returns a JavaScript function, the wrapper keeps the original
+receiver so `this` is preserved when calling it.
+
+## Conversion Rules
+
+Values passed from Ruby to JavaScript are made JavaScript-friendly:
+
+- `nil` becomes JavaScript `undefined`.
+- `Array` becomes a JavaScript array.
+- `Hash` becomes a plain JavaScript object.
+- `Proc` becomes a stable JavaScript function.
+- `Opal::JS` wrappers unwrap to their JavaScript value.
+
+Hash keys are not camelized during normal Ruby-to-JavaScript conversion.
+Unsupported keys raise `Opal::JS::ConversionError`. Use string, symbol, or
+numeric keys for plain JavaScript object conversion.
+
+JavaScript objects coming back to Ruby stay live by default. Call `to_h` or
+`to_a` only when you want a snapshot.
+
+```ruby
+config = Opal::JS.wrap(`{name: 'opal', nested: {ok: true}}`)
+config[:nested][:ok] # => true
+config.to_h         # => { 'name' => 'opal', 'nested' => { 'ok' => true } }
+```
+
+Snapshot conversion raises `Opal::JS::ConversionError` for cyclic object graphs.
+
+## Callbacks
+
+Ruby blocks and procs can be passed to JavaScript callbacks.
+
+```ruby
+callback = proc do
+  $js.console.log('later')
+end
+
+$js.set_timeout(callback, 100)
+```
+
+During a synchronous JavaScript callback, `Opal::JS.this` exposes the JavaScript
+receiver.
+
+```ruby
+$js.button.add_event_listener('click') do
+  Opal::JS.this[:disabled] = true
+end
+```
+
+The same Ruby callable maps to the same JavaScript function, so listener removal
+can use the same proc object.
+
+## Wrapper Classes
+
+Use `Opal::JS::Wrapper` to give a JavaScript class a Ruby API.
+
+```ruby
+class Document
+  include Opal::JS::Wrapper
+
+  js_object
+  js_register_wrapper $js.Document
+
+  js_method :query, 'querySelector'
+end
+
+$js.document.query_selector('#app') # dynamic access works directly
+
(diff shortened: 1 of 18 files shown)


docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+class Document
+  include Opal::JS::Wrapper
+
+  js_object
+  js_register_wrapper $js.Document
+
+  js_method :query, 'querySelector'
+end
+
+doc = Document.wrap($js.document)
```
hmdne, line 108: `$js.document` itself should also work

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+`js_method` can reshape Ruby arguments for JavaScript APIs.
+
+```ruby
+module Timers
+  include Opal::JS::Wrapper
+
+  js_method :set_timeout, 'setTimeout', args: [:&, :*]
+end
+```
+
```
hmdne, line 177: Nice presentation, but perhaps we should also describe how to use it.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+Keyword keys are converted only when requested with `kwargs: :convert`.
+
+## JavaScript to Ruby: `Opal.$`
+
+`Opal.$` exposes Ruby constants and objects to JavaScript through a proxy facade.
+
+```javascript
+const Parser = Opal.$.Parser;
+const parser = new Parser({strictMode: true});
```
hmdne: But - we don't translate hashes, only method names.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+
+parser.source = 'text';      // Ruby: parser.source = 'text'
+parser.parse();             // Ruby: parser.parse
+parser.convert('input');    // Ruby: parser.convert('input')
+```
+
+JavaScript names are mapped to Ruby names:
+
+```javascript
```
hmdne, line 209: This is a surprising thing. Needs to be emphasized.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+Uppercase names look up Ruby constants first:
+
+```javascript
+Opal.$.Asciidoctor.Converters.HTML
+```
+
+Use escape helpers for exact Ruby access:
+
+```javascript
+obj.__send__('empty?');
```
hmdne, line 224: Or we could use `obj['empty?']()`

docs/opal_js.md
hmdne: Also module inclusion should be somehow documented.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+
+obj.__send_raw__('method_name', function () { return 'block'; });
+obj.__send_raw__('method_name', {kwarg: 123});
+```
+
+`Object.keys`, `in`, and property descriptors expose JavaScript-facing names.
+For example, Ruby `my_property` appears as `myProperty`. Exact Ruby names remain
+available through `__send__`.
+
+`then` is hidden on non-Promise facades so Ruby objects do not accidentally
```
hmdne: Not needed. This is rather implementation detail.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Wrapper Classes
+
+Use `Opal::JS::Wrapper` to give a JavaScript class a Ruby API.
+
+```ruby
+class Document
+  include Opal::JS::Wrapper
+
+  js_object
```
hmdne, line 100: Describe what js_object does and why it's highly recommended to use it.

Also briefly describe js_array.

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
+# Opal::JS
+
+`Opal::JS` is Opal's Ruby API for working with JavaScript values. It is meant
```
hmdne: It is basically a go-to thing to access JavaScript world.

Needs to be more beginner friendly. Mention about `native` must be secondary.

stdlib/opal/js/conversion.rb
```diff
@@ -0,0 +1,288 @@
...
+              Opal.JS.$$this = previous;
+            }
+          };
+
+          proc.$$opal_js_function = fn;
+          return fn;
+        }
+      end
+
+      def this
```
hmdne, line 218: Add a comment to describe what it is.

docs/opal_js.md
janbiedermann: i agree, don't mention native

docs/opal_js.md
janbiedermann, line 108: Briefly describe what #wrap does

docs/opal_js.md
janbiedermann, line 177: yes, a description and some more examples would be nice to have here

docs/opal_js.md
```diff
@@ -0,0 +1,194 @@
...
+
+parser.source = 'text';      // Ruby: parser.source = 'text'
+parser.parse();             // Ruby: parser.parse
+parser.convert('input');    // Ruby: parser.convert('input')
+```
+
+JavaScript names are mapped to Ruby names:
+
+```javascript
+parser.myProperty(); // Ruby: parser.my_property
```
janbiedermann: This is confusing, use different property names for the exampels

docs/opal_js.md
```diff
@@ -0,0 +1,239 @@
...
+
+```javascript
+Opal.$.Asciidoctor.Converters.HTML
+```
+
+Use escape helpers for exact Ruby access:
+
+```javascript
+obj.__send__('empty?');
+obj['empty?']();
```
hmdne: This is not a valid fix. This section describes escape helpers.

docs/opal_js.md
```diff
@@ -0,0 +1,239 @@
...
+
+doc = Document.wrap($js.document)
+doc.query('#app')                   # typed wrapper access
+```
+
+`.wrap` takes an existing JavaScript value, allocates the Ruby wrapper, and
+stores that JS value inside it. Use `.wrap` when the object already exists, such
+as `document`, an event target, or a value returned from another JS API. Use
+`.new` only when the wrapper has a `js_constructor` and should construct a new
+JavaScript object.
```
hmdne: Note `initialize_wrapped`

hmdne: Also ensure you describe that perhaps you wouldn't really need to use it. Because js_constructor/js_register_wrapper will automatically do that for you.

PR #2791, opened 2026-08-01
Resolve requires with a leading ./

### LINKED ISSUE #778: An inconsistency with requiring in MRI vs. Opal
So, I was recently having some problems with `require` in Opal. In MRI, requiring files works like so:
`main.rb`
```
require "./file1.rb"
require "./file2.rb"
```
`file1.rb`
```
puts "file1"
```
`file2.rb`
```
puts "file2"
```
If you run the above code, you'll get `file1` and `file2` printed out. Also note that if you remove the `./` from the `require` statements, the above code will not work. But, in Opal (as of 0.7.1), using the `./` syntax does not work and requiring only works without using `./`. In Opal 0.6.x, it seems to be that you can `require` with either `./` or without it. As far as I can tell, this problem seems to be an inconsistency between MRI and Opal.

elia:
Makes `require "./foo"` resolve instead of raising `MissingRequire`, closing a ten-year-old inconsistency with MRI.

## The bug

`PathReader#expand` returned any `./`-prefixed path verbatim rather than resolving it at all, so resolution only succeeded when the process CWD happened to contain the file. The issue's original reproduction now works, and without needing `-I.`:

```sh
echo 'puts "file1"' > file1.rb
echo 'require "./file1.rb"' > main.rb
opal main.rb   # => file1
```

## `Builder#cwd`

Following @hmdne's review, this adds a `Builder#cwd` property defaulting to `nil`:

- **No cwd** (the default): a leading `./` is stripped and looked up in the load path, so `require './foo'` finds the same asset as `require 'foo'`.
- **cwd set**: `./foo` resolves against it first, as MRI resolves against the process CWD, falling back to the load path.

The CLI sets it to `Dir.pwd`, since running a file is the one entry point with a meaningful working directory. A build task or Sprockets leaves it `nil`, so their output stays independent of the directory the build was invoked from.

## Why a cwd is safe

An earlier revision of this PR claimed resolving against `Dir.pwd` was impossible, because the runtime would not find the module again. **That was wrong**, and the reasoning has been corrected in the code, the commits and this description.

The module key does not depend on where the file was found. The compiler derives it from the *require string* via `Compiler.module_name`, which cleanpaths `./foo` to `foo`, and the runtime's `Opal.normalize` strips the `./` too. Both sides land on `Opal.modules["foo"]` whichever way `expand` located the file — so a cwd only decides which file is read, never how it is named.

Absolute paths and `../` paths cannot be expressed as load path entries and are still returned verbatim.

## Divergences worth knowing

Pinned by specs so they aren't later mistaken for bugs:

- A `./foo` is never resolved against the **requiring file's** own directory, so one inside a nested file does not pick up that file's sibling. `require_relative` remains the way to reach one. (MRI wouldn't find the sibling either; it resolves against the CWD.)
- Without a cwd, a `./foo` that exists only in the working directory and not on the load path still raises `MissingRequire`, where MRI would load it.
- A cwd lookup accepts an extension-less file, as a load path lookup already does. MRI only tries its own known extensions.
- With a cwd, `./foo` (in the cwd) and `foo` (a different file on the load path) share the module key `foo`, and whichever is required first wins. MRI keys `$LOADED_FEATURES` by absolute path and would load both. The compile cache is content-keyed, so this cannot serve the wrong output.

## Preparatory work

Deduplication of requires was keyed by the raw require string, but `foo`, `foo.rb` and `./foo` all compile into the same `Opal.modules` entry. Once `./foo` became resolvable, the same module could be emitted more than once — verified against the CLI, where it was emitted three times. The first commit re-keys the set by `Compiler.module_name` and routes the sequential, threaded and prefork schedulers (which each kept their own bookkeeping) through a shared `#already_processed?`/`#mark_as_processed` pair.

## Documentation

The require-resolution rules now live in `docs/explanation/require_resolution.md` rather than `AGENTS.md` — background a contributor wants, not an instruction to an agent. What stays in `AGENTS.md` is the local test workflow: `bin/rake rspec` exits non-zero on a clean checkout because opal-repl needs readline, and the reported example count varies run to run because the prefork scheduler forks mid-suite, so neither is the signal it looks like.

## How to test

```sh
bundle exec rspec spec/lib/builder_spec.rb spec/lib/path_reader_spec.rb spec/lib/cli_spec.rb
```

Specs cover load-path resolution of `./`, cwd resolution of a file absent from the load path, resolution being independent of the *process* directory, the cwd surviving `dup` and a path-reader swap, the nested-file case, and single emission across all three schedulers.

Full suites run clean locally: rspec, `mspec_opal_nodejs` (684), `mspec_ruby_nodejs` (14619), `minitest_nodejs` (208 runs).

Closes #778

---

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


elia:
Rebased onto #2792, which extracts an unrelated pre-existing bug this branch's specs happened to surface (`MissingRequire` reporting an empty message under the threaded scheduler, i.e. on JRuby/TruffleRuby).

This PR now shows only the `./` require work. **Merge #2792 first**, or merge it and this base will retarget to `master` automatically.

AGENTS.md
```diff
@@ -32,8 +32,50 @@ codebase healthy.
...
+  `N examples, N failures` lines rather than the task's exit status.
+- Always run `bundle install` before the first build in a fresh checkout.
+- Run `bundle exec` from the repository root; invoking it from another working
+  directory breaks Bundler's Gemfile resolution.
+
 ## Linting
 - Run `bin/rake lint` to check code style. This builds the corelib and stdlib then executes RuboCop and ESLint.
+- To lint only what you touched: `bundle exec rubocop <files>`.
+
+## The require system
```
hmdne: Is this a good place for documenting every single PR?

hmdne:
The idea that Opal's equivalent of Dir.pwd is $LOAD_PATH union makes sense, especially in terms of CLI use. But... this has a drawback of creating an incompatible contract that we would need to support risking regressions, especially further down the line if we wish better MRI compat.

Perhaps more betterer solution would be to add a `Builder#cwd` property defaulting to `nil` (preserving current behavior), but CLI runner would set it to real Dir.pwd.

elia:
>Perhaps more betterer solution would be to add a Builder#cwd property defaulting to nil (preserving current behavior), but CLI runner would set it to real Dir.pwd.

I like that!

**About documenting stuff in AGENTS.md**

That's a good point, risks becoming a kitchen sink of info, I'll move that stuff somewhere else… my honest take is that it's ok to change it often and freely, when it's too big we'll compact and redirect the info elsewhere, that's why I take it lightly.

elia:
Done — pushed, rebased onto current master (#2792 is in, so this targets `master` directly now).

`Builder#cwd` defaults to `nil` and preserves the previous behaviour exactly; the CLI runner sets it to `Dir.pwd`. With a cwd, `./foo` resolves against it first and falls back to the load path, so setting one never makes a previously resolvable require stop working. `opal-build` and Sprockets leave it `nil`, which keeps their output independent of the invocation directory.

One correction while implementing this: my earlier justification for *not* using `Dir.pwd` was wrong. I'd claimed the runtime could never find a module resolved that way. It can — the key comes from the require string via `Compiler.module_name`, not from the resolved path, so `./foo` and `foo` both land on `Opal.modules["foo"]` regardless of which file `expand` found. A cwd only decides which file is read. Your suggestion was safe for a reason I'd argued against, so I've rewritten that reasoning in the code comment, the commit message and the PR description rather than leave the wrong version around to be rediscovered.

Two things worth flagging as deliberate:

- A cwd lookup accepts an extension-less file, matching how the load path lookup already behaves. MRI is stricter here and only tries its known extensions. I went with Opal's existing behaviour over MRI's for consistency, but happy to tighten it if you'd rather match MRI.
- With a cwd, `./foo` in the cwd and a different `foo` on the load path share a module key, first one wins. MRI keys `$LOADED_FEATURES` by absolute path and would load both. Documented rather than coded around, since it's the same first-one-wins that `foo` vs `foo.rb` already has, and the compile cache is content-keyed so nothing can serve the wrong output.

On AGENTS.md — moved the require-resolution part to `docs/explanation/require_resolution.md`. What stayed is the test-workflow bit, plus one thing that cost me time this round: the reported rspec example count varies between identical runs, because the prefork scheduler forks mid-suite and children report partial tallies. `0 failures` means something, the count doesn't.

Suites all clean: rspec, mspec_opal (684), mspec_ruby (14619), minitest (208 runs).

---

*🤖 This comment was drafted by Claude Opus 5 (1M context).*

hmdne (approved):

PR #2793, opened 2026-08-02
Make the development environment work out of the box

elia:
Splitting this out of a larger documentation branch: these are the things that stop a fresh clone from working, independent of any docs changes and much smaller to review.

## `bin/setup` did not finish

It ended with `bin/yarn install`, and `bin/yarn` unconditionally exec'd `yarnpkg`. Under a version manager that provides a `yarnpkg` shim without a pinned version, that aborts — so setup exited nonzero with no `node_modules`, and `bin/rake lint` then failed too, since `lint:eslint` shells out to yarn.

`bin/yarn` now probes for a Yarn 1.x on `PATH` and falls back to `npx yarn@1.22.22`. Pinning to 1.x is deliberate: `yarn.lock` is a v1 lockfile, and CI installs with the runner image's preinstalled yarn, which is also 1.x. Letting Yarn 4 rewrite the lockfile would be an unrelated migration.

Setup also ran `bundle update` rather than `bundle install` — the one command that rewrites `Gemfile.lock`. Simply setting up the project could produce a diff before any change was made.

## `yarn.lock` had drifted

Far enough that `--frozen-lockfile` failed: it still described the eslint 7 dependency tree while `package.json` asks for `^9.10.0`. Every install rewrote it and left the tree dirty. Refreshed, so installs are reproducible — eslint now resolves to 9.39.5.

## `rake lint` was invisible

It is the documented pre-commit check and the task CI runs, but it had no `desc`, so it never appeared in `rake -T`. It also bypassed `bin/yarn`, and failed with yarn's own error rather than anything actionable when `node_modules` was missing.

## `HACKING.md` commands that could not be copy-pasted

- `git clone git://github.com/…` — GitHub disabled the `git://` protocol in 2022.
- `rake bench:opal_node` (three occurrences) — the task is `bench:opal_nodejs`; one occurrence was also missing `rake` entirely.
- A benchmark example referencing `benchmark/bm_array_add.rb`, which does not exist.
- Bare `rake`, which fails with a `Gem::LoadError` when the system rake is newer than the one in `Gemfile.lock`. Uses `bin/rake` throughout now, as CI and `AGENTS.md` already do.

Also dropped the pasted `rake -T | grep bench` output: it listed 8 tasks where there are now 15, with descriptions that no longer match. Points at the command instead.

Documented a few things that were previously only discoverable by trial: that the default task needs a browser, that the Chrome runner finds Chrome inside `/Applications` on macOS (which is why the suite appears to need a manual install and then works anyway), that `bundle exec rspec spec/lib/compiler_spec.rb` is the fast inner loop for compiler work, and that `performance:compare` needs bun and a non-shallow clone.

Plus a short pull request template — `CONTRIBUTING.md` asks for docs updates, but nothing prompted for them at the point the PR is opened.

## Testing

```sh
bin/setup                                   # exits 0, node_modules present
bin/yarn install --frozen-lockfile          # "Already up-to-date"
bin/rake lint                               # 326 files inspected, no offenses
bundle exec rspec spec/lib/compiler_spec.rb # 79 examples, 0 failures
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,30 +1,85 @@
 # Hacking
 
+## Prerequisites
+
+- **Ruby.** The gemspec requires `>= 2.3`, and CI covers 3.0 through 4.0 plus JRuby.
+  CI's `DEFAULT_RUBY` is **3.4**, so that is the version to develop against if you
+  want the closest match to the reference build.
+- **Node.js** with `npm`. Needed for the Node-based spec runners and for the JS
+  linting step. There is no pinned version; CI uses whatever its runner image ships.
+- **A browser**, if you intend to run the full suite. See
+  [Running the tests](#running-the-tests) below.
+
 ## Quick Start
 
 [Fork opal/opal on GitHub](https://github.com/opal/opal/fork), then clone the fork to your machine:
 
-```
-$ git clone git://github.com/<YOUR-GITHUB-USERNAME>/opal.git
+```sh
+$ git clone https://github.com/<YOUR-GITHUB-USERNAME>/opal.git
 ```
 
-Setup the project:
+Setup the project. This installs gems, checks out the git submodules (`spec/mspec`,
+`spec/ruby`, `test/cruby`) and installs the JS dependencies:
 
-```
+```sh
 $ bin/setup
 ```
 
-Run the test suite:
+Then check that things work, starting with the fastest useful command:
 
+```sh
+$ bundle exec rspec spec/lib/compiler_spec.rb
 ```
-$ bundle exec rake
-```
+
+That runs 79 examples in well under a second and needs no browser and no Node.js, so
+it is the best signal that your checkout is healthy. It is also the inner loop to reach
+for while working on the compiler itself.
 
 You are now ready to make your first contribution to Opal! At a high level, your workflow will be to:
 
 1. Make changes to Opal source code
 2. Run the test suite to make sure it still passes
-3. Submit a pull request
+3. Run `bin/rake lint` before pushing
+4. Submit a pull request
+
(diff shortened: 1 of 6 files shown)


HACKING.md
```diff
@@ -1,30 +1,85 @@
...
-Run the test suite:
+Then check that things work, starting with the fastest useful command:
 
+```sh
+$ bundle exec rspec spec/lib/compiler_spec.rb
 ```
-$ bundle exec rake
-```
+
+That runs 79 examples in well under a second and needs no browser and no Node.js, so
```
elia: @claude Avoid using specific numbers that might change over time here

HACKING.md
```diff
@@ -1,30 +1,85 @@
...
 1. Make changes to Opal source code
 2. Run the test suite to make sure it still passes
-3. Submit a pull request
+3. Run `bin/rake lint` before pushing
+4. Submit a pull request
+
+Note on invoking rake: this repo ships a `bin/rake` wrapper that loads `bundler/setup`
+for you, so **prefer `bin/rake <task>`**. Bare `rake` fails with a `Gem::LoadError` when
+your system rake differs from the one in `Gemfile.lock`; `bundle exec rake` works too.
+CI uses `bin/rake`.
```
elia: No need to explain this stuff, experts will know, everyone else should use bin/rake as already advertised

HACKING.md
elia: Dropped the count — the sentence is about the command being fast and dependency-free, and "79" goes stale on every spec added or removed.

HACKING.md
elia: Removed. The doc already uses `bin/rake` in every command it shows, which is the part that matters for a newcomer.

